PR #1148, opened 2020-12-04
python gunittest: Add a new assertion for raster equivalence

marisn:
Existing raster difference tests might miss corner cases when rasters differ within specified statistics (e.g. min and max match although rasters are not equal due to presence of NULL values).
assertRastersNoDifference is reporting rasters being equal on clearly different rasters – one having NULL values and second having no NULL values as NULLs in diff raster does not affect its min/max range.

### DIFF
--- a/lib/python/gunittest/case.py
+++ b/lib/python/gunittest/case.py
@@ -282,9 +282,9 @@ class TestCase(unittest.TestCase):
 
     def assertRasterFitsInfo(self, raster, reference,
                              precision=None, msg=None):
-        r"""Test that raster map has the values obtained by r.univar module.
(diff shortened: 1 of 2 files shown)


wenzeslaus (approved):
This looks good and helpful. Thanks. More convenient comparisons is one of the two main areas testing framework should improve (the other is output for GH Actions type of CIs).

marisn:
> This looks good and helpful. Thanks. More convenient comparisons is one of the two main areas testing framework should improve (the other is output for GH Actions type of CIs).

Without this change my r.in.pdal tests (a441ba632c474daf0363e771c8401e0e0c6da1bc) were not failing on wrong values.

PR #1158, opened 2020-12-06
g.extension: add tests for download from various sources

ninsbl:
This PR adds test for downloading addons from:
- github
- gitlab
- bitbucket and
- the official addon repository as well as
- for MS Windows

Cause I was not sure if tests that download content from the internet are acceptable I deactivated tests by default. So devs would have to activate them by setting skip to False.

### DIFF
--- a/scripts/g.extension/testsuite/test_addons_download.py
+++ b/scripts/g.extension/testsuite/test_addons_download.py
@@ -121,6 +121,29 @@ class TestModuleDownloadFromDifferentSources(TestCase):
             self.assertFileExists(file)
 
     @unittest.skipIf(skip, "Skipping as chosen...")
+    def test_github_install_official_multimodule(self):
+        """Test installing multi-module extension from official addons repository"""
+        i_sentinel_files = [
+            os.path.join(
+                self.install_prefix, "scripts", "i.sentinel.parallel.download"
+            ),
+            os.path.join(
+                self.install_prefix, "docs", "html", "i.sentinel.parallel.download.html"
+            ),
+            os.path.join(self.install_prefix, "scripts", "i.sentinel.import"),
+            os.path.join(self.install_prefix, "docs", "html", "i.sentinel.import.html"),
+        ]
+        self.assertModule(
+            "g.extension", extension="i.sentinel", url=None, prefix=self.install_prefix
+        )
+
+        for file in i_sentinel_files:
+            self.assertFileExists(file)
+            if not file.endswith("html"):
+                print(file)
+                self.assertModule(file, **{"help": True})
+
+    @unittest.skipIf(skip, "Skipping as chosen...")
     def test_windows_install(self):
         """Test function for installing extension on MS Windows"""
 
--- a/scripts/g.extension/testsuite/test_addons_download.py
+++ b/scripts/g.extension/testsuite/test_addons_download.py
@@ -59,7 +59,7 @@ class TestModuleDownloadFromDifferentSources(TestCase):
 
         self.assertModule(
             "g.extension",
-            extension="r.plus.example",
+            extension="r.example.plus",
             url="https://github.com/wenzeslaus/r.example.plus",
             prefix=self.install_prefix,
         )
@@ -67,7 +67,7 @@ class TestModuleDownloadFromDifferentSources(TestCase):
         # Modules with non-standard branch would be good for testing...
         self.assertModule(
             "g.extension",
-            extension="r.plus.example",
+            extension="r.example.plus",
             url="https://github.com/wenzeslaus/r.example.plus",
             prefix=self.install_prefix,
             branch="main",
@@ -81,7 +81,7 @@ class TestModuleDownloadFromDifferentSources(TestCase):
         """Test installing extension from gitlab"""
         self.assertModule(
             "g.extension",
-            extension="r.plus.example",
+            extension="r.example.plus",
             url="https://gitlab.com/vpetras/r.example.plus",
             prefix=self.install_prefix,
         )
@@ -114,7 +114,7 @@ class TestModuleDownloadFromDifferentSources(TestCase):
             os.path.join(self.install_prefix, "docs", "html", "r.clip.html"),
         ]
         self.assertModule(
-            "g.extension", extension="r.clip", url=None, prefix=self.install_prefix
+            "g.extension", extension="v.in.gbif", url=None, prefix=self.install_prefix
         )
 
         for file in r_clip_files:
@@ -165,7 +165,7 @@ class TestModuleDownloadFromDifferentSources(TestCase):
             "prefix": self.install_prefix,
         }
         g_extension.build_platform = "x86_64"
-        g_extension.version = "784"
+        g_extension.version = "785"
         g_extension.TMPDIR = self.install_prefix
         g_extension.install_extension_win("v.in.gbif")
 
--- a/scripts/g.extension/testsuite/test_addons_download.py
+++ b/scripts/g.extension/testsuite/test_addons_download.py
@@ -1,5 +1,5 @@
 """
-TEST:      test_addons_modules.py
+TEST:      test_addons_download.py
 
 AUTHOR(S): Stefan Blumentrath <stefan dot blumentrath at nina dot no)
            based on test_addons_modules.py by
--- a/scripts/g.extension/testsuite/test_addons_download.py
+++ b/scripts/g.extension/testsuite/test_addons_download.py
@@ -1,0 +1,156 @@
+"""
+TEST:      test_addons_modules.py
+
+AUTHOR(S): Stefan Blumentrath <stefan dot blumentrath at nina dot no)
+           based on test_addons_modules.py by
+           Vaclav Petras <wenzeslaus gmail com>
+
+PURPOSE:   Test for g.extension individual modules/extensions download
+
+COPYRIGHT: (C) 2020 Stefan Blumentrath, Vaclav Petras,
+           and by the GRASS Development Team
+
+           This program is free software under the GNU General Public
+           License (>=v2). Read the file COPYING that comes with GRASS
+           for details.
+"""
+
+import os
+import imp
+from distutils.spawn import find_executable
+import unittest
+
+from grass.gunittest.case import TestCase
+from grass.gunittest.main import test
+from grass.gunittest.utils import silent_rmtree
+
+# Set to True if test is supposed to run
+skip = False
+
+
+class TestModuleDownloadFromDifferentSources(TestCase):
+
+    url = "https://github.com/wenzeslaus/r.example.plus"
+    path = os.path.join("data", "sample_modules")
+    # MS Windows install function requires absolute paths
+    install_prefix = os.path.abspath("gextension_test_install_path")
+
+    files = [
+        os.path.join(install_prefix, "scripts", "r.example.plus"),
+        os.path.join(install_prefix, "docs", "html", "r.example.plus.html"),
+    ]
+
+    def setUp(self):
+        """Make sure we are not dealing with some old files"""
+        if os.path.exists(self.install_prefix):
+            files = os.listdir(self.install_prefix)
+            if files:
+                RuntimeError(
+                    "Install prefix path '{}' contains files {}".format(
+                        self.install_prefix, files
+                    )
+                )
+
+    def tearDown(self):
+        """Remove created files"""
+        silent_rmtree(self.install_prefix)
+
+    @unittest.skipIf(skip, "Skipping as chosen...")
+    def test_github_install(self):
+        """Test installing extension from github"""
+
+        self.assertModule(
+            "g.extension",
+            extension="r.plus.example",
+            url="https://github.com/wenzeslaus/r.example.plus",
+            prefix=self.install_prefix,
+        )
+
+        # Modules with non-standard branch would be good for testing...
+        self.assertModule(
+            "g.extension",
+            extension="r.plus.example",
+            url="https://github.com/wenzeslaus/r.example.plus",
+            prefix=self.install_prefix,
+            branch="main",
+        )
+
+        for file in self.files:
+            self.assertFileExists(file)
+
+    @unittest.skipIf(skip, "Skipping as chosen...")
+    def test_gitlab_install(self):
+        """Test installing extension from gitlab"""
+        self.assertModule(
+            "g.extension",
+            extension="r.plus.example",
+            url="https://gitlab.com/vpetras/r.example.plus",
+            prefix=self.install_prefix,
+        )
+
+        for file in self.files:
+            self.assertFileExists(file)
+
+    @unittest.skipIf(skip, "Skipping as chosen...")
+    def test_bitbucket_install(self):
+        """Test installing extension from bitbucket"""
+        d_rast_files = [
+            os.path.join(self.install_prefix, "scripts", "r.sim.stats"),
+            os.path.join(self.install_prefix, "docs", "html", "r.sim.stats.html"),
+        ]
+        self.assertModule(
+            "g.extension",
+            extension="r.sim.stats",
+            url="https://bitbucket.org/lrntct/r.sim.stats",
+            prefix=self.install_prefix,
+        )
+
+        for file in d_rast_files:
+            self.assertFileExists(file)
+
+    @unittest.skipIf(skip, "Skipping as chosen...")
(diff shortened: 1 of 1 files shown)


ninsbl:
See also #625

ninsbl:
> Cause I was not sure if tests that download content from the internet are acceptable I deactivated tests by default. So devs would have to activate them by setting skip to False.

Any opinion on this?

neteler:
Testing it with Python 3.9.4 shows:

```
testsuite/test_addons_download.py:19: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
```

Could that be updated?

Then some tests fail since `branch` has a default setting but `url=None`, e.g.

https://github.com/OSGeo/grass/pull/1158/files#diff-826354815781b87e6d667944c79e46e0ae51bc1b9debf80295293494201951ddR137

which leads to

```
  File "/home/mneteler/software/grass_master/scripts/g.extension/testsuite/test_addons_download.py", line 136, in test_github_install_official_multimodule
    self.assertModule(
  File "/home/mneteler/software/grass_master/dist.x86_64-pc-linux-gnu/etc/python/grass/gunittest/case.py", line 1419, in assertModule
AssertionError: Running <g.extension> module ended with non-zero return code (1)
Called: g.extension(extension='i.sentinel', operation='add', prefix='/home/mneteler/software/grass_master/scripts/g.extension/testsuite/gextension_test_install_path', branch='main')
See the following errors:
...

        url   URL or directory to get the extension from (supported only on Linux and Mac)
...
     branch   Specific branch to fetch addon from (only used when fetching from git)
              default: main

ERROR: Option <branch> requires <url>
```

ninsbl:
> Then some tests fail since `branch` has a default setting but `url=None`, e.g.
> 
> https://github.com/OSGeo/grass/pull/1158/files#diff-826354815781b87e6d667944c79e46e0ae51bc1b9debf80295293494201951ddR137
> 
> ERROR: Option <branch> requires <url>

Using `run_command` from `grass.script` does not throw errors related to `branch` and `url`, but any pygrass based call to `g.extension` does.

We should probably remove the requirement to specify `url` in `g.extension` if `branch` option is used (which comes with `main` as the default which causes the issues in the tests).

wenzeslaus:
> Using `run_command` from `grass.script` does not throw errors related to `branch` and `url`, but any pygrass based call to `g.extension` does.

That sounds like pygrass is broken, not the g.extension interface.

wenzeslaus:
> I was not sure if tests that download content from the internet are acceptable I deactivated tests by default.

This is tricky either way. Not running the tests when we have them? Unfortunate. Not for datasets and normal tests, but in general, downloading is okay in the CI and we do it (dependencies, sample dataset). It may fail time to time due to 3rd party issues and that's fine.

I would say, let's have it enabled and fix it when it doesn't work. (An alternative is having a separate GitHub Action or job which will do this special test.)

The URLs it is reaching should be somewhat reliable and perhaps under our control. For example, we have the alternative GRASS org on GitHub.

wenzeslaus (changes requested):
There is still couple of `r.plus.example`.

I don't get the reasons for the MS Win test. Can you comment more in the code?

r.clip, perhaps a candidate to be moved to core like v.clip? Maybe a different module would be a better match for this test. Also one Python and one C module might be a good idea.

ninsbl:
> There is still couple of `r.plus.example`.

Yet all tests pass locally. I think the module name is not relevant if you install from single module repositories, as the URL is build just with the repo name... But I wil change that so it is more clean...
 
> I don't get the reasons for the MS Win test. Can you comment more in the code?

When writing the tests, I thought there were no tests on MS Windows. On MS Windows, installation is mainly downloading and extracting a zip file, so the tests checks the windows specific functions for that on Linux.
An issue with running this test on Windows may still be that addons are pre-compiled and installation may fail because of the version mismatch between GRASS compiled for testing and GRASS version compiled to build addon binaries... Which I guess is why OSGeo uses this patch:
https://github.com/jef-n/OSGeo4W/blob/a372e6fd3a5c9962146d1967a067c326ee814ef3/src/grass/osgeo4w/patch#L57-L64
Yet, I am not sure if that kicks in during tests of g.extension... Maybe the best way is to comment it out for now, then see how the testcase behaves on MS Windows (will try to test locally) and remove it if we can run the testcase on MS Widows as well...

> r.clip, perhaps a candidate to be moved to core like v.clip? Maybe a different module would be a better match for this test. Also one Python and one C module might be a good idea.

Yes, agreed. Will change that.

wenzeslaus:
> I think the module name is not relevant if you install from single module repositories

Agreed.

> > I don't get the reasons for the MS Win test. Can you comment more in the code?
> 
> When writing the tests, I thought there were no tests on MS Windows.

Thanks, that makes sense.

> An issue with running this test on Windows may still be that addons are pre-compiled and installation may fail because of the version mismatch between GRASS compiled for testing and GRASS version compiled to build addon binaries...

I didn't think about that. The check is performed when a module runs, but I don't think modules run during Windows installation, so we may be okay as long as the download links are resolved properly.

> Which I guess is why OSGeo uses this patch:
> https://github.com/jef-n/OSGeo4W/blob/a372e6fd3a5c9962146d1967a067c326ee814ef3/src/grass/osgeo4w/patch#L57-L64

Unfortunate situation. Well, good to know about it.

> Maybe the best way is to comment it out for now, then see how the testcase behaves on MS Windows (will try to test locally) and remove it if we can run the testcase on MS Widows as well...

It is definitively more advanced test than I would write. I guess I would just add some secret flag (not a g.parser one) to g.extension, but this seems too be more general and it does not mess up the interface and tests can be outside of the already long module code. There might be value in testing things this way, e.g., you can test locally on Linux. On the other hand, if it makes things more complicated and the same functionality is easily tested elsewhere, I would not keep it for simplicity.

wenzeslaus (commented):
The commit message says add C module, but it changes the module tested in the official module test instead of having two tests. Intentional?

ninsbl:
The failing test does not seem to be related to this PR.
Suggest to remove the relatively complex windows test function later if it should causes trouble.
The rest is hopefully addressed!?

wenzeslaus (changes requested):
This seems to have a lot of files which are not related to the PR and likely not really part of the source code at all. Accidental `git -A`?

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,164 @@
...
+            url="https://gitlab.com/vpetras/r.example.plus",
+            prefix=self.install_prefix,
+        )
+
+        for file in self.files:
+            self.assertFileExists(file)
+
+    def test_bitbucket_install(self):
+        """Test installing extension from bitbucket"""
+        d_rast_files = [
```
wenzeslaus, line 87:
```suggestion
        files = [
```

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,164 @@
...
+            url="https://bitbucket.org/lrntct/r.sim.stats",
+            prefix=self.install_prefix,
+        )
+
+        for file in d_rast_files:
+            self.assertFileExists(file)
+
+    def test_github_install_official(self):
+        """Test installing C-extension from official addons repository"""
+        r_gdd_files = [
```
wenzeslaus, line 103:
```suggestion
        files = [
```

Not wrong unlike the other case, but the code is short thus the variable clear. Confusion with self.files is possible, but easily discovered.

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,164 @@
...
+    def test_github_install_official_multimodule(self):
+        """Test installing multi-module extension from official addons repository"""
+        i_sentinel_files = [
+            os.path.join(
+                self.install_prefix, "scripts", "i.sentinel.parallel.download"
+            ),
+            os.path.join(
+                self.install_prefix, "docs", "html", "i.sentinel.parallel.download.html"
+            ),
+            os.path.join(self.install_prefix, "scripts", "i.sentinel.import"),
```
wenzeslaus, line 121: self.install_prefix as Path may make this code much shorter. We are using it more and more in the code as is everybody else, so it should be quite readable.

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,164 @@
...
+        ]
+        self.assertModule(
+            "g.extension", extension="i.sentinel", prefix=self.install_prefix
+        )
+
+        for file in i_sentinel_files:
+            self.assertFileExists(file)
+            if not file.endswith("html"):
+                print(file)
+                self.assertModule(file, **{"help": True})
```
wenzeslaus, line 132: help=True doesn't work or is some linter giving you a hard time for that?

```python
>>> # works
>>> def a(help):
...     print(help)
... 
>>> a(help="A")
A
```

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,164 @@
...
+        ]
+        g_extension_path = find_executable("g.extension")
+        g_extension = SourceFileLoader("g.extension", g_extension_path).load_module()
+        g_extension.options = {
+            "extension": "v.in.gbif",
+            "operation": "add",
+            "prefix": self.install_prefix,
+        }
+        g_extension.build_platform = "x86_64"
+        g_extension.version = "785"
```
wenzeslaus, line 155: You could use g.version/grass.script.version for this.

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+
+        for file in i_sentinel_files:
+            self.assertFileExists(file)
+            if not file.endswith("html"):
+                print(file)
+                self.assertModule(file, **{"help": True})
+
+    def test_windows_install(self):
+        """Test function for installing extension on MS Windows"""
```
wenzeslaus, line 135: This function really needs more info/motivation why it is here. You can put it in the docstring.

wenzeslaus (changes requested):

ninsbl:
With OSGeo4W tests in CI, I am all fine with removing the MS Windows test function, if that is preferable.

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,141 @@
...
+            prefix=str(self.install_prefix),
+        )
+
+        # Modules with non-standard branch would be good for testing...
+        self.assertModule(
+            "g.extension",
+            extension="r.example.plus",
+            url="https://github.com/wenzeslaus/r.example.plus",
+            prefix=str(self.install_prefix),
+        )
```
wenzeslaus, line 72: Is this a placeholder for a future test?

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,141 @@
+"""
+TEST:      test_addons_download.py
+
+AUTHOR(S): Stefan Blumentrath <stefan dot blumentrath at nina dot no)
+           based on test_addons_modules.py by
+           Vaclav Petras <wenzeslaus gmail com>
```
wenzeslaus, line 6: I appreciate giving credit and tracking origin, but perhaps not even needed in this case, esp. after all the modernization you did (same for copyright line).

Maybe mention in the class or module docstring that it is using the presence of the files as a test like the other test does. (You will also create a useful docstring and Pylint will be happier.)

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,141 @@
...
+ms_windows = sys.platform == "win32"
+
+
+class TestModuleDownloadFromDifferentSources(TestCase):
+
+    # MS Windows install function requires absolute paths
+    install_prefix = Path("gextension_test_install_path").absolute()
+
+    files = [
+        install_prefix.joinpath("scripts", "r.example.plus"),
```
wenzeslaus, line 36: When I was suggesting Path, I was really thinking:

```suggestion
        install_prefix / "scripts" / "r.example.plus",
```

For single line, it is shorter. I'm not sure if it is readable for multiple lines, _joinpath_ might be better when it needs a line break.

Totally optional, though.

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,141 @@
...
+    ]
+
+    def setUp(self):
+        """Make sure we are not dealing with some old files"""
+        if self.install_prefix.exists():
+            files = list(path.name for path in self.install_prefix.iterdir())
+            if files:
+                RuntimeError(
+                    "Install prefix path '{}' contains files {}".format(
+                        str(self.install_prefix), files
```
wenzeslaus, line 47: Now, f-string is the way to go here.

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,141 @@
+"""
+TEST:      test_addons_download.py
+
+AUTHOR(S): Stefan Blumentrath <stefan dot blumentrath at nina dot no)
+           based on test_addons_modules.py by
+           Vaclav Petras <wenzeslaus gmail com>
+
+PURPOSE:   Test for g.extension individual modules/extensions download
+
+COPYRIGHT: (C) 2020 Stefan Blumentrath, Vaclav Petras,
```
wenzeslaus, line 10: Oh well, let's update to 2022.

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,141 @@
...
+            if files:
+                RuntimeError(
+                    "Install prefix path '{}' contains files {}".format(
+                        str(self.install_prefix), files
+                    )
+                )
+
+    def tearDown(self):
+        """Remove created files"""
+        silent_rmtree(str(self.install_prefix))
```
wenzeslaus, line 53: Out of scope for this PR, but silent_rmtree should support Path (same for all path handling functions). I recall something about shutil not supporting it initially, but later on adding it.

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -0,0 +1,141 @@
...
+            self.install_prefix.joinpath("scripts", "i.sentinel.import"),
+            self.install_prefix.joinpath("docs", "html", "i.sentinel.import.html"),
+        ]
+        self.assertModule(
+            "g.extension", extension="i.sentinel", prefix=str(self.install_prefix)
+        )
+
+        for file in files:
+            self.assertFileExists(file)
+            if not str(file).endswith("html"):
```
wenzeslaus, line 136: This should work the same and is more explicit in what the goal is:

```suggestion
            if not file.suffix == "html":
```

wenzeslaus (changes requested):
Thanks. I have nothing too serious except perhaps one item. It could go in as is, but I think some things could be modernized further.

scripts/g.extension/testsuite/test_addons_download.py
ninsbl: In a way, but as it is now it is only a duplication. So I removed it in [b0e1215](https://github.com/OSGeo/grass/pull/1158/commits/b0e12150274791f3dd7f03c62a47245095145a15)

scripts/g.extension/testsuite/test_addons_download.py
ninsbl: Was not aware of this possibility. Good to know and implemented in [b0e1215](https://github.com/OSGeo/grass/pull/1158/commits/b0e12150274791f3dd7f03c62a47245095145a15)

scripts/g.extension/testsuite/test_addons_download.py
ninsbl: Changed in [b0e1215](https://github.com/OSGeo/grass/pull/1158/commits/b0e12150274791f3dd7f03c62a47245095145a15)

scripts/g.extension/testsuite/test_addons_download.py
ninsbl: Implemented

scripts/g.extension/testsuite/test_addons_download.py
ninsbl: Implemented too

wenzeslaus (approved):

PR #1162, opened 2020-12-08
g.extension: support token authentication for gitlab/hub

ninsbl:
It seems gitlab and github only support token authentication for private repos. This PR implements such authentication to private github/gitlab repositories, so they can be accessed by g.extension.

My main question (and reason this is WIP) however is: How to we access the users Private Access Token (PAT) in a reliable and appropriate way. Current implementation assumes the token is stored in an environment variable named "PRIVATE_ACCESS_TOKEN" as environment variables are recommended for tokens. Would it be better to read from a file with the token or stdin?

See also: #625

### DIFF
--- a/scripts/g.extension/g.extension.html
+++ b/scripts/g.extension/g.extension.html
@@ -88,11 +88,22 @@ For well known general hosting services, namely GitHub, GitLab and Bitbucket,
 <em>g.extension</em> supports the download of a repository as a ZIP file.
 Here the user only needs to provide a base URL to the repository web page
 (with or without the <tt>https://</tt> part).
-For GitLab and Bitbucket, the latest source code in the default branch is 
-downloaded, for GitHub, the latest source code in the master branch is downloaded.
-Of course, a user can still specify the full URL of a ZIP file
-and install a specific branch or release in this way (ZIP file mechanism
-will be applied).
+If not otherwise specified in the <b>branch</b> option, the default branch 
+(main or master) is downloaded. Of course, a user can still specify the full 
+URL of a ZIP file and install a specific branch, commit or release in this way 
+(ZIP file mechanism will be applied).
+
+<p>
+For GitHub and GitLab authentication with a Personal Access Token (PAT) 
+is supported in order to access private repositories. For PAT authentication, 
+first a Personal Access Token needs to be acquired from either
+<a href="https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html">
+GitLab</a> or <a href="https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token">
+GitHub</a>. For security reasons it is recommended to pass PATs either 
+through environment variables or files to the command. <em>g.extension</em> 
+will automatically check for an environment variable named
+PERSONAL_ACCESS_TOKEN and use it for authentication, unless a PAT is 
+provided through the <b>token</b> option.
 
 <p>
 For the official repository, <em>g.extension</em> supports listing available
@@ -192,6 +203,13 @@ Simple URL to GitHub, GitLab, Bitbucket repositories from a specific (e.g. devel
 
 <div class="code"><pre>
 g.extension r.example.plus url="https://github.com/wenzeslaus/r.example.plus" branch=master
+</pre></div>
+
+Private GitHub or GitLab repository with token authentication:
+
+<div class="code"><pre>
+export PERSONAL_ACCESS_TOKEN=my_personal_access_token
+g.extension my_private_repo url="https://github.com/my_user/my_private_repo"
 </pre></div>
 
 Simple URL to OSGeo Trac (downloads a ZIP file, requires download to be enabled in Trac):
--- a/scripts/g.extension/g.extension.html
+++ b/scripts/g.extension/g.extension.html
@@ -102,7 +102,7 @@ GitLab</a> or <a href="https://docs.github.com/en/free-pro-team@latest/github/au
 GitHub</a>. For security reasons it is recommended to pass PATs either 
 through environment variables or files to the command. <em>g.extension</em> 
 will automatically check for an environment variable named
-PERSONAL_ACCESS_TOKEN and use it for authentication, unless a PAT is 
+<code>PERSONAL_ACCESS_TOKEN</code> and use it for authentication, unless a PAT is 
 provided through the <b>token</b> option.
 
 <p>
--- a/scripts/g.extension/g.extension.html
+++ b/scripts/g.extension/g.extension.html
@@ -88,7 +88,7 @@ For well known general hosting services, namely GitHub, GitLab and Bitbucket,
 <em>g.extension</em> supports the download of a repository as a ZIP file.
 Here the user only needs to provide a base URL to the repository web page
 (with or without the <tt>https://</tt> part).
-If not otherwise specified in the <b>branch</b> option, the default branch 
+If not specified otherwise in the <b>branch</b> option, the default branch 
 (main or master) is downloaded. Of course, a user can still specify the full 
 URL of a ZIP file and install a specific branch, commit or release in this way 
 (ZIP file mechanism will be applied).
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -2202,6 +2202,17 @@ def resolve_source_code(url=None, name=None, branch=None):
                 pass
             # Test if github repo exists (need to use API)
             try:
+                # Check token scopes (permission)
+                open_url = urlopen("https://api.github.com/users/{}".format(
+                    *url.split('/')[-2:-1]))
+                scopes = open_url.getheader('X-OAuth-Scopes')
+                if 'repo' not in scopes:
+                    grass.fatal(
+                        _("Your access token don't have permission to "
+                          "read private repository info. Add 'repo' "
+                          "scope which define the access for personal token "
+                          "on the url <https://github.com/settings/tokens>")
+                    )
                 open_url = urlopen('https://api.github.com/repos/{}/{}'.format(*url.split("/")[-2:]))
                 open_url.close()
                 url_validated = True
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -2334,17 +2334,16 @@ def main():
     if token and hosting:
         HEADERS['Authorization'] = "{} {}".format(hosting[0], token)
 
-
     # Get token from input option
     token_input = False
     if options['token']:
         if options['token'] == "-":
-            token_input = sys.stdin.read().rstrip()
+            token_input = sys.stdin.readline().rstrip()
             if not token_input:
                 grass.fatal(_('No token specified. Please try again.'))
         else:
             try:
-                with open(options['token'], 'r') as tokenfile:
+                with open(options['token']) as tokenfile:
                     token_input = tokenfile.readline().rstrip()
                     if not token_input:
                         grass.fatal(
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -2216,6 +2216,10 @@ def resolve_source_code(url=None, name=None, branch=None):
                 open_url = urlopen('https://api.github.com/repos/{}/{}'.format(*url.split("/")[-2:]))
                 open_url.close()
                 url_validated = True
+            except HTTPError as err:
+                if (err.code == 403 and
+                    err.msg == 'rate limit exceeded'):
+                    gscript.warning(_('GitHub API rate limit exceeded.'))
             except:
                 pass
             # Test if gitlab repo exists (need to use API)
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -2200,6 +2200,20 @@ def resolve_source_code(url=None, name=None, branch=None):
                 url_validated = True
             except:
                 pass
+            # Test if github repo exists (need to use API)
+            try:
+                open_url = urlopen('https://api.github.com/repos/{}/{}'.format(*url.split("/")[-2:]))
+                open_url.close()
+                url_validated = True
+            except:
+                pass
+            # Test if gitlab repo exists (need to use API)
+            try:
+                open_url = urlopen('https://gitlab.com/api/v4/projects/{}%2F{}'.format(*url.split("/")[-2:]))
+                open_url.close()
+                url_validated = True
+            except:
+                pass
         else:
             try:
                 open_url = urlopen('http://' + url)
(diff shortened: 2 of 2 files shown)


scripts/g.extension/g.extension.py
```diff
@@ -2200,6 +2200,20 @@ def resolve_source_code(url=None, name=None, branch=None):
                 url_validated = True
             except:
                 pass
+            # Test if github repo exists (need to use API)
+            try:
+                open_url = urlopen('https://api.github.com/repos/{}/{}'.format(*url.split("/")[-2:]))
```
tmszi, line 2205: Before checking if private repo url exists, it will be good checking if token has scope (permission) to get list of private repos (I'm going to add it in next commit). 

I'm using double quotes in python string interpolation in case `open_url = urlopen("https://api.github.com/repos/{}/{}".format(` and simple quotes in case `*url.split('/')`.

scripts/g.extension/g.extension.py
```diff
@@ -2200,6 +2200,20 @@ def resolve_source_code(url=None, name=None, branch=None):
...
+                open_url.close()
+                url_validated = True
+            except:
```
tmszi, line 2208: It will be good catch 'GitHub API rate limit exceeded.' exception and inform user. (I'm going to add it in next commit).

scripts/g.extension/g.extension.py
```diff
@@ -2200,6 +2200,20 @@ def resolve_source_code(url=None, name=None, branch=None):
...
+                pass
+            # Test if gitlab repo exists (need to use API)
+            try:
+                open_url = urlopen('https://gitlab.com/api/v4/projects/{}%2F{}'.format(*url.split("/")[-2:]))
```
tmszi, line 2212: I'm using double quotes in python string interpolation in case `open_url = urlopen("https://api.github.com/repos/{}/{}".format(` and simple quotes in case `*url.split('/')`.

scripts/g.extension/g.extension.py
```diff
@@ -2281,6 +2295,18 @@ def main():
...
+        "gitlab": "Bearer",
+        "github": "token"
+    }
+
+    hosting = [token_tags[key] for key in token_tags if key in original_url]
+    token = os.getenv('PRIVATE_ACCESS_TOKEN')
+
+    if token and hosting:
+        global HEADERS
+        HEADERS["Authorization"] = "{} {}".format(hosting[0], token)
```
tmszi, line 2308: double quotes -> single quotes  'Authorization'

scripts/g.extension/g.extension.py
```diff
@@ -2281,6 +2295,18 @@ def main():
         opener = urlrequest.build_opener(proxy)
         urlrequest.install_opener(opener)
 
+    token_tags = {
+        "gitlab": "Bearer",
```
tmszi, line 2299: double quotes -> single quotes

tmszi (changes requested):

tmszi:
> Would it be better to read from a file with the token or stdin?

My opinion is combination stdin (module option) with shell env var `PRIVATE_ACCESS_TOKEN`.

tmszi:
So far, I have tested the add-on installation from the private GitHub repository and it works correctly (I currently don't have an account on the GitLab server).

scripts/g.extension/g.extension.py
ninsbl: Thanks for adding that!

ninsbl:
> My opinion is combination stdin (module option) with shell env var `PRIVATE_ACCESS_TOKEN`.
Implemented.

ninsbl:
Documentation added. This should be ready for merging now (except maybe (hopefully) for proofreading of the manual changes).

scripts/g.extension/g.extension.html
```diff
@@ -88,11 +88,22 @@ <h4>Online repositories: GitHub, GitLab and Bitbucket</h4>
 <em>g.extension</em> supports the download of a repository as a ZIP file.
 Here the user only needs to provide a base URL to the repository web page
 (with or without the <tt>https://</tt> part).
-For GitLab and Bitbucket, the latest source code in the default branch is 
-downloaded, for GitHub, the latest source code in the master branch is downloaded.
-Of course, a user can still specify the full URL of a ZIP file
-and install a specific branch or release in this way (ZIP file mechanism
-will be applied).
+If not otherwise specified in the <b>branch</b> option, the default branch
```
neteler, line 91:
```suggestion
If not specified otherwise in the <b>branch</b> option, the default branch 
```

scripts/g.extension/g.extension.html
```diff
@@ -88,11 +88,22 @@ <h4>Online repositories: GitHub, GitLab and Bitbucket</h4>
...
+<p>
+For GitHub and GitLab authentication with a Personal Access Token (PAT) 
+is supported in order to access private repositories. For PAT authentication, 
+first a Personal Access Token needs to be acquired from either
+<a href="https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html">
+GitLab</a> or <a href="https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token">
+GitHub</a>. For security reasons it is recommended to pass PATs either 
+through environment variables or files to the command. <em>g.extension</em> 
+will automatically check for an environment variable named
+PERSONAL_ACCESS_TOKEN and use it for authentication, unless a PAT is
```
neteler, line 105:
```suggestion
<code>PERSONAL_ACCESS_TOKEN</code> and use it for authentication, unless a PAT is 
```

scripts/g.extension/g.extension.py
```diff
@@ -2281,6 +2317,42 @@ def main():
...
+    # Set authorization method according to supported hosting service
+    if token and hosting:
+        HEADERS['Authorization'] = "{} {}".format(hosting[0], token)
+
+
+    # Get token from input option
+    token_input = False
+    if options['token']:
+        if options['token'] == "-":
+            token_input = sys.stdin.read().rstrip()
```
tmszi, line 2342: `sys.stdin.readline()`

scripts/g.extension/g.extension.py
```diff
@@ -2281,6 +2317,42 @@ def main():
...
+        else:
+            with open(options['token'], 'r') as tokenfile:
```
tmszi: It will be good handle some file exception FileNotFoundError... (I'm going to add in the next commit).

scripts/g.extension/g.extension.py
```diff
@@ -2281,6 +2317,42 @@ def main():
...
+    hosting = [token_tags[key] for key in token_tags if key in original_url]
+
+    # Get token from hardcoded environment variable
+    token = os.getenv('PERSONAL_ACCESS_TOKEN')
+
+    global HEADERS
+    # Set authorization method according to supported hosting service
+    if token and hosting:
+        HEADERS['Authorization'] = "{} {}".format(hosting[0], token)
+
```
tmszi, line 2336: Remove one line space, please.

scripts/g.extension/g.extension.py
tmszi: simplier `open(options['token']) as tokenfile:`

tmszi (changes requested):
Great work. I tested it and it works correctly.

scripts/g.extension/g.extension.py
ninsbl: Applied in af21fdc

scripts/g.extension/g.extension.py
ninsbl: Applied in af21fdc

scripts/g.extension/g.extension.py
ninsbl: Thanks!

scripts/g.extension/g.extension.py
ninsbl: Applied in af21fdc

tmszi:
- [ ] test the installation of the add-on from the gitlab private repository

tmszi (approved):

tmszi:
I tested the Add-On installation from a GitLab private repository (token has all scopes).

```
GRASS nc_basic_spm_grass7/PERMANENT:~ > g.extension test url=https://gitlab.com/tmszi/test token='-'
token
Fetching <test> from
<https://gitlab.com/tmszi/test/-/archive/main/test-main.zip> (be
patient)...
Expected default branch not found. Trying again from
<https://gitlab.com/tmszi/test/-/archive/master/test-master.zip>...
ERROR: Extension <test> not found. Please check 'url' and 'branch' options
```
URL `https://gitlab.com/tmszi/test/-/archive/main/test-main.zip` exists (I tried download archive from web browser).

Test with curl:
```
curl -H 'User-Agent: Mozilla/5.0' -H 'Authorization: Bearer token' -X GET https://gitlab.com/tmszi/test/-/archive/main/test-main.zip -o /tmp/test-main.zip

file /tmp/test-main.zip 
test-main.zip: HTML document, ASCII text, with no line terminators

cat /tmp/test-main.zip 
<html><body>You are being <a href="https://gitlab.com/users/sign_in">redirected</a>.</body></html>
```

Test with curl but with GitLab API URL endpoint for download repository archive:
```
curl -H 'User-Agent: Mozilla/5.0' -H 'Authorization: Bearer token' -X GET https://gitlab.com/api/v4/projects/:project_id/repository/archive.zip

file /tmp/test-main.zip
/tmp/test-main.zip: Zip archive data, at least v1.0 to extract 
```

It looks like the address `https://gitlab.com/tmszi/test/-/archive/main/test-main.zip` for downloading the archive of the private repository doesn't work with the token.

tmszi (changes requested):
Installing the Add-On from the GitLab private repository doesn't seem to work as expected.

ninsbl:
Thanks for testing, Tomas (and all your support on this issue). The problem with private repos in gitlab is probably related to:
https://gitlab.com/gitlab-org/gitlab/-/issues/28978
The URL works for public repos. For private repos, I am afraid we would have to use the gitlab API until the above bug is fixed.

Using the API would require to fetch the repo ID first:
`curl -L -H 'Authorization: Bearer my_token' https://gitlab.com/api/v4/projects/11027718/repository/archive.zip -o test.zip`
Which makes it unfortunately a major change regarding how urls are handled right now...

Thi is how you get the repo ID through the API:
https://gitlab.com/api/v4/projects/ninsbl%2Fr.example.plus

tmszi:
> Thanks for testing, Tomas (and all your support on this issue). The problem with private repos in gitlab is probably related to:
> https://gitlab.com/gitlab-org/gitlab/-/issues/28978
> The URL works for public repos. For private repos, I am afraid we would have to use the gitlab API until the above bug is fixed.

You're welcome, Stefan. Thanks for analysis. I implemented it in the commit  [25a34c7](https://github.com/OSGeo/grass/pull/1162/commits/25a34c7e30e17011f6c3f4a48d45ab2d097a3778). I tested it and it works properly, could you test it too?

ninsbl:
> I implemented it in the commit [25a34c7](https://github.com/OSGeo/grass/pull/1162/commits/25a34c7e30e17011f6c3f4a48d45ab2d097a3778). I tested it and it works properly, could you test it too?

Nice. I can confirm that it works now. Furthermore, I had a look at validation of the token scope from gitlab. Unfortunately, this does not look straight forward either. One can get a list of tokens through the API: https://gitlab.com/api/v4/personal_access_tokens

However, one can have several tokens, and I cannot see how one can find out wich one is used...

So, I think we can merge with the state now...

tmszi:
> > I implemented it in the commit [25a34c7](https://github.com/OSGeo/grass/pull/1162/commits/25a34c7e30e17011f6c3f4a48d45ab2d097a3778). I tested it and it works properly, could you test it too?
> 
> Nice. I can confirm that it works now. Furthermore, I had a look at validation of the token scope from gitlab. Unfortunately, this does not look straight forward either. One can get a list of tokens through the API: https://gitlab.com/api/v4/personal_access_tokens
> 
> However, one can have several tokens, and I cannot see how one can find out wich one is used...

I also didn't find it in GitLab API endpoints.

In the commit [13ff843](https://github.com/OSGeo/grass/pull/1162/commits/13ff84325b83a8cfcec2e39a0d588cb975d73f5c), I use GitHub API for downloading repo archive too (same as for GitLab), because it allows you to specify commit, and allows you to installation Add-On from user specified commit (for public and private repo).

`g.extension <your extension name> url=<your extension url> branch=<your extension commit>`

tmszi:
Could @ninsbl test commit [13ff843](https://github.com/OSGeo/grass/pull/1162/commits/13ff84325b83a8cfcec2e39a0d588cb975d73f5c) again, please (GitLab/GitHub public/private repository installation from different branch and commit. I tested it, and works as expected.)? Thanks for your time.

ninsbl:
> Could @ninsbl test commit [13ff843](https://github.com/OSGeo/grass/pull/1162/commits/13ff84325b83a8cfcec2e39a0d588cb975d73f5c) again, please (GitLab/GitHub public/private repository installation from different branch and commit. I tested it, and works as expected.)? 

Thanks! Yes, that works fine for me as well. 

I considered adding test for the existence of gitlab tokens (and their minimal scope) but it seems one needs a valid token to list tokens, so that check would be rather pointless...

Cause the solution with the API now also supports commits, and not only branches, I am wondering f the option should be renamed from "branch" to "reference" or something like that (for GRASS 8)?
At least we should probably mention the possibility to specify a commit hash in the branch option?

tmszi:
> Cause the solution with the API now also supports commits, and not only branches, I am wondering f the option should be renamed from "branch" to "reference" or something like that (for GRASS 8)?

Yes I agree ("branch" to "reference"). Will you make this change?

> At least we should probably mention the possibility to specify a commit hash in the branch option?

Yes, I also agree

ninsbl:
> > Cause the solution with the API now also supports commits, and not only branches, I am wondering f the option should be renamed from "branch" to "reference" or something like that (for GRASS 8)?
> 
> Yes I agree ("branch" to "reference"). Will you make this change?

@neteler , any objections agains renaming the recently added option `branch` to `reference` (in GRASS 8 only I guess)?

neteler:
> > > Cause the solution with the API now also supports commits, and not only branches, I am wondering f the option should be renamed from "branch" to "reference" or something like that (for GRASS 8)?
> > 
> > Yes I agree ("branch" to "reference"). Will you make this change?
> 
> @neteler , any objections agains renaming the recently added option `branch` to `reference` (in GRASS 8 only I guess)?

No strong opinion here. @anikaweinmann what do you think?

tmszi:
In the [fff3103](https://github.com/OSGeo/grass/pull/1162/commits/fff3103c61f0743aa77d4aaa7a684c750853c1a4) fixed GitHub API repository archive URL repository path (I had hardcoded my test repository name in this URL).

ninsbl:
With the switch to sparse checkout, this is no longer relevant. Authentication to private repositories could be achieved by using SSH, I guess... So, an update of the documentation seems sufficient...

PR #1166, opened 2020-12-10
Change default color for rendering vectors (gray->blue) [news]

wenzeslaus:
Use light and dark blue as defaults in d.vect and in the default GUI settings
for rendering vector maps.

The d.vect values are applied in d.mon and with direct rendering,
the GUI values when user does not have user settings yet and
when user settings is saved for the first time (since then it is driven
by value there, so this needs settings to not exists to take its effect).

## Points (default marker x)

points_only_x

## Lines (default width 0/1)

lines

## Areas (default border 0/1)

areas

## Points with marker point (small circle)

points_only_o

## Lines with widths 1 and 3

lines_wide

## Comparison of old and new in action

Screenshot from 2020-12-09 23-01-23

### DIFF
--- a/display/d.vect/main.c
+++ b/display/d.vect/main.c
@@ -118,11 +118,12 @@ int main(int argc, char **argv)
     /* Colors */
     color_opt = G_define_standard_option(G_OPT_CN);
     color_opt->label = _("Feature color");
+    color_opt->answer = "0:29:57";
     color_opt->guisection = _("Colors");
     
     fcolor_opt = G_define_standard_option(G_OPT_CN);
     fcolor_opt->key = "fill_color";
-    fcolor_opt->answer = "200:200:200";
+    fcolor_opt->answer = "0:103:204";
     fcolor_opt->label = _("Area fill color");
     fcolor_opt->guisection = _("Colors");
 
--- a/gui/wxpython/core/settings.py
+++ b/gui/wxpython/core/settings.py
@@ -339,13 +339,13 @@ class Settings:
             #
             'vectorLayer': {
                 'featureColor': {
-                    'color': (0, 0, 0),
+                    'color': (0, 29, 57),
                     'transparent': {
                         'enabled': False
                     }
                 },
                 'areaFillColor': {
-                    'color': (200, 200, 200),
+                    'color': (0, 103, 204),
                     'transparent': {
                         'enabled': False
                     }


mlennert:
Not an opposition, but curiosity: is there a reasoning behind this that goes beyond personal aesthetics ? ;-)

veroandreo:
Aren't blue tones somehow reserved for water features? I wouldn't like to see that for keeping these new default colors, users create maps in which roads can be confused with canals/rivers, for example. 

Maybe a cartographer can enlight us here. Anyone?

wenzeslaus:
> is there a reasoning behind this that goes beyond personal aesthetics ? ;-)

We already switched from gray scale to pretty colors for rasters. As for the particular choice:

* With green it is difficult to get a nice one. Yellow is (at least currently) selection. Red is danger. Violet, pink, purple... too controversial. So, if I need to pick only one here, I go with blue.
* The particular combo of blues is just something which seemed nice to me.
* However, they actually work well for both light and dark background (white nice, gray nice, black okayish). Additionally, blue is a common default color schemes with red partially because the blue-red combo is easier to get color-blind friendly than the green-red one (not tested here, but one is darker than the other so that's a good start).

This commit gives clear idea where to change it if there is a better suggestion later (along the lines of a simple change of defaults), so I would just merge this soon and leave it open for criticism in new issues.

wenzeslaus:
Ideally, the colors would be different and (seemingly) random for each newly added layer (actually, selected from a pool of nice ones according to type of features in the vector if only one is present), but that's far beyond what I want to do in this PR.

nilason:
I like this suggestion. I'm also looking forward to a random selection of colour (from normal/dark mode aware and attuned sets).
I don't think there is a cartographer who would leave a job with default colours only anyway. I know I wouldn't.
Let's make these modern monitors with support of millions of colours work :-)

wenzeslaus:
> I wouldn't like to see that for keeping these new default colors, users create maps in which roads can be confused with canals/rivers, for example.

I prefer to risk that, than the gray and black as the defaults, but I don't think that's likely to happen and certainly there is a lot of users who create streams with GRASS, so one could make an opposite claim too. :smile:

wenzeslaus:
Merged with blues. If you want to change the colors or pursue a more advanced coloring system, please open a new issue or PR.

PR #1170, opened 2020-12-11
g.proj: fix segfault issue #1163

metzm:
Affects only master, no backport needed


petrasovaa (approved):
Tested with both raster and vector and with valgrind, works fine, thanks!

PR #1173, opened 2020-12-13
wxGUI: Remove user mapset from demolocation

lindakladivova:
Now, we have had two mapsets in demolocation - PERMANENT and mapset named after a user (current). Now the current mapset will be PERMANENT and user mapset will be removed. This solution will be clearer for newcomers.


wenzeslaus (approved):
The `_create_startup_mapset()` may be still needed once we decide how to handled in-use/deleted mapset, but I'm for merging the PR as is. Works as expected.

PR #1177, opened 2020-12-15
g.extension: allow to install extensions from a fork of the official repo

ninsbl:
This PR adds an o-flag to specify that the `url` points to a fork of the official repo. In combination with the `branch` option, this allows easy testing of changes in addons by others...

See #625

### DIFF
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -128,6 +128,10 @@
 #% suppress_required: yes
 #%end
 #%flag
+#% key: o
+#% description: url refers to a fork of the official extension repository
+#%end
+#%flag
 #% key: j
 #% description: Generates JSON file containing the download URLs of the official Addons
 #% guisection: Install
@@ -139,6 +143,8 @@
 #% required: extension, -l, -c, -g, -a, -j
 #% exclusive: extension, -l, -c, -g
 #% exclusive: extension, -l, -c, -a
+#% requires: -o, url
+#% requires: branch, url
 #%end
 
 # TODO: solve addon-extension(-module) confusion
@@ -1451,6 +1457,10 @@ def download_source_code(source, url, name, outdev,
         gscript.message(_("Fetching <%s> from "
                           "GRASS GIS Addons repository (be patient)...") % name)
         download_source_code_official_github(url, name, outdev, directory)
+    elif source == 'official_fork':
+        gscript.message(_("Fetching <{name}> from "
+                          "<{url}> (be patient)...").format(name=name, url=url))
+        download_source_code_official_github(url, name, outdev, directory)
     elif source == 'svn':
         gscript.message(_("Fetching <{name}> from "
                           "<{url}> (be patient)...").format(name=name, url=url))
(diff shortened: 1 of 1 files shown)


scripts/g.extension/g.extension.py
```diff
@@ -139,6 +143,8 @@
 #% required: extension, -l, -c, -g, -a, -j
 #% exclusive: extension, -l, -c, -g
 #% exclusive: extension, -l, -c, -a
+#% requires: -o, url
+#% requires: branch, url
```
tmszi, line 147: It is not necessary to define the relationship between the `url` and the `branch` option. Branch option has default answer `main`.

tmszi (changes requested):
Great work, I tested it and it works fine.

tmszi (approved):

ninsbl:
Thanks @tmszi for reviewing.

> It is not necessary to define the relationship between the `url` and the `branch` option. Branch option has default answer `main`.

Yes. That is right. However, as it is right now, the `branch` option will be ignored if no `url` is given...
The official addon repository currently does not really have branches anyway (in contrast to feature branches in forks).

In the long run and for GRASS 8 we should probably aim at throwing out SVN and use the github/gitlab API to fetch addons (esp. as we already started using it for branches)...

Something along this lines:
```
curl https://api.github.com/repos/ninsbl/grass-addons/contents/grass7/imagery/i.sentinel/i.sentinel.import
```

This results in the following response (with a download URL for every file (folders have null)):
```
[
  {
    "name": "Makefile",
    "path": "grass7/imagery/i.sentinel/i.sentinel.import/Makefile",
    "sha": "50dfe89f5ecbd917b2e2d9019cf42bc55a4b23a0",
    "size": 115,
    "url": "https://api.github.com/repos/ninsbl/grass-addons/contents/grass7/imagery/i.sentinel/i.sentinel.import/Makefile?ref=master",
    "html_url": "https://github.com/ninsbl/grass-addons/blob/master/grass7/imagery/i.sentinel/i.sentinel.import/Makefile",
    "git_url": "https://api.github.com/repos/ninsbl/grass-addons/git/blobs/50dfe89f5ecbd917b2e2d9019cf42bc55a4b23a0",
    "download_url": "https://raw.githubusercontent.com/ninsbl/grass-addons/master/grass7/imagery/i.sentinel/i.sentinel.import/Makefile",
    "type": "file",
    "_links": {
      "self": "https://api.github.com/repos/ninsbl/grass-addons/contents/grass7/imagery/i.sentinel/i.sentinel.import/Makefile?ref=master",
      "git": "https://api.github.com/repos/ninsbl/grass-addons/git/blobs/50dfe89f5ecbd917b2e2d9019cf42bc55a4b23a0",
      "html": "https://github.com/ninsbl/grass-addons/blob/master/grass7/imagery/i.sentinel/i.sentinel.import/Makefile"
    }
  },
  {
    "name": "i.sentinel.import.html",
    "path": "grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html",
    "sha": "e0e3826fb48c721523643144608799fb575e4f6d",
    "size": 8244,
    "url": "https://api.github.com/repos/ninsbl/grass-addons/contents/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html?ref=master",
    "html_url": "https://github.com/ninsbl/grass-addons/blob/master/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html",
    "git_url": "https://api.github.com/repos/ninsbl/grass-addons/git/blobs/e0e3826fb48c721523643144608799fb575e4f6d",
    "download_url": "https://raw.githubusercontent.com/ninsbl/grass-addons/master/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html",
    "type": "file",
    "_links": {
      "self": "https://api.github.com/repos/ninsbl/grass-addons/contents/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html?ref=master",
      "git": "https://api.github.com/repos/ninsbl/grass-addons/git/blobs/e0e3826fb48c721523643144608799fb575e4f6d",
      "html": "https://github.com/ninsbl/grass-addons/blob/master/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html"
    }
  },

```

But that is a different PR and more work...

tmszi:
> Thanks @tmszi for reviewing.

You are welcome.

> > It is not necessary to define the relationship between the `url` and the `branch` option. Branch option has default answer `main`.
> 
> Yes. That is right. However, as it is right now, the `branch` option will be ignored if no `url` is given...
> The official addon repository currently does not really have branches anyway (in contrast to feature branches in forks).

Yes, you're right.

> In the long run and for GRASS 8 we should probably aim at throwing out SVN and use the github/gitlab API to fetch addons (esp. as we already started using it for branches)...
> 

We can use it, but the problem is that you need to download each file separately, number of files == number of requests/responses. It would be better if GitHub API had one REST endpoint to download any repository directory as an archive (zip or another format) with one request/response. I haven't found that yet. My opinion is that it will therefore be better to continue to use `svn export`.

PR #1182, opened 2020-12-17
Stop ignoring .git folder for docker builds

mmacata:
Currently the `.git` folder is ignored during docker builds. This masks the revision in the GRASS GIS version:
```
/src/actinia_core # grass --tmp-location epsg:4326 --exec g.version -rge
Starting GRASS GIS...
Creating new GRASS GIS location <tmploc>...
Cleaning up temporary files...
Executing <g.version -rge> ...
version=7.9.dev
date=2020
revision=exported
build_date=2020-12-16
build_platform=x86_64-pc-linux-musl
build_off_t_size=8
libgis_revision=2020-12-16T00:36:32+00:00
libgis_date=2020-12-16T00:36:32+00:00
proj=7.0.1
gdal=3.1.4
geos=3.8.1
sqlite=3.32.1
Execution of <g.version -rge> finished.
Cleaning up temporary files...
```
Looking for the cause, it behaves like this when no `.git` folder is present: https://github.com/OSGeo/grass/blob/master/configure.in#L149. Therefore this PR suggests to no longer ignore the `.git` folder during docker builds. Local test was successful.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,7 +6,7 @@ docker
 .travis.yml
 
 # The following git files are needed by GRASS GIS to extract the revision
-# during compilation. If you are not using one of the Dockerimgages from this
+# during compilation. If you are not using one of the Dockerimages from this
 # repository, delete the .git folder in your Dockerfile after compilation.
 .git
 !.git/HEAD
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,7 +2,7 @@ docker
 !docker/testdata
 !docker/alpine/alpine-py38-ctypes.patch
 !docker/alpine/alpine-py38-gisinit.patch
-.git
+!.git
 .gitignore
 .github
 .travis
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,9 +1,15 @@
 docker
 !docker/testdata
-!docker/alpine/alpine-py38-ctypes.patch
-!docker/alpine/alpine-py38-gisinit.patch
-!.git
 .gitignore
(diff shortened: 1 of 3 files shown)


mmacata:
After some objections the PR was reworked a bit. Now the `.git` folder is removed in Ubuntu and Debian Dockerimages after compilation. The alpine Dockerimage uses multistage build so the `.git` folder is not included in the final image.
The `.dockerignore` file now keeps ignoring the `.git` folder as a whole for docker builds but explicitly requests necessary files. The `.git` directory tree which is copied in the Docker now looks like this (depending on local branches):
```
.git
├── HEAD
├── objects
└── refs
    └── heads
        ├── docker-g-version
        ├── master
        └── releasebranch_7_8
```
And there is no sensitive information contained:
```
/src/grass_build # for i in `tree -fi .git`;do if [ -f $i ];then echo -n "$i: ";cat $i;fi;done
.git/HEAD: ref: refs/heads/docker-g-version
.git/refs/heads/docker-g-version: 79616a34b70e59cf472226e453016f9c279d21a6
.git/refs/heads/master: 8596c7d6a446a8ac1bf7a176dc71b67c15f2cc2d
.git/refs/heads/releasebranch_7_8: dfec2b62dbdbc0dab85c6abc494899c1e7b350dd
```
(`objects` was cleaned with `.dockerignore` file but needs to exist for `git rev-parse --short HEAD` to succeed.)

neteler (approved):

PR #1183, opened 2020-12-17
wxGUI: displaying infobar focused on first-time user (part 2)

lindakladivova:
After creating a new location when a user is in demolocation, second Info Bar appears offering a user to import data using Import vector data or Import raster data buttons.

A the same time, when creating a new Location, GRASS switches to the Location that a user created to its PERMANENT mapset.

### DIFF
--- a/gui/wxpython/datacatalog/catalog.py
+++ b/gui/wxpython/datacatalog/catalog.py
@@ -57,6 +57,8 @@ class DataCatalog(wx.Panel):
         # infobar manager for data catalog
         self.infoManager = DataCatalogInfoManager(infobar=self.infoBar,
                                                   giface=self.giface)
+        self.tree.showImportDataInfo.connect(self.showImportDataInfo())
+
         # some layout
         self._layout()
 
@@ -125,10 +127,7 @@ class DataCatalog(wx.Panel):
     def OnCreateLocation(self, event):
         """Create new location"""
         db_node, loc_node, mapset_node = self.tree.GetCurrentDbLocationMapsetNode()
-        location_created, is_user_in_demolocation = self.tree.CreateLocation(db_node)
-        if location_created and is_user_in_demolocation:
-            # show data import infobar for first-time user with proper layout
-            wx.CallLater(2000, self.showImportDataInfo)
+        self.tree.CreateLocation(db_node)
 
     def OnDownloadLocation(self, event):
         """Download location to current grass database"""
--- a/gui/wxpython/datacatalog/catalog.py
+++ b/gui/wxpython/datacatalog/catalog.py
@@ -80,6 +80,9 @@ class DataCatalog(wx.Panel):
     def showDataStructureInfo(self):
         self.infoManager.ShowDataStructureInfo(self.OnCreateLocation)
 
+    def showImportDataInfo(self):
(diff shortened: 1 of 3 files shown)


lindakladivova:
Info Bar 2:
infobar2

Import vector data dialog:
infobar2_1

GRASS after vector data import:
infobar_3

gui/wxpython/datacatalog/tree.py
```diff
@@ -994,10 +1002,15 @@ def CreateLocation(self, grassdb_node):
...
+            self.SwitchMapset(grassdatabase, location, mapset)
+            location_created = True
+        return (location_created, is_user_in_demolocation)
 
     def OnCreateLocation(self, event):
         """Create new location"""
-        self.CreateLocation(self.selected_grassdb[0])
+        location_created, is_user_in_demolocation = (
+            self.CreateLocation(self.selected_grassdb[0])
+        )
```
wenzeslaus, line 1013: There is no need to assign the variables if you are not using them, but more importantly, this should be solved in a different way as context menu on the database node does not trigger the info bar. It also makes the two functions in tree and catalog diverge too much (two different non-trivial functions which should be the same and are not).

What do you think @petrasovaa @landam, leave for another PR or resolve now?

wenzeslaus (changes requested):

gui/wxpython/datacatalog/infomanager.py
```diff
@@ -43,5 +43,19 @@ def ShowDataStructureInfo(self, onCreateLocationHandler):
...
         self.infoBar.ShowMessage(message, wx.ICON_INFORMATION, buttons)
 
+    def ShowImportDataInfo(self, OnImportOgrLayersHandler, OnImportGdalLayersHandler):
+        """Show info about the data import focused on the first-time user"""
+        buttons = [("Import vector data", OnImportOgrLayersHandler),
+                   ("Import raster data", OnImportGdalLayersHandler)]
+        message = _(
+            "You have successfully created a new Location {loc}. "
+            "Currently you are in its PERMANENT Mapset which is used for "
+            "storing your base maps to make it readily avaliable in other "
```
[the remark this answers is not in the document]
lindakladivova: Solved.

gui/wxpython/datacatalog/catalog.py
```diff
@@ -57,6 +57,8 @@ def __init__(self, parent, giface=None, id=wx.ID_ANY,
         # infobar manager for data catalog
         self.infoManager = DataCatalogInfoManager(infobar=self.infoBar,
                                                   giface=self.giface)
+        self.tree.showImportDataInfo.connect(self.showImportDataInfo())
```
wenzeslaus, line 60: You are calling the `self.showImportDataInfo()` function here, you just need to pass it to connect just like with `ShowImportDataInfo` or at line 52.

wenzeslaus (changes requested):
I like these changes much better than the previous ones. The code is more straightforward and all decisions are kept in tree where is the main functionality now.

gui/wxpython/datacatalog/catalog.py
lindakladivova: Has been corrected, now works as expected.

wenzeslaus (approved):

PR #1204, opened 2020-12-22
wxGUI: New management icons for vector and raster data import

lindakladivova:
In order to improve the second Info Bar for data import, management icons were created.
icons

The third Info Bar focusing on the explanation of the Modules and Display tabs will respond to the import using these management icons.


gui/wxpython/datacatalog/catalog.py
```diff
@@ -110,20 +110,26 @@ def OnCreateMapset(self, event):
         db_node, loc_node, mapset_node = self.tree.GetCurrentDbLocationMapsetNode()
         self.tree.CreateMapset(db_node, loc_node)
 
-    def OnImportOgrLayers(self, event, cmd=None):
+    def ImportOgrLayers(self, cmd=None):
```
landam, line 113: `cmd` argument is not used

gui/wxpython/datacatalog/catalog.py
```diff
@@ -110,20 +110,26 @@ def OnCreateMapset(self, event):
...
         from modules.import_export import OgrImportDialog
         dlg = OgrImportDialog(parent=self, giface=self.giface)
         dlg.CentreOnScreen()
         dlg.Show()
 
-    def OnImportGdalLayers(self, event, cmd=None):
+    def OnImportOgrLayers(self, event):
+        self.ImportOgrLayers()
+
+    def ImportGdalLayers(self, cmd=None):
```
landam, line 123: `cmd` argument is not used

gui/wxpython/datacatalog/catalog.py
```diff
@@ -110,20 +110,26 @@ def OnCreateMapset(self, event):
         db_node, loc_node, mapset_node = self.tree.GetCurrentDbLocationMapsetNode()
         self.tree.CreateMapset(db_node, loc_node)
 
-    def OnImportOgrLayers(self, event, cmd=None):
```
landam, line 113: Do we need `Import*Layers()`? What about:

```
    def OnImportOgrLayers(self, event):
        from modules.import_export import OgrImportDialog
        dlg = OgrImportDialog(parent=self, giface=self.giface)
        dlg.CentreOnScreen()
        dlg.Show()
```

landam (changes requested):

gui/wxpython/datacatalog/catalog.py
lindakladivova: If we called it later in frame.py it would make sense to have only ImportOgrLayers however as we decided to not connect Info Bars to File menu events, we do not need it...

gui/wxpython/datacatalog/catalog.py
lindakladivova: solved

gui/wxpython/datacatalog/catalog.py
lindakladivova: solved

landam (approved):

petrasovaa (changes requested):
We talked about this in the last call with @lindakladivova, the import icon from LM toolbar needs to be moved here and the options changed appropriately (have raster, vector icons and the icon for other options).

lindakladivova:
These commits are only for information, work is still in progress, now I am working on onMenuCmdEmit which does not work yet.

lindakladivova:
The situation looks as follows:
Výstřižek

lindakladivova:
What do we prefer?

    def OnRasterOutputFormat(self, event):
        """Set raster output format handler"""
        from modules.import_export import GdalOutputDialog
        dlg = GdalOutputDialog(parent=self, ogr=False)
        dlg.CentreOnScreen()
        dlg.Show()

Or in lmgr/frame.py, there is:

    def OnRasterOutputFormat(self, event):
        """Set raster output format handler"""
        self.OnMenuCmd(cmd=['r.external.out'])

veroandreo:
The commands next to unpack raster map and unpack vector map are switched

lindakladivova:
> 
> 
> The commands next to unpack raster map and unpack vector map are switched

Thanks Vero! Corrected. :-)

lindakladivova:
Corrected version: 

Výstřižek

petrasovaa:
Let's change the raster/vector ASCII points import labels to something more understandable.

lindakladivova:
Ok, how about this?
Výstřižek

wenzeslaus:
@lindakladivova I don't think these are more understandable nor that they are expressing what these do.

Everybody, any suggestions on how to succinctly refer to v.unpack and r.unpack? Packing seems unclear to me, it could refer to compressing. It is an import of files in GRASS custom/own/native exchange format, isn't it? Perhaps we need a name for the format itself (and an extension), not just _packed map_, _unpack packed map_, etc. Or is it clear that packing is export as opposed to, e.g., compression?

wenzeslaus:
What do you think about the order @lindakladivova? Did you order them by how often you think these will be used? I would put Link at the very bottom (before the More/Doc item we talked about) because it is kind of the odd one here.

lindakladivova:
Yes, makes sense. I also changed the labels. I think ASCII imports are now better explained. 
Výstřižek

And what about adding r.in.gdal and v.in.ogr? They are also widely used, I think.

petrasovaa:
> @wenzeslaus Yes, makes sense. I also changed the labels. I think ASCII imports are now better explained. Other options refer to https://grass.osgeo.org/grass79/manuals/topic_import.html.

I would simplify it even more:
* Create raster map from x,y,z data
* Create vector map from x,y,z data
> 
> And what about adding r.in.gdal and v.in.ogr? They are also widely used, I think.

r.import and v.import (the icons you have there) are wrappers for r.in.gdal and v.in.ogr

lindakladivova:
Anicka's suggestion was taken into consideration:
Výstřižek

gui/wxpython/lmgr/frame.py
```diff
@@ -1302,20 +1302,6 @@ def _popupMenu(self, data):
...
-                         ('vectImport', self.OnImportOgrLayers),
-                         (None, None),
-                         ('rastUnpack', self.OnUnpackRaster),
-                         ('vectUnpack', self.OnUnpackVector),
-                         (None, None),
-                         ('rastLink', self.OnLinkGdalLayers),
-                         ('vectLink', self.OnLinkOgrLayers),
-                         ('rastOut', self.OnRasterOutputFormat),
-                         ('vectOut', self.OnVectorOutputFormat)))
-
```
petrasovaa, line 1318: Some of these methods are not used anymore so they should be removed.

gui/wxpython/datacatalog/catalog.py
```diff
@@ -110,33 +113,141 @@ def OnCreateMapset(self, event):
...
 
     def Filter(self, text, element=None):
         self.tree.Filter(text=text, element=element)
+
+    def OnImportMenu(self, event):
+        """Import maps menu (import, link)
+        """
+        self._popupImportMenu()
+
+    def _popupImportMenu(self):
```
petrasovaa, line 199: If _popupImportMenu is not used elsewhere, why to have it as a separate method, I would just merge those 2 methods.

gui/wxpython/datacatalog/catalog.py
```diff
@@ -110,33 +113,141 @@ def OnCreateMapset(self, event):
...
+        """Create popup menu for other import options"""
+
+        # create submenu
+        subMenu = Menu()
+
+        subitem = wx.MenuItem(subMenu, wx.ID_ANY, _("&Link external raster data  [r.external]"))
```
petrasovaa, line 205: Remove all the & from the labels, they are not very useful here and they anyway shouldn't use the same characters.

gui/wxpython/datacatalog/catalog.py
```diff
@@ -110,33 +113,141 @@ def OnCreateMapset(self, event):
...
+        dlg.CentreOnScreen()
+        dlg.Show()
+
+    def OnImportRasterAscii(self, event):
+        """Create raster map from x,y,z data handler"""
+        GUI(parent=self, giface=self.giface).ParseCommand(cmd=['r.in.xyz'])
+
+    def OnImportVectorAscii(self, event):
+        """Create vector map from x,y,z data handler"""
+        GUI(parent=self, giface=self.giface).ParseCommand(cmd=['v.in.ascii'])
```
petrasovaa, line 182: I would make a generic handler for these and the r/v.pack ones and use lambda function in Bind().

petrasovaa (changes requested):
Looks good, just some small changes.

gui/wxpython/lmgr/frame.py
lindakladivova: They are used in RunSpecialCmd function in lgmr/frame.py. Not sure if they can be removed from "elif" parts in RunSpecialCmd.

gui/wxpython/datacatalog/catalog.py
lindakladivova: Merged.

gui/wxpython/datacatalog/catalog.py
lindakladivova: Removed.

gui/wxpython/datacatalog/catalog.py
lindakladivova: Good point. I made the generic function GuiParseCommand which is then used in Bind() for all four menu items.

gui/wxpython/lmgr/frame.py
petrasovaa: Not the unpack ones.

petrasovaa (changes requested):
Just remove the 2 handlers that are not used and this should be good to go.

petrasovaa (approved):

PR #1205, opened 2020-12-22
wxGUI: displaying infobar focused on first-time user (part 3)

lindakladivova:
Implementation of a third Info Bar for the first-time user which is focused on Display and Modules tabs. 
After success import using _Import vector data_ or _Import raster data_ button, this message appears:
third infobar

A user can click on the button _See documentation_ and then is redirected to https://grass.osgeo.org/grass79/manuals/index.html


gui/wxpython/datacatalog/catalog.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, parent, giface=None, id=wx.ID_ANY,
         self.parent = parent
         self.baseTitle = title
         self.giface = giface
+        self.first_time_mode = False
```
wenzeslaus, line 42: This should be `show_info_after_import`. I think it best expresses what it does and how it is used. The new name is explicit without introducing any new concepts.

gui/wxpython/modules/import_export.py
```diff
@@ -407,6 +408,7 @@ def __init__(self, parent, giface, link=False):
         self.link = link
 
         self.layersData = []
+        self.showModulesDisplayInfo = Signal('GdalImportDialog.showModulesDisplayInfo')
```
wenzeslaus, line 411: This signal name should primarily express what happened. This class does not know and does not need to know what happens when the signal is emitted. Name like `importSuccessful` is a better fit.

gui/wxpython/modules/import_export.py
```diff
@@ -559,6 +563,8 @@ def __init__(self, parent, giface, link=False):
 
         self.layersData = []
```
wenzeslaus, line 565: No need for an empty line here.

gui/wxpython/modules/import_export.py
```diff
@@ -559,6 +563,8 @@ def __init__(self, parent, giface, link=False):
...
 
+        self.showModulesDisplayInfo = Signal('OgrImportDialog.showModulesDisplayInfo')
```
wenzeslaus, line 566: It looks like both import dialogs have the same signal and both already inherit from the same parent class. It seems that the signal should be part of the interface of the parent class, i.e., any import dialog. There seems to be some code duplication in OnCmdDone, but the signal can be emitted in the child classes, that's fine.

(No harm done if the DXF dialog has that part not implemented.)

gui/wxpython/datacatalog/infomanager.py
```diff
@@ -57,5 +57,18 @@ def ShowImportDataInfo(self, OnImportOgrLayersHandler, OnImportGdalLayersHandler
         ).format(loc=gisenv()['LOCATION_NAME'])
         self.infoBar.ShowMessage(message, wx.ICON_INFORMATION, buttons)
 
+    def ShowModulesDisplayInfo(self):
+        """Show info about the Display and Modules tabs focused on the first-time user"""
+        buttons = [("See documentation", self._onSeeDocumentation)]
+        message = _(
+            "You have successfully imported data. To change layer properties, "
+            "go to Display tab below and double click over layer name.\n\n"
```
wenzeslaus, line 65: Maybe the double click is not needed here.

wenzeslaus (changes requested):
This looks pretty good.

wenzeslaus (changes requested):
When the import runs, it switches to Console tab, so the user won't actually see the info bar. Either the info bar needs to be placed elsewhere (not in the Data tab) or it should not switch to the Console tab. The latter might be simpler to do, but the idea behind that behavior, I think, is that import is important process and thus the user should know about the details or be informed that there is a place to get these details. Some other ideas @landam ?

gui/wxpython/modules/import_export.py
lindakladivova: renamed to showImportSuccessfulInfo

gui/wxpython/modules/import_export.py
lindakladivova: Signal is now created in the ImportDialog.

lindakladivova:
It would be probably needed to make another instance of info bar in the Console tab. But I would let it for now as it is - switch to info bar which is in the Data tab. It would switch only in the case of successful import. What do you think? We can change the text: "  "You have successfully imported data. **The process of data import is described in the Console tab below.** \n\n To change layer properties, go to **the** Display tab and click over **the** layer name. \n\n To analyze your data, search for a module (tool) in the Modules tab below."

gui/wxpython/datacatalog/catalog.py
```diff
@@ -92,6 +92,9 @@ def showImportSuccessfulInfo(self):
             self.infoManager.ShowImportSuccessfulInfo()
             self.show_info_after_import = False
 
+    def switchToDataTab(self):
+        wx.CallAfter(self.parent.SetSelectionByName, 'catalog')
```
wenzeslaus, line 96: Looks like candidate for a lambda function.

gui/wxpython/modules/import_export.py
```diff
@@ -60,7 +60,8 @@ def __init__(self, parent, giface, itype,
         self.options = dict()   # list of options
         self.options_par = dict()
 
-        self.showImportSuccessfulInfo = Signal('GdalImportDialog.showImportSuccessfulInfo')
+        self.switchToDataTab = Signal('ImportDialog.switchToDataTab')
+        self.showImportSuccessfulInfo = Signal('ImportDialog.showImportSuccessfulInfo')
```
wenzeslaus: What about one Signal for both? They seem to be pretty much the same. The point of the signals is that the class which emits them doesn't know anything about the actual usage. Here it seems really just as calling specific functions, but doing it through the signals.

wenzeslaus (changes requested):

gui/wxpython/modules/import_export.py
lindakladivova: Makes sense, but the behaviour should be consistent. We need to decide whether we want to show the third infobar and switch to Data tab when calling specific functions from File menu. Or do not want to change lgmr.py at all and focus only on calling using buttons in infobar and Management icons. 

This PR deals with the third Infobar, so probably the best way would be to not solve File tab here at all. So, the switch to Data tab and show of infobar would be conditioned by calling the functions either from infobar buttons, or from Management icons for import (https://github.com/OSGeo/grass/pull/1204 (still open)).

lindakladivova: This waits for #1205.

gui/wxpython/datacatalog/infomanager.py
```diff
@@ -57,5 +57,18 @@ def ShowImportDataInfo(self, OnImportOgrLayersHandler, OnImportGdalLayersHandler
         ).format(loc=gisenv()['LOCATION_NAME'])
         self.infoBar.ShowMessage(message, wx.ICON_INFORMATION, buttons)
 
+    def ShowImportSuccessfulInfo(self):
+        """Show info about the Display and Modules tabs focused on the first-time user"""
+        buttons = [("See documentation", self._onSeeDocumentation)]
+        message = _(
+            "You have successfully imported data. To change layer properties, "
+            "go to Display tab below and click over layer name.\n\n"
```
petrasovaa, line 65: either 'double-click on a layer' or 'right-click on a layer'

gui/wxpython/datacatalog/infomanager.py
```diff
@@ -57,5 +57,18 @@ def ShowImportDataInfo(self, OnImportOgrLayersHandler, OnImportGdalLayersHandler
         ).format(loc=gisenv()['LOCATION_NAME'])
         self.infoBar.ShowMessage(message, wx.ICON_INFORMATION, buttons)
 
+    def ShowImportSuccessfulInfo(self):
+        """Show info about the Display and Modules tabs focused on the first-time user"""
+        buttons = [("See documentation", self._onSeeDocumentation)]
```
petrasovaa, line 62: This button here seems unnecessary and not helpful, I would just remove it.

petrasovaa (changes requested):

gui/wxpython/datacatalog/catalog.py
```diff
@@ -88,6 +90,12 @@ def showDataStructureInfo(self):
     def showImportDataInfo(self):
         self.infoManager.ShowImportDataInfo(self.OnImportOgrLayers, self.OnImportGdalLayers)
 
+    def showImportSuccessfulInfo(self):
+        if self.show_info_after_import:
+            wx.CallAfter(self.parent.SetSelectionByName, 'catalog')
```
wenzeslaus, line 95: This is a bad dependency making data catalog dependent on lmgr. The dependency is and should be the other way around. lmgr depends on data catalog while `self.parent.SetSelectionByName, 'catalog'` makes assumption that the parent has a `SetSelectionByName` function which sets focus/shows things and that data catalog is called `catalog`.

How to achieve this is to have signal showMe in data catalog. lmgr can connect to it and do what is needed. This will make data catalog work smoothly in e.g. g.gui.datacatalog where this signal will simple not be connected to anything.

If CallAfter is really needed, the `emit` should be wrapped in it, i.e., it should stay here. It is the data catalog which needs to finish things first, so lmgr should not have the responsibility to do things right.

gui/wxpython/datacatalog/infomanager.py
wenzeslaus, line 62: The reason for the button is unclear to me. The modules are in the Modules tab, so there is no need for the button if that's what it is drawing attention to.

wenzeslaus (changes requested):

gui/wxpython/datacatalog/infomanager.py
lindakladivova: I have changed it to 'right-click on a layer name'

gui/wxpython/datacatalog/infomanager.py
lindakladivova: Agree, removed

gui/wxpython/datacatalog/catalog.py
lindakladivova: I get it. I have created signal showTab, and I no longer use CallAfter - seems useless.

lindakladivova:
Now, the situation is as follows: Výstřižek
I think it is ready for merge.

petrasovaa (approved):

gui/wxpython/datacatalog/catalog.py
```diff
@@ -39,9 +39,11 @@ def __init__(self, parent, giface=None, id=wx.ID_ANY,
                  title=_("Data catalog"), name='catalog', **kwargs):
         """Panel constructor  """
         self.showNotification = Signal('DataCatalog.showNotification')
+        self.showPage = Signal('DataCatalog.showPage')
```
wenzeslaus, line 42: The signals serve us in two ways, they remove direct code dependence and also remove conceptual dependence. This one removes the code dependence, but does not address the conceptual dependence. The show notification signal above just says, "Hi there, I have a some sort of notification, so dear class user, if you think the end user wants to see them, here is one you may want to show now." The show page signal, on the other hand, says "Hi there, switch the page if you can, I have no idea what it is, but try to switch it anyway."

In other words, the signal (name, parameters) should be relevant to the object emitting the signals. `DataCatalog.showNotification` says to show notification generated by the data catalog. `DataCatalog.showPage` makes assumptions about surrounding code and imports this assumption into its API resulting in an some _page_ being part of data catalog.

One test of this is the standalone data catalog (g.gui...). Obviously, I don't have any page there to show and ignoring a signal is fine (at least unless otherwise noted). I can choose to ignore even the notification signal, but there it is clear what I'm ignoring. With the show page signal, it is not clear what event I'm ignoring and what should I do in case I want to react to what happened in the data catalog.

wenzeslaus (changes requested):

lindakladivova:
@wenzeslaus @petrasovaa  In the Single-Window, the Layers pane as well as the Modules pane are visible at once. I think that in that case the suggested info message is no longer needed.  What do you think?

petrasovaa:
> @wenzeslaus @petrasovaa In the Single-Window, the Layers pane as well as the Modules pane are visible at once. I think that in that case the suggested info message is no longer needed. What do you think?

I agree it's probably not needed now.

lindakladivova:
Ok, I am closing this PR for now.

PR #1216, opened 2020-12-29
wxGUI: New, simplified, unified window titles

wenzeslaus:
The 'lmgr' window is considered the main one, so it is no longer
called (GRASS GIS) Layer Manager, but just GRASS GIS. This fits well with this
window being the main GRASS GIS component for many users, although it is not
technically completely precise.

The lmgr window title previously included only workspace name after its base title.
Now it includes workspace name (if applicable), followed by
mapset/location (previously in map display title) and that is followed by the base title,
separated by dashes. This follows common practice of document before application name,
or from most specific information to most general. This also brings the titles closer
to what a single window interface would use. It also may improve the situation reported by a beginner:
*Once I open the GRASS GIS console...it opens another application called Layer Manager*, reported in
[Trac Ticket 3474](https://trac.osgeo.org/grass/ticket/3474).

The version number of is no longer included in the title. This is again a common practice
and it makes making release screenshots easier. Bug reports need a more complete info anyway.
Users which have multiple installations are likely advanced anyway and can always check if needed.

Map display does not include mapset/location at the end as before,
nor GRASS GIS as prefix. Titles of tabs in the Display tab are fully synchronized
with the titles regardless of the title being the default or custom,
so the tabs now say Map Display, not just Display. This simplifies the code
significantly and clarifies the interface. (Small change to rename item
in menu is included for further clarification.)

Database manager which includes vector name in the title follows the same practice
as the main window lmgr, i.e., document before tool/app name.
The read-only info is included with more common spelling and standard case
(applied in the table tab title).

Dialogs started from main GUI don't show GRASS GIS as first part of the title.
This basic title is usually set as default in the frame.
The corresponding g.gui.* tools (if any) set their own title using
a tool - program name syntax and are considered to be somewhat standalone,
so they include GRASS GIS after a dash.

The GUI forms/dialogs for modules no longer include keywords in (square) brackets.
Module name is therefore more highlighted as the actual tool name.
Keywords are not meant to be used this way. (A useful title might be menu item
label, but that's not readily available.)

wxgui.py now sets the application name and vendor which improves how the program
is titled on some platforms. (Making no distinction between internal and display
name because there is no current use case for that.)

Screenshot from 2020-12-29 00-37-08

### DIFF
--- a/gui/wxpython/psmap/g.gui.psmap.py
+++ b/gui/wxpython/psmap/g.gui.psmap.py
@@ -46,7 +46,10 @@ def main():
     from psmap.frame import PsMapFrame
 
     app = wx.App()
-    frame = PsMapFrame(parent=None)
+    frame = PsMapFrame(
+        parent=None,
+        title=_("Cartographic Composer - GRASS GIS"),
+    )
     frame.Show()
 
     if options['file']:
--- a/gui/wxpython/rlisetup/g.gui.rlisetup.py
+++ b/gui/wxpython/rlisetup/g.gui.rlisetup.py
@@ -41,7 +41,11 @@ def main():
     from rlisetup.frame import RLiSetupFrame
 
     app = wx.App()
-    frame = RLiSetupFrame(parent=None, giface=StandaloneGrassInterface())
+    frame = RLiSetupFrame(
+        parent=None,
+        giface=StandaloneGrassInterface(),
+        title=_("Setup for r.li modules - GRASS GIS"),
+    )
     frame.Show()
     frame.CenterOnScreen()
 
--- a/gui/wxpython/timeline/g.gui.timeline.py
+++ b/gui/wxpython/timeline/g.gui.timeline.py
@@ -58,7 +58,10 @@ def main():
     view3d = flags['3']
 
     app = wx.App()
-    frame = TimelineFrame(None)
+    frame = TimelineFrame(
+        parent=None,
+        title=_("Timeline Tool - GRASS GIS"),
+    )
     frame.SetDatasets(datasets)
     frame.Show3D(view3d)
     frame.Show()
--- a/gui/wxpython/tplot/g.gui.tplot.py
+++ b/gui/wxpython/tplot/g.gui.tplot.py
@@ -162,7 +162,11 @@ def main():
     if options['csv']:
         csvfile = options['csv']
     app = wx.App()
-    frame = TplotFrame(parent=None, giface=StandaloneGrassInterface())
+    frame = TplotFrame(
+        parent=None,
+        giface=StandaloneGrassInterface(),
+        title=_("Temporal Plot Tool - GRASS GIS"),
+    )
     if flags['l']:
         frame.linRegRaster.SetValue(state=True)
         frame.linRegVector.SetValue(state=True)
--- a/gui/wxpython/vdigit/g.gui.vdigit.py
+++ b/gui/wxpython/vdigit/g.gui.vdigit.py
@@ -64,7 +64,7 @@ def main():
         def __init__(self, vectorMap):
             MapFrame.__init__(
                 self, parent=None, Map=Map(), giface=DMonGrassInterface(None),
-                title=_("GRASS GIS Vector Digitizer"), size=(850, 600))
+                title=_("Vector Digitizer - GRASS GIS"), size=(850, 600))
             # this giface issue not solved yet, we must set mapframe aferwards
             self._giface._mapframe = self
             # load vector map
--- a/gui/wxpython/animation/frame.py
+++ b/gui/wxpython/animation/frame.py
@@ -48,7 +48,7 @@ gcore.set_raise_on_error(True)
 
 class AnimationFrame(wx.Frame):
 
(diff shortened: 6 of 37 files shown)


gui/wxpython/psmap/g.gui.psmap.py
```diff
@@ -46,7 +46,10 @@ def main():
     from psmap.frame import PsMapFrame
 
     app = wx.App()
-    frame = PsMapFrame(parent=None)
+    frame = PsMapFrame(
+        parent=None,
+        title=_("Cartographic Composer - GRASS GIS"),
```
veroandreo, line 51: Why adding GRASS GIS here? If it follows the same logic than the others, shouldn't it be `Cartographic Composer` only?

gui/wxpython/rlisetup/g.gui.rlisetup.py
```diff
@@ -41,7 +41,11 @@ def main():
     from rlisetup.frame import RLiSetupFrame
 
     app = wx.App()
-    frame = RLiSetupFrame(parent=None, giface=StandaloneGrassInterface())
+    frame = RLiSetupFrame(
+        parent=None,
+        giface=StandaloneGrassInterface(),
+        title=_("Setup for r.li modules - GRASS GIS"),
```
veroandreo, line 47: idem as before

gui/wxpython/timeline/g.gui.timeline.py
```diff
@@ -58,7 +58,10 @@ def main():
     view3d = flags['3']
 
     app = wx.App()
-    frame = TimelineFrame(None)
+    frame = TimelineFrame(
+        parent=None,
+        title=_("Timeline Tool - GRASS GIS"),
```
veroandreo, line 63: idem

gui/wxpython/vdigit/g.gui.vdigit.py
```diff
@@ -64,7 +64,7 @@ class VDigitMapFrame(MapFrame):
         def __init__(self, vectorMap):
             MapFrame.__init__(
                 self, parent=None, Map=Map(), giface=DMonGrassInterface(None),
-                title=_("GRASS GIS Vector Digitizer"), size=(850, 600))
+                title=_("Vector Digitizer - GRASS GIS"), size=(850, 600))
```
veroandreo, line 67: idem

gui/wxpython/tplot/g.gui.tplot.py
```diff
@@ -162,7 +162,11 @@ def main():
     if options['csv']:
         csvfile = options['csv']
     app = wx.App()
-    frame = TplotFrame(parent=None, giface=StandaloneGrassInterface())
+    frame = TplotFrame(
+        parent=None,
+        giface=StandaloneGrassInterface(),
+        title=_("Temporal Plot Tool - GRASS GIS"),
```
veroandreo, line 168: idem

veroandreo (approved):
I like the changes proposed. Maybe I missed something, but there is GRASS GIS added now in some titles, those are my only comments

gui/wxpython/psmap/g.gui.psmap.py
wenzeslaus: These are the `g.gui.*` tools and the idea here is that they stand on its own, so they include GRASS GIS in their title unlike when the same component is started as part of the main GUI. This requires customizing the title (i.e., setting a separate title) in all `g.gui.*` (which is the line we are looking at), so it is more code than just keeping the same ones.

Unfortunately, this is not consistent with starting module GUI from command line, but that would be more work as there is no mechanism in place to set title from outside.

wenzeslaus:
Another thing which is out of scope for this PR, besides more advanced titles for module/tool forms/dialogs, is applying this to the image2target and photo2image directories (too much code duplication and unclear use for me to change the code).

neteler:
In case I run more than one GRASS sessions in parallel, will i still understand to which session the windows belong to?

veroandreo:
> In case I run more than one GRASS sessions in parallel, will i still understand to which session the windows belong to?

Ah, this is a good question. The layer manager will contain the location and mapset name so that should be fine, but what about the map displays?

wenzeslaus:
> > In case I run more than one GRASS sessions in parallel, will i still understand to which session the windows belong to?

The ultimate answer depends on how your system (window managers) handles applications and windows, but yes, if you were able to recognizance it before, you will be able to do it with this as well.

> The layer manager will contain the location and mapset name so that should be fine,

Right. Originally, it was (repeated) in (each) map display. Now it is in the main window (aka Layer Manager).

> but what about the map displays?

Now the map display does not contain location/mapset. It did not seem to me that this was a problem. If map display should be distinguished somehow, should also be all the other windows and dialogs (import, animation tool, histogram, ...)?

wenzeslaus:
I have included additional changes to sync how d.histogram is mentioned in the title with what you see in the menu or Modules tab. I also changed Histogramming Tool to Histogram Tool because sentences such as *What a great morning for some histogramming!* and *I would rather do some histogramming tonight.* are not as common as the previous title assumed.

PR #1237, opened 2021-01-09
parser: fixup more messages

infrastation:


wenzeslaus:
I think we use angle brackets for _values entered by the user_. A name of an option is an edge case, because you type it in the CLI, but not in GUI and it is not really a value in the first place. The second `%s` is neither entered by user nor it is a single value in the general case., So I would say the message is technically correct, although it does not look good. Maybe the rule is wrong? What do you think @neteler?

infrastation:
These changes do not block any of my work, but I thought it would be useful to make the messages consistent. Would it help to reduce these changes to only these that fix obvious issues?

neteler:
Sorry, I overlooked that I was "called in" here (I'll try to go through asap).

However, we have a dedicated "Message Standardization" page which might be helpful for orientation (and may also receive improvements as needed):
https://trac.osgeo.org/grass/wiki/MessageStandardization

HuidaeCho:
> 
> 
> These changes do not block any of my work, but I thought it would be useful to make the messages consistent. Would it help to reduce these changes to only these that fix obvious issues?

`describe_rule()` already adds `<>`, so your change in `parser_dependencies.c` would print leading and trailing *extra* `<>`.

The `r.info` related fixes look fine to me.

HuidaeCho:
> 
> 
> I think we use angle brackets for _values entered by the user_. A name of an option is an edge case, because you type it in the CLI, but not in GUI and it is not really a value in the first place. The second `%s` is neither entered by user nor it is a single value in the general case., So I would say the message is technically correct, although it does not look good. Maybe the rule is wrong? What do you think @neteler?

I don't remember we have discussed any rules about angle brackets if I'm not missing anything. _Values entered by the user_... I think it depends. Options and flags, even from GUI, are _user selected (GUI) or entered (CLI)_. Options are long enough, but flags definitely need some attention with extra `<>`, so I think the current practice looks good to me. This PR would add extra angle brackets. For example,
```
Option <-a> is mutually exclusive with all of <<-b>, <-c> or <-d>>
```

`describe_rule()` doesn't add `<>` for a single rule. That's the problem.

HuidaeCho:
#1399 for consistency: Add angle brackets to single-option/flag rules.

infrastation:
The commit that tried to fix the problem that belongs to `describe_rule()` is now gone. The commit that fixes `r.info` remains and has been rebased on the current master branch.

infrastation:
Rebased on the latest master branch.

infrastation:
Rebased on the latest master branch.

PR #1245, opened 2021-01-13
gis/parser: disable file existence check for stdout '-'

### LINKED ISSUE #1242: [Bug] Profile tool fails in GUI
**Describe the bug**
I just compiled and installed GRASS 7.8.5. When I tried the GUI profile tool, it fails with the error message:
"ERROR: option <output>: <-> exists. To overwrite, use the --overwrite flag. A blank profile window is generated. 

**To Reproduce**
Display "elevation" from the SC demo data set. 
Launch the profile tool
Make a rubber band line to create a profile

- Operating System: Mac OS 11.1
- GRASS GIS version 7.8.5

<!---

version=7.8.5
date=2020
revision=exported
build_date=2021-01-11
build_platform=x86_64-apple-darwin20.2.0
build_off_t_size=8
libgis_revision=2021-01-11T17:09:27+00:00
libgis_date=2021-01-11T17:09:27+00:00
proj=7.1.1
gdal=3.1.4
geos=3.8.1
sqlite=3.33.0

python3 -c "import sys, wx; print(sys.version); print(wx.version())"
3.8.6 | packaged by conda-forge | (default, Nov 27 2020, 19:17:44) 
[Clang 11.0.0 ]
4.0.7.post2 osx-cocoa (phoenix) wxWidgets 3.0.5
GRASS 7.8.5 (nc_spm_08_grass7):~ >

tmszi:
Fixes #1242.

For testing purpose:

1. Create file '-' in home dir `echo "test" > '-'`
2. Try launch wxGUI Profile tool or using r.profile, r.horizon, r.out.bin module

### DIFF
--- a/lib/gis/parser.c
+++ b/lib/gis/parser.c
@@ -1592,16 +1592,17 @@ int check_overwrite(void)
 		for (i = 0; opt->answers[i]; i++) {
 		    found = FALSE;
 		    if (strcmp(element, "file") == 0) {
-                        if (access(opt->answers[i], F_OK) == 0)
-                            found = TRUE;
-		    }
-                    else if (strcmp(element, "mapset") != 0) {
-                        /* TODO: also other elements should be
-                           probably skipped */
-                        if (G_find_file(element, opt->answers[i], G_mapset())) {
-                            found = TRUE;
-                        }
-		    }
+                if (strcmp(opt->answers[i], "-") != 0 &&
+                    access(opt->answers[i], F_OK) == 0)
+                    found = TRUE;
+            }
+            else if (strcmp(element, "mapset") != 0) {
+                /* TODO: also other elements should be
+                probably skipped */
+                if (G_find_file(element, opt->answers[i], G_mapset())) {
+                    found = TRUE;
+                }
+            }
                     
 		    if (found) {	/* found */
 			if (!st->overwrite && !over) {


neteler:
I made a test with the unpatched G78git: Strangely I do not get the "overwrite" issue:

```
grass78 ~/grassdata/nc_spm_08_grass7/user1/
...
Welcome to GRASS GIS 7.8.6dev (af07fc916)
...

GRASS :~ > r.out.bin input=elevation output=-
Exporting raster as floating values (bytes=4)
�- 0%
 100%
C�
  �B
```

But the file `-` is there (as I created it previously as written above):

```
GRASS :~ > ls -la '-' 
-rw-rw-r-- 1 mneteler mneteler 5 Jan 14 22:11 -
```

Shoudn't I see an error without this PR? I also tested in G79, same behaviour.

tmszi:
> I made a test with the unpatched G78git: Strangely I do not get the "overwrite" issue:
> 
> ```
> grass78 ~/grassdata/nc_spm_08_grass7/user1/
> ...
> Welcome to GRASS GIS 7.8.6dev (af07fc916)
> ...
> 
> GRASS :~ > r.out.bin input=elevation output=-
> Exporting raster as floating values (bytes=4)
> �- 0%
>  100%
> C�
>   �B
> ```
> 
> But the file `-` is there (as I created it previously as written above):
> 
> ```
> GRASS :~ > ls -la '-' 
> -rw-rw-r-- 1 mneteler mneteler 5 Jan 14 22:11 -
> ```
> 
> Shoudn't I see an error without this PR? I also tested in G79, same behaviour.
output parameter 

Sorry, my mistake I took a closer look at the code r.out.bin and it has an otherwise defined [output parameter](https://github.com/OSGeo/grass/blob/master/raster/r.out.bin/main.c#L293) (isn't affected with this bug) than [r.horizon](https://github.com/OSGeo/grass/blob/master/raster/r.horizon/main.c#L298) and [r.profile](https://github.com/OSGeo/grass/blob/master/raster/r.profile/main.c#L61).

neteler:
I see. I can replicate the original behavior with `r.profile`:

```
# no overwrite enabled
GRASS nc_spm_08_grass7/user1:~ > r.profile elevation resolution=1000 file=- << EOF
> 641712,226095
> 641546,224138
> 641546,222048
> 641049,221186
> EOF
ERROR: option <output>: <-> exists. To overwrite, use the --overwrite flag

# --overwrite
GRASS nc_spm_08_grass7/user1:~ > export GRASS_OVERWRITE=1
GRASS nc_spm_08_grass7/user1:~ > r.profile elevation resolution=1000 file=- << EOF
641712,226095
641546,224138
641546,222048
641049,221186
EOF
Using resolution: 1000 [meters]
Output columns:
Along track dist. [meters], Elevation
Approx. transect length: 1964.027749 [meters]
 0.000000 78.522743
 1000.000000 78.522743
Approx. transect length: 2090.000000 [meters]
 1964.027749 78.522743
 2964.027749 78.522743
 3964.027749 78.522743
Approx. transect length: 995.014070 [meters]
 4054.027749 78.522743
```

However, isn't `-` a legal file name (I agree, a bit unexpected)?

Not sure if we may happily overwrite it to fix the GUI issue shown in #1242.
An alternative might be:

- when in GUI:
    - move `-` to a tmp name
    - do the plotting
    - move tmp name back to `-`

Disadvantage: it is a kind of race condition in case the user wants to access the real file `$HOME/-` in the very moment that the GUI is plotting stuff.

Am I right that this PR kind of intercept `output=-` and avoids the real file `$HOME/-` being affected? If yes, that sounds reasonable to me.

tmszi:
> However, isn't `-` a legal file name (I agree, a bit unexpected)?

Yes it's allowed char https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282

> Not sure if we may happily overwrite it to fix the GUI issue shown in #1242.
> An alternative might be:
> 
>     * when in GUI:
>       
>       * move `-` to a tmp name
>       * do the plotting
>       * move tmp name back to `-`
> 
> 
> Disadvantage: it is a kind of race condition in case the user wants to access the real file `$HOME/-` in the very moment that the GUI is plotting stuff.
> 

The error is that even though this char `-`  means stdout, but module parser takes it as a file and checks if it exists.

> Am I right that this PR kind of intercept `output=-` and avoids the real file `$HOME/-` being affected? If yes, that sounds reasonable to me.

PR adds the [condition](https://github.com/OSGeo/grass/pull/1245/commits/4b4de2817ad2de3d927c84f9b2d8999dc228ab8b#diff-014394a922e77c10db6dd304c9241278f5da8cab037bee80269881e51837cd21R1595) if the output file parameter is stdout `output=-`  and then do not [check the existence](https://github.com/OSGeo/grass/pull/1245/commits/4b4de2817ad2de3d927c84f9b2d8999dc228ab8b#diff-014394a922e77c10db6dd304c9241278f5da8cab037bee80269881e51837cd21R1596) of this 'file', because it is not a file but a stdout indicator.

neteler (dismissed):
Tested successfully with `r.profile`.

neteler:
Could @cmbarton or @nilason make a test of this PR on Mac, please?

cmbarton:
Sure. I can try it tomorrow or Wednesday

Michael

______________
Michael Barton

Sent from  my iPhone
Please excusr any typoz

nilason:
I couldn't reproduce this issue. Following log without this PR:
```
GRASS : grass > touch ~/-                                               Mapset <user1> in <nc_spm_08>
GRASS : grass > r.profile elevation resolution=1000 file=- << EOF       Mapset <user1> in <nc_spm_08>
641712,226095
641546,224138
641546,222048
641049,221186
EOF
Using resolution: 1000 [meters]
Output columns:
Along track dist. [meters], Elevation
Approx. transect length: 1964.027749 [meters]
 0.000000 84.511765
 1000.000000 98.179062
Approx. transect length: 2090.000000 [meters]
 1964.027749 83.541069
 2964.027749 89.507294
 3964.027749 77.892715
Approx. transect length: 995.014070 [meters]
 4054.027749 73.837120
GRASS : grass >                                                         Mapset <user1> in <nc_spm_08>
```

nilason:
I did also test in GUI without issue.

cmbarton:
I tried to test this today and could not get GRASS to launch. Since this is an unmarked PR, I just downloaded the zip rather than clone it, to make sure I only had the code with this PR. I compiled it using the procedure that @nilason automated and which has worked very well. It compiled without error. But when I tried to launch GRASS, I got this error in the terminal. (I compiled and tried it twice). No idea what the problem is, since I compiled 7.9dev a few days ago and it runs fine. 

```
Last login: Wed Jan 20 14:05:14 on ttys001
/usr/bin/env -i HOME=/Users/cmbarton LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 BASH_SILENCE_DEPRECATION_WARNING=1 SHELL=/bin/bash PATH=/usr/bin:/bin:/usr/sbin:/etc:/usr/lib /Applications/GRASS-7.9.app/Contents/MacOS/Grass.sh
cmbarton@TCLASHE-C02V435YJ1GQ ~ % /usr/bin/env -i HOME=/Users/cmbarton LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 BASH_SILENCE_DEPRECATION_WARNING=1 SHELL=/bin/bash PATH=/usr/bin:/bin:/usr/sbin:/etc:/usr/lib /Applications/GRASS-7.9.app/Contents/MacOS/Grass.sh
Rebuilding Addon HTML manual pages index...
Rebuilding Addon menu...
usage: grass79 [-h] [-v] [--text] [--gtext] [--gui] [-c [CRS]] [-e]
               [--config [CONFIG [CONFIG ...]]] [--tmp-mapset]
               [--tmp-location CRS] [-f] [--exec ...]
               [PATH]
grass79: error: unrecognized arguments: -gui
```

nilason:
> I tried to test this today and could not get GRASS to launch. Since this is an unmarked PR, I just downloaded the zip rather than clone it, to make sure I only had the code with this PR. I compiled it using the procedure that @nilason automated and which has worked very well. It compiled without error. But when I tried to launch GRASS, I got this error in the terminal. (I compiled and tried it twice). No idea what the problem is, since I compiled 7.9dev a few days ago and it runs fine.


I don't know what went wrong at the moment. But please, do test with a working copy (without this PR) to see if you can reproduce the issue at all. I couldn't.

cmbarton:
When I deleted a file named '-' (who knows where it came from) in my home folder, as suggested by @tmszi the problem went away. I now have slightly newer versions on all my machines and do not have a problem, even without the PR and adding back a file named '-'.

nilason:
@cmbarton See https://trac.osgeo.org/grass/wiki/HowToGit#Checkingoutpullrequestslocally for how-to checkout a GitHub PR for testing.

cmbarton:
> @cmbarton See https://trac.osgeo.org/grass/wiki/HowToGit#Checkingoutpullrequestslocally for how-to checkout a GitHub PR for testing.

Right. But since I'm not contributing code, I have a much simpler setup for now. I just clone and update the main repository to make binaries, rather than keeping my own fork in sync. My understanding is that I should have been able to just download or clone the PR fork to test. If this is not correct, let me know. Thanks.

wenzeslaus:
> > However, isn't `-` a legal file name (I agree, a bit unexpected)?
> 
> Yes it's allowed char https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282

I think our approach should be, `-` always means stdout (or stdin). If module creates (or accepts) files named `-`, it is a bug. `-` should either result in using stdout (or stdin) or result in an error that -/stdin/stdout is not supported. In other words, we don't support file name named `-` even when stdout (or stdin) is not supported. This may not be the behavior of all modules right now, but it is aligned  with this PR.

ninsbl:

HuidaeCho:
> I couldn't reproduce this issue. Following log without this PR:
> 
> ```
> GRASS : grass > touch ~/-                                               Mapset <user1> in <nc_spm_08>
> GRASS : grass > r.profile elevation resolution=1000 file=- << EOF       Mapset <user1> in <nc_spm_08>
> 641712,226095
> 641546,224138
> 641546,222048
> 641049,221186
> EOF
> Using resolution: 1000 [meters]
> Output columns:
> Along track dist. [meters], Elevation
> Approx. transect length: 1964.027749 [meters]
>  0.000000 84.511765
>  1000.000000 98.179062
> Approx. transect length: 2090.000000 [meters]
>  1964.027749 83.541069
>  2964.027749 89.507294
>  3964.027749 77.892715
> Approx. transect length: 995.014070 [meters]
>  4054.027749 73.837120
> GRASS : grass >                                                         Mapset <user1> in <nc_spm_08>
> ```

That is because this bug only affects "new" files.

https://github.com/OSGeo/grass/blob/9cb4745b6c4abfeaf542ef05468060d68af72703/lib/gis/parser.c#L1717-L1726

HuidaeCho:
> > > However, isn't `-` a legal file name (I agree, a bit unexpected)?
> > 
> > 
> > Yes it's allowed char https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282
> 
> I think our approach should be, `-` always means stdout (or stdin). If module creates (or accepts) files named `-`, it is a bug. `-` should either result in using stdout (or stdin) or result in an error that -/stdin/stdout is not supported. In other words, we don't support file name named `-` even when stdout (or stdin) is not supported. This may not be the behavior of all modules right now, but it is aligned with this PR.

I agreed. I think currently different modules behave differently with `-` filenames. We need to standardize it. For that, we need to handle existing `-` files as well (maybe, spit out a warning message saying the module will use stdin, not the existing `-` file), not just "new" ones.

What do we do with `-` files then? `input=- < -` ;-)

Issue #1247, opened 2021-01-13
[Feat] Fix compiler warnings


nilason:
The number of compiler warnings are very high, with gcc there are about 140 and with clang about 300 warnings. Most are perhaps trivial and just annoyance, but who knows...
They not only add a lot of noise to build log (you might add to these some 200+ OpenGL deprecation warnings for Mac, and the ctypes warnings addressed in #507), they also makes it very difficult to discover any new potential problems.

**Describe the solution you'd like**
The best would be to address the warnings in the source.

**Describe alternatives you've considered**
It is possible, but in general not advisable, to suppress the warnings with compiler flags.

**Additional context**

*Warning Type	(No. of Warnings)*
*Apple clang version 12.0.0 (clang-1200.0.32.21)*
- [ ] -W#warnings	(16)
- [ ] -Wabsolute-value	(41)
- [ ] -Warray-bounds	(2)
- [ ] -Wc++11-compat-deprecated-writable-strings	(4)
- [ ] -Wdangling-else	(1)
- [ ] -Wdeprecated-declarations	(2)
- [ ] -Wformat	(83)
- [ ] -Wformat-extra-args	(2)
- [ ] -Wformat-invalid-specifier	(4)
- [ ] -Wheader-guard	(8)
- [ ] -Wincompatible-pointer-types	(1)
- [ ] -Wincompatible-pointer-types-discards-qualifiers	(15)
- [ ] -Wlogical-not-parentheses	(5)
- [ ] -Wmacro-redefined	(3)
- [ ] -Wnon-literal-null-conversion	(7)
- [ ] -Wparentheses	(2)
- [ ] -Wparentheses-equality	(3)
- [ ] -Wpointer-bool-conversion	(51)
- [ ] -Wpointer-compare	(4)
- [ ] -Wpointer-sign	(19)
- [ ] -Wswitch	(1)
- [ ] -Wtautological-compare	(1)
- [ ] -Wtautological-pointer-compare	(1)
- [ ] -Wundefined-inline	(16)
- [ ] -Wunused-result	(8)
- [ ] -Wwritable-strings	(5)

HuidaeCho:
I don't think we want to suppress *any* warnings. Ideally, yes, we should fix all of them in the source code. I think we can tackle them one by one (a PR per warning type).

nilason:
> I don't think we want to suppress _any_ warnings. Ideally, yes, we should fix all of them in the source code. I think we can tackle them one by one (a PR per warning type).

[...]

PR #1248, opened 2021-01-13
Fix compiler warnings, part 1

nilason:
Fixes compiler warnings:
- -Wabsolute-value
- -Warray-bounds
- -Wc++11-compat-deprecated-writable-strings
- -Wdangling-else

First part addressing #1247.

There are a couple of ambivalent cases and I'm not sure how to correctly address them.
I hope for input from experienced C programmers and leave this open as a draft.

Update:

- imagery/i.aster.toar (partly rewritten)
- imagery/i.evapo.time (partly rewritten)
- lib/gmath
- lib/imagery
- lib/vector/neta
- raster/r.grow.distance
- raster/r.latlong
- raster/r.stream.extract (commented out a part)
- raster/r.surf.idw
- raster/r.viewshed
- vector/v.generalize
- vector/v.label.sa
- vector/v.patch

### DIFF
--- a/imagery/i.aster.toar/main.c
+++ b/imagery/i.aster.toar/main.c
@@ -80,7 +80,7 @@ int main(int argc, char *argv[])
 
     /* For some strange reason infd[0] cannot be used later */
     /* So nfiles is initialized with nfiles = 1 */
-    int nfiles = 1;
+    int nfiles = 0;
     int i = 0, j = 0;
     int radiance = 0;
     void *inrast[MAXFILES];
@@ -227,23 +227,20 @@ int main(int argc, char *argv[])
 
     /********************/
     for (; *ptr != NULL; ptr++) {
-        if (nfiles > MAXFILES)
+        if (nfiles == MAXFILES)
             G_fatal_error(_("Too many input maps. Only %d allowed."),
                           MAXFILES);
         name = *ptr;
         /* Allocate input buffer */
-        in_data_type[nfiles - 1] = Rast_map_type(name, "");
-        /* For some strange reason infd[0] cannot be used later */
-        /* So nfiles is initialized with nfiles = 1 */
+        in_data_type[nfiles] = Rast_map_type(name, "");
         infd[nfiles] = Rast_open_old(name, "");
 
         Rast_get_cellhd(name, "", &cellhd);
-        inrast[nfiles - 1] = Rast_allocate_buf(in_data_type[nfiles - 1]);
+        inrast[nfiles] = Rast_allocate_buf(in_data_type[nfiles]);
         nfiles++;
     }
-    nfiles--;
     if (nfiles < MAXFILES)
-        G_fatal_error(_("The input band number should be 15"));
+        G_fatal_error(_("The input band number should be %d"), MAXFILES);
 
     /***************************************************/
     /* Allocate output buffer, use input map data_type */
@@ -253,21 +250,21 @@ int main(int argc, char *argv[])
     for (i = 0; i < MAXFILES; i++)
         outrast[i] = Rast_allocate_buf(out_data_type);
 
-    outfd[1] = Rast_open_new(result0, 1);
-    outfd[2] = Rast_open_new(result1, 1);
-    outfd[3] = Rast_open_new(result2, 1);
-    outfd[4] = Rast_open_new(result3, 1);
-    outfd[5] = Rast_open_new(result4, 1);
-    outfd[6] = Rast_open_new(result5, 1);
-    outfd[7] = Rast_open_new(result6, 1);
-    outfd[8] = Rast_open_new(result7, 1);
-    outfd[9] = Rast_open_new(result8, 1);
-    outfd[10] = Rast_open_new(result9, 1);
-    outfd[11] = Rast_open_new(result10, 1);
-    outfd[12] = Rast_open_new(result11, 1);
-    outfd[13] = Rast_open_new(result12, 1);
-    outfd[14] = Rast_open_new(result13, 1);
-    outfd[15] = Rast_open_new(result14, 1);
+    outfd[0] = Rast_open_new(result0, 1);
+    outfd[1] = Rast_open_new(result1, 1);
+    outfd[2] = Rast_open_new(result2, 1);
+    outfd[3] = Rast_open_new(result3, 1);
+    outfd[4] = Rast_open_new(result4, 1);
+    outfd[5] = Rast_open_new(result5, 1);
+    outfd[6] = Rast_open_new(result6, 1);
+    outfd[7] = Rast_open_new(result7, 1);
+    outfd[8] = Rast_open_new(result8, 1);
+    outfd[9] = Rast_open_new(result9, 1);
+    outfd[10] = Rast_open_new(result10, 1);
+    outfd[11] = Rast_open_new(result11, 1);
+    outfd[12] = Rast_open_new(result12, 1);
+    outfd[13] = Rast_open_new(result13, 1);
+    outfd[14] = Rast_open_new(result14, 1);
     /* Process pixels */
 
     DCELL dout[MAXFILES];
@@ -276,8 +273,8 @@ int main(int argc, char *argv[])
     for (row = 0; row < nrows; row++) {
         G_percent(row, nrows, 2);
         /* read input map */
-        for (i = 1; i <= MAXFILES; i++)
-            Rast_get_row(infd[i], inrast[i - 1], row, in_data_type[i - 1]);
+        for (i = 0; i < MAXFILES; i++)
+            Rast_get_row(infd[i], inrast[i], row, in_data_type[i]);
 
         /*process the data */
         for (col = 0; col < ncols; col++) {
@@ -306,13 +303,13 @@ int main(int argc, char *argv[])
                 outrast[i][col] = dout[i];
             }
         }
-        for (i = 1; i <= MAXFILES; i++)
-            Rast_put_row(outfd[i], outrast[i - 1], out_data_type);
+        for (i = 0; i < MAXFILES; i++)
+            Rast_put_row(outfd[i], outrast[i], out_data_type);
     }
-    for (i = 1; i <= MAXFILES; i++) {
-        G_free(inrast[i - 1]);
+    for (i = 0; i < MAXFILES; i++) {
+        G_free(inrast[i]);
         Rast_close(infd[i]);
-        G_free(outrast[i - 1]);
+        G_free(outrast[i]);
         Rast_close(outfd[i]);
     }
     exit(EXIT_SUCCESS);
--- a/imagery/i.evapo.time/main.c
+++ b/imagery/i.evapo.time/main.c
@@ -79,7 +79,7 @@ int main(int argc, char *argv[])
     module->description =_("Computes temporal integration of satellite "
 			   "ET actual (ETa) following the daily ET reference "
 			   "(ETo) from meteorological station(s).");
-    
+
     /* Define the different options */
     input = G_define_standard_option(G_OPT_R_INPUTS);
     input->key = "eta";
@@ -100,7 +100,7 @@ int main(int argc, char *argv[])
     input3->type = TYPE_DOUBLE;
     input3->required = YES;
     input3->description = _("Value of DOY for ETo first day");
-    
+
     input4 = G_define_option();
     input4->key = "start_period";
     input4->type = TYPE_DOUBLE;
@@ -114,11 +114,11 @@ int main(int argc, char *argv[])
     input5->description = _("Value of DOY for the last day of the period studied");
 
     output = G_define_standard_option(G_OPT_R_OUTPUT);
-    
+
     /* init nfiles */
-    nfiles = 1;
-    nfiles1 = 1;
-    nfiles2 = 1;
+    nfiles = 0;
+    nfiles1 = 0;
+    nfiles2 = 0;
 
     /********************/
 
@@ -157,13 +157,13 @@ int main(int argc, char *argv[])
 	inrast[nfiles] = Rast_allocate_d_buf();
 	nfiles++;
     }
-    nfiles--;
-    if (nfiles <= 1)
+
+    if (nfiles < 2)
 	G_fatal_error(_("The min specified input map is two"));
-	
+
 	/****************************************/
     for (; *ptr1 != NULL; ptr1++) {
-	if (nfiles1 > MAXFILES)
+	if (nfiles1 == MAXFILES)
 	    G_fatal_error(_("Too many ETa_doy files. Only %d allowed."),
 			  MAXFILES);
 	name1 = *ptr1;
@@ -173,8 +173,8 @@ int main(int argc, char *argv[])
 	inrast1[nfiles1] = Rast_allocate_d_buf();
 	nfiles1++;
     }
-    nfiles1--;
-    if (nfiles1 <= 1)
+
+    if (nfiles1 < 2)
 	G_fatal_error(_("The min specified input map is two"));
 
 
@@ -185,7 +185,7 @@ int main(int argc, char *argv[])
 	/****************************************/
 
     for (; *ptr2 != NULL; ptr2++) {
-	if (nfiles > MAXFILES)
+	if (nfiles2 == MAXFILES)
 	    G_fatal_error(_("Too many ETo files. Only %d allowed."),
 			  MAXFILES);
 	name2 = *ptr2;
@@ -195,8 +195,8 @@ int main(int argc, char *argv[])
 	inrast2[nfiles2] = Rast_allocate_d_buf();
 	nfiles2++;
     }
-    nfiles2--;
-    if (nfiles2 <= 1)
+
+    if (nfiles2 < 2)
 	G_fatal_error(_("The min specified input map is two"));
 
     /* Allocate output buffer, use input map data_type */
@@ -222,13 +222,13 @@ int main(int argc, char *argv[])
 	G_percent(row, nrows, 2);
 
 	/* read input map */
-	for (i = 1; i <= nfiles; i++) 
+	for (i = 0; i < nfiles; i++)
 	    Rast_get_d_row(infd[i], inrast[i], row);
-	
-	for (i = 1; i <= nfiles1; i++) 
+
+	for (i = 0; i < nfiles1; i++)
 	    Rast_get_d_row(infd1[i], inrast1[i], row);
 
-	for (i = 1; i <= nfiles2; i++) 
+	for (i = 0; i < nfiles2; i++)
 	    Rast_get_d_row (infd2[i], inrast2[i], row);
 
 	/*process the data */
@@ -236,14 +236,14 @@ int main(int argc, char *argv[])
         {
             int	d1_null=0;
             int	d_null=0;
-	    for (i = 1; i <= nfiles; i++) 
+	    for (i = 0; i < nfiles; i++)
             {
 		    if (Rast_is_d_null_value(&((DCELL *) inrast[i])[col]))
 		    	d_null=1;
 		    else
 	                d[i] = ((DCELL *) inrast[i])[col];
 	    }
-	    for (i = 1; i <= nfiles1; i++) 
+	    for (i = 0; i < nfiles1; i++)
             {
 		    if (Rast_is_d_null_value(&((DCELL *) inrast1[i])[col]))
 			d1_null=1;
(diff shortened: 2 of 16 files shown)


raster/r.stream.extract/streams.c
```diff
@@ -598,6 +598,12 @@ int extract_streams(double threshold, double mont_exp, int internal_acc)
 	    
 	/* set main drainage direction to A* path if possible */
 	if (mfd_cells > 0 && max_side != np_side) {
+	    /* UNCERTAIN SOLUTION
+	    either:
+	    if (abs(wat_nbr[np_side] >= max_acc)) {
+	    or:
+	    if (fabs(wat_nbr[np_side]) >= max_acc) {
+	    */
 	    if (fabs(wat_nbr[np_side] >= max_acc)) {
```
nilason, line 607: Please, see comment in code, which is the correct solution?

lib/raster3d/test/test_put_get_value_large_file.c
```diff
@@ -130,6 +130,10 @@ int test_large_file(int depths, int rows, int cols, int tile_size)
             for(x = 0; x < region.cols; x++) {
                 /* Check the counter as cell value */
                 Rast3d_get_value(map, x, y, z, &value, DCELL_TYPE);
+                /* UNCERTAIN SOLUTION
+                warning: using floating point absolute value
+                function 'fabs' when argument is of integer type 
+                -- either rearrange parenthesis or use abs() */
                 if(fabs(value - (double)(count) > EPSILON)) {
```
nilason, line 137: Please, see comment in code, what is the correct solution?

raster/r.viewshed/main.cpp
```diff
@@ -501,7 +501,7 @@ parse_args(int argc, char *argv[], int *vpRow, int *vpCol,
     obsElevOpt->required = NO;
     obsElevOpt->key_desc = "value";
     obsElevOpt->description = _("Viewing elevation above the ground");
-    obsElevOpt->answer = "1.75";
+    obsElevOpt->answer = strdup("1.75");
```
nilason, line 504: String from strdup() should be freed. If this is a problem, an alternative solution must be found.

wenzeslaus: The `strdup()` call is there likely to avoid the issue of assigning `const char *` to `char *` which in general can be problem and causes a "discarding const qualifiers" warning. I think that's the underlying problem here. The GRASS solution to that is `G_store` which I think suffers from the same issues as `strdup`.

Since this is setting a default which, if I'm not mistaken, is only read, not used for storage, it is safe to recast. In C++ (where this is a warning on Linux with GCC for a long time already), I was using `const_cast<char*>("...")` for the answer. I assume, but never needed to try, that in C, a `(char *)"..."` cast should do the trick, i.e., express that here it is indeed okay and intended to assign literal to `char *`.

I think the ultimate solution (to this and several other issues with these structures) would be a function-based API for the options rather than directly assigning to the member variables.

As a heads-up, the freeing will be/is an issue in many other places.

lib/raster3d/test/test_put_get_value_large_file.c
wenzeslaus: I didn't examine the broader context, but given that value is DCELL, that count is casted to double, the statement itself, the fact that `>` poperator returns 0 or 1, and the following message, it looks like the intention was a classic `|a - b| > threshold` where `|x|` signifies absolute value of `x`. Conclusion: the parenthesis is wrong. It is a test, so perhaps the problem it never surfaced.

raster/r.stream.extract/streams.c
metzm: The correct solution is the second suggestion: `if (fabs(wat_nbr[np_side]) >= max_acc) {`
However, this if block should be commented out as in [r.watershed](https://github.com/OSGeo/grass/blob/master/raster/r.watershed/seg/do_cum.c#L508).

lib/raster3d/test/test_put_get_value_large_file.c
metzm: Following the logic of the code, the correct solution must be `if (fabs(value - (double)(count)) > EPSILON) {`

raster/r.viewshed/main.cpp
metzm: I think explicitly free'ing the string from strdup() is not needed because the default answer is allocated only once and automatically free'd when the program terminates.

imagery/i.aster.toar/main.c
```diff
@@ -307,13 +307,13 @@ int main(int argc, char *argv[])
             }
         }
         for (i = 1; i <= MAXFILES; i++)
```
HuidaeCho, line 309: I would change the for loop instead of its index. `for (i = 0; i < MAXFILES; i++)`.

imagery/i.aster.toar/main.c
```diff
@@ -307,13 +307,13 @@ int main(int argc, char *argv[])
...
-            Rast_put_row(outfd[i], outrast[i - 1], out_data_type);
+            Rast_put_row(outfd[i-1], outrast[i - 1], out_data_type);
     }
     for (i = 1; i <= MAXFILES; i++) {
         G_free(inrast[i - 1]);
         Rast_close(infd[i]);
```
HuidaeCho, line 314: Is `infd[i]` still correct?

imagery/i.aster.toar/main.c
```diff
@@ -307,13 +307,13 @@ int main(int argc, char *argv[])
             }
         }
         for (i = 1; i <= MAXFILES; i++)
-            Rast_put_row(outfd[i], outrast[i - 1], out_data_type);
+            Rast_put_row(outfd[i-1], outrast[i - 1], out_data_type);
```
HuidaeCho, line 310: `i - 1` (spacing) to be consistent.

imagery/i.aster.toar/main.c
metzm: It seems that this is C code translated from Fortran code without taking care of the indices. E.g. MAXFILES is set to [15](https://github.com/OSGeo/grass/blob/master/imagery/i.aster.toar/main.c#L23), and outfd is an array of size MAXFILES. In https://github.com/OSGeo/grass/blob/master/imagery/i.aster.toar/main.c#L256 ff, the outfd array is set for indices 1 - 15 instead of 0 - 14, which can cause a segmentation fault because outfd[15] has never been allocated, only outfd[0] through to outfd[14].

The array indexing needs to be fixed first to avoid index out of array bounds errors.

nilason: > E.g. MAXFILES is set to 15, and outfd is an array of size MAXFILES. In https://github.com/OSGeo/grass/blob/master/imagery/i.aster.toar/main.c#L256 ff, the outfd array is set for indices 1 - 15 instead of 0 - 14, which can cause a segmentation fault because outfd[15] has never been allocated, only outfd[0] through to outfd[14].

Yes this is what I meant to to address. (Please see commit message for suggested fix for each compiler warning type).

Big change re. to this seem to have committed with 
https://github.com/OSGeo/grass/commit/8661e3430d65af4acc76090ef92e5766a7ba6ad3.

Thanks for taking this up both of you, I should have brought this up to attention from the beginning, as I realised this might have real functional implications.

The question remain, where and how did the indices go wrong?

I’ll look into it, but I’m more than happy if someone beat me to it.  :)

raster/r.stream.extract/streams.c
nilason: That would be my take also, re. logical test.

But comment out, why not delete?

lib/raster3d/test/test_put_get_value_large_file.c
nilason: Again, figured `|a - b| > threshold` must be correct. Thanks for confirmation.

imagery/i.aster.toar/main.c
nilason: Please see comment  above.

imagery/i.aster.toar/main.c
wenzeslaus: > The question remain, where and how did the indices go wrong?

I would like to understand that too. Is that likely that you translate it from one language to another and miss an obvious well-known difference? Sometimes people use some padding to make some loops easy and what not, but perhaps it was really just forcing Fortran indices on C.

metzm: Set `int nfiles = 0;` in [L83](https://github.com/OSGeo/grass/pull/1248/files#diff-b73bcf9f5e3fdccbde25ae12cb5dcbb27d2af079a54944cb1994dd9a02d39d3fR83). 
The block from L229 to L243 should be
```
    for (; *ptr != NULL; ptr++) {
        if (nfiles == MAXFILES)
            G_fatal_error(_("Too many input maps. Only %d allowed."),
                          MAXFILES);
        name = *ptr;
        /* Allocate input buffer */
        in_data_type[nfiles] = Rast_map_type(name, "");
        infd[nfiles] = Rast_open_old(name, "");

        Rast_get_cellhd(name, "", &cellhd);
        inrast[nfiles] = Rast_allocate_buf(in_data_type[nfiles]);
        nfiles++;
    }
```
`nfiles--;` in L244 must then be removed.

The lines L279, L280 should be
```
        for (i = 0; i < MAXFILES; i++)
            Rast_get_row(infd[i], inrast[i], row, in_data_type[i]);
```
L309, L310 should be
```
        for (i = 0; i < MAXFILES; i++)
            Rast_put_row(outfd[i], outrast[i], out_data_type);
```
L312 - L317:
```
    for (i = 0; i < MAXFILES; i++) {
        G_free(inrast[i]);
        Rast_close(infd[i]);
        G_free(outrast[i]);
        Rast_close(outfd[i]);
    }
```
All array indices are now going from 0 to MAXFILES - 1. I could not add these suggestions directly to the code, thus here. The suggestions are untested.

nilason (commented):
Updated `imagery/i.aster.toar/main.c` according to @metzm suggestion, hopefully I got it right.

raster/r.viewshed/main.cpp
HuidaeCho: > I believe this is not the place to start a discussion of memory handling in grass. That been said, It is my understanding that harmless memory leaks are accepted in GRASS (due to short program/module life) — that’s why I suggested it in the first place. I’d prefer strdup() or rather G_store (thanks, that was new to me) over simple cast (however harmless that seem to be).

I wouldn't rely on *harmless memory leaks* because we cannot predict the life time of each module. I think here we're only talking about C++ modules, right? I tested this short code with `cc -Wall -otest test.c` and `clang -Wall -otest test.c`:
```c
#include <stdio.h>
int main(){
    char *a = "test";
    printf("%s\n", a);
}
```
I got no warnings or errors. Then, the same code (`test.cpp`) with `c++ -Wall -otest test.cpp` and `clang++ -Wall -otest test.cpp` produced:
```
test.cpp: In function ‘int main()’:
test.cpp:3:15: warning: deprecated conversion from string constant to ‘char*’ [-Wwrite-strings]
     char *a = "test";
               ^
```
and
```
test.c:3:15: warning: conversion from string literal to 'char *' is deprecated [-Wc++11-compat-deprecated-writable-strings]
    char *a = "test";
              ^
```
But it was a matter of simple type casting to `(char *)` to get rid of these warnings.
```c
#include <stdio.h>
int main(){
    char *a = (char *)"test";
    printf("%s\n", a);
}
```
If I'm not missing something, can't we just cast it to `(char *)` explicitly only in C++ modules?

raster/r.stream.extract/streams.c
nilason: updated as suggested and commented out

lib/raster3d/test/test_put_get_value_large_file.c
nilason: updated as suggested

raster/r.viewshed/main.cpp
nilason: If it would "only" be an argument to a printf I would agree totally. But there is no guarantee `Option->answer` will not be altered at some point, or is there.. ?

In any case both the string literal as well as a mutable string will be freed at r.viewshed execution finishes.

HuidaeCho: > If it would "only" be an argument to a printf I would agree totally. But there is no guarantee `Option->answer` will not be altered at some point, or is there.. ?

I see. At this point, a lot of modules use static `char *` to assign default answers. Once parsed, they can either come directly from the default (read only) or from the user (`G_malloc` from `lib/gis/parser.c`, but is usually never explicitly freed by modules). It would be a developer's mistake if s/he treated `->answer` as modifiable when its default was not dynamically allocated.

> 
> In any case both the string literal as well as a mutable string will be freed at r.viewshed execution finishes.

Yes, any allocated memory will be freed when a module completes and it's a relatively small amount of memory.

Because `parser.c` `G_malloc` memory for user-passed options, which again is rarely explicitly freed by developers, I think `G_store` is safe and even more consistent than static `char *`.

Maybe, I just don't like asymmetry where we only allocate memory, but never free it in the *same code*. If we have to, we can always use `G_add_error_handler` to free manually allocated memory for `->answer`. Again, at the end of module execution, there is not much gain in doing so other than making us satisfied ;-).

nilason: Implemented with `G_store()`. I believe this is the least bad solution.

imagery/i.aster.toar/main.c
nilason: Code updated.

imagery/i.aster.toar/main.c
nilason: Code updated.

imagery/i.aster.toar/main.c
nilason: Updated a suggested.

raster/r.viewshed/main.cpp
HuidaeCho: > Implemented with `G_store()`. I believe this is the least bad solution.

Looking at the code, answer is actually copied to another buffer and is never modified directly. I mean, let's consider the developer's intention. If s/he wanted to modify answer in place, `G_store` might not be a bad solution. However, if answer (default or not) never changes, simply casting will be fine. Do we want to update all C modules that do this (`->answer = "default";`)? I think not.

nilason:
This is now ready for review. Please pay special attention to `i.aster.toar`, which has been partly re-written by @metzm and `r.stream.extract` where a part of code has been commented out (per suggestion by @metzm).

imagery/i.aster.toar/main.c
```diff
@@ -306,14 +303,14 @@ int main(int argc, char *argv[])
...
+        for (i = 0; i < MAXFILES; i++)
+            Rast_put_row(outfd[i], outrast[i], out_data_type);
     }
-    for (i = 1; i <= MAXFILES; i++) {
-        G_free(inrast[i - 1]);
-        Rast_close(infd[i]);
-        G_free(outrast[i - 1]);
-        Rast_close(outfd[i]);
+    for (i = 0; i < MAXFILES; i++) {
+            G_free(inrast[i]);
```
HuidaeCho, line 310: L310-L313: Are these tabs? Why extra whitespaces than before?

imagery/i.evapo.time/main.c
```diff
@@ -316,7 +316,7 @@ int main(int argc, char *argv[])
 		}	
 	    }
 
-	    sum[MAXFILES] = 0.0;
+	    sum[MAXFILES-1] = 0.0;
 	    for (i = 1; i <= nfiles1; i++)
```
HuidaeCho, line 320: There is more off-by-1 indexing in this module. Do you want to fix it as well?

raster/r.viewshed/main.cpp
```diff
@@ -501,7 +501,7 @@ parse_args(int argc, char *argv[], int *vpRow, int *vpCol,
     obsElevOpt->required = NO;
     obsElevOpt->key_desc = "value";
     obsElevOpt->description = _("Viewing elevation above the ground");
-    obsElevOpt->answer = "1.75";
+    obsElevOpt->answer = G_store("1.75");
```
HuidaeCho, line 504: I really think that we should use `G_store()` on a case-by-case basis, not globally. If a module never changes `->answer`, simple casting should do just fine. In this case, it's just a warning, not a segfault error from trying to alter static strings. I would prefer `(char *)"1.75"` over `G_store("1.75")`. Maybe, that's just me?

imagery/i.evapo.time/main.c
nilason: You’re right, we have another index bug there! Bigger than the warning.

nilason: Looking into this a bit more and testing the indexing with sprinkling in some G_debug(), the indexing seems to me to be fine. But I come to the belief that line `sum[MAXFILES] = 0.0` is not necessary at all. That line of code never contributed in any meaningful way.
It's an attempt to zero out an initial value, which will be done anyway in row 334 (`sum[i]=0.0;`) if reached number of `MAXFILES`.

imagery/i.aster.toar/main.c
nilason: You're right, will correct that.

raster/r.viewshed/main.cpp
nilason: This is the only warning for assigning a string to Option->answer, this is because of this being c++ code (e.g. `i.atcorr` already uses G_store). There is no need to change other c code files. I think this is being consistent with existing code.

imagery/i.evapo.time/main.c
metzm: The indexing seems to be wrong to me. E.g. there is
```
    int infd[MAXFILES], infd1[MAXFILES], infd2[MAXFILES];
```
but in L149ff nfiles starts with 1 going to MAXFILES whereas allowed indices for infd and inrast are [0, MAXFILES - 1]
Also in L213 is `double sum[MAXFILES];`, largest allowed index is thus `MAXFILES - 1` and L319 `sum[MAXFILES] = 0.0;` is out of array bounds.
In this module, all `for` loops seem to use invalid indices.

nilason: Thanks! Although I did test the loops and it seemed to look ok for me, I must admit I never tested reaching end of array (`MAXFILE`). I'll look into this again and try come up with a solution...

nilason: Now I updated with new indexing. I believe I got it right, I need to some more actual testing (before and after output). Meanwhile, if you have the time please do take a look.

I'm not quite sure though, about the line, from where this all started: `sum[MAXFILES] = 0.0;`, exactly what was intended to be zeroed out.

metzm:
Looks good to me, ready to merge.

There are probably some other modules with wrong array indexing, but I recommend to put these issues in a separate PR if needed.

nilason:
> Looks good to me, ready to merge.

Thanks a lot!

I did rearrange the commits: separated the fixes for `i.aster.toar` and `i.evapo.time` to one commit each.

> There are probably some other modules with wrong array indexing, but I recommend to put these issues in a separate PR if needed.

At the moment these two are the only places reported by clang.

metzm (approved):
Interesting how many occurrences of `abs()` needed to be replaced with `fabs()`!
Thanks for your work, IMHO now ready to merge.

nilason:
If there are no objections, I will take the liberty to merge in the next couple of days.

wenzeslaus (approved):
+1, but I'm also wondering about that commented out code in stream extract. I don't see anything useful in r.watershed blame.

I think, if it is just wrong, it should be deleted with a proper commit message (it is in the history after all if needed). If it is servers as a reminder of something, a comment should be added saying why it is there instead of leaving the reader wondering.

Commenting out all code would indeed fix all or most of compiler warnings, but that would be certainly an unexpected thing to find behind a "fix warnings commit". :smile:

nilason:
> I think, if it is just wrong, it should be deleted with a proper commit message (it is in the history after all if needed). If it is servers as a reminder of something, a comment should be added saying why it is there instead of leaving the reader wondering.
> 
> Commenting out all code would indeed fix all or most of compiler warnings, but that would be certainly an unexpected thing to find behind a "fix warnings commit". 😄

Good point. Although I can live with commented out code (for readily available reminder in case...), it certainly need a commit (with message) on its own. Do you have any suggestion,  @metzm?

nilason:
I separated the commented out code in `r.stream.extract` to a single distinct commit. The commit message may of course be improved...

metzm:
> > I think, if it is just wrong, it should be deleted with a proper commit message (it is in the history after all if needed). If it is servers as a reminder of something, a comment should be added saying why it is there instead of leaving the reader wondering.
> > Commenting out all code would indeed fix all or most of compiler warnings, but that would be certainly an unexpected thing to find behind a "fix warnings commit". smile
> 
> Good point. Although I can live with commented out code (for readily available reminder in case...), it certainly need a commit (with message) on its own. Do you have any suggestion, @metzm?

Delete the commented out code, no need to keep it.

nilason:
Apparently there is no way for me to merge this without sqashing all commits.
It would be preferable to keep the commits separate. Suggestion anyone?

neteler:
We may temporarily disable squashing for this PR

nilason:
> We may temporarily disable squashing for this PR

If you have authority to merge without sqashing, if you do not mind please go ahead and merge.
This will (probably) only be necessary for these warning fix PRs.

neteler:
> > We may temporarily disable squashing for this PR
> 
> If you have authority to merge without sqashing, if you do not mind please go ahead and merge.
> This will (probably) only be necessary for these warning fix PRs.

I tried by temporarily setting in https://github.com/OSGeo/grass/settings

[x] Allow merge commits

but perhaps that's then only valid for new PRs opened after such a change? Do @landam or @wenzeslaus know?

nilason:
Updated commit messages with addition of PR no.

nilason:
It might work if I cherry pick to master locally and then push. What do you think, shall I give it a try?

nilason:
Merged through cherry picking. Closing this manually.
Thanks a bunch for all the help!

PR #175, opened 2019-11-01
json parser: change + to @ for URLs

neteler:
Since `+` is used in algebraic expressions incl. v.db.update, this char is unsuitable
for the indication of external URL resourcs as used by the importer of
actinia (https://github.com/mundialis/actinia_core/).

This change replaces `+` with `@`, hence this JSON output is now produced correctly (example):

```
GRASS 7.9.dev (utm32N_25832):~ > v.db.update map=antenna_position column="z_antenna" qcolumn="cat + 2.00" --json
{
  "module": "v.db.update",
  "id": "v.db.update_1804289383",
  "inputs":[
     {"param": "map", "value": "antenna_position"},
     {"param": "layer", "value": "1"},
     {"param": "column", "value": "z_antenna"},
     {"param": "query_column", "value": "cat + 2.00"}
   ]}
```

Fixes https://trac.osgeo.org/grass/ticket/3928

PR #1252, opened 2021-01-14
json parser: fix mapsets in tokenizing

neteler:
This PR fixes the broken tokenizing of `map@mapset`.

Wrong "import_descr":

```
GRASS :~ > r.info globcover@globcover --json
{
  "module": "r.info",
  "id": "r.info_1804289383",
  "inputs":[
     {"import_descr": {"source":"globcover", "type":"raster"},
      "param": "map", "value": "globcover"}
   ]}
```

Fixed with this PR:

```
GRASS :~ > r.info globcover@globcover --json
{
  "module": "r.info",
  "id": "r.info_1804289383",
  "inputs":[
     {"param": "map", "value": "globcover@globcover"}
   ]}
```

And this remains functional:

```
GRASS :~ > r.slope.aspect elevation="elevation@https://storage.googleapis.com/graas-geodata/elev_ned_30m.tif" slope="slope+COG" aspect="aspect+COG" --json
{
  "module": "r.slope.aspect",
  "id": "r.slope.aspect_1804289383",
  "inputs":[
     {"import_descr": {"source":"https://storage.googleapis.com/graas-geodata/elev_ned_30m.tif", "type":"raster"},
      "param": "elevation", "value": "elevation"},
     {"param": "format", "value": "degrees"},
     {"param": "precision", "value": "FCELL"},
     {"param": "zscale", "value": "1.0"},
     {"param": "min_slope", "value": "0.0"}
   ],
  "outputs":[
     {"export": {"format":"COG", "type":"raster"},
      "param": "slope", "value": "slope"},
     {"export": {"format":"COG", "type":"raster"},
      "param": "aspect", "value": "aspect"}
   ]
}
```

It completes the earlier fix in #175.

Applied `tools/grass_indent.sh ` in 04962a3

Co-authored-by: @anikaweinmann
Co-authored-by: @metzm

### DIFF
--- a/lib/gis/parser_json.c
+++ b/lib/gis/parser_json.c
@@ -342,7 +342,7 @@ char *G__json(void)
  */
 void check_create_import_opts(struct Option *opt, char *element, FILE *fp)
 {
-    int i = 0, urlfound;
+    int i = 0, urlfound = 0;
     int has_import = 0;
     char **tokens;
 
--- a/lib/gis/parser_json.c
+++ b/lib/gis/parser_json.c
@@ -32,157 +32,157 @@ char *check_mapset_in_layer_name(char *, int);
 
    r.slope.aspect elevation="elevation@https://storage.googleapis.com/graas-geodata/elev_ned_30m.tif" slope="slope+GTiff" aspect="aspect+GTiff" --json
 
-   {
-   "module": "r.slope.aspect",
-   "id": "r.slope.aspect_1804289383",
-   "inputs":[
-   {"import_descr": {"source":"https://storage.googleapis.com/graas-geodata/elev_ned_30m.tif", "type":"raster"},
-   "param": "elevation", "value": "elevation"},
-   {"param": "format", "value": "degrees"},
-   {"param": "precision", "value": "FCELL"},
-   {"param": "zscale", "value": "1.0"},
-   {"param": "min_slope", "value": "0.0"}
-   ],
-   "outputs":[
-   {"export": {"format":"GTiff", "type":"raster"},
-   "param": "slope", "value": "slope"},
-   {"export": {"format":"GTiff", "type":"raster"},
-   "param": "aspect", "value": "aspect"}
-   ]
-   }
+    {
+      "module": "r.slope.aspect",
+      "id": "r.slope.aspect_1804289383",
+      "inputs":[
+         {"import_descr": {"source":"https://storage.googleapis.com/graas-geodata/elev_ned_30m.tif", "type":"raster"},
+          "param": "elevation", "value": "elevation"},
+         {"param": "format", "value": "degrees"},
+         {"param": "precision", "value": "FCELL"},
+         {"param": "zscale", "value": "1.0"},
+         {"param": "min_slope", "value": "0.0"}
+       ],
+      "outputs":[
+         {"export": {"format":"GTiff", "type":"raster"},
+          "param": "slope", "value": "slope"},
+         {"export": {"format":"GTiff", "type":"raster"},
+          "param": "aspect", "value": "aspect"}
+       ]
+    }
 
    v.out.ascii input="hospitals@PERMANENT" output="myfile+TXT" --json
 
-   {
-   "module": "v.out.ascii",
-   "id": "v.out.ascii_1804289383",
-   "inputs":[
-   {"param": "input", "value": "hospitals@PERMANENT"},
-   {"param": "layer", "value": "1"},
-   {"param": "type", "value": "point,line,boundary,centroid,area,face,kernel"},
-   {"param": "format", "value": "point"},
-   {"param": "separator", "value": "pipe"},
-   {"param": "precision", "value": "8"}
-   ],
-   "outputs":[
-   {"export": {"format":"TXT", "type":"file"},
-   "param": "output", "value": "$file::myfile"}
-   ]
-   }
+    {
+      "module": "v.out.ascii",
+      "id": "v.out.ascii_1804289383",
+      "inputs":[
+         {"param": "input", "value": "hospitals@PERMANENT"},
+         {"param": "layer", "value": "1"},
+         {"param": "type", "value": "point,line,boundary,centroid,area,face,kernel"},
+         {"param": "format", "value": "point"},
+         {"param": "separator", "value": "pipe"},
+         {"param": "precision", "value": "8"}
+       ],
+      "outputs":[
+         {"export": {"format":"TXT", "type":"file"},
+          "param": "output", "value": "$file::myfile"}
+       ]
+    }
 
    v.info map="hospitals@PERMANENT" -c --json
 
-   {
-   "module": "v.info",
-   "id": "v.info_1804289383",
-   "flags":"c",
-   "inputs":[
-   {"param": "map", "value": "hospitals@PERMANENT"},
-   {"param": "layer", "value": "1"}
-   ]
-   }
+    {
+      "module": "v.info",
+      "id": "v.info_1804289383",
+      "flags":"c",
+      "inputs":[
+         {"param": "map", "value": "hospitals@PERMANENT"},
+         {"param": "layer", "value": "1"}
+       ]
+    }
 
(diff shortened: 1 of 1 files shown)


lib/gis/parser_json.c
```diff
@@ -28,7 +28,7 @@ char *check_mapset_in_layer_name(char *, int);
 
   The following commands will create according JSON output:
 
-    r.slope.aspect elevation="elevation+https://storage.googleapis.com/graas-geodata/elev_ned_30m.tif" slope="slope+GTiff" aspect="aspect+GTiff" --json
+    r.slope.aspect elevation="elevation@https://storage.googleapis.com/graas-geodata/elev_ned_30m.tif" slope="slope+GTiff" aspect="aspect+GTiff" --json
```
wenzeslaus, line 31: `slope+GTiff` and `aspect+GTiff` are fine?

lib/gis/parser_json.c
```diff
@@ -340,19 +340,27 @@ char *G__json(void)
...
         G_chop(tokens[i]);
         i++;
     }
 
+    /* check if it contains an URL, if ":" present, then URL, map@mapset otherwise */
+    urltokens = G_tokenize(tokens[1], ":");
+    while (urltokens[j]) {
+        G_chop(urltokens[j]);
+        j++;
+    }
```
wenzeslaus, line 359: Why not simple `strchr()` for char in string check?

lib/gis/parser_json.c
```diff
@@ -340,19 +340,27 @@ char *G__json(void)
...
+    char **tokens, **urltokens;
 
+    G_debug(3,"tokenize string: <%s>", opt->answer);
     tokens = G_tokenize(opt->answer, "@");
     while (tokens[i]) {
         G_chop(tokens[i]);
         i++;
     }
 
+    /* check if it contains an URL, if ":" present, then URL, map@mapset otherwise */
```
wenzeslaus, line 354: This does not sound dangerous at all.

wenzeslaus (commented):
I think `--json` lacks documentation and this issue seems to me as a consequence of that missing definition.

That could clarify also some interactions with `@` operator in `r.mapcalc` and the `@OGR` pseudo-mapset in commands like these:

```
v.info map="/tmp/file.gpkg@OGR" layer=abc
v.info map="/tmp/user:file.gpkg@OGR" layer=abc
```

Although `/tmp/user:file.gpkg` is the vector name, a message is generated saying `Topology not available for vector map <abc@/tmp/user:file.gpkg>.`

Colon is unlikely for a file name, but some mounts do include colons. `ogrinfo file:///tmp/file.gpkg` is not possible either at this point, but in context of trickeries such `@OGR` or `@http://` is seems possible this would be added at some level. Then there is `PG:`.

Anyway, I think figuring out how to make this work well is worth the trouble. The syntax could potentially replace `@OGR` or even facilitate need for explicit r.external, better `--exec`, or exposing things like `@OGR` in GUI.

lib/gis/parser_json.c
```diff
@@ -340,19 +340,27 @@ char *G__json(void)
...
     int has_import = 0;
-    char **tokens;
+    char **tokens, **urltokens;
 
+    G_debug(3,"tokenize string: <%s>", opt->answer);
     tokens = G_tokenize(opt->answer, "@");
     while (tokens[i]) {
         G_chop(tokens[i]);
         i++;
     }
```
metzm, line 352: Check if there are more than two tokens, if yes, fatal error. Or split at the first occurrence of "@".

lib/gis/parser_json.c
metzm: A simple `strchr()` would be ok if more than one ":" is allowed, then split at the first occurrence of ":". Otherwise, as above, check if there are more than two tokens, if yes, fatal error.

lib/gis/parser_json.c
neteler: Strangely apparently yes.

lib/gis/parser_json.c
neteler: Thanks for the idea, added in b601c83

lib/gis/parser_json.c
neteler: Thanks for the fatal error idea, added in b601c83

wenzeslaus: So no port number in the URL? Colon can also be part of URL in general, e.g., https://grasswiki.osgeo.org/wiki/Category:Documentation. Back to my other comment: Regular expressions would be quite appropriate for JSON/web use case here, but some manual check for a sequence of characters would likely do the trick here as well.

neteler: OK, entirely rewritten in 19d8065, now using `strstr()` instead of the tokenizer.

lib/gis/parser_json.c
```diff
@@ -340,19 +341,30 @@ char *G__json(void)
...
+    G_debug(2,"tokenize opt string: <%s> with '@'", opt->answer);
     tokens = G_tokenize(opt->answer, "@");
     while (tokens[i]) {
         G_chop(tokens[i]);
         i++;
     }
 
+    /* check if tokens[1] contains an URL or name@mapset */
+    G_debug(2,"tokens[1]: <%s>", tokens[1]);
+    if (strstr(tokens[1], "http://") != NULL || strstr(tokens[1], "https://") != NULL || strstr(tokens[1], "ftp://") != NULL ) {
```
metzm, line 357: if you use `strstr() != NULL`, you don't care where the substring you are searching for occurs in `tokens[1]`. Is this intended? Like `strstr("blablablahttp://", "http://")` is not NULL and an url was found.
Why not
```
    if (strncmp(tokens[1], "http://", 7) == 0 || strncmp(tokens[1], "https://", 8) == 0 || strncmp(tokens[1], "ftp://", 6) == 0 ) {
```
?

neteler: Suggestion happily accepted in d7114ac9a

lib/gis/parser_json.c
```diff
@@ -340,19 +341,30 @@ char *G__json(void)
...
     char **tokens;
 
+    G_debug(2,"tokenize opt string: <%s> with '@'", opt->answer);
     tokens = G_tokenize(opt->answer, "@");
     while (tokens[i]) {
         G_chop(tokens[i]);
         i++;
     }
 
+    /* check if tokens[1] starts with an URL or name@mapset */
```
metzm, line 355: This block needs to be enclosed in a `if (i > 1) {`, i.e. do not investigate token[1] if there is only one token.
Fatal error if i > 2?

neteler: OK, updated in 7d0d68ecb

lib/gis/parser_json.c
```diff
@@ -340,19 +342,34 @@ char *G__json(void)
  */
 void check_create_import_opts(struct Option *opt, char *element, FILE *fp)
 {
-    int i = 0;
+    int i = 0, urlfound;
```
metzm, line 345:
```suggestion
    int i = 0, urlfound = 0;
```

metzm: Do not use uninitialized urlfound.

metzm:
Please use [grass_indent.sh](https://github.com/OSGeo/grass/blob/master/tools/grass_indent.sh)

neteler:
> Please use [grass_indent.sh](https://github.com/OSGeo/grass/blob/master/tools/grass_indent.sh)

Done so in 04962a392

neteler:
Question: besides http/s and ftp, do we also need to add support for `s3://`?

metzm:
Apparently, `grass_indent.sh` can not format JSON code in C comments. This is up to you if you want to adjust the JSON code in the comments. For me, the C code is ready to merge.

neteler:
> Apparently, `grass_indent.sh` can not format JSON code in C comments. This is up to you if you want to adjust the JSON code in the comments. 

Sure: previous formatting restored.

> For me, the C code is ready to merge.

Thanks for your review!

PR #63, opened 2019-07-26
WIP Band references implementation

landam:
This PR implements basic functionality related to image collections (https://developers.google.com/earth-engine/tutorial_api_04).

It's dedicated for testing, various improvements are still planned.

Development has been supported mundialis GmbH & Co.KG

# How to test

## 1. Create GRASS location, download sample data

```
./bin.x86_64-pc-linux-gnu/grass77 -c EPSG:32633 /tmp/image-collection-test
> g.extension ext=i.sentinel
> i.sentinel.download settings=sentinel.txt uuid=a12d7d6d-0c41-4f5e-8789-7e9cff3529df output=/tmp
> i.sentinel.import input=/tmp
```

## 2. Create STRDS

```
> t.create test title=test description=test
```

### 3.a Create input file

(note that similar functionality is planned to be implemented in `i.sentinel.import`)

Run PyGRASS script in order to create input file for `t.register`:

```python
#!/usr/bin/env python

from grass.pygrass.raster import RasterRow
from grass.pygrass.gis import Mapset

from datetime import datetime

def list_rasters():
    m = Mapset()
    return m.glist('raster')

def do_list():
    for name in list_rasters():
        with RasterRow(name) as rast:
            band = rast.name.split('_', -1)[-1].lstrip('B').lstrip('0')
            band_ref = "S2_{}".format(band)

            date = datetime.strptime(
                rast.name.split('_')[1],
                '%Y%m%dT%H%M%S'
            )
            print ('{}|{}|{}'.format(name, date, band_ref))

if __name__ == "__main__":
    do_list()
```

Expected result:

```
T33UYP_20190331T094039_B01|2019-03-31 09:40:39|S2_1
T33UYP_20190331T094039_B10|2019-03-31 09:40:39|S2_10
T33UYP_20190331T094039_B02|2019-03-31 09:40:39|S2_2
T33UYP_20190331T094039_B05|2019-03-31 09:40:39|S2_5
T33UYP_20190331T094039_B11|2019-03-31 09:40:39|S2_11
T33UYP_20190331T094039_B08|2019-03-31 09:40:39|S2_8
T33UYP_20190331T094039_B12|2019-03-31 09:40:39|S2_12
T33UYP_20190331T094039_B8A|2019-03-31 09:40:39|S2_8A
T33UYP_20190331T094039_B06|2019-03-31 09:40:39|S2_6
T33UYP_20190331T094039_B04|2019-03-31 09:40:39|S2_4
[...]

PR #1272, opened 2021-01-26
Simplify raster band reference management

marisn:
PR #63 introduced imagery raster band reference support.
This PR simplifies raster band management by moving band references into simple metadata items. Band references are not any more tied to a known sensor (e.g. "vi_NDVI" now is a valid band reference).
Band reference is now printed by r.info and can be managed with metadata editing module r.support.


marisn:
On a second though current approach in band management is not flexible enough to untangle signature files from groups. I'll try to add some extra functionality to raster band references needed for further progress.

wenzeslaus:
> Long term goal – untangle signature files from groups and tie to band references.

Which brings back question which was not sufficiently answered before I think. What is the relation of band references and groups?

wenzeslaus:
Since you are thinking about this and adding interface, any opinion on using the word band here, esp. in names of modules without the word reference? Given how the term _band_ is usually used, i.band seems like referring to an object (the band reference) by adjective rather than a noun, i.e., going against the established meaning of band. I would just like to avoid another _map-layer_ mess ("Check which _layer_ of _vector map_ you use when adding a new _vector map layer_ to _map_ in _map display_").

marisn:
> Since you are thinking about this and adding interface, any opinion on using the word band here, esp. in names of modules without the word reference? Given how the term _band_ is usually used, i.band seems like referring to an object (the band reference) by adjective rather than a noun, i.e., going against the established meaning of band. I would just like to avoid another _map-layer_ mess ("Check which _layer_ of _vector map_ you use when adding a new _vector map layer_ to _map_ in _map display_").

Woosh! (The sound your comment made flying high above my head)

marisn:
> > Long term goal – untangle signature files from groups and tie to band references.
> 
> Which brings back question which was not sufficiently answered before I think. What is the relation of band references and groups?

None (sort of).
The way *I* see band references as an optional dimension name of *any* raster map. Such universal dimension identifier is required to make reusable signature files. Currently signatures are mapped to raster files by their order in group REF file. If map order in REF file changes, signature files become erroneous; There is no way to tell that the signature of group item 1 generated from map "a.red" is also valid for maps "b.red" and "c.red" (assuming a, b, c are form the same sensor). Thus my proposal would be to:
1. each raster map to have an optional band (dimension) name
2. map signatures to band (dimension) names (a.k.a. natural keys)

Thus a single signature file would be valid for any group as long as group members have the same band (dimension) names. Groups would stay as a mean of organizing a realization of set of bands (dimensions). 

Downsides:
* existing signature files will become obsolete
* impossible to do classification of rasters without band names
Impact of this issue could be minimised by:
  * automatic assigning of band name on import (if it is known)
  * automatic assigning of band name for various raster maps generated by GRASS tools (e.g. band "GRASS_slope_deg", "RI_NDVI")
  * improvement of metadata management tools (this is a good thing regardless of band management)

include/defs/raster.h
```diff
@@ -40,6 +40,7 @@ int Rast_has_band_reference(const char *, const char *);
 int Rast_read_band_reference(const char *, const char *, char **, char **);
 int Rast_write_band_reference(const char *, const char *, const char *);
 int Rast_remove_band_reference(const char *);
+int R_legal_band_name(const char *);
```
landam, line 43: Why not `Rast_legal_band_name()` ?

marisn: Forget about this. This is a bad idea. I am now playing with moving band IDs to map metadata block (a.k.a. struct History) + making sensor JSON file optional. Such approach would make band IDs tightly integrated and easy to use.

neteler:
Something went wrong here, @marisn :  Files changed 1,169?!
Forgot to update before push?

nilason:
> Something went wrong here, @marisn : Files changed 1,169?!
> Forgot to update before push?

I think this is what happens after a rebase. Just normal behaviour as far as I know.

wenzeslaus:
> > Something went wrong here, @marisn : Files changed 1,169?!
> > Forgot to update before push?
> 
> I think this is what happens after a rebase. Just normal behaviour as far as I know.

Unfortunately, something indeed went wrong as @neteler suggests. This is behavior when you rebase all other commits on top of yours. What you want when rebasing is rebasing your commits on top the other ones. `git rebase upstream/master` on your feature branch does that assuming your upstream/master is matching the OSGeo:master branch. The "right" rebase done without squash does not show anything special on GitHub except that you force pushed. With squash, it destroys commit history in the PR which may or may not be what you want (think e.g. review). Doing merge instead of the rebase is fine too. GitHub understands that and does shows only relevant changes, however it shows the merge commit. The merge commit disappears during _Squash and merge_.

I'm not sure what merging this would do, i.e., if _Squash and merge_ would remove the extra commits. You can see on GitHub that @marisn is now additional author for each of these commits. However, I think it needs to be redone anyway because of the review.

I think the fix should be straightforward, identify your commits, create a new branch from updated master branch, cherry pick your commits, force push that branch to your remote branch bandrefs. When you check that all is fine, you can delete the local branch called bandrefs. (This is perfectly fine approach to completely redoing code of a PR without opening a new PR especially applicable when the original code _nor the history_ does not have any value or the good parts will be transferred to the new branch and the original history is of no value for the review.)

marisn:
I significantly simplified raster band references based on the fact that only ID part is used everywhere and filename part is used only to print band metadata and thus can be replaced with a search at runtime.

@landam what do you think about such approach? (This is still WIP)

Tests for some temporal modules are failing and I have hard time to understand why. For some strange reason during registration code takes a wrong branch here: https://github.com/OSGeo/grass/blob/802b3c84beaad21ad15056257e50f19ff6cb951c/python/grass/temporal/register.py#L176

I also haven't understood what to do with i.band module. Band management now can be done via r.support (except printing). On one hand it could be kept as is (+ code upgrade) with additional explanations in documentation (+ link to r.support).

marisn:
> Tests for some temporal modules are failing and I have hard time to understand why. For some strange reason during registration code takes a wrong branch here:
> 
> https://github.com/OSGeo/grass/blob/802b3c84beaad21ad15056257e50f19ff6cb951c/python/grass/temporal/register.py#L176

Ah, I see that test `temporal/t.rast.aggregate/testsuite/test_aggregation_relative.py` also fails in master. I can fix it by following code, but, as I do not understand whole codebase, I can not judge if it is the right thing to do.

```
--- a/python/grass/temporal/register.py
+++ b/python/grass/temporal/register.py
@@ -173,7 +173,7 @@ def register_maps_in_space_time_dataset(
                 start_time_in_file = True
                 # Check if last column is an end time or a band reference
                 time_object = check_datetime_string(line_list[2])
-                if not sp.is_time_relative() and isinstance(time_object, datetime):
+                if sp.is_time_relative():
                     end_time_in_file = True
                     band_reference_in_file = False
                 else:
```

lib/raster/raster_metadata.c
```diff
@@ -81,6 +82,86 @@ void Rast_write_vdatum(const char *name, const char *str)
...
+ *
+ * \return 1 success
+ * \return -1 failure
+ */
+int Rast_legal_bandref(const char *bandref)
+{
+    char **tokens;
+    int ntok;
+
+    if (strlen(bandref) >= GNAME_MAX - 1) {
```
nilason, line 141: `strlen(bandref) >= GNAME_MAX` should be more correct.

`GNAME_MAX=256` which *includes* null string (i.e. 255 + 1), but strlen() returns length *without* the terminating character. If strlen() results in 256 or more the file name is too long.

raster/r.support/main.c
```diff
@@ -112,28 +120,48 @@ int main(int argc, char *argv[])
     map_opt->type = TYPE_STRING;
     map_opt->required = NO;
     map_opt->gisprompt = "old,cell,raster";
+    map_opt->guisection = _("Import / export");
```
nilason, line 123: Changing "Import / export" to  "Import / Export" looks better in my eyes.

raster/r.support/main.c
```diff
@@ -252,10 +284,21 @@ int main(int argc, char *argv[])
...
+    }
+
+    if (bandref_rm_flag->answer)
+        G_remove_misc("cell_misc", "bandref", infile);
 
     if (title_opt->answer || history_opt->answer || units_opt->answer
 	|| vdatum_opt->answer || datasrc1_opt->answer || datasrc2_opt->answer
-	|| datadesc_opt->answer || map_opt->answer)
+	|| datadesc_opt->answer || map_opt->answer
+    || bandref_opt->answer || bandref_rm_flag->answer)
```
nilason, line 301: This block has mixed tab/space indentation.

raster/r.support/testsuite/test_r_support.py
```diff
@@ -0,0 +1,66 @@
...
+from grass.lib.gis import G_remove_misc
+from grass.lib.raster import Rast_read_bandref
+
+
+class RSupportBandHandlingTestCase(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.map = tempname(10)
+        cls.mapset = Mapset().name
+        cls.bandref = "The_Doors"
```
nilason, line 30: 😄

python/grass/temporal/c_libraries_interface.py
```diff
@@ -1768,9 +1743,7 @@ def available_mapsets(self):
         """
         self.check_server()
         self.client_conn.send(
-            [
-                RPCDefs.AVAILABLE_MAPSETS,
-            ]
+            [RPCDefs.AVAILABLE_MAPSETS,]
```
nilason, line 1746: Black wants to revert this change.

python/grass/temporal/c_libraries_interface.py
```diff
@@ -1803,9 +1776,7 @@ def get_mapset(self):
         """
         self.check_server()
         self.client_conn.send(
-            [
-                RPCDefs.G_MAPSET,
-            ]
+            [RPCDefs.G_MAPSET,]
```
nilason, line 1779: Black wants to revert this change.

python/grass/temporal/c_libraries_interface.py
```diff
@@ -1816,9 +1787,7 @@ def get_location(self):
         """
         self.check_server()
         self.client_conn.send(
-            [
-                RPCDefs.G_LOCATION,
-            ]
+            [RPCDefs.G_LOCATION,]
```
nilason, line 1790: Black wants to revert this change.

python/grass/temporal/c_libraries_interface.py
```diff
@@ -1829,9 +1798,7 @@ def get_gisdbase(self):
         """
         self.check_server()
         self.client_conn.send(
-            [
-                RPCDefs.G_GISDBASE,
-            ]
+            [RPCDefs.G_GISDBASE,]
```
nilason, line 1801: Black wants to revert this change.

nilason:
Compilation and test files went fine. I mainly made a code and spelling review. Other than noted comments it looks good to me.

python/grass/temporal/c_libraries_interface.py
nilason: Reverting these will also solve Flake8's failure.

include/grass/defs/raster.h
```diff
@@ -550,8 +544,11 @@ DCELL Rast_get_d_value(const void *, RASTER_MAP_TYPE);
 /* raster_metadata.c */
 char *Rast_read_units(const char *, const char *);
 char *Rast_read_vdatum(const char *, const char *);
+char *Rast_read_bandref(const char *, const char *);
 void Rast_write_units(const char *, const char *);
 void Rast_write_vdatum(const char *, const char *);
+void Rast_write_bandref(const char *, const char *);
+int Rast_legal_bandref(const char *);
```
nilason, line 551: Why not bool as return value type?

raster/r.support/main.c
marisn: Most of C code base does not confirm to GRASS indentation rules. I made a proposal to re-indent whole code for 8.0, but so far only Python part has materialized. I could run grass_indent.sh on r.support, but that would change all files significantly thus cluttering the band reference part with formatting noise.

raster/r.support/testsuite/test_r_support.py
marisn: Wait till you'll see signature file tests :D

python/grass/temporal/c_libraries_interface.py
marisn: This is a problem, as this change **is** made by running black on the file... I'll try to ignore black and edit file manually.

raster/r.support/main.c
nilason: I absolutely agree it is high time to batch run grass_indent.sh, with #1391 in place at least the whitespace changes of such operation may be "hidden".

Meanwhile, the main problem with this line (although minor) is not with the mixed use of space/tab, but the added line should have 8 spaces to line up with the tab on previous line.

python/grass/temporal/c_libraries_interface.py
nilason: Testing this locally with Black 20.8b1, gives the same result as on CI's Black 21.5b1 (#1605).

raster/r.support/main.c
```diff
@@ -112,28 +120,48 @@ int main(int argc, char *argv[])
...
+	_("Band reference in form shortcut_name e.g. S2_8A");
+
+    bandref_rm_flag = G_define_flag();
+    bandref_rm_flag->key = 'b';
+    bandref_rm_flag->guisection = _("Band reference");
+    bandref_rm_flag->description = _("Delete the band reference");
+
     stats_flag = G_define_flag();
     stats_flag->key = 's';
+    stats_flag->guisection = _("Maintenece");
```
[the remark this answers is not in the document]
marisn: Thanks!

lib/raster/raster_metadata.c
marisn: Can't remember my rationale behind the "-1" + no notes => adjusted to your recommendation.

python/grass/temporal/c_libraries_interface.py
marisn: The latest version of black shipped for my distribution (Ubuntu 20.10; black, version 19.10b0) considers my version to be the correct one. This is a more serious issue as we expect contributors to run black on source before commit, but the outcome will depend on the version of black in use. And this is not a case of me running some ancient distribution. Latest Ubuntu LTS ships black 19.3b0 and thus will be around for a long time. @wenzeslaus comments?
I could edit the file manually, but AFAIU avoiding manual formatting of code was the main reason of black'ifying in the first place.

include/grass/defs/raster.h
marisn: Consistency.
https://github.com/OSGeo/grass/blob/019f18d61acf6e94a7afe653b13aa12a95076de6/lib/db/dbmi_base/legal_dbname.c#L31
https://github.com/OSGeo/grass/blob/019f18d61acf6e94a7afe653b13aa12a95076de6/lib/gis/legal_name.c#L34

If you want to see a real history, look into imagery lib as many functions return int just because void functions are not yet available in C at the time of writing...

raster/r.support/main.c
marisn: I replaced with a tab for consistency. Let the grass_indent.sh do the dirty job.

raster/r.support/main.c
marisn: My Latvian language thinking always finds a way into my otherwise flawless and excellent Baltic English language :D

python/grass/temporal/c_libraries_interface.py
wenzeslaus: By default, Black is _using trailing commas as a reason to split lines_ (disabled by `--skip-magic-trailing-comma` which we are currently not using). There was some change between v19 and v20 which causes this loop of satisfying one but not the other, but I didn't investigate further. Black is in beta, so this is expected, but we commit to using Black because it is better than the alternative. Did you try to remove the trailing comma?

What you can do with older systems is to install Black with `pip` into `~/.local` (my current choice) or get a Docker image and set a alias (my previous choice). What we could do in the CI is to apply Black and commit to the PR's branch, but that's much more complicated and potentially confusing for beginners.

nilason: This is a bit of annoyance we have to live with until there is a stable release of Black (which is also widely distributed). The version on CI is the latest and should have precedence. The log files of the CIs are a treasure trove, if you look at *details* on "Python Black Formatting / Check" you will see the diff of Black's suggested changes, the eventual minor deviations of local Black formatting and the CIs may be applied manually.

include/grass/defs/raster.h
nilason: Even though I have a lot yet to learn on GRASS API, I very much aware of the situation with using int for all kind of cases, also for boolean values. Recently the possibility of using `bool` has been introduced to gis.h (see #1567 for background and related discussion). I firmly believe that code comprehension beat consistency in this case. I'm not advocating converting all boolean functions to using bool (although I'd welcome such a change), but for new code or reworked code why not use bool.


Consider:
```
int Rast_legal_bandref(const char *);
...
if (Rast_legal_bandref(bandref) < 0) {}
...
```
vs.
```
bool Rast_legal_bandref(const char *);
...
if (Rast_legal_bandref(bandref) == false) {}
...
```
the latter case is self-explanatory, both the declaration and the use of the function.

marisn: Using bool for Rast_legal_bandref is impossible. I changed the code as you suggested, C part compiles and works just fine, but Python is a dead horse:
```
Error: dist.x86_64-pc-linux-gnu/include/grass/defs/raster.h:551: Syntax error at 'Rast_legal_bandref'
```
Unless you can spot an error in my diff, bool is a no-go:
```diff
diff --git a/include/grass/defs/raster.h b/include/grass/defs/raster.h
index 821a2fc2a..60fd4b575 100644
--- a/include/grass/defs/raster.h
+++ b/include/grass/defs/raster.h
@@ -548,7 +548,7 @@ char *Rast_read_bandref(const char *, const char *);
 void Rast_write_units(const char *, const char *);
 void Rast_write_vdatum(const char *, const char *);
 void Rast_write_bandref(const char *, const char *);
-int Rast_legal_bandref(const char *);
+bool Rast_legal_bandref(const char *);
 
 /* rast_to_img_string.c */
 int Rast_map_to_img_str(char *, int, unsigned char*);
diff --git a/lib/raster/raster_metadata.c b/lib/raster/raster_metadata.c
index c9daed88b..77e8bafa2 100644
--- a/lib/raster/raster_metadata.c
+++ b/lib/raster/raster_metadata.c
@@ -121,7 +121,7 @@ void Rast_write_bandref(const char *name, const char *str)
  * They are in format <shortcut>_<bandname>.
  * Band identifiers are capped in legth to GNAME_MAX.
  *
- * This function will return -1 if provided band id is not considered
+ * This function will return false if provided band id is not considered
  * to be valid.
  * This function does not check if band id maps to any entry in band
  * metadata files as not all band id's have files with extra metadata.
@@ -130,37 +130,37 @@ void Rast_write_bandref(const char *name, const char *str)
  *
  * \param bandref band reference to check
  *
- * \return 1 success
- * \return -1 failure
+ * \return true on success
+ * \return false on failure
  */
-int Rast_legal_bandref(const char *bandref)
+bool Rast_legal_bandref(const char *bandref)
 {
     char **tokens;
     int ntok;
 
     if (strlen(bandref) >= GNAME_MAX) {
         G_warning(_("Band reference is too long"));
-        return -1;
+        return false;
     }
 
     if (G_legal_filename(bandref) != 1)
-        return -1;
+        return false;
 
     tokens = G_tokenize(bandref, "_");
     ntok = G_number_of_tokens(tokens);
     if (ntok < 2) {
         G_warning(_("Band reference must be in form <shortcut>_<bandname>"));
         G_free_tokens(tokens);
-        return -1;
+        return false;
     }
 
     if (strlen(tokens[1]) < 1) {
         G_free_tokens(tokens);
-        return -1;
+        return false;
     }
 
     G_free_tokens(tokens);
-    return 1;
+    return true;
 }
 
 /*!
diff --git a/raster/r.support/main.c b/raster/r.support/main.c
index e1bc8472c..9bae8e5f2 100644
--- a/raster/r.support/main.c
+++ b/raster/r.support/main.c
@@ -285,7 +285,7 @@ int main(int argc, char *argv[])
     }
 
     if (bandref_opt->answer) {
-        if (Rast_legal_bandref(bandref_opt->answer) < 0)
+        if (Rast_legal_bandref(bandref_opt->answer) == false)
             G_fatal_error(_("Provided band reference is not valid. "
                             "See documentation for valid examples"));

```

nilason: Can’t test at the moment, but I’d look into c-python bridge for possible solution.

nilason: It seems the problem is caused by the Ctypesgen version embedded with GRASS source. It may probably be fixed, but that is for another PR. Let's keep the int version for the time being.

nilason: Now I can confirm that the problem lies with embedded ctypesgen. I tried using (modified) generated code using  [ctypesgen](https://github.com/davidjamesca/ctypesgen) version under active development.
Inserting:
```py
for _lib in six.itervalues(_libs):
    if not hasattr(_lib, 'Rast_legal_bandref'):
        continue
    Rast_legal_bandref = _lib.Rast_legal_bandref
    Rast_legal_bandref.argtypes = [String]
    Rast_legal_bandref.restype = c_bool
    break
```
into `dist.XXX/etc/python/grass/lib/raster.py` file and the c bool type works.

I believe the long-term solution would be to update ctypesgen with upstream code, meanwhile I try to fix add bool support to existing code base.

At any rate, I withdraw this suggestion of using bool at this point.

nilason: Reported with #1647.

python/grass/temporal/c_libraries_interface.py
marisn: Fixed with black in a virtualenv.

marisn:
I have no idea why python parallel region test is failing as this PR is not touching anything in that area.

nilason:
> I have no idea why python parallel region test is failing as this PR is not touching anything in that area.

Rebase maybe helps.

marisn:
> Not this time, but in these cases you can also try rerunning the jobs, may help this time and would lead to the correct conclusion about unrelated fluctuation in tests (which of course can work both ways).

I can not reproduce the issue locally on Ubuntu 20.10. Unfortunately this could indicate on a Heisenbug in the code. It might be almost impossible to catch as the code could look file but e.g. something considered (and look like) to be atomic is not atomic at all and thus special timing or system settings would cause code to fail. Unless one of developers manages to reproduce it locally, it will remain with us for a long time (unless it is fixed by the PR #1627).

wenzeslaus:
I don't want to distract from the PR topic too much, so just to conclude, so that someone can pick this up elsewhere:

> > Not this time, but in these cases you can also try rerunning the jobs...

I really meant rerunning the CI jobs. There is a button for it in the top right when you view job details.

> I can not reproduce the issue locally on Ubuntu 20.10.

And I don't expect you would. My guess is it is file system related. I executed that test on my machine many times when working on #638.

> ...it will remain with us for a long time...

I think the big question will be if the specific conditions are something the code or the test should account for.

> ...unless it is fixed by the PR #1627.

I fixed the race condition related to that in #638, so I don't think there is another one related to region, at least not in the area of #1627.

marisn:
Should I (finally) merge this PR?

veroandreo:
> Should I (finally) merge this PR?

all checks passed, +1 for merging so we can more easily test

nilason:
> Should I (finally) merge this PR?

No objections from me. +1

PR #1276, opened 2021-01-27
Fix compiler warnings, part 8

nilason:
Fixes compiler warnings:

- -Wincompatible-pointer-types-discards-qualifiers
- -Wunused-result
- -Wheader-guard
- "too many arguments in call to function"

Eighth part addressing #1247.

Note: there are a few cases of OGR error handling added, please do pay special attention to them and their implementation.

- lib/raster
- lib/imagery
- lib/nviz
- lib/vector/Vlib
- db/drivers/ogr
- general/g.parser
- raster/r.fill.stats
- raster3d/r3.retile
- raster/r.horizon
- raster/r.random
- vector/v.lrs/v.lrs.create
- vector/v.lrs/v.lrs.label
- vector/v.lrs/v.lrs.segment
- vector/v.lrs/v.lrs.where
- vector/v.rectify

### DIFF
--- a/db/drivers/ogr/execute.c
+++ b/db/drivers/ogr/execute.c
@@ -109,7 +109,9 @@ int db__driver_execute_immediate(dbString * sql)
 	    }
 	    OGR_F_SetFieldString(hFeature, cols[i].index, value);
 	}
-	OGR_L_SetFeature(hLayer, hFeature);
+	if (OGR_L_SetFeature(hLayer, hFeature) != OGRERR_NONE)
+	    G_warning(_("\tOGR failed to write feature fid=%ld to layer <%s>"),
+		      OGR_F_GetFID(hFeature), OGR_L_GetName(hLayer));
 	OGR_F_Destroy(hFeature);
     }
     
--- a/lib/nviz/draw.c
+++ b/lib/nviz/draw.c
@@ -118,7 +118,7 @@ int sort_surfs_max(int *surf, int *id_sort, int *indices, int num)
 
    \return 1
  */
-int Nviz_draw_all_vect()
+int Nviz_draw_all_vect(void)
 {
     /* GS_set_cancel(0); */
 
@@ -143,7 +143,7 @@ int Nviz_draw_all_vect()
 
    \return 1
  */
-int Nviz_draw_all_site()
+int Nviz_draw_all_site(void)
 {
     int i;
     int *site_list, nsites;
@@ -172,7 +172,7 @@ int Nviz_draw_all_site()
 
    \return 1
  */
-int Nviz_draw_all_vol()
+int Nviz_draw_all_vol(void)
 {
     int *vol_list, nvols, i;
 
@@ -230,13 +230,13 @@ int Nviz_draw_all(nv_data * data)
 	Nviz_draw_all_surf(data);
 
     if (draw_vect)
-	Nviz_draw_all_vect(data);
+	Nviz_draw_all_vect();
 
     if (draw_site)
-	Nviz_draw_all_site(data);
+	Nviz_draw_all_site();
 
     if (draw_vol)
-	Nviz_draw_all_vol(data);
+	Nviz_draw_all_vol();
 	
     for(i = 0; i < data->num_fringes; i++) {
 	struct fringe_data * f = data->fringe[i];
--- a/lib/raster/put_row.c
+++ b/lib/raster/put_row.c
@@ -514,7 +514,7 @@ static void write_null_bits_compressed(const unsigned char *flags,
     compressed_buf = G_malloc(cmax);
 
     /* compress null bits file with LZ4, see lib/gis/compress.h */
-    nwrite = G_compress(flags, size, compressed_buf, cmax, 3);
+    nwrite = G_compress((unsigned char *)flags, size, compressed_buf, cmax, 3);
 
     if (nwrite > 0 && nwrite < size) {
 	if (write(fcb->null_fd, compressed_buf, nwrite) != nwrite)
--- a/lib/vector/Vlib/open_ogr.c
+++ b/lib/vector/Vlib/open_ogr.c
@@ -100,8 +100,13 @@ int V1_open_old_ogr(struct Map_info *Map, int update)
     G_debug(2, "OGR layer %d opened", layer);
 
     ogr_info->layer = Ogr_layer;
-    if (update && OGR_L_TestCapability(ogr_info->layer, OLCTransactions))
-	OGR_L_StartTransaction(ogr_info->layer);
+    if (update && OGR_L_TestCapability(ogr_info->layer, OLCTransactions) &&
+	(OGR_L_StartTransaction(ogr_info->layer) != OGRERR_NONE)) {
+	OGR_DS_Destroy(Ogr_ds);
+	G_warning(_("OGR transaction with layer <%s> failed to start"),
+		  ogr_info->layer_name);
+	return -1;
+    }
     
     switch(Ogr_geom_type) {
     case wkbPoint25D: case wkbLineString25D: case wkbPolygon25D:
--- a/lib/vector/Vlib/write_ogr.c
+++ b/lib/vector/Vlib/write_ogr.c
@@ -347,8 +347,12 @@ int create_ogr_layer(struct Map_info *Map, int type)
 		      "Unable to write attributes."));
     }
     
-    if (OGR_L_TestCapability(ogr_info->layer, OLCTransactions))
-	OGR_L_StartTransaction(ogr_info->layer);
+    if (OGR_L_TestCapability(ogr_info->layer, OLCTransactions) &&
+	(OGR_L_StartTransaction(ogr_info->layer) != OGRERR_NONE)) {
+	G_warning(_("OGR transaction with layer <%s> failed to start"),
+		  ogr_info->layer_name);
+	return -1;
+    }
 
     return 0;
 }
--- a/raster/r.horizon/main.c
+++ b/raster/r.horizon/main.c
@@ -129,7 +129,7 @@ double TOLER;
 const char *str_step;
 
 int mode;
-int isMode()
+int isMode(void)
 {
     return mode;
 }
@@ -393,7 +393,7 @@ int main(int argc, char *argv[])
 	sscanf(parm.direction->answer, "%lf", &single_direction);
 
 
-    if (isMode(WHOLE_RASTER)) {
+    if (WHOLE_RASTER == isMode()) {
 	if ((parm.direction->answer == NULL) && (parm.step->answer == NULL)) {
 	    G_fatal_error(
 		_("You didn't specify a direction value or step size. Aborting."));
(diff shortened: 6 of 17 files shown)


db/drivers/ogr/execute.c
```diff
@@ -109,7 +109,9 @@ int db__driver_execute_immediate(dbString * sql)
 	    }
 	    OGR_F_SetFieldString(hFeature, cols[i].index, value);
 	}
-	OGR_L_SetFeature(hLayer, hFeature);
+	if (0 != OGR_L_SetFeature(hLayer, hFeature))
```
metzm, line 112:
```suggestion
	if (OGR_L_SetFeature(hLayer, hFeature) != OGRERR_NONE)
```

lib/nviz/draw.c
```diff
@@ -230,13 +230,13 @@ int Nviz_draw_all(nv_data * data)
 	Nviz_draw_all_surf(data);
 
     if (draw_vect)
-	Nviz_draw_all_vect(data);
+	Nviz_draw_all_vect();
```
metzm, line 233: The fn definition `int Nviz_draw_all_vect()` in L121 should be changed to `int Nviz_draw_all_vect(void)`

lib/nviz/draw.c
```diff
@@ -230,13 +230,13 @@ int Nviz_draw_all(nv_data * data)
...
 
     if (draw_site)
-	Nviz_draw_all_site(data);
+	Nviz_draw_all_site();
```
metzm, line 236: The fn definition `int Nviz_draw_all_site()` in L146 should be changed to `int Nviz_draw_all_site(void)`

lib/nviz/draw.c
```diff
@@ -230,13 +230,13 @@ int Nviz_draw_all(nv_data * data)
...
 
     if (draw_vol)
-	Nviz_draw_all_vol(data);
+	Nviz_draw_all_vol();
```
metzm, line 239: The fn definition `int Nviz_draw_all_vol()` in L175 should be changed to `int Nviz_draw_all_vol(void)`

lib/raster/put_row.c
```diff
@@ -514,7 +514,7 @@ static void write_null_bits_compressed(const unsigned char *flags,
     compressed_buf = G_malloc(cmax);
 
     /* compress null bits file with LZ4, see lib/gis/compress.h */
-    nwrite = G_compress(flags, size, compressed_buf, cmax, 3);
+    nwrite = G_compress((char *)flags, size, compressed_buf, cmax, 3);
```
metzm, line 517: This cast from `const unsigned char *` to `char *` fixes compiler warnings but masks the need to synchronize the affected GRASS function definitions. Something for GRASS 8.

lib/vector/Vlib/open_ogr.c
```diff
@@ -100,8 +100,13 @@ int V1_open_old_ogr(struct Map_info *Map, int update)
     G_debug(2, "OGR layer %d opened", layer);
 
     ogr_info->layer = Ogr_layer;
-    if (update && OGR_L_TestCapability(ogr_info->layer, OLCTransactions))
-	OGR_L_StartTransaction(ogr_info->layer);
+    if (update && OGR_L_TestCapability(ogr_info->layer, OLCTransactions) &&
+	(0 != OGR_L_StartTransaction(ogr_info->layer))) {
```
metzm, line 104:
```suggestion
	( OGR_L_StartTransaction(ogr_info->layer) != OGRERR_NONE)) {
```

lib/vector/Vlib/write_ogr.c
```diff
@@ -347,8 +347,12 @@ int create_ogr_layer(struct Map_info *Map, int type)
 		      "Unable to write attributes."));
     }
     
-    if (OGR_L_TestCapability(ogr_info->layer, OLCTransactions))
-	OGR_L_StartTransaction(ogr_info->layer);
+    if (OGR_L_TestCapability(ogr_info->layer, OLCTransactions) &&
+	(0 != OGR_L_StartTransaction(ogr_info->layer))) {
```
metzm, line 351:
```suggestion
	(OGR_L_StartTransaction(ogr_info->layer) != OGRERR_NONE)) {
```

raster/r.horizon/main.c
```diff
@@ -393,7 +393,7 @@ int main(int argc, char *argv[])
 	sscanf(parm.direction->answer, "%lf", &single_direction);
 
 
-    if (isMode(WHOLE_RASTER)) {
+    if (WHOLE_RASTER == isMode()) {
```
metzm, line 396: `int  isMode()` in L132 should be changed to `int  isMode(void)`

vector/v.lrs/v.lrs.create/main.c
```diff
@@ -206,14 +206,14 @@ int main(int argc, char **argv)
     driver_opt->required = NO;
     driver_opt->description = _("Driver name for reference system table");
     driver_opt->options = db_list_drivers();
-    driver_opt->answer = db_get_default_driver_name();
+    driver_opt->answer = (char *) db_get_default_driver_name();
```
metzm, line 209: Maybe `option->answer` in [include/gis.h](https://github.com/OSGeo/grass/blob/master/include/gis.h#L541) can be changed to `const char *`? Together with `option->answers` in  [include/gis.h](https://github.com/OSGeo/grass/blob/master/include/gis.h#L543). Something for GRASS 8.

IMHO, these type casts avoid compiler warnings but do not fix the real problem.

db/drivers/ogr/execute.c
nilason: Good. I actually did consider using `OGRERR_NONE`,  but sided with using example from surrounding code.
Updated.

lib/nviz/draw.c
nilason: updated

lib/nviz/draw.c
nilason: updated

lib/nviz/draw.c
nilason: updated

lib/raster/put_row.c
nilason: That should have been `const unsigned char *` to `unsigned char *` cast, `const` being the "only" difference. Updated accordingly.

I guess the G_compress() family of functions could be updated from:
```
int G_compress(unsigned char *, int, unsigned char *, int, int);
```

to
```
int G_compress(const unsigned char *, int, unsigned char *, int, int);
```
where the source is a const pointer. Is this what you were thinking of?

lib/vector/Vlib/open_ogr.c
nilason: updated

lib/vector/Vlib/write_ogr.c
nilason: updated

raster/r.horizon/main.c
nilason: updated

vector/v.lrs/v.lrs.create/main.c
nilason: I don't know the reason Option->answer is not a const string. But I'm afraid changing that wouldn't be a stroll in the park :)  These casts are all over. A alternative now, for this `db_get_default_driver_name()` case, could be to use G_store() (with the memory leak that comes with it).

nilason: Just out of curiosity I tried compile with Option->answer defined as `const char *answer`.
This resulted in some 660 warnings of type `-Wincompatible-pointer-types-discards-qualifiers`.

db/drivers/ogr/execute.c
metzm: I recommend following the OGR API for [OGR_L_SetFeature()](https://gdal.org/doxygen/ogr__api_8h.html#a5b380520d0999422987f06b6e31b9b00)

lib/nviz/draw.c
metzm: What about the suggested change?

lib/nviz/draw.c
metzm: What about the suggested change?

lib/nviz/draw.c
metzm: What about the suggested change?

lib/raster/put_row.c
metzm: Some fns of external libraries used by the `G_compress()` family of fns want `(char *)` as input, e.g. `BZ2_bzBuffToBuffCompress()`. According casts are done in the respective `lib/gis/cmpr*.c` files.

Your updated change makes sense.

lib/nviz/draw.c
metzm: Apparently resolved.

lib/nviz/draw.c
metzm: Apparently resolved.

lib/nviz/draw.c
metzm: Apparently resolved.

vector/v.lrs/v.lrs.create/main.c
metzm: OK for your change, but still, type casts avoid compiler warnings but do not fix the real problem.

nilason:
Updated commit messages and I believe it is ready for merge.

I did split the original "-Wincompatible-pointer-types-discards-qualifiers" fix commit into two, depending on the cause of the warning (and nature of fix).

A note regarding the cast of the result of `db_get_default_driver_name()`. The function is defined as: 
```c
const char *db_get_default_driver_name(void)
```
but in fact returns a `char *`:

https://github.com/OSGeo/grass/blob/51fe070ae52ba63ab789bf30adf5808ce80dd5e9/lib/db/dbmi_base/default_name.c#L32

which is here then re-casted back to `char *`.

metzm:
> A note regarding the cast of the result of `db_get_default_driver_name()`. The function is defined as:
> 
> ```c
> const char *db_get_default_driver_name(void)
> ```
> 
> but in fact returns a `char *`:
> 
> https://github.com/OSGeo/grass/blob/51fe070ae52ba63ab789bf30adf5808ce80dd5e9/lib/db/dbmi_base/default_name.c#L32
> 
> which is here then re-casted back to `char *`.

IMHO, `db_get_default_driver_name()` is correct, it provides the default driver name which should not be modified, thus `const`. Within the scope of this PR, I think re-casting back to `char *` is fine.

metzm (approved):
Ready to merge, thanks a lot!

nilason:
> IMHO, db_get_default_driver_name() is correct, it provides the default driver name which should not be modified, thus const. Within the scope of this PR, I think re-casting back to char * is fine.

Oh, I do understand the logic that db_get_default_driver_name() is a const and agree it's correct. Only brought it up because I suggested the return value could be wrapped in a G_store() instead of casting, but that wouldn't make much sense in light of its implementation.

nilason:
Cherry picked and pushed to master.
Thank you!

nilason:
@metzm 
Regarding the structure member Option->answer and the casting problems it seem to cause. I noticed that originally (at least with G5) all (string) members were `char *`, but this [changed in 2008](https://trac.osgeo.org/grass/changeset/31906) (G6.4) to what we see today. Only `answer` (and `answers`) stayed mutable. The question is: do you think it's worth the effort to even explore the possibility to change ´answer´ to a const? Or will that set me off in a one-way road?

PR #1291, opened 2021-02-01
r3.out.vtk: replace 'void *' with 'RASTER3D_Map *' for clarity

nilason:
This replaces anonymous `void *` pointer to typed `RASTER3D_Map *` in `r3.out.vtk`.


metzm (approved):
Much better, thanks!

nilason:
> Much better, thanks!

I agree and will update #1256 accordingly.
Thanks for checking this out!

PR #1188, opened 2020-12-18
GitHub Action: Activate security analysis with CodeQL

neteler:
This PR adds Code-QL which is a tool for

```
Code scanning alerts
Automatically detect common vulnerability and coding errors 
```

It will run for master branch and protected branches and for MRs into the master branch. 
For more details see https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/enabling-code-scanning-for-a-repository

Developed with support by @mmacata (https://github.com/mundialis/actinia_core/pull/129)

PR #1300, opened 2021-02-03
GitHub Action: Completion of security analysis with CodeQL

mmacata:
See closed PR ["Activation security analysis with CodeQL"](https://github.com/OSGeo/grass/pull/1188):

> This PR adds Code-QL which is a tool for
> 
> Code scanning alerts
> Automatically detect common vulnerability and coding errors 
> 
> It will run for master branch and protected branches and for MRs into the master branch.
> For more details see https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/enabling-code-scanning-for-a-repository
> 

Output can be seen in [fork](https://github.com/mmacata/grass/security/code-scanning?query=tool%3ACodeQL+ref%3Arefs%2Fheads%2Fcodeql). To generate it, `codeql` branch was also activated in codeql-analysis.yml but now removed.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -40,10 +40,6 @@ jobs:
       run: |
         mkdir $HOME/install
 
-    - name: Ensure one core for compilation
-      run: |
-        echo "MAKEFLAGS=-j1" >> $GITHUB_ENV
-
     - name: Set LD_LIBRARY_PATH for compilation
       run: |
         echo "LD_LIBRARY_PATH=$HOME/install/lib" >> $GITHUB_ENV
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   analyze:
-    name: ${{ matrix.language }} - Analyze
+    name: ${{ matrix.language }}
     runs-on: ubuntu-20.04
 
     strategy:
@@ -35,15 +35,19 @@ jobs:
         sudo apt-get install -y wget git gawk findutils
         xargs -a <(awk '! /^ *(#|$)/' ".github/workflows/apt.txt") -r -- \
             sudo apt-get install -y --no-install-recommends --no-install-suggests
+
     - name: Create installation directory
       run: |
         mkdir $HOME/install
+
     - name: Ensure one core for compilation
       run: |
         echo "MAKEFLAGS=-j1" >> $GITHUB_ENV
+
     - name: Set LD_LIBRARY_PATH for compilation
       run: |
         echo "LD_LIBRARY_PATH=$HOME/install/lib" >> $GITHUB_ENV
+
     - name: Build
       env:
         CFLAGS: "-std=gnu11"
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,7 +2,7 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master,  releasebranch_7_8, codeql]
+    branches: [ master,  releasebranch_7_8]
   pull_request:
     branches: [ master ]
   schedule:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -6,17 +6,31 @@ on:
   pull_request:
     branches: [ master ]
   schedule:
+    # Check every Saturday at 18:36
+    #         ┌───────────── minute (0 - 59)
+    #         │ ┌───────────── hour (0 - 23)
+    #         │ │ ┌───────────── day of the month (1 - 31)
+    #         │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+    #         │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+    #         │ │ │ │ │
+    #         │ │ │ │ │
+    #         │ │ │ │ │
+    #         * * * * *
     - cron: '36 18 * * 6'
 
 jobs:
   analyze:
-    name: ${{ matrix.c }} & ${{ matrix.cpp }} - Analyze
+    name: ${{ matrix.c }} & ${{ matrix.cpp }} - ${{ matrix.language }}
     runs-on: ubuntu-20.04
 
     strategy:
       fail-fast: false
       matrix:
+        # c is included in cpp.
         language: [ 'cpp', 'python' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
+        # Learn more:
+        # https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
         c:
         - gnu99
         - gnu11
@@ -52,9 +66,7 @@ jobs:
         echo "LD_LIBRARY_PATH=$HOME/install/lib" >> $GITHUB_ENV
     - name: Build
       env:
-        # TODO: -pedantic-errors here won't go through ./configure (with GNU C)
         CFLAGS: "-std=${{ matrix.c }}"
-        # TODO: -pedantic-errors here won't compile
         CXXFLAGS: "-std=${{ matrix.cpp }}"
       run: .github/workflows/build.sh $HOME/install
 
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -7,38 +7,18 @@ on:
     branches: [ master ]
   schedule:
     # Check every Saturday at 18:36
-    #         ┌───────────── minute (0 - 59)
-    #         │ ┌───────────── hour (0 - 23)
-    #         │ │ ┌───────────── day of the month (1 - 31)
-    #         │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
-    #         │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
-    #         │ │ │ │ │
-    #         │ │ │ │ │
-    #         │ │ │ │ │
-    #         * * * * *
     - cron: '36 18 * * 6'
 
 jobs:
   analyze:
-    name: ${{ matrix.c }} & ${{ matrix.cpp }} - ${{ matrix.language }}
+    name: ${{ matrix.language }} - Analyze
     runs-on: ubuntu-20.04
 
     strategy:
       fail-fast: false
       matrix:
-        # c is included in cpp.
+        # C is included in cpp, no separate C language available on CodeQL
         language: [ 'cpp', 'python' ]
-        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
-        # Learn more:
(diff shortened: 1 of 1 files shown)


neteler:
... maybe of interest with respect to the latest compiler fixes by @nilason and others.

wenzeslaus (changes requested):
It actually works!

I get the 404 Jedi message for the fork. I assume it is visible to repo owners only since it is security. Here is the link for [this PR/repo](https://github.com/OSGeo/grass/security/code-scanning?query=ref%3Arefs%2Fpull%2F1300%2Fmerge+tool%3ACodeQL) (again, you will need to be part of one of the teams to see that).

Most of the 244 messages is "Multiplication result converted to larger type". Do you know if it handles duplicates from different runs? Couple of the other I checked made sense to me.

I didn't see any Python, does it work?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,62 @@
...
+      matrix:
+        language: [ 'cpp', 'python' ]
+        c:
+        - gnu99
+        - gnu11
+        - gnu17
+        cpp:
+        - c++11
+        - c++14
+        - c++17
```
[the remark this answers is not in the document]
mmacata: I am not sure and wanted to test all possibilities. Unfortunately, the Code Scanning alert doesn't show the matrix that was used for creation, so it is not possible to see which combinations led to which alert. But no duplicates are shown, so I guess the job recognizes when an alert already exists.
I would like to reduce to one but am no expert in this. Which one would you recommend? gnu17 + c++17 ? Else I will change to #1288 approach.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,62 @@
...
+        echo "MAKEFLAGS=-j1" >> $GITHUB_ENV
+    - name: Set LD_LIBRARY_PATH for compilation
+      run: |
+        echo "LD_LIBRARY_PATH=$HOME/install/lib" >> $GITHUB_ENV
+    - name: Build
+      env:
+        # TODO: -pedantic-errors here won't go through ./configure (with GNU C)
+        CFLAGS: "-std=${{ matrix.c }}"
+        # TODO: -pedantic-errors here won't compile
+        CXXFLAGS: "-std=${{ matrix.cpp }}"
```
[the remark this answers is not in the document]
mmacata: I wanted to be close to the `gcc.yml`. Do you mean this by "GCC test"? I will remove the comments if they are misleading in this context.

mmacata:
> I get the 404 Jedi message for the fork. I assume it is visible to repo owners only since it is security. Here is the link for this PR/repo (again, you will need to be part of one of the teams to see that).

I cannot see it for this repo and I didn't know it was triggered already, but even the better. Then it is not bad that you cannot see the security alerts for my fork as I assume they are identical.

> Most of the 244 messages is "Multiplication result converted to larger type". Do you know if it handles duplicates from different runs? Couple of the other I checked made sense to me.

As I did already 2 commits to this branch, I assume no duplicates will be created. Also experiences with other repos show no duplicates.

> I didn't see any Python, does it work?

First 2 alerts are for python code.

.github/workflows/codeql-analysis.yml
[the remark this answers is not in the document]
nilason: There is a discussion at [dev-ml](https://lists.osgeo.org/pipermail/grass-dev/2021-January/094899.html) regarding min. standard for G8. No decision has been made yet, but I personally would say now gnu11/c++11 would be sufficient for the time being. This can always be updated. Also no need to overstrain Github's resources with 10s of extra builds.

nilason:
This is great. Neither did I see the first fork, but now with the one @wenzeslaus provided I could take a peek. I see a lot of warnings I also see with clang (default mode), but some also new.

.github/workflows/codeql-analysis.yml
```diff
@@ -6,17 +6,31 @@ on:
...
     runs-on: ubuntu-20.04
 
     strategy:
       fail-fast: false
       matrix:
+        # c is included in cpp.
         language: [ 'cpp', 'python' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
+        # Learn more:
+        # https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
```
wenzeslaus, line 33: This is not relevant. `# c is included in cpp.` strikes the right balance between too little information and unnecessary information. You could add that there is no separate C language on CodeQL. That would clearly document of what what you considered when you wrote `[ 'cpp', 'python' ]`.

.github/workflows/codeql-analysis.yml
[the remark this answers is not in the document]
wenzeslaus: I agree with gnu11/c++11. How relevant the versions are to the CodeQL test? Your comment seems to imply that they influence something, @mmacata. Is there something in the doc about that? If they are not relevant, you could remove them from the matrix and just hardcode them to the `env: CFLAGS...` part. Even if you keep them in the matrix, there is only job based on these (gnu11 & c++11), so no need to include them in the job name (see how the items in Checks look like).

wenzeslaus (changes requested):

.github/workflows/codeql-analysis.yml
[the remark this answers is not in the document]
wenzeslaus: Yes, right, `gcc.yml`. That tests different C and C++ versions/standards supported by GCC, so it does not seem to be that much relevant here (see my other comment), but setting some standard explicitly here makes sense.

wenzeslaus:
> > I didn't see any Python, does it work?
> 
> First 2 alerts are for python code.

They must be sorted differently, but I found those. However, they are the same:

```
Insecure temporary file      dist.x86_64-pc-linux-gnu/etc/python/... Detected yesterday
Insecure temporary file      lib/python/...                          Detected on Dec 18, 2020
```

I'm not sure about the `Detected on Dec 18` piece, but it is still there, so I assume it is still re-created by the current checks (it says *First appeared in commit 54cee7c on Dec 18, 2020*). So, some ignoring or de-duplication would be nice, although does not seem necessary since there is only one error.

.github/workflows/codeql-analysis.yml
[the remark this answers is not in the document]
mmacata: I don't know if they are relevant at all. Unfortunately I cannot find any documentation about it. I added them because I thought there might be an indication in the security alerts but as there is none, I will reduce to suggested gnu11/c++11.

.github/workflows/codeql-analysis.yml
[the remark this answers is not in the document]
mmacata: Ok I removed the comments. Change of matrix is related to other discussion, so I will resolve this one.

mmacata:
>    First 2 alerts are for python code.

>> They must be sorted differently, but I found those. However, they are the same:

It might be possible that CodeQL does not recognize them as the same because the paths looks different.

>> I'm not sure about the Detected on Dec 18 piece, but it is still there, so I assume it is still re-created by the current checks (it says First appeared in commit 54cee7c on Dec 18, 2020). So, some ignoring or de-duplication would be nice, although does not seem necessary since there is only one error.

This was likely triggered by #1188 as creation date of it is also Dec 18, although I cannot see the exact commit related to it. The duplication might have appeared only because of different file paths - as I mentioned, in other repositories (python) we don't have any problems with duplications. I can also image that the "autobuild approach" which was tried first led to the different file paths compared to the current approach.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,54 @@
...
+      with:
+        languages: ${{ matrix.language }}
+
+    - name: Get dependencies
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y wget git gawk findutils
+        xargs -a <(awk '! /^ *(#|$)/' ".github/workflows/apt.txt") -r -- \
+            sudo apt-get install -y --no-install-recommends --no-install-suggests
+    - name: Create installation directory
```
wenzeslaus, line 38: Use either an empty line between steps or no empty line. Not both styles. Internal consistency is probably more important than convenience of some theoretical Meld-like operation in the future.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,54 @@
...
+      run: |
+        mkdir $HOME/install
+    - name: Ensure one core for compilation
+      run: |
+        echo "MAKEFLAGS=-j1" >> $GITHUB_ENV
```
wenzeslaus, line 43: I think reason for this in gcc.yml was just that compilation errors are not mixed. This does not apply here. Using all cores is preferred because of speed, so use nproc like in, e.g., flake8.yml or check what is the best way to do this in GH Actions in combination with make.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,54 @@
+name: "CodeQL"
```
wenzeslaus, line 1: `name: CodeQL` syntax should be sufficient or is there some YAML camel case special case?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,54 @@
...
+    branches: [ master,  releasebranch_7_8]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    # Check every Saturday at 18:36
+    - cron: '36 18 * * 6'
+
+jobs:
+  analyze:
+    name: ${{ matrix.language }} - Analyze
```
wenzeslaus, line 14: I don't see reason for " - Analyze" here. Isn't that just a clutter since there is nothing else than analyze? It doesn't seem to add much info to CodeQL if you don't already know what that is. _CodeQL / cpp_ in checks seems quite sufficient to me, but feel free to modify. For flake.yml, I just rewrote it as _Python Flake8 Code Quality_ / {directory}.

wenzeslaus (changes requested):
Just some smaller issues.

.github/workflows/codeql-analysis.yml
mmacata: The mix was done intentionally because this way the GRASS GIS specific part appears as one block, separating it from the CodeQL specific steps. I agree on Meld-like convenience, so I will update it.

.github/workflows/codeql-analysis.yml
mmacata: As this was auto-generated by GitHub [during the initialisation](https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/setting-up-code-scanning-for-a-repository), I would like to keep it this way as it is the suggested way.

.github/workflows/codeql-analysis.yml
mmacata: Same here - `name: Analyze` appears as default name during auto-generation. For me either way is fine. Do you mean `python / cpp` above?

nilason (approved):
Looks good to me!

I tried to come up with a solution to name the jobs like:

Screenshot 2021-02-09 at 17 43 07

but I failed at that.

Would have been nice if there is a way...

mmacata:
@wenzeslaus do you see any more issues? Or can this be merged?

wenzeslaus (approved):

PR #1305, opened 2021-02-04
r.external speed-up

metzm:
Use `GDALComputeRasterMinMax()` for faster range computation.
Add super-fast mode without range computation, and a warning for this mode.

`r.info` will not report the range if not existing. If the range is explicitly requested with -r or -s, the range and stats are computed on-the-fly if not existing.

`r.support -s` is now computing the correct range for FCELL /DCELL maps.

Tested with a tiled GeoTIFF with 7 billion cells, linking with `r.external` drops from 2:06 minutes to 46 seconds. In super-fast mode, linking takes only 0.05 seconds.

Regarding the super-fast mode without data range computation, a number of modules will not work correctly. If the range is needed, it can be computed for the linked map with `r.support`, but linking again with `r.external` would be faster, at least for large, tiled maps.

### DIFF
--- a/lib/raster/range.c
+++ b/lib/raster/range.c
@@ -122,7 +122,7 @@ int Rast_read_fp_range(const char *name, const char *mapset,
     }
     else {
 	/* "f_range" file does not exist */
-	G_warning(_("Missing fp range file for <%s>"),
+	G_warning(_("Missing fp range file for <%s> (run r.support)"),
 		  G_fully_qualified_name(name, mapset));
 	return -1;
     }
--- a/lib/raster/range.c
+++ b/lib/raster/range.c
@@ -122,7 +122,7 @@ int Rast_read_fp_range(const char *name, const char *mapset,
     }
     else {
 	/* "f_range" file does not exist */
-	G_warning(_("Missing fp range file for <%s> (run r.support)"),
+	G_warning(_("Missing fp range file for <%s> (run r.support -s)"),
 		  G_fully_qualified_name(name, mapset));
 	return -1;
     }
@@ -245,7 +245,7 @@ int Rast_read_range(const char *name, const char *mapset, struct Range *range)
     }
     else {
 	/* "range" file does not exist */
-	G_warning(_("Missing range file for <%s> (run r.support)"),
+	G_warning(_("Missing range file for <%s> (run r.support -s)"),
 		  G_fully_qualified_name(name, mapset));
 	return -1;
     }
--- a/lib/raster/range.c
+++ b/lib/raster/range.c
@@ -120,6 +120,12 @@ int Rast_read_fp_range(const char *name, const char *mapset,
 	Rast_update_fp_range(dcell2, drange);
 	close(fd);
     }
+    else {
+	/* "f_range" file does not exist */
+	G_warning(_("Missing fp range file for <%s>"),
+		  G_fully_qualified_name(name, mapset));
+	return -1;
+    }
 
     return 1;
 }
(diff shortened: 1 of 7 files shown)


lib/raster/range.c
```diff
@@ -120,6 +120,12 @@ int Rast_read_fp_range(const char *name, const char *mapset,
 	Rast_update_fp_range(dcell2, drange);
 	close(fd);
     }
+    else {
+	/* "f_range" file does not exist */
+	G_warning(_("Missing fp range file for <%s>"),
```
neteler, line 125:
```suggestion
	G_warning(_("Missing fp range file for <%s> (run r.support)"),
```

lib/raster/range.c
```diff
@@ -237,6 +243,12 @@ int Rast_read_range(const char *name, const char *mapset, struct Range *range)
 	}
 	fclose(fd);
     }
+    else {
+	/* "range" file does not exist */
+	G_warning(_("Missing range file for <%s>"),
```
neteler, line 248:
```suggestion
	G_warning(_("Missing range file for <%s> (run r.support)"),
```

metzm:
A case for the super-fast mode is raster reprojection with `r.proj`. That means `r.import` can in most cases be faster, if reprojection is needed, if `r.in.gdal` is replaced with `r.external -r` in [r.import](https://github.com/OSGeo/grass/blob/master/scripts/r.import/r.import.py#L226). However, there must be a reason why `r.import` uses `r.in.gdal` instead of `r.external` when the input raster needs to be reprojected. IIRC, there were some problems with `r.external` that did not occur with `r.in.gdal`.

neteler:
> However, there must be a reason why `r.import` uses `r.in.gdal` instead of `r.external` when the input raster needs to be reprojected. IIRC, there were some problems with `r.external` that did not occur with `r.in.gdal`.

I traced the snippet back to GRASS GIS addons, r.in.proj (commit from Jan 27, 2015):
https://github.com/OSGeo/grass-addons/commit/59b69764c890b6b7513817e12d0b23f4073c0bf7#diff-d06b709384f210f1f01d068364921d06ba692c21040323997220dad9e1d9d0beR182

Already back then it was based on `r.in.gdal`.

(For those here not knowing the story: the `r.in.proj` addon was in https://github.com/OSGeo/grass-addons/commit/d26fd4dd2cd219b9eacd5cb08b354cc0fa8d5ca7 renamed to `r.import` and later moved to GRASS-core)

metzm:
> > However, there must be a reason why `r.import` uses `r.in.gdal` instead of `r.external` when the input raster needs to be reprojected. IIRC, there were some problems with `r.external` that did not occur with `r.in.gdal`.
> 
> I traced the snippet back to GRASS GIS addons, r.in.proj (commit from Jan 27, 2015):
> [OSGeo/grass-addons@59b6976#diff-d06b709384f210f1f01d068364921d06ba692c21040323997220dad9e1d9d0beR182](https://github.com/OSGeo/grass-addons/commit/59b69764c890b6b7513817e12d0b23f4073c0bf7#diff-d06b709384f210f1f01d068364921d06ba692c21040323997220dad9e1d9d0beR182)
> 
> Already back then it was based on `r.in.gdal`.
> 
`r.import` uses some options and flags with `r.in.gdal` that are not available in `r.external`, thus in `r.import`, `r.in.gdal` can not be simply replaced with `r.external` when it comes to the actual import of the data. Worth a separate issue/PR.

neteler:
Due to the containing bugfix I suggest to backport this after a while

PR #1316, opened 2021-02-09
Fix compiler warnings, part 9

nilason:
Use platform independent int64 format specifier, with the `PRId64` macro from <inttypes.h> available with c99.

Lifted out of PR #1256, awaiting community decision on C standard usage within the GRASS GIS project, see [discussion on grass-dev mail list](https://lists.osgeo.org/pipermail/grass-dev/2021-January/094899.html). 

Addresses `-Wformat` compiler warnings.

Ninth part addressing #1247.

### DIFF
--- a/lib/vector/Vlib/build_nat.c
+++ b/lib/vector/Vlib/build_nat.c
@@ -16,6 +16,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <sys/types.h>
+#include <inttypes.h>
 #include <grass/glocale.h>
 #include <grass/vector.h>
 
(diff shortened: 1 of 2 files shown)


metzm (approved):
Fine with me.

I am not sure if the community understands that accepting C99 or something higher as C standard does not mean we need to change the remaining code base.

C code written for an older C standard works fine with newer C standards.

This is different with Python: if we require a certain Python version, code written for a previous Python version might no longer work with a newer Python version.

nilason:
> I am not sure if the community understands that accepting C99 or something higher as C standard does not mean we need to change the remaining code base.
> 
> C code written for an older C standard works fine with newer C standards.
> 
> This is different with Python: if we require a certain Python version, code written for a previous Python version might no longer work with a newer Python version.

Indeed.

Re: `<inttypes.h>`: it is included with the (only partly and still so) [C99 support of Visual Studio 2013](https://devblogs.microsoft.com/cppblog/c99-library-support-in-visual-studio-2013/). This being the platform historically least supportive of C standards, I'm quite convinced this should be safe in 2021.

metzm (approved):
OK, let's go for it!

nilason:
> OK, let's go for it!

I agree, merged. On the off chance this causes trouble, it is easy to revert.

PR #1321, opened 2021-02-11
wxGUI: Make Display related toolbars part of Display tab

lindakladivova:
1. The toolbar Data Toolbar and Vector Toolbar only for Display tab were created in that order.
- avoids the new issue created by Data tab re-using the "pencil" icon to allow editing of other mapsets (now user sees two of these icons above each other which is strange)
- avoids the confusion when Data tab is active and a user clicks the Show attribute... button which is associated with selected layer in Displays, not selected vector map in Data tab.

2. We can have NewDisplay button in new Data Toolbar.
- main toolbar will be then a bit shorter than Display toolbar.

### DIFF
--- a/gui/wxpython/lmgr/frame.py
+++ b/gui/wxpython/lmgr/frame.py
@@ -298,12 +298,12 @@ class GMFrame(wx.Frame):
     def _createDisplayPanel(self):
         """Creates display panel"""
         # create superior display panel
-        display_manager = wx.Panel(self.notebook, id=wx.ID_ANY)
+        display_panel = wx.Panel(self.notebook, id=wx.ID_ANY)
         # create display toolbar
-        dmgrToolbar = DisplayPanelToolbar(guiparent=display_manager,
+        dmgrToolbar = DisplayPanelToolbar(guiparent=display_panel,
                                           parent=self)
         # create display notebook
-        notebookLayers = GNotebook(parent=display_manager,
+        notebookLayers = GNotebook(parent=display_panel,
                                    style=globalvar.FNPageStyle)
         notebookLayers.SetTabAreaColour(globalvar.FNPageColor)
         menu = self._createTabMenu()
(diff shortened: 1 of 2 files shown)


wenzeslaus:
I think it would still be worth it to do this change without doing any refactoring, i.e., minimizing the changes in the related code to a bare minimum.

lindakladivova:
So, now it looks as follows. :-) :

Data Tab:
novatoolbar=data

Display tab:
nova_toolbar-displaytab

I have tried to make it simple. It is not an ideal solution, I still think we should separate display things from general ones in future.

gui/wxpython/lmgr/frame.py
```diff
@@ -304,6 +295,32 @@ def _createTabMenu(self):
...
+    def _createDisplayPanel(self):
+        """Creates display panel"""
+        # create superior display panel
+        display_panel = wx.Panel(self.notebook, id=wx.ID_ANY)
+        # create display toolbar
+        dmgrToolbar = DisplayPanelToolbar(guiparent=display_panel,
+                                          parent=self)
+        # create display notebook
+        notebookLayers = GNotebook(parent=display_panel,
+                                   style=globalvar.FNPageStyle)
```
wenzeslaus, line 307: Since you are changing this, can you use the Black formatting here (dmgrToolbar and notebookLayers)? It will result in this line staing the same after Black is applied globally possibly making navigating history easier. (I was going to suggest the same for DisplayPanelToolbar, but it seems it is actually just misplaced old code.)

gui/wxpython/lmgr/toolbars.py
```diff
@@ -68,71 +62,6 @@ def _toolbarData(self):
                                      ))
 
 
-class LMDataToolbar(BaseToolbar):
```
wenzeslaus, line 71: Unnecessary large change in the code layout. DisplayPanelToolbar is more or less LMDataToolbar with addition of stuff from LMVectorToolbar, so why new class on a new place. Just put DisplayPanelToolbar in the same place as LMDataToolbar.

gui/wxpython/lmgr/frame.py
```diff
@@ -337,13 +354,9 @@ def _createNoteBook(self):
...
         # create displays notebook widget
-        self.notebookLayers = GNotebook(parent=self.notebook,
-                                        style=globalvar.FNPageStyle)
-        self.notebookLayers.SetTabAreaColour(globalvar.FNPageColor)
-        menu = self._createTabMenu()
-        self.notebookLayers.SetRightClickMenu(menu)
+        self.display_panel, self.notebookLayers = self._createDisplayPanel()
         self.notebook.AddPage(
-            page=self.notebookLayers,
+            page=self.display_panel,
```
wenzeslaus, line 359: Please just double check that this is the only place where this replacement is needed.

wenzeslaus (changes requested):
This looks great and it works for me. (I have a problem with rendering when a group is added, but it seems unrelated. Can someone confirm?)

I have only some minor comments.

gui/wxpython/lmgr/frame.py
lindakladivova: I think it is all right. display_panel is just a widget "wrapper".

wenzeslaus (approved):

wenzeslaus:
For me, this makes the GUI much easier to navigate. Thanks!

Issue #1257, opened 2021-01-17
v.rast.stats: warning instead of an error when no categories are found in raster map


pesekon2:
**Describe the issue**

When you run `v.rast.stats` on a vector that is not in your computational region and does not have an overlap with the queried raster, it raises an error saying that `No categories found in raster map`. However, the behaviour of other modules is not to do so and to finish (e.g., `v.what.rast`). I would expect the same, systematic behaviour, i.e. to populate the columns with `NULL` values.

However, as the `No categories found in raster map` message is definitely an interesting information, I propose to change the error to a warning. What do you think?

**To Reproduce**

In:

```
g.region n=1 s=0 w=0 e=1 nsres=1 ewres=1
r.random.surface out=rast
g.region n=2 s=1 w=0 e=1 nsres=1 ewres=1
v.random out=vect npoints=1
v.db.addtable vect
g.region n=1 s=0 w=0 e=1 nsres=1 ewres=1
v.rast.stats  map=vect raster=rast column_prefix=rast
```

Out:

```
Generating raster map <rast>...
 100%
Writing raster map <rast>...
 100%
r.random.surface complete.
Generating points...
 100%
Building topology for vector map <vect@PERMANENT>...
Registering primitives...
WARNING: Values in column <cat> will be overwritten
Reading features...
 100%
Updating database...
 100%
1 categories read from vector map (layer 1)
1 categories read from vector map don't exist in selection from table
1 records updated/inserted (layer 1)
WARNING: No areas selected from vector map <vect>
ERROR: No categories found in raster map
```

An example with `v.what.rast`:

In: 

```
g.region n=1 s=0 w=0 e=1 nsres=1 ewres=1
r.random.surface out=rast
g.region n=2 s=1 w=0 e=1 nsres=1 ewres=1
v.random out=vect npoints=1
v.db.addtable vect
g.region n=1 s=0 w=0 e=1 nsres=1 ewres=1
v.what.rast  map=vect raster=rast column=rast
```

Out:

```
r.random.surface out=rast
g.region n=2 s=1 w=0 e=1 nsres=1 ewres=1
v.random out=vect npoints=1
v.db.addtable vect
g.region n=1 s=0 w=0 e=1 nsres=1 ewres=1
v.what.rast  map=vect raster=rast column=rast
Generating raster map <rast>...
 100%
Writing raster map <rast>...
 100%
r.random.surface complete.
Generating points...
 100%
Building topology for vector map <vect@PERMANENT>...
[...]

PR #1323, opened 2021-02-12
v.what.rast: no features message wording, switched to be a warning

landam:
From #1257 :

```
g.region n=1 s=0 w=0 e=1 nsres=1 ewres=1
r.random.surface out=rast
g.region n=2 s=1 w=0 e=1 nsres=1 ewres=1
v.random out=vect npoints=1
v.db.addtable vect
g.region n=1 s=0 w=0 e=1 nsres=1 ewres=1
v.what.rast  map=vect raster=rast column=rast
...
Reading features from vector map...
No features of type (point) found in vector map <vect>
```

This PR improves two issues:

* message "No features..." switched to be a warning (to be consistent with `v.what.rast`)
* message clarified (*)

(*)

Input vector map can contain features, as in example above.

```
v.info -t vect
nodes=0
points=1
...
```

Now the module ends up with:

```
v.what.rast  map=vect raster=rast column=rast
Reading features from vector map...
WARNING: No features of type (point) found in the current computational
         region
```

### DIFF
--- a/vector/v.what.rast/main.c
+++ b/vector/v.what.rast/main.c
@@ -271,8 +271,8 @@ int main(int argc, char *argv[])
     }
 
     if (point_cnt < 1) {
-        G_important_message(_("No features of type (%s) found in vector map <%s>"),
-                            opt.type->answer, opt.vect->answer);
+        G_warning(_("No features of type (%s) found in the current computational region"),
+                  opt.type->answer);
         exit(EXIT_SUCCESS);
     }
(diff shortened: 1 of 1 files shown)


pesekon2 (approved):

neteler:
@landam Is that backport relevant?

pesekon2:
As it is more of an enhancement and slight change of original behaviour and not a bugfix, I would not recommend to backport it.

PR #1326, opened 2021-02-12
i18n: sync from Transifex

HuidaeCho:


neteler (approved):

neteler:
@HuidaeCho let me suggest to also update G78, but as a separate PR instead of a backport. I can do that if you prefer

HuidaeCho:
> @HuidaeCho let me suggest to also update G78, but as a separate PR instead of a backport. I can do that if you prefer

@neteler Is there a reason for a separate PR? I guess conflicts?

neteler:
We may give it a try but the message "pools" in the .po files differ a bit, that's why I went for separate syncs in the past.

HuidaeCho:
> 
> 
> We may give it a try but the message "pools" in the .po files differ a bit, that's why I went for separate syncs in the past.

Makes sense. But isn't the source Transifex the same? Let me try it first. Working on it now...

neteler:
> Makes sense. But isn't the source Transifex the same? Let me try it first. Working on it now...

Yes, the source Transifex is the same but perhaps (just guessing) the `gettext` merging part ends up differently.

PR #1328, opened 2021-02-13
v.import: Add layer to the projection check

wenzeslaus:
Layers in some formats (e.g., GeoPackage) can have different projections, so a test with v.in.ogr
without layer may report a difference although the user specified layer with matching projection for v.import.


metzm:
This is a bug fix, +1 from me for merging and backporting.

marisn (approved):
Seems good to me. As code change is trivial, +1 for backport.

wenzeslaus:
Given the reviews, I backported right away.

For the record, here is how the backport worked out: I used the [standard procedure](https://trac.osgeo.org/grass/wiki/HowToGit#Backportingtoreleasebranches). The `git cherry-pick` reported conflict. The conflict was lines which were modified by Black in master branch (prior to the cherry-picked commit). I simply resolved it by accepting my changes (the new code) and rejecting the old code since there is nothing wrong with having some of the Black modifications leak to v78 with the new code. `git diff` afterwards reports reversion of the Black changes (code from master branch is `-` and v78 code is `+`) - this is not that confusing [if you think about commits](https://m.xkcd.com/1597/) as trees. After `git add` of the file, `git diff --staged` shows exactly what you want, so it is time to do `git commit` and copy-paste the commit message from the original commit. You can test this in some other branch first which is what I did. Obviously, you use `git status` a lot, too. It is not clear to me how it works with authorship here, so that's something to explore before someone applies these steps.

PR #1333, opened 2021-02-14
wxGUI/dbmgr: fix insert record for each category into table

tmszi:
**To Reproduce**

1. Create some new vector map without DB table e.g. `v.random output=test npoints=5`
2. Launch Attribute Table Manager `g.gui.dbmgr map=test`
3. Switch to Manage Layers page (tab) -> Add layer page (tab),  create new DB table with name 'test' and add new layer with this table (check 'Insert record for each category into table' checkbox)
5. Record for each category have not been added into table

**Error message**

wxgui_dbmgr_error

**Expected behavior**

The record for each category should be inserted in the table

### DIFF
--- a/gui/wxpython/dbmgr/base.py
+++ b/gui/wxpython/dbmgr/base.py
@@ -3635,7 +3635,8 @@ class LayerBook(wx.Notebook):
                        layer=layer,
                        qlayer=layer,
                        option='cat',
-                       columns=key)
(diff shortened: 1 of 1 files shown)


petrasovaa (approved):
Thanks for catching this, it apparently relates to changes in v.to.db column option. I assume backport is needed.

PR #1336, opened 2021-02-14
v.net.visibility: fix coordinates parameter

### LINKED ISSUE #1335: [Bug] v.net.visibility doesn't use the additional points coords when the visibility option is present
**Describe the bug**
According to the [docs](https://grass.osgeo.org/grass78/manuals/v.net.visibility.html), v.net.visibility should work in three modes:
1) `v.net.visibility input=input_layer output=graph` -> computes the visibility graph between points/vertices of input_layer
2) `v.net.visibility input=input_layer coordinates=x1,y1,x2,y2,... output=graph` -> computes the visibility graph between points/vertices of input_layer plus the additional (x1,y1), (x2,y2), ... points
3) `v.net.visibility input=input_layer coordinates=x1,y1,x2,y2,... visibility=old_graph output=new_graph` -> copies the previously computed visibility graph and then computes only additional visibility graph between points/vertices of input_layer and the additional (x1,y1), (x2,y2), ... extra points

While modes 1 and 2 seems to work well, mode 3 seems not working properly: the new output graph (new_graph) is always identical to the previously computed visibility graph (old_graph) specified with the `visibility` parameter as it seems that the coordinates of the point specified with the `coordinates` parameter are not taken into account.

**To Reproduce**

Follow Example 4 in [v.net.visibility docs](https://grass.osgeo.org/grass78/manuals/v.net.visibility.html):
```
v.extract input=zipcodes_wake output=areas_7_11_25 cats=7,11,25
g.region vector=zipcodes_wake
d.mon wx0
d.vect areas_7_11_25
v.net.visibility input=areas_7_11_25 output=graph
v.net.visibility input=areas_7_11_25 visibility=graph output=new_graph \
      coordinates=669547.97,208348.20
```

**Expected behavior**
The vector map graph is computed then a new visibility graph is computed from it with the extra point 669547.97,208348.20

- Operating System: Windows
- GRASS GIS version 7.8.5

**Additional context**

It seems to me there is a flaw in the module logic:
in the main function the `n` int variable is initially set to `0`
https://github.com/OSGeo/grass/blob/8a9a7876dd9a990b7fce5f5b0b3acd1c8c5426f6/vector/v.net.visibility/main.c#L36

then the variable `n` is not properly set to the number of additional extra points (specified with the "coordinates" parameter) before the `visibility_points` function is called:
https://github.com/OSGeo/grass/blob/8a9a7876dd9a990b7fce5f5b0b3acd1c8c5426f6/vector/v.net.visibility/main.c#L113
so the outermost loop in the `visibility_points` function is never performed:
https://github.com/OSGeo/grass/blob/8a9a7876dd9a990b7fce5f5b0b3acd1c8c5426f6/vector/v.net.visibility/visibility.c#L328-L329

Maybe @mmaldacker could give better insights.

agiudiceandrea:
Fixes #1335

### DIFF
--- a/vector/v.net.visibility/main.c
+++ b/vector/v.net.visibility/main.c
@@ -94,8 +94,10 @@ int main(int argc, char *argv[])
     count(&in, &num_points, &num_lines);
 
     /* modify the number if we have new points to add */
-    if (coor->answers != NULL)
-	num_points += count_new(coor->answers);
+    if (coor->answers != NULL) {
+	n = count_new(coor->answers);
+	num_points += n;
+    }
 
     /* and allocate */
     points = G_malloc(num_points * sizeof(struct Point));


metzm (approved):
Looks good to me.

PR #1331, opened 2021-02-13
datum: add support for D_Rauenberg_1983

neteler:
The user data provided in https://lists.osgeo.org/pipermail/grass-user/2021-February/082215.html (download at 
https://github.com/amlobat2/Coal-Test-Shapefile) contain a yet unrecognized datum: `Rauenberg_Datum_83` (EPSG:6745, see https://epsg.io/6745-datum).

This PR adds recognition of this datum string:

```
testepsg coal_test.prj
PROJCS["RD/83 / 3-degree Gauss-Kruger zone 5",
    GEOGCS["RD/83",
        DATUM["Rauenberg_Datum_83",
            SPHEROID["Bessel 1841",6377397.155,299.1528128,
                AUTHORITY["EPSG","7004"]],
            AUTHORITY["EPSG","6745"]],
        PRIMEM["Greenwich",0],
        UNIT["Degree",0.0174532925199433]],
    PROJECTION["Transverse_Mercator"],
    PARAMETER["latitude_of_origin",0],
    PARAMETER["central_meridian",15],
    PARAMETER["scale_factor",1],
    PARAMETER["false_easting",5500000],
    PARAMETER["false_northing",0],
    UNIT["metre",1,
        AUTHORITY["EPSG","9001"]],
    AXIS["Easting",EAST],
    AXIS["Northing",NORTH]]
```

My test case:

```
# create location from dataset
grass79 -c coal_test.shp ~/grassdata/RD_83_GK5

GRASS RD_83_GK5/PERMANENT:Coal-Test-Shapefile > g.proj -w
PROJCRS["RD/83 / 3-degree Gauss-Kruger zone 5",
    BASEGEOGCRS["RD/83",
        DATUM["Rauenberg Datum/83",
            ELLIPSOID["Bessel 1841",6377397.155,299.1528128,
                LENGTHUNIT["metre",1]]],
        PRIMEM["Greenwich",0,
            ANGLEUNIT["degree",0.0174532925199433]],
        ID["EPSG",4745]],
    CONVERSION["3-degree Gauss-Kruger zone 5",
        METHOD["Transverse Mercator",
            ID["EPSG",9807]],
        PARAMETER["Latitude of natural origin",0,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8801]],
        PARAMETER["Longitude of natural origin",15,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8802]],
        PARAMETER["Scale factor at natural origin",1,
[...]

PR #1337, opened 2021-02-14
libproj: change warning to debug message

neteler:
- as suggested in the #1331 review, datum transformation is handled exclusively by PROJ 6+ (no backport)
- update of projlib.dox: added note on datum transformation; added missing new functions; fix PROJ4 -> PROJ
- description of PROJ_EPSG file

### DIFF
--- a/lib/proj/projlib.dox
+++ b/lib/proj/projlib.dox
@@ -36,7 +36,7 @@ versions of GRASS).
 File lib/proj/convert.c contains functions for converting GRASS
 co-ordinate system descriptions to and from the formats used by other
 GIS. Heavy use is made of the OGR library, part of GDAL
-(http://www.gdal.osgeo.org/ogr/).
+(https://gdal.org).
 
 \subsection datum_transformation Datum transformation
 
--- a/lib/proj/projlib.dox
+++ b/lib/proj/projlib.dox
@@ -22,7 +22,11 @@ corrections to account for different datums.
 
 This is transparent for the user as input and output parameters are
 either read from PROJ_INFO and PROJ_UNITS files (<tt>g.proj</tt>,
-<tt>v.proj</tt>, <tt>r.proj</tt>).
+<tt>v.proj</tt>, <tt>r.proj</tt>). In addition, a PROJ_EPSG file is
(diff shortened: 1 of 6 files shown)


neteler:
@metzm I have added a draft description of PROJ_EPSG file (based on a best guess), please feel free to update/correct

lib/proj/projlib.dox
```diff
@@ -36,15 +40,18 @@ versions of GRASS).
...
 
 GRASS generally supports datum transformation (3 parameters, 7 parameters
 and grid).
 <p>
 
-If a warning appears that a certain datum is not recognised by GRASS and
+Note that datum transformation is handled exclusively by PROJ 6 and later.
+In the case that GRASS is compiled with PROJ 5 or older, the datum
+management is still done within GRASS GIS. Only in that case,
+if a warning appears that a certain datum is not recognised by GRASS and
```
wenzeslaus, line 54: I know this is programmer and not user doc, but anyway, the wording seems to be geared towards reader with 10-30 years of GRASS projection handling history in mind instead of just describing the current state. Leaving out things like _Note_, _exclusively_, and _still_, would be probably sufficient fix, but in addition to that, adding a proper condition to the first sentence may be even better, something like: _...is handled exclusively by PROJ if GRASS is compiled with PROJ 6 and later._ Alternatively, the whole version info can be left out from the first sentence because the exception to that rule follows immediately.

metzm: The important change here is that PROJ 6 introduced a far better handling of datum transformations. Instead of datum X -> WGS84 -> datum Y, PROJ can now do datum X -> datum Y directly. This change was introduced in PROJ with the official release of PROJ 6.0.0 on 1 Mar 2019, less than 2 years ago. The suggested text exactly reflects this transition, and its full implementation in GRASS master, to be backported to GRASS 7.8.

metzm (approved):
These changes were on my TODO list, thanks for taking care of them!

lib/proj/projlib.dox
wenzeslaus: and it would make for a good commit message.

PR #1341, opened 2021-02-15
db.univar: Ignore rows with NULLs with -e

wenzeslaus:
When value in a row is NULL, basic statistics ignore the rows
(e.g., not counted in N), however,
prior to this PR, the extended stats (-e) were processing all rows.
This was causing two issues:
1. db.univar/v.db.univar failed with 'could not convert string to float', or
2. wrong results when number of NULLs happeded to be less than N/4.

Now the NULL values (represented as empty strings) are always skipped in
the same was as for the basic stats,
i.e., not being counted towards the quartiles anymore (which is what was
causing the wrong results when number of NULL values was low).
With where conditions such as 'IS NOT NULL' neither of these issues emerged.
Now a similar condition is in the code for extended stasts too.

This PR refactors the rstrip calls, so they are in the code only once per loop.

The test now tests a run with both basic and extended stats with data
which contains NULL values. Actual values are not tested as before,
but setup and teardown steps are cleaned up (e.g. not trying to delete
elevation raster).


ninsbl (approved):
I have not tested, but the changes look good to me.

More generally however, I wonder if it would not be more efficient to read the column directly into a numpy array and aggregate there instead of writing it to file and reading it again? 

Memory should not be an issue now adays, would it!?

wenzeslaus:
> More generally however, I wonder if it would not be more efficient to read the column directly into a numpy array and aggregate there instead of writing it to file and reading it again?

That would make sense. Might be faster. I think it was written before numpy was widely available, so a custom implementation was safer in case the dependency is not met. numpy as a dependency is still not really needed for most of the modules.

PR #1344, opened 2021-02-15
wxGUI: fix 'wxPyDeprecationWarning: Call to deprecated item. Use FindItem instead.'

tmszi:
**To Reproduce**

1. Launch e.g. Ground Control Points Manager `g.gui.gcp`
2. Increase value for `SpinCtrl` widget (plus button), which is inside `StatusBar` widget
3. See warning message

**Py deprecation warning message**

```
/usr/lib64/grass79/gui/wxpython/gcp/statusbar.py:64: wxPyDeprecationWarning: Call to deprecated item. Use FindItem instead.
```


petrasovaa (approved):
I haven't actually tested it, but looks ok!

PR #1347, opened 2021-02-16
Apply Black to all scripts

wenzeslaus:
This applies Black 20.8b1 to all Python modules in scripts
and adds an action which checks the formatting and produces diff if needed.

It modifies the .flake8 file so that Flake8 ignores an error which
is considered to be a false positive (PEP8 says 'in a slice the colon acts like a binary operator'
and thus suggests/allows spaces around it while Flake8 marks that as an error
as if it would be a colon in if or for.


wenzeslaus:
The two commits and modules and tests. The new GitHub Action (workflow) is applied to the first one already, so it shows how it looks like when the formatting is wrong.

wenzeslaus:
This unifies the formatting and makes it quite easy to keep as only `black .` needs to be executed. It also radically changes the current formatting of the code. Remember one of the user testimonials: [Black is opinionated so you don't have to be.](https://github.com/psf/black)

Notably, Black is still beta, but already careful about changing things and some comments suggest non-beta release is close. Occasional Black compliance change seems like a small price for completely removing all formatting questions.

scripts/g.bands/testsuite/test_g_bands.py
```diff
@@ -18,18 +19,18 @@ def test_number_system_defined(self):
...
         # test number of S2 band identifiers (hardcoded, no changes expected)
         #
-        nbands = self._number_of_bands(pattern='S2')
+        nbands = self._number_of_bands(pattern="S2")
 
         self.assertEqual(nbands, 13)
 
-
     def test_number_s2_1(self):
         # test if S2_1 is defined (lower + upper case)
```
neteler, line 35: pls fix `'S2_1'` --> `"S2_1"` two lines below which I broke in the merge conflict resolution (I cannot reach the line here to edit it myself)

neteler:
Do you have an idea about this failing in Ubuntu 18.04?

```
...
Executing <python3 -m grass.gunittest.main --grassdata /home/runner --location nc_spm_full_v2alpha2 --location-type nc --min-success 80> ...
unittests_temporal_algebra_mixed_stds from ./lib/python/temporal failed (1 test failed)
unittests_temporal_raster3d_algebra from ./lib/python/temporal failed (1 test failed)
unittests_temporal_raster_conditionals_complement_else from ./lib/python/temporal failed (2 tests failed)
unittests_temporal_raster_algebra_equal_ts from ./lib/python/temporal failed (2 tests failed)
test_gunitest_doctests from ./lib/python/gunittest failed (1 test failed)
test_gmodules from ./lib/python/gunittest failed (3 tests failed)
test_script_doctests from ./lib/python/script failed (1 test failed)
...
```

scripts/g.bands/testsuite/test_g_bands.py
wenzeslaus: You might be looking for _Jump to... > {filename} > ⋯ > Edit file_.

wenzeslaus:
> Do you have an idea about this failing in Ubuntu 18.04?

If you mean the fail specific to Ubuntu 18.04 in the second commit in this PR, that's unrelated to this PR. Both Ubuntu tests time out for some reason time to time. It is actually not direct fail, it is a time out (the output ends with `Error: The operation was canceled.`; the symbol is exclamation mark, not red x). We need updates to testing framework to debug this. You can always re-run the jobs manually if needed.

wenzeslaus:
Checks are passing and as of writing there is no conflict with the master branch. Good time to merge it. (It is much easier to update an open PR to conform to Black than to update this PR with newly merged changes.) 

Any reviews?

marisn (approved):
It's a bit too much to check it all, but at spot checks everything seems to be fine. Besides I haven't experienced yet any problems with Black'ifying old code.

PR #1350, opened 2021-02-17
Move headers from include to include/grass

wenzeslaus:
The include contained the header files are there directly while distribution
has them in include/grass which is an unnecessary difference.
Having them in a subdirectory can make it easier for code editing tools
and other operations directly running on the source tree. See also:
* https://trac.osgeo.org/grass/wiki/G8SourceLayout

This creates the new grass directory, moves the header files, and
updates build system to use this new layout.
Makefile for makefiles and for header files is now split into two.


marisn:
Could you, please, provide some examples for "Common source code layout nowadays is to have a directory named as the resulting directory in the distribution" regarding includes? (This is not a rejection, just a request for clarification)
QGIS: no separate include directory (but that's C++)
GDAL: no separate include directory (but that's C++)
MapServer: no separate include directory (but much smaller code base)
MapCache: plain include without a subdirectory (but also a quite old code base)
PostGIS: no separate include directory
Linux kernel: plain include without a subdirectory (ok, this is a really specific case)
gtk: no separate include directory

wenzeslaus:
> Could you, please, provide some examples for "Common source code layout nowadays is to have a directory named as the resulting directory in the distribution" regarding includes?

This was perhaps a too simplified - a condition is missing: _if header files are in a separate directory_ and/or _if there is a include directory_. Examples are [PROJ](https://github.com/OSGeo/PROJ/) where include contains proj and makefiles and [{fmt}](https://github.com/fmtlib/fmt) (a modern C++ library) where include directory contains only fmt subdirectory. The `include/mylib/*.h*` is also recommended layout for [modern CMake projects](https://hsf-training.github.io/hsf-training-cmake-webpage/06-projectstructure/index.html).

However, the main idea behind the `include/*.h` ->  `include/grass/*.h` is mainly that this reflects how it is actually structured in the distribution (dist dir or actual install) and it helps you to use the include directory from the source code in an IDE.

marisn (approved):
I'm not a Make guru to judge that part. The rest seems to be good for me.

nilason:
`make`and `make distclean` at least, work without issues for me.

metzm (approved):
Makes sense, GRASS headers are (have always been) in `include/grass` after compilation.

PR #1351, opened 2021-02-17
v.to.rast: Print the correct option name

HuidaeCho:


marisn (approved):

neteler:
@HuidaeCho, @marisn: this is a perfect candidate to also update the .po files since `labelcolumn` is unique therein.

I.e.,

- search/replace  `labelcolumn` -> in  `label_column`  the .po files
- backport this PR + .po file update to keep translations in sync.

Advantage: the existing translations would not break (as they do now).

I can do the .po file updates if needed.

HuidaeCho:
> 
> 
> @HuidaeCho, @marisn: this is a perfect candidate to also update the .po files since `labelcolumn` is unique therein.
> 
> I.e.,
> 
>     * search/replace  `labelcolumn` -> in  `label_column`  the .po files
> 
>     * backport this PR + .po file update to keep translations in sync.
> 
> 
> Advantage: the existing translations would not break (as they do now).
> 
> I can do the .po file updates if needed.

@neteler I can take care of that. Let me understand the Transifex process correctly. If po files are updated in the source code, Transifex will pull them, not overwrite them? How does it know if the files need to be overwritten or read in?

HuidaeCho:
Just tried `make pot` and got this encoding error:
```
xgettext: Non-ASCII string at ../locale/scriptstrings/r.green.biomassfor.ito_translate.c:22.
          Please specify the source encoding through --from-code.
Makefile:47: recipe for target 'pot' failed
make: *** [pot] Error 1
```
Yes, that file has a non-ASCII character (superscript 3). 
image

I think it needs to be fixed, but had `make pot` ever worked before? README says `make pot` and `make update-po` are `TODO OLD`, but what does that mean? @neteler Would you recommend regenerating all the PO files from the source code or just updating `label_column` manually? I want to be very careful and conservative to not lose any translated messages.

neteler:
I thought that fixing strings the .po file in the source code would be then fetched by Transifex. Seems I am wrong as I see in https://trac.osgeo.org/grass/ticket/3539:

Needed might be to run the `tx push` command (push translations from git to Transifex). Is that correct, @lucadelu?

neteler:
> Just tried `make pot` and got this encoding error:
> 
> ```
> xgettext: Non-ASCII string at ../locale/scriptstrings/r.green.biomassfor.ito_translate.c:22.

Note that this is from an addon. Prior to `make pot` I always run `make distclean`. Not sure why addon msgs show up at your end.


> I think it needs to be fixed, but had `make pot` ever worked before?

Yes, it is also used in the server cronjob to generate the pot files which feed into Transifex:
https://github.com/OSGeo/grass-addons/blob/5d6f018fc00f7cd72fa9f12c0ee388677e039f60/tools/cronjobs_osgeo_lxd/cron_grass78_releasebranch_78_build_bins.sh#L185

which are stored here: https://grass.osgeo.org/grass79/binary/linux/snapshot/transifex/

> README says `make pot` and `make update-po` are `TODO OLD`, but what does that mean?

In this README do you see that?

> @neteler Would you recommend regenerating all the PO files from the source code or just updating `label_column` manually? I want to be very careful and conservative to not lose any translated messages.

I thought that running `make pot` actually deletes the translation in question since it doesn't match any more - that's why I suggested to edit the .po files directly, just in this case.

HuidaeCho:
> 
> 
> I thought that fixing strings the .po file in the source code would be then fetched by Transifex. Seems I am wrong as I see in https://trac.osgeo.org/grass/ticket/3539:
> 
> Needed might be to run the `tx push` command (push translations from git to Transifex). Is that correct, @lucadelu?

That's interesting. If we have to `tx push`, what is that automatic download from Transifex settings for?

HuidaeCho:
> 
> 
> > Just tried `make pot` and got this encoding error:
> > ```
> > xgettext: Non-ASCII string at ../locale/scriptstrings/r.green.biomassfor.ito_translate.c:22.
> > ```
> 
> Note that this is from an addon. Prior to `make pot` I always run `make distclean`. Not sure why addon msgs show up at your end.

I see. Still I would fix that message to something like "^3". So you don't include messages from addons?

> 
> > I think it needs to be fixed, but had `make pot` ever worked before?
> 
> Yes, it is also used in the server cronjob to generate the pot files which feed into Transifex:
> https://github.com/OSGeo/grass-addons/blob/5d6f018fc00f7cd72fa9f12c0ee388677e039f60/tools/cronjobs_osgeo_lxd/cron_grass78_releasebranch_78_build_bins.sh#L185
> 
> which are stored here: https://grass.osgeo.org/grass79/binary/linux/snapshot/transifex/
> 
> > README says `make pot` and `make update-po` are `TODO OLD`, but what does that mean?
> 
> In this README do you see that?

https://github.com/OSGeo/grass/tree/master/locale

> 
> > @neteler Would you recommend regenerating all the PO files from the source code or just updating `label_column` manually? I want to be very careful and conservative to not lose any translated messages.
> 
> I thought that running `make pot` actually deletes the translation in question since it doesn't match any more - that's why I suggested to edit the .po files directly, just in this case.

Yes, it would be safer to manually edit the .po files. Then, what if we have new messages? How can we *update* the .po files without deleting already translated messages? I thought `make pot` and `make update-po` were just for that.

lucadelu:
> > I thought that fixing strings the .po file in the source code would be then fetched by Transifex. Seems I am wrong as I see in https://trac.osgeo.org/grass/ticket/3539:
> > Needed might be to run the `tx push` command (push translations from git to Transifex). Is that correct, @lucadelu?
> 
> That's interesting. If we have to `tx push`, what is that automatic download from Transifex settings for?

I think that automatic settings should update Transifex version, I never tried to check maybe we can test this time?

HuidaeCho:
PR for non-ASCII in grass-addons: https://github.com/OSGeo/grass-addons/pull/437

HuidaeCho:
Updated PO files: https://github.com/OSGeo/grass/pull/1361

HuidaeCho:
`labelcolumn` => `label_column` was a minor change in msgid and [fuzzy entries](http://www.gnu.org/savannah-checkouts/gnu/gettext/manual/gettext.html#Fuzzy-Entries) worked!

HuidaeCho:
> https://grass.osgeo.org/grass79/binary/linux/snapshot/transifex/

Currently, we have https://grass.osgeo.org/grass78/binary/linux/snapshot/transifex/, not grass79. Should I update the URL in Transifex?

HuidaeCho:
> > > I thought that fixing strings the .po file in the source code would be then fetched by Transifex. Seems I am wrong as I see in https://trac.osgeo.org/grass/ticket/3539:
> > > Needed might be to run the `tx push` command (push translations from git to Transifex). Is that correct, @lucadelu?
> > 
> > 
> > That's interesting. If we have to `tx push`, what is that automatic download from Transifex settings for?
> 
> I think that automatic settings should update Transifex version, I never tried to check maybe we can test this time?

I agree. I'm going to `tx push` this time and see if Transifex is updated automatically after merging #1361.

lucadelu:
> > https://grass.osgeo.org/grass79/binary/linux/snapshot/transifex/
> 
> Currently, we have https://grass.osgeo.org/grass78/binary/linux/snapshot/transifex/, not grass79. Should I update the URL in Transifex?

We have only the last stable in Transifex, to simplify translator's work

veroandreo:
> > > https://grass.osgeo.org/grass79/binary/linux/snapshot/transifex/
> > 
> > 
> > Currently, we have https://grass.osgeo.org/grass78/binary/linux/snapshot/transifex/, not grass79. Should I update the URL in Transifex?
> 
> We have only the last stable in Transifex, to simplify translator's work

Yes, IIRC we discussed this in Bonn 2018 code sprint and decided it was better to have only the last stable version in Transifex

PR #1356, opened 2021-02-17
Move grass package content to directory named grass

### LINKED ISSUE #1100: Move lib/python contents into directory named grass
## Is your feature request related to a problem? Please describe.
The code of the Python package called grass which comes with GRASS GIS is hidden among C libraries in `lib`. It may look like it is a C library to use/communicate with Python.

If you are somewhat familiar with writing Python packages, you will be looking for directory which matches the name of the package. 

It is not possible to import files directly from the source code which is usually possible for Python packages.

## Describe the solution you'd like
Move the contents of the `lib/python` directory to a directory named `grass`.

This `grass` directory can be anywhere. I think it will likely be nested under some other directory, possibly as the only subdirectory.

Directory called `python` at the top level is one option. It would good to see what other mixed C-Python projects are doing (C code usually lives in src and include directories for these projects).

## Describe alternatives you've considered
The new path could be also `lib/python/grass` addressing part of the issue.

## Additional context
This relates to a similar issue with the wxGUI code which currently lives separately. They could be in the same directory in the source code or in the distribution.

A parallel issue exists also with the C code which lives in `include`, but in distribution it is in `grass`.

A broader discussion of the source code layout changes is here:
https://trac.osgeo.org/grass/wiki/G8SourceLayout

Importing the Python code directly may turn out to be more complex than just this issue, but let's get the directory part first.

Despite its significance and the number of files moved, this might be a _good first issue_.

wenzeslaus:
Before this change, the code of the Python package called grass which comes
with GRASS GIS is hidden among C libraries in lib/python.
This may look like it is a C library to use or to communicate with Python,
not a Python package.

For Python packages, the standard naming is a directory which matches the name of the package.
This allows, for example, to import files directly from the source code.

This PR moves content of lib/python to python/grass so that the name of the directory
matches the name of the resulting package (currently placed in etc/python).

This does not change Sphinx which continues to ignore the top-level grass package
(this is future work).

Fixes #1100.


wenzeslaus:
The Code scanning results / CodeQL failure is from the file move (1 fix, 1 new).

nilason (approved):
Seems to work good, without issues on Mac.
Nice improvement!

neteler:
> The Code scanning results / CodeQL failure is from the file move (1 fix, 1 new).

image

wenzeslaus:
> > The Code scanning results / CodeQL failure is from the file move (1 fix, 1 new).
> 
> image

Perhaps I was not clear: This is unrelated to this PR. CodeQL reports that here only because it sees that as a new file with a new issue (while the old file was removed and CodeQL considers that to be a resolved issue). If you want to address the `mktemp` problem, open a new issue.

neteler:
> Perhaps I was not clear: This is unrelated to this PR. CodeQL reports that here only because it sees that as a new file with a new issue (while the old file was removed and CodeQL considers that to be a resolved issue). If you want to address the `mktemp` problem, open a new issue.

Sure: I just wanted to explain to others what it means since the CodeQL access is restricted to a certain user group (I suppose to those with write access in this repo).

marisn (approved):
Didn't test the code, but I like the idea.

PR #1359, opened 2021-02-18
init: Apply Black to the init script

wenzeslaus:
An already long file made even longer, but hopefully this will facilitate refactoring.


wenzeslaus:
Will need rebase after #1347 is merged.

PR #1363, opened 2021-02-18
utils: Small fixes of Bash util scripts

wenzeslaus:
A selection of small fixes by Mega-Linter (combination of bash-exec, shellcheck, and esp. shfmt).

### DIFF
--- a/utils/dep_tree2sql.sh
+++ b/utils/dep_tree2sql.sh
@@ -37,7 +37,7 @@
 # You can then use simple queries such as:
 #
 # 	grass=> SELECT object FROM obj_imp WHERE symbol = 'I_get_target' ;
-# 	                               object                               
+# 	                               object
 # 	--------------------------------------------------------------------
 # 	 imagery/i.ortho.photo/photo.2image/OBJ.i686-pc-linux-gnu/target.o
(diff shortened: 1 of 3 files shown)


neteler:
Portability question:
the scripts might have been bourne shell scripts originally. Is the change to bash-style safe, even on systems which offer e.g. dash instead of bash?

wenzeslaus:
> the scripts might have been bourne shell scripts originally. Is the change to bash-style safe, even on systems which offer e.g. dash instead of bash?

The scripts have shebangs, so that should define what they are actually written in. One has `bash`, the other two have `sh`, so the language should be resolved by the shebang even for the linter (and checked) AFAIU and definitively when running. The `$(pwd)` syntax works in Dash too if you are asking that.

wenzeslaus:
> Portability question...

See [ShellCheck > Portability](https://github.com/koalaman/shellcheck#portability):

> ShellCheck will warn when using features not supported by the shebang. For example, if you set the shebang to #!/bin/sh, ShellCheck will warn about portability issues similar to checkbashisms.

wenzeslaus:
None of these is used by complication, so I think there is low risk in merging this.

PR #1367, opened 2021-02-18
Format YAML in GH workflows

wenzeslaus:
Apply consistent formatting to YAML files in GitHub workflows.
Using the defaults for yamllint, i.e., 2 spaces and indenting lists.
Applied using Mega-Linter (prettier, yamllint).


wenzeslaus:
I would actually prefer:

```yaml
on:
- push
- pull_request
```

over:

```yaml
on:
  - push
  - pull_request
```

But doing indent before `- ` seems more accessible to beginners and it is the default, so no configuration needed.

wenzeslaus:
@mmacata Let me know what you think.

mmacata:
@wenzeslaus I agree with a consistent formatting. I don't have a strong opinion on the list indentation. I am also a fan of using the defaults where possible, to not add unimportant exceptions. The indent before ` - ` is at least consistent to all other indents after a key. But as I said - no strong opinion. If you like to keep lists without indents, it would also be reasonable.

PR #1370, opened 2021-02-19
wxGUI/dbmgr: fix calculate column values

tmszi:
**To Reproduce**

1. Launch Attribute Table Manager `g.gui.dbmgr map=railroads`
2. Add some new column e.g. 'line_length' (double precision)
3. Right mouse click on the new added column 'line_length'
4. From the menu choose Calculate (only numeric columns) -> Line length
5. See error message

**Error message**

wx_dbmgr_calculate_column_values

Same as in the PR https://github.com/OSGeo/grass/pull/1333.

**Expected behavior**

Calculated values of the lines length in the column.


petrasovaa (approved):
Thanks!

PR #1372, opened 2021-02-19
Dockerfile_alpine_wxgui: drop file

neteler:
Rationale: Building it with wxPython is overly complicated and not working anyway.


wenzeslaus (approved):
+1 Alpine+wxPython does not seem as a high priority to me.

Issue #1327, opened 2021-02-13
[Bug] ld.lld: error: duplicate symbol: columns when building r.mapcalc


attilaolah:
**Describe the bug**

When building GRASS 7.8.5 using LLVM+Clang 11.0.1, I get the following error:

```
ld.lld: error: duplicate symbol: columns
>>> defined at map.c
>>>            OBJ.x86_64-pc-linux-gnu/map.o:(columns)
>>> defined at calc.c
>>>            calc.o:(.bss+0x4) in archive /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/352/execroot/wasm/bazel-out/k8-fastbuild/bin/lib/grass/grass_lib.build_tmpdir/dist.x86_64-pc-linux-gnu/lib/libgrass_calc.7.8.a
```

The full compile + link line is as follows:

```
[...]

PR #1377, opened 2021-02-19
Add missing #include <stdlib.h> to display/d.where/main.c

attilaolah:
It is required for `exit()`, and more importantly for `EXIT_SUCCESS`. Without this, compilation fails with with Clang 11.0.1 with the flags listed in #1327.

### DIFF
--- a/display/d.where/main.c
+++ b/display/d.where/main.c
@@ -19,6 +19,7 @@
  *               for details.
  *
  *****************************************************************************/
+#include <stdlib.h>
 #include <string.h>
 #include <grass/gis.h>
 #include <grass/gprojects.h>


nilason:
I'd say it's better to replace `exit()` with `return`, thus no need for additional header include.

attilaolah:
PTAL, updated.

HuidaeCho:
> I'd say it's better to replace `exit()` with `return`

@nilason C++ may be different, but do you have any good reference for C for this claim?
* http://www.iso-9899.info/n1570.html#5.1.2.2.3p1
* https://stackoverflow.com/questions/461449/return-statement-vs-exit-in-main
* Semantically, I prefer `exit()` especially when you can always call `main()` recursively or multiple times (I know who would do that, but still). Also, we can use `atexit()`.
  * Is `return 0` success or failure?
  * `exit(EXIT_SUCCESS)` and `exit(EXIT_FAILURE)` are success and failure, respectively. No doubt.
* I cannot find a good reason for trying to avoid including `stdlib.h`. It makes our intention clearer.
* What if you want to exit from a subroutine without a fatal error? We have to use `exit()` anyway and `return`ing `main()` would cause inconsistency.
* There are already 1,051 `exit()` calls in GRASS. `find -name *.c -exec grep 'exit(.*);' {} \; | wc`

attilaolah:
Personally I'd prefer the use of `return`, with the macros from `stdlib.h`, i.e.:

```c
#include <stdlib.h>

…
return EXIT_SUCCESS;
return EXIT_FAILURE;
```

That way exit/success is clear to the reader, as is the fact that the function returns at that point, and does not execute the remaining code.

But up to you, I'm happy to change this PR to whatever you prefer.

nilason:
In fact `return EXIT_SUCCESS;` and `return EXIT_FAILURE;` was what I had in mind. Only too late I realised the return macros are defined in the same header!  There is nothing that prevents including `stdlib.h`. I'm very sorry for this misunderstanding from my side, if you don't mind @attilaolah please put back the header and use the return macros. The least invasive fix for this is what you did from the start: just add the include header line, no really need for changing exit() to return (even though the latter is what I'd use if starting with a blank sheet).

nilason:
> > I'd say it's better to replace `exit()` with `return`
> 
> @nilason C++ may be different, but do you have any good reference for C for this claim?

@HuidaeCho  This is a rather a matter of taste, why return a function with a function when there is build in semantic for this in `return`? That's what I always use personally. (Concerning the return value, please see above comment.)

attilaolah:
I updated the pull request. I can also separate this into two changes if you wish: one for inclusion of `<stdlib.h>`, which will have to be applied anyway, and a second one for the `return` vs `exit()` discussion.

metzm:
Functions should use `return`. Programs, or any function that terminates a program such as `G_fatal_error()` should use `exit()`. That means you should use `exit()` in `main`.

HuidaeCho:
> I updated the pull request. I can also separate this into two changes if you wish: one for inclusion of `<stdlib.h>`, which will have to be applied anyway, and a second one for the `return` vs `exit()` discussion.

Yes, please. `return` vs. `exit()` needs more discussions if we really want to `return` from `main()`. It would be an impactful commit.

nilason:
@HuidaeCho , @metzm After doing my homework, now properly, I see why exit() should return from main() functions in C. Up till lately I mainly used C in Objective C programming and things are bit different there I realise. Thanks for enlighten me!

@attilaolah I can only repeat my suggestion to revert to your initial commit, inserting only the header include line and leave the lines of return as it was (with exit(EXIT_SUCCESS) etc). Now, not only because it the least invasive, but that is the correct way.

attilaolah:
PTAL.

nilason (approved):
@attilaolah Thank you!

nilason:
@HuidaeCho If you are OK with this, please push the "Approved" button or even merge if you like. No need for this to wait.

PR #1382, opened 2021-02-19
Apply Black to grass package, use Python 3 style everywhere

wenzeslaus:
This applies Black 20.8b1 formatting to grass Python package and all related tests.
ctypes is not formatted by Black and two additional files are ignored because
of bug in Black which breaks strings with escape sequence for tab.

This adds pyproject.toml file with Black configuration for the project.
Black in CI is now executed from the top dir just to be clear that the
top dir configuration is used (although Black picks config from parent
project dirs too).

Newly, only Python 3 versions are specified as targets for Black,
so also string literals starting with u (Python 2) are replaced by
simple strings.

Instead of mutliple Black runs, just run once for the whole tree and ignore directories which are not formatted yet. Black is very fast (so no need for multiple job) and only one configuration is needed (unlike Flake8)

Apply Python 3 targeted Black to already formatted code. This replaces unicode literals (Python 2) by plain Python 3 strings and adds commas to kwargs in function calls.


wenzeslaus:
Some examples of changes, i.e., what to expect in the source code now:

### No long lines, even if just sort of long 

```diff
             s_var_exprA = self.check_stds(t[1])
             s_var_exprB = self.check_stds(t[3])
             # Evaluate temporal comparison operator.
-            relations, temporal, function, aggregate = self.eval_toperator(t[2], optype='boolean')
+            relations, temporal, function, aggregate = self.eval_toperator(
+                t[2], optype="boolean"
+            )
```

### No figuring out style for a long parameter list indent

```diff
             # Build conditional values based on topological relationships.
-            complist = self.build_spatio_temporal_topology_list(s_var_exprA,
-                                                                s_var_exprB,
-                                                                topolist=relations,
-                                                                compare_cmd=True,
-                                                                compop=function,
-                                                                aggregate=aggregate)
+            complist = self.build_spatio_temporal_topology_list(
+                s_var_exprA,
+                s_var_exprB,
+                topolist=relations,
+                compare_cmd=True,
+                compop=function,
+                aggregate=aggregate,
+            )
             # Set temporal extent based on topological relationships.
-            resultlist = self.set_temporal_extent_list(complist,
-                                                       topolist=relations,
-                                                       temporal=temporal)
+            resultlist = self.set_temporal_extent_list(
+                complist, topolist=relations, temporal=temporal
+            )
 
             t[0] = resultlist
```

### No "this two parameters go on this line and these other two on the next" 

```diff
		 ta = tgis.TemporalAlgebraParser(run=True, debug=True, dry_run=True)
-        pc = ta.parse(expression="R = A {:,during} stvds(C)",
-                      stdstype='str3ds', basename="r", overwrite=True)
+        pc = ta.parse(
+            expression="R = A {:,during} stvds(C)",
+            stdstype="str3ds",
+            basename="r",
+            overwrite=True,
+        )
```

### No backslashes for continued lines, operator at the beginning of the row

```diff
			 if extended is True:
				 if "first_quartile" in stats:
-                    string += fs + str(stats["first_quartile"]) + fs + \
-                        str(stats["median"]) + fs + \
-                        str(stats["third_quartile"]) + fs + \
-                        str(stats["percentile_90"])
+                    string += (
+                        fs
+                        + str(stats["first_quartile"])
+                        + fs
+                        + str(stats["median"])
+                        + fs
+                        + str(stats["third_quartile"])
+                        + fs
+                        + str(stats["percentile_90"])
+                    )
```

This is what to expect when reading. However, when writing, most of the issues should be simply solved by running `black file.py`. No need for manual formatting except for splitting long strings and couple other things which are checked by `flake8 file.py`.

wenzeslaus:
This PR assumes that Python is Python 3, so Python 2 unicode strings are changed to standard Python 3 strings (done by Black).

```diff
-        self.assertEqual(D.get_granularity(), u'1 month')
+        self.assertEqual(D.get_granularity(), "1 month")
```

Additionally, Python 3.6 is set as a minimum for Black, so already formatted code is now reformatted with comma when `**kwargs` is at the end of a parameter list:

```diff
              x=x,
              y=y,
              z=z,
-             **addl_opts
+             **addl_opts,
        )
```

This is because previous Black changes used "Python version per file" while now it is set globally.

wenzeslaus:
@nilason This is making Python 3.6 actual minimal requirement, so if you can +1 this aspect that would be great. No need to review the tons of changes by Black, but you can read through the commit messages for context.

nilason (approved):
Browsing through the changes, this seems to be good. I don't expect anything to break because of this so I didn't test.

I'm really glad for this big spring cleaning you are doing. Will make it a lot easier keeping it tidy going forward. I suspect I'm not alone on this one:)

nilason:
Forgot to add that I was a little surprised to see so much Python 2 stuff still there.

wenzeslaus:
> Forgot to add that I was a little surprised to see so much Python 2 stuff still there.

Yes, and there are for example a lot Python 2 compatibility in the imports (which we probably should _isort_ at some point).

Issue #1297, opened 2021-02-03
[Bug] v.colors cannot properly copy color table


cmbarton:
**Describe the bug**
The 'raster=' argument does not seem to work correctly. It colors the vector map but not at all like the raster map from which it is copied. Thereafter, the color cannot be set in d.vect unless the color table is removed using v.colors. 

**To Reproduce**
Using the NC dataset, make a raster map from TOTAL_POP from the SW Wake census map, and give it a nice histogram stretched blue color ramp (It does not matter what colors are used, the results are similarly random)
```
v.to.rast input=censusblk_swwake@PERMANENT output=censusblk_swwake_vect@spatialtech2021 use=attr attribute_column=TOTAL_POP --o
r.colors -e map=censusblk_swwake_vect@spatialtech2021 color=blues
```

Try to copy this raster color table to the original vector map 
```
v.colors map=censusblk_swwake@PERMANENT use=attr column=TOTAL_POP raster=censusblk_swwake_vect@spatialtech2021
```

**Expected behavior**
The color table in the vector map should be the same as in the raster map but it is not

The incorrect colors are also written to a selected RGB column if the rgb_column= argument is used

**Screenshots**
Raster version of census blocks
![census_raster](https://user-images.githubusercontent.com/8967599/106678479-15229180-6578-11eb-800c-093ceb8616cb.png)

Vector of census blocks after running v.color to copy the raster color table
![census_vector](https://user-images.githubusercontent.com/8967599/106678486-1a7fdc00-6578-11eb-8ecb-729aa16125f6.png)

**System description (please complete the following information):**

- Operating System: Mac OS 11
- GRASS GIS version 7.8.6 dev

**Additional context**
Possibly the problem is that a raster color table has colors assigned to raster values (TOTAL_POP here). But a vector color table has colors assigned to the CAT value, not the attribute value that it is supposed to match. I think that in this case, v.colors is assigning raster colors to whatever vector objects have CAT values that match raster values (TOTAL_POP here) instead of objects that have matching attribute values, even though the v.color arguments use=attr column=TOTAL_POP  are specified. 

Also note that v.colors is lacking histogram equalization.

cmbarton:
Unlike importing a raster color table, if you export a raster color table to a **rules** file (r.colors.out) and import the rules file into v.colors, this does create a good color table that matches the raster.

cmbarton:
[...]

PR #1389, opened 2021-02-20
v.colors: Implement rules from stdin

HuidaeCho:
Make #1297 more efficient by supporting rules from stdin.

### DIFF
--- a/vector/v.colors/v.colors.html
+++ b/vector/v.colors/v.colors.html
@@ -65,7 +65,7 @@ instead of creating color table.
 v.colors map=soils_general layer=1 color=wave use=attr column=AREA rgb_column=GRASSRGB
 
 # See some GRASSRGB values:
-v.db.select map=soils_general where="cat < 4"
+v.db.select map=soils_general where="cat &lt; 4"
 cat|OBJECTID|AREA|PERIMETER|GSLNC250_|GSLNC250_I|GSL_NAME|GRASSRGB
 1|1|0|164616.125|2|1|NC113|212:42:127
 2|2|0|30785.529297|3|2|NC096|212:42:127
@@ -85,6 +85,17 @@ assigned (check
 by <em><a href="v.colors.out.html">v.colors.out</a></em>) together
 with GRASSRGB attribute column. Also note that color table is preferred
 over RGB values stored in attribute table.
+
+<h3>Transfer raster colors to vector</h3>
+
+<div class="code"><pre>
+# create an example raster
+v.to.rast input=censusblk_swwake use=attr attribute_column=TOTAL_POP output=censusblk_swwake_total_pop
+r.colors -e map=censusblk_swwake_total_pop color=blues
+
+# transfer raster colors to vector using an attribute column
+r.colors.out map=censusblk_swwake_total_pop rules=- | v.colors map=censusblk_swwake use=attr column=TOTAL_POP rules=-
+</pre></div>
 
 <h3>Remove existing color table</h3>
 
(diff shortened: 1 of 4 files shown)


vector/v.colors/v.colors.html
```diff
@@ -86,6 +86,17 @@ <h3>Convert RGB attribute values into color table</h3>
 with GRASSRGB attribute column. Also note that color table is preferred
 over RGB values stored in attribute table.
 
+<h3>Transfer raster colors to vector</h3>
+
+<div class="code"><pre>
+# create an example raster
+v.to.rast input=censusblk_swwake use=attr attribute_column=TOTAL_POP output=censusblk_swwake_total_pop
```
neteler, line 93:
```suggestion
# create an example raster from census blocks (10m pixel resolution)
g.region vector=censusblk_swwake res=10 -ap
v.to.rast input=censusblk_swwake use=attr attribute_column=TOTAL_POP output=censusblk_swwake_total_pop
```

neteler (approved):
Nice, works as expected.

cmbarton:
This is good. Fixing the raster= argument #1297 would make it even more efficient

HuidaeCho:
> This is good. Fixing the raster= argument #1297 would make it even more efficient

I wouldn't say "fix" because it seems that that usage in #1297 was never intended. It would be a new feature.

cmbarton:
> > This is good. Fixing the raster= argument #1297 would make it even more efficient
> 
> I wouldn't say "fix" because it seems that that usage in #1297 was never intended. It would be a new feature.

From v.colors manual. "The raster option allows user to specify a raster map from which to copy the color table" (this is the only explanation given)

How is this intended to be used?

HuidaeCho:
> > > This is good. Fixing the raster= argument #1297 would make it even more efficient
> > 
> > 
> > I wouldn't say "fix" because it seems that that usage in #1297 was never intended. It would be a new feature.
> 
> From v.colors manual. "The raster option allows user to specify a raster map from which to copy the color table" (this is the only explanation given)
> 
> How is this intended to be used?

Yes, it's not clear. Based on the code, looks like that sentence means raster value to vector cat color mapping (not that useful?), not value to attribute (more useful and intuitive). Actually, we can have both.

cmbarton:
> > > > This is good. Fixing the raster= argument #1297 would make it even more efficient
> > > 
> > > 
> > > I wouldn't say "fix" because it seems that that usage in #1297 was never intended. It would be a new feature.
> > 
> > 
> > From v.colors manual. "The raster option allows user to specify a raster map from which to copy the color table" (this is the only explanation given)
> > How is this intended to be used?
> 
> Yes, it's not clear. Based on the code, looks like that sentence means raster value to vector cat color mapping (not that useful?), not value to attribute (more useful and intuitive). Actually, we can have both.

I agree. That would be much more useful

HuidaeCho:
> > > > > This is good. Fixing the raster= argument #1297 would make it even more efficient
> > > > 
> > > > 
> > > > I wouldn't say "fix" because it seems that that usage in #1297 was never intended. It would be a new feature.
> > > 
> > > 
> > > From v.colors manual. "The raster option allows user to specify a raster map from which to copy the color table" (this is the only explanation given)
> > > How is this intended to be used?
> > 
> > 
> > Yes, it's not clear. Based on the code, looks like that sentence means raster value to vector cat color mapping (not that useful?), not value to attribute (more useful and intuitive). Actually, we can have both.
> 
> I agree. That would be much more useful

Implemented in #1392.

PR #1392, opened 2021-02-21
v.colors: EXPERIMENTAL histogram equalization

HuidaeCho:
Lightness (in HSL)-based color histogram equalization. Calculates probability using feature areas (no cells in vector), but I may not merge it if I fail to fix the following issue.

```bash
v.colors map=censusblk_swwake use=attr column=TOTAL_POP color=blues -e
```
seems to work,
image

but
```bash
v.colors map=censusblk_swwake use=attr column=TOTAL_POP color=rainbow -e
```
doesn't work.

Issue #1280, opened 2021-01-28
[Feat] Option for v.legend to use vcolr2 file like d.legend or let d.legend make a legend from a vcolor2 file


cmbarton:
Making legends in GRASS has always been difficult. It is doable now, but is still not an easy process. However, since vectors can now have color tables similar to raster color tables, it seems like it would not take much to be able to create a legend using a vector color table. Either have an option for d.legend.vect to read the color table and set the interval as in d.legend, or expand d.legend to create legends for vectors just like those for rasters (i.e., from a single attribute). That way, d.legend.vect could stay the way it is to create legends for multiple vector layers.

veroandreo:
Would it be possible to integrate [d.vect.thematic](https://grass.osgeo.org/grass79/manuals/d.vect.thematic.html) functionality in d.legend.vect? Ideally without asking users to define the RGB values mandatorily but reading the color table instead

cmbarton:
This would be nice

PR #1393, opened 2021-02-21
v.colors: raster= with use=attr

HuidaeCho:
Implement #1297.


HuidaeCho:
Related TODO: #1392.

cmbarton (approved):

cmbarton:
This works great! Behaves as one might expect now and gives very good results. Thanks much. 

Now, if  v.legend could be enhanced to work like d.legend #1280 and have d.vect.thematic write to a vector color table, GRASS would have a really nice and powerful vector thematic mapping capability.

PR #1274, opened 2021-01-26
Fix compiler warnings, part 6

nilason:
Fixes compiler warnings:

- -Wformat-extra-args
- -Wformat-invalid-specifier
- -Wparentheses
- -Wparentheses-equality
- -Wtautological-compare

Sixth part addressing #1247.

There are a couple of ambivalent cases and I'm not sure how to correctly address them. See comments.

Modules / code parts directly affected:
- lib/dspf
- lib/imagery/
- lib/vector/Vlib/
- display/d.vect.thematic
- raster/r.compress
- raster/r.fill.stats
- raster/r.in.gridatb
- raster3d/r3.info

PR #1395, opened 2021-02-21
Fix compiler warnings, part 10

nilason:
Remove argument not used by format string
Fixes -Wformat-extra-args compiler warnings.

This was originally part of #1274, but separated in hope for finding a better solution.

Tenth part addressing #1247.

Compiler warning:
```
main.c:166:30: warning: data argument not used by format string [-Wformat-extra-args]
                             WEIGHTS[i][j]);
                             ^
```

### DIFF
--- a/raster/r.fill.stats/main.c
+++ b/raster/r.fill.stats/main.c
@@ -28,7 +28,7 @@
    if (default) cells size specification is used, then the neighborhood
    shape is a less exact rectangle; in that case the neighborhood dimensions
    will also be different for x and y if the cell dimensions are different
-   - lat/lon data is allowed, but distance measure for -m(ap units) is straight line! 
+   - lat/lon data is allowed, but distance measure for -m(ap units) is straight line!
    - cell weights are now normalized to be in range [0;1]
    - center cell weight for "wmean" is now "1.0"
 
@@ -159,12 +159,6 @@ void print_weights_matrix(long int rows, long int cols)
             if (WEIGHTS[i][j] != -1.0) {
                 snprintf(weight_matrix_weight_buf, weight_matrix_line_length,
                          "%06.2f ", WEIGHTS[i][j]);
-                if (strlen(weight_matrix_weight_buf) >
-                    (weight_matrix_weight_length)) {
-                    snprintf(weight_matrix_weight_buf,
-                             weight_matrix_line_length, "[????] ",
-                             WEIGHTS[i][j]);
-                }
             }
             else {
                 snprintf(weight_matrix_weight_buf, weight_matrix_line_length,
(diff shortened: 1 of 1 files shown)


nilason:
> I prefer to keep this code active as long as there is a reason to test for `if (WEIGHTS[i][j] != -1.0)`

_Originally posted by @metzm in https://github.com/OSGeo/grass/pull/1274#discussion_r579716760_

I think you perhaps misread the code. By design `WEIGHTS[i][j]` may be `-1.0` if map units is used (with `-m` flag) printed with `......`:

https://github.com/OSGeo/grass/blob/4e7606fac92ca8418af976ba4cbb84dc4f10113a/raster/r.fill.stats/r.fill.stats.html#L300-L310


However, that part doesn't affect the troublesome snprintf call and its own `if` condition.


https://github.com/OSGeo/grass/blob/4e7606fac92ca8418af976ba4cbb84dc4f10113a/raster/r.fill.stats/main.c#L159-L172

The `if` condition on line 162-163 can by definition *never* be true, as the size of `weight_matrix_weight_buf ` is set to `weight_matrix_weight_length + 1`.

I'm quite confident now this is dead code (L162-167) which can be removed all together. Pushed changes according to this.

raster/r.fill.stats/main.c
```diff
@@ -159,12 +159,6 @@ void print_weights_matrix(long int rows, long int cols)
             if (WEIGHTS[i][j] != -1.0) {
                 snprintf(weight_matrix_weight_buf, weight_matrix_line_length,
                          "%06.2f ", WEIGHTS[i][j]);
```
metzm, line 161:
```suggestion
                         "%.15g ", WEIGHTS[i][j]);
```

metzm: `weight_matrix_weight_length` should be `#define `d. Alternatively, use `G_asprintf()` with `char *weight_matrix_line_buf = NULL;` instead of `snprintf()`.

There are numerous other compiler warnings for this module. Are they addressed by your other PRs or do you intend to get rid of compiler warnings for this module?

nilason: The `print_weights_matrix()` is a "print nice" function with the `-w` argument which outputs nicely (and precisely calculated) in a 80 column row (if it fits). To use `%.15g` turns this upside down:

before (with `%06.2f`):

```
r.fill.stats -m -w input=cfactorbare_1m@PERMANENT output=asdf distance=20 mode=wmean power=2.0 cells=8
W-E size of neighborhood is 5 cells.
S-N size of neighborhood is 5 cells.
Spatial weights neighborhood (cells):
...... 000.00 000.01 000.00 ...... 
000.00 000.14 000.31 000.14 000.00 
000.01 000.31 001.00 000.31 000.01 
000.00 000.14 000.31 000.14 000.00 
...... 000.00 000.01 000.00 ...... 
```

after (with `%.15g`):
```
r.fill.stats -m -w input=cfactorbare_1m@PERMANENT output=asdf distance=20 mode=wmean power=2.0 cells=8
W-E size of neighborhood is 5 cells.
S-N size of neighborhood is 5 cells.
Spatial weights neighborhood (cells):
...... 0 0.0111456180001682 0 ...... 
0 0.135088935932648 0.305572809000084 0.135088935932648 0 
0.0111456180001682 0.305572809000084 1 0.305572809000084 0.0111456180001682 
0 0.135088935932648 0.305572809000084 0.135088935932648 0 
...... 0 0.0111456180001682 0 ...... 
```

I seem to me that this function worked well, there was only an unnecessary part which is removed with this PR.

nilason: > There are numerous other compiler warnings for this module. Are they addressed by your other PRs or do you intend to get rid of compiler warnings for this module?

They are already dealt with, with #1276.

metzm: > The `print_weights_matrix()` is a "print nice" function with the `-w` argument which outputs nicely (and precisely calculated) in a 80 column row (if it fits). To use `%.15g` turns this upside down:

OK, please ignore my suggestion.

metzm (approved):
The variable `weight_matrix_weight_length` needs to be removed in a separate PR.

nilason:
@metzm Thanks!

PR #1396, opened 2021-02-21
i18n: fix duplicate messsages in po files

neteler:
Current issue in releasebranch_7_8 branch:

```
cd locale/
make
...
WARNING: file <grassmods_cs.po> has some problems: <b'po/grassmods_cs.po:60512: duplicate message definition...\npo/grassmods_cs.po:31759: ...this is the location of the first definition\nmsgfmt: found 1 fatal error'>
WARNING: file <grassmods_de.po> has some problems: <b'po/grassmods_de.po:60550: duplicate message definition...\npo/grassmods_de.po:31802: ...this is the location of the first definition\nmsgfmt: found 1 fatal error'>
WARNING: file <grassmods_es.po> has some problems: <b'po/grassmods_es.po:61273: duplicate message definition...\npo/grassmods_es.po:32070: ...this is the location of the first definition\nmsgfmt: found 1 fatal error'>
WARNING: file <grassmods_it.po> has some problems: <b'po/grassmods_it.po:60025: duplicate message definition...\npo/grassmods_it.po:31693: ...this is the location of the first definition\nmsgfmt: found 1 fatal error'>
WARNING: file <grassmods_tr.po> has some problems: <b'po/grassmods_tr.po:59702: duplicate message definition...\npo/grassmods_tr.po:31491: ...this is the location of the first definition\nmsgfmt: found 1 fatal error'>
make[1]: Leaving directory '/home/mneteler/software/grass78/locale'
```

This PR deduplicates these messages. I used the following command:

```
for i in cs de es it tr; do
    msguniq --use-first grassmods_${i}.po > tmp.$i
    rm -f grassmods_${i}.po
    mv tmp.$i grassmods_${i}.po
done
```

Sorry for the linebreak noise in this PR but I don't know how to avoid it (maybe another `msguniq` flag is needed?).

PR #1397, opened 2021-02-23
i18n: Remove duplicate messages from po files

HuidaeCho:
Redo #1396.

```bash
for i in cs de es it tr; do
    sed '/^#~/d' grassmods_$i.po | msguniq --use-first > tmp.$i
    mv tmp.$i grassmods_$i.po
done
```


neteler (approved):
No more dupes, thanks!

PR #1400, opened 2021-02-23
wxGUI: startup screen removed

lindakladivova:
**STATE BEFORE THIS PR:**

Before this PR, when the last mapset was not usable, the startup screen still appeared.

**EXPECTED STATE AFTER THIS PR:**

This PR suggests the entire removal of the startup screen and using the default location EPSG:4326 with Data Catalog for the user's data hierarchy definition.

It uses Infobar that informs the user about the reason why they are in the demolocation and suggests further process - creating or finding another usable mapset through Data Catalog.

To be clearer, I will explain here which situations can occur and how the new startup copes with them.

**First-time startup**
Still the same as before this PR. The data structure info is given to the user.
first_time_user_mapset

**Second GRASS GUI startup (without closing the first session)**
GRASS was not closed, therefore, the user is still considered a beginner. No settings were saved to the RC file. Whereas the PERMANENT mapset is used, the startup creates a spare mapset called after a user. 
first_time_user_mapset_if_no_rc_file

**Third GRASS GUI startup (without closing the second session)**
 Whereas the user mapset is used, the startup creates a spare mapset called "user_1"...
 
We closed all three sessions.
Then we start up GRASS again:

**Last used mapset usable**
The normal situation in which the last used mapset in RC file is usable (most common)
last_used_mapset_usable

Now we will try to start GRASS again without closing the previous session:

**Last used mapset is locked**
last_used_mapset_locked
Yes. The last used mapset is locked (running in the previous session). The infobar is telling us the option for switching and removing gislock or we can set usable mapset through Data Catalog.

To complete the list of situations, two other situations can occur:

**Last usable mapset is invalid**
invalid_mapset

**Last usable mapset is owned by different user**
different_user_mapset

Here, GRASS chooses PERMANENT mapset, if not capable, then "user" mapset, if "user" mapset not capable, then it chooses or creates user_2 mapset and so on...
In all situations, the default location is used with applicable advice provided in Infobar.

All comments welcome! :-) This is very important PR so the more discussion the better!


veroandreo:
Shouldn't we say something about how to remove all these mapsets that could eventually be created when the user fails to close GRASS properly or opens multiple sessions...? Especially because s/he needs to activate the pencil first to be able to remove mapsets. Just wondering...  

Maybe it's just my OCD that makes my mind crash when I see all those user, user_1, user_2, etc 😬

lindakladivova:
> 
> 
> What's not clear to me is where a new mapset is created. In the screenshots, you used the default demo location only. Is a new mapset created in this demo location only, not in the requested location (makes more sense)?
> 
> Still, I'm sorry, but I really don't understand the rationale behind this idea of "choose or create _automatically_ a mapset every time GRASS runs." We cannot exactly predict the user's intention when they start GRASS. For me, if a mapset was locked and I accidentally started the same mapset, I would like to be kicked out or offered to remove the lock. Must we always start GRASS without user intervention at all, creating a new mapset sometimes, which in most of the case will have to be _manually_ cleaned up because they are more of temporary fixes from my perspective?
> 
> If it has to be done, please add an option to turn it off. Related issue #1403.

Yet, I work only with the default location. We discussed it with Anna, and so far it seems to us clearer to always end up the nonstandard GRASS startup in the default location. But it is quite open.

As you write "For me, if a mapset was locked and I accidentally started the same mapset, I would like to be kicked out or offered to remove the lock. " - You _are_ now kicked off and in the Infobar there is the button "Switch to last used mapset" and you can switch to it. 
However, I have an idea about generating user mapsets - one nonstandard situation is the locked mapset. If a user switches to the last used mapset using the Infobar button, we could automatically remove the user mapset since it is not needed anymore.
What do you think about it?

I know that there is a lack in generating user mapsets, but IMHO I am still convinced that it is a smaller problem compared to keeping and maintaining the user-unfriendly and old startup screen...

veroandreo:
> > What's not clear to me is where a new mapset is created. In the screenshots, you used the default demo location only. Is a new mapset created in this demo location only, not in the requested location (makes more sense)?
> > Still, I'm sorry, but I really don't understand the rationale behind this idea of "choose or create _automatically_ a mapset every time GRASS runs." We cannot exactly predict the user's intention when they start GRASS. For me, if a mapset was locked and I accidentally started the same mapset, I would like to be kicked out or offered to remove the lock. Must we always start GRASS without user intervention at all, creating a new mapset sometimes, which in most of the case will have to be _manually_ cleaned up because they are more of temporary fixes from my perspective?
> > If it has to be done, please add an option to turn it off. Related issue #1403.
> 
> Yet, I work only with the default location. We discussed it with Anna, and so far it seems to us clearer to always end up the nonstandard GRASS startup in the default location. But it is quite open.
> 
> As you write "For me, if a mapset was locked and I accidentally started the same mapset, I would like to be kicked out or offered to remove the lock. " - You _are_ now kicked off and in the Infobar there is the button "Switch to last used mapset" and you can switch to it.
> However, I have an idea about generating user mapsets - one nonstandard situation is the locked mapset. If a user switches to the last used mapset using the Infobar button, we could automatically remove the user mapset since it is not needed anymore.
> What do you think about it?

I like this solution, it will avoid cluttering the location with empty (?) `user_*` mapsets - my OCD is happy :)

wenzeslaus:
@HuidaeCho @veroandreo About all this cluttering with empty mapsets, how often do you think it happens? The current code in the PR tries PERMANENT mapset and steps through other candidate (mapset) names till it hits one which is usable. If it is the PERMANENT mapset, no mapset is created. If one of these candidates is usable, it uses that and no other mapset is created, etc. As far as I understand the code, the cluttering happens when you keep starting new sessions. Is this a realistic use case which we need to specifically address in the design?

Note also that the locking part is at this point irrelevant for Windows where the locking is not enabled (I don't know why).

> ...If it has to be done, please add an option to turn it off.

@HuidaeCho If turned of, the question would be what happens instead. For example, making location/mapset optional is an interesting way to address this issue, but the question is how to implement that.

> ...we could automatically remove the user mapset since it is not needed anymore.

@lindakladivova I consider this nice to have, but not necessary. Depends on how hard this would be to implement. You started to heavily leverage the "gisrc" file, similarly to that, perhaps the VAR file in the mapset is a place for marking mapset as temporary. The concept of a  temporary mapset is already there (`grass --tmp-mapset ...`), but this would require completely different implementation.

HuidaeCho:
> how often do you think it happens?

It just happened to me 10 minutes ago. I already opened the NC sample mapset in one desktop and tried to open it again in another, all from the same computer. With this PR, I would have ended up in the world demo location, which has nothing to do with the NC sample. I would be lucky that I won't have to delete a new temporary mapset because the default demo mapset was not locked. In some sense, it's better to create all these temporary mapsets only in the demo location to avoid *cluttering* user locations. Cleaning would be easier.

So this PR is mainly for those who start GRASS using the icon, but I start it from the terminal. I think I found a solution. If I add `g.gui` in `~/.grass7/bashrc` and start `grass -text`, it will fail if a mapset is locked (we're not changing this CLI behavior right?). If not, GUI will start in the requested mapset because it is usable or locked? I'm not sure about this.
1. A mapset is usable
2. Start GRASS in this mapset from the terminal
3. Start g.gui
4. To GUI, is this current mapset locked by the same CLI session or usable?

If it's usable because GUI is in the same session, that will be great. This solution will allow a delayed GUI startup only after the CLI session started successfully.

wenzeslaus:
> > how often do you think it [_this cluttering with empty mapsets_] happens?
> 
> It just happened to me 10 minutes ago. I already opened the NC sample mapset in one desktop and tried to open it again in another, all from the same computer. With this PR, I would have ended up in the world demo location, which has nothing to do with the NC sample. I would be lucky that I won't have to delete a new temporary mapset because the default demo mapset was not locked.

You were not lucky. This is how it should behave. No mapset was (would be) created, so no clutter is accumulating which is exactly my point: The scenario when you end up with location cluttered by these mapsets is unlikely.

```
 _________________________
| No clutter, no problem. |
 -------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
```

> In some sense, it's better to create all these temporary mapsets only in the demo location to avoid _cluttering_ user locations. Cleaning would be easier.

Right, it is a compromise. We could create completely temporary location, but that brings in a lot of other issues like what should happen when a user starts to import data there. Similar, although smaller, issue is with the temporary mapset.

> So this PR is mainly for those who start GRASS using the icon,

Yes, something like that. And that's the whole issue here. GRASS power user starting from command line can probably deal with GUI (and GRASS itself) not starting at all in case mapset is locked. However, we are trying to have find a solution which works for all users who use GUI and not just when the mapset is locked. We need to provide something like data catalog/startup screen which allows all kinds of management (force unlock, all database, ..., view content) which leads us to delivering the Data tab which needs an active session, so we need to pick a reasonable mapset.

> but I start it from the terminal.

Well, we know when we are running in interactive terminal, so we could behave differently in that case. However, Windows users seem to insist on CLI (#1221), so this distinction could be applied only on Linux and maybe macOS for the time being and that assumes that #1221 gets merged and actually applied in distributions/installations.

> start `grass -text`, it will fail if a mapset is locked (we're not changing this CLI behavior right?).

`grass --text` and `grass --exec` should fail with an error message in these situations. (Only partially related: It is still not 100% clear if `grass --exec` should require mapset always.)

> ...To GUI, is this current mapset locked by the same CLI session or usable?...If it's usable because GUI is in the same session, that will be great.

Yes, at least that part should work. GUI behaves like any other GRASS GIS module (which sounds nice on the surface, but it is partially the source of this whole starting GRASS issue; if we change that, we have potentially more options available to us).

veroandreo:
> @HuidaeCho @veroandreo About all this cluttering with empty mapsets, how often do you think it happens? The current code in the PR tries PERMANENT mapset and steps through other candidate (mapset) names till it hits one which is usable. If it is the PERMANENT mapset, no mapset is created. If one of these candidates is usable, it uses that and no other mapset is created, etc. As far as I understand the code, the cluttering happens when you keep starting new sessions. 

Thanks much for this explanation! I have not understood this before and hence my concern. But like this, to get the location cluttered with `user_*` mapsets, seems indeed not highly likely.

lindakladivova:
> 
> 
> @HuidaeCho @veroandreo About all this cluttering with empty mapsets, how often do you think it happens? The current code in the PR tries PERMANENT mapset and steps through other candidate (mapset) names till it hits one which is usable. If it is the PERMANENT mapset, no mapset is created. If one of these candidates is usable, it uses that and no other mapset is created, etc. As far as I understand the code, the cluttering happens when you keep starting new sessions. Is this a realistic use case which we need to specifically address in the design?
> 
> Note also that the locking part is at this point irrelevant for Windows where the locking is not enabled (I don't know why).
> 
> > ...If it has to be done, please add an option to turn it off.
> 
> @HuidaeCho If turned of, the question would be what happens instead. For example, making location/mapset optional is an interesting way to address this issue, but the question is how to implement that.
> 
> > ...we could automatically remove the user mapset since it is not needed anymore.
> 
> @lindakladivova I consider this nice to have, but not necessary. Depends on how hard this would be to implement. You started to heavily leverage the "gisrc" file, similarly to that, perhaps the VAR file in the mapset is a place for marking mapset as temporary. The concept of a temporary mapset is already there (`grass --tmp-mapset ...`), but this would require completely different implementation.

@wenzeslaus I also think that it is not necessary. I realize that there is quite a heavy change in gisrc :-).  I was thinking about how to code this PR very long time and finally, used gisrc file because the last used mapset, location, grassdb here are closely related to this issue, so everything would be at the same place. And also reading gisrc in catalog.py is quite elegant. What do you think about the concept I came up with?

HuidaeCho:
@lindakladivova Please make the location of the default database (`$HOME/grassdata`) configurable at least through `$HOME/.grass7/rc` manually. I have my own folder structure.

lindakladivova:
> 
> 
> @lindakladivova Please make the location of the default database (`$HOME/grassdata`) configurable at least through `$HOME/.grass7/rc` manually. I have my own folder structure. Or maybe, please move it to `$HOME/.grass7` so it's hidden from `ls`.

@HuidaeCho I like your note about extending the `$HOME/.grass7/rc` file with the default db. But I am not sure about making the `world_latlong_wgs84` location directly in rc. This would mean that a user cannot treat it as a normal db/loc (it would be something like temporary db/loc). I think that the "grassdata" db folder is quite good. And also if the first-time user will see this path `$HOME/.grass7/rc/world_latlong_wgs84/PERMANENT`, I am not sure if it wouldn't be rather confusing.... However, I do not have any really strong opinion about this. It depends, how we perceive the default location. (@landa, @wenzeslaus ?)
Maybe the _default location_ is a bit more correct collocation here than _demolocation_ (note that _default location_ is also used in Info Bar) since we provide a common EPSG: 4326 option which a user can immediately start using.

HuidaeCho:
> > @lindakladivova Please make the location of the default database (`$HOME/grassdata`) configurable at least through `$HOME/.grass7/rc` manually. I have my own folder structure. Or maybe, please move it to `$HOME/.grass7` so it's hidden from `ls`.
> 
> @HuidaeCho I like your note about extending the `$HOME/.grass7/rc` file with the default db. But I am not sure about making the `world_latlong_wgs84` location directly in rc. This would mean that a user cannot treat it as a normal db/loc (it would be something like temporary db/loc). I think that the "grassdata" db folder is quite good. And also if the first-time user will see this path `$HOME/.grass7/world_latlong_wgs84/PERMANENT`, I am not sure if it wouldn't be rather confusing.... However, I do not have any really strong opinion about this. It depends, how we perceive the default location. (@landa, @wenzeslaus ?)
> 

I agree with you. It all depends on how we look at this database. Is it a demo (probably could be hidden in `$HOME/.grass7`) or a *default* database (just like a regular database). Since some people may decide to use this default database for their EPSG4326 locations, the latter seems to make more sense. I'm fine with either, but *any* hardcoded paths will get in anyone's way when they want to keep their folder structure clean in their own way. Maybe, it's just me, but I just feel that the Linux home directory seems to be different from Windows' for that matter. I rarely look into my Windows home directory because random software tends to create random folders there and I stop caring about what are in there and I just don't use it. However, in Linux, that's my true home for my daily use, probably, for most power users as well, who will most likely care about this issue.

> Maybe the _default location_ is a bit more correct collocation here than _demolocation_ (note that _default location_ is also used in Info Bar) since we provide a common EPSG: 4326 option which a user can immediately start using.

Absolutely, the `default_gisdbase` setting as you mentioned in #1419 will address my concern.

neteler:
> ... but _any_ hardcoded paths will get in anyone's way when they want to keep their folder structure clean in their own way. Maybe, it's just me, but I just feel that the Linux home directory seems to be different from Windows' for that matter. I rarely look into my Windows home directory because random software tends to create random folders there and I stop caring about what are in there and I just don't use it. However, in Linux, that's my true home for my daily use, probably, for most power users as well, who will most likely care about this issue.

Today and in the last 10+ years we used and use network drives for storing "grassdata/" (in IRST, in FEM, in some HPC infrastructure, in mundialis, ...). Hence it would be generally important for us to avoid any hardcoded paths.

petrasovaa:
> > > @lindakladivova Please make the location of the default database (`$HOME/grassdata`) configurable at least through `$HOME/.grass7/rc` manually. I have my own folder structure. Or maybe, please move it to `$HOME/.grass7` so it's hidden from `ls`.
> > 
> > 
> > @HuidaeCho I like your note about extending the `$HOME/.grass7/rc` file with the default db. But I am not sure about making the `world_latlong_wgs84` location directly in rc. This would mean that a user cannot treat it as a normal db/loc (it would be something like temporary db/loc). I think that the "grassdata" db folder is quite good. And also if the first-time user will see this path `$HOME/.grass7/world_latlong_wgs84/PERMANENT`, I am not sure if it wouldn't be rather confusing.... However, I do not have any really strong opinion about this. It depends, how we perceive the default location. (@landa, @wenzeslaus ?)
> 
> 
> Absolutely, the `default_gisdbase` setting as you mentioned in #1419 will address my concern.

This doesn't make much sense to me, rc is not a place users should be editing. Why not create it in current database? Disadvantage is that if you use multiple databases, it would eventually be created in each.

Alternatively, we could create a temporary empty location, but that would be probably little more complex, e.g. maybe we would want to hide it from data catalog and when user would switch to other location, it would need to be removed. I think we wouldn't have issues with user importing data into it as long as we clearly inform them in the infobar.

wenzeslaus:
Let me comment on what I think the rc (`.grassX/rc`) file is:

* The name rc suggest it is a configuration loaded during start of the program relevant to the while runtime of the program similarly to `.bashrc` or `.pylintrc`. However, the current rc file is far from these large configurations given its focus on the last used db/location/mapset. 
* The file saves the last used "file" (db/location/mapset - aka last used _project_), so it works like what, e.g., wxPython calls file history which is used to implement (list of) most recent files (in menu). Just in this case, it is only the most recent one and, quite importantly, it is not the only information stored there.
* It looks little bit like user settings, but it is not directly modified by the user, instead it stores the most recent used values. This more applies to the value of GUI used (text, wx) than to the db/loc/mapset which, being a file, would not be typically found in settings.
* The GISRC variable with its associated file and grass.py code referring to `.grassX/rc` as `gisrcrc` suggests there was some historical development in what the file was doing possibly having only one rc file around persistent over what would be now different sessions.
* There seems to be three concepts, a settings/configuration file, last used file/data/project file, and current session file. The rc (`.grassX/rc`) trying to be the first two while GISRC is the (current) session file. Some untangling might be needed.

HuidaeCho:
> This doesn't make much sense to me, rc is not a place users should be editing.

Not necessarily directly. Users can use `g.gisenv` to add/change variables in `.grassX/rc`.

Alternatively, an environment variable could be another option. It would be more permanent... 

> Why not create it in current database? Disadvantage is that if you use multiple databases, it would eventually be created in each.
> 

That will be a *big* disadvantage.

> Alternatively, we could create a temporary empty location, but that would be probably little more complex, e.g. maybe we would want to hide it from data catalog and when user would switch to other location, it would need to be removed. I think we wouldn't have issues with user importing data into it as long as we clearly inform them in the infobar.

Only if we can guarantee a clean removal of the temp location. e.g., multiple sessions in the same temp location and exit one session at a time? Or temp location per session?

lindakladivova:
> 
> 
> > This doesn't make much sense to me, rc is not a place users should be editing.
> 
> Not necessarily directly. Users can use `g.gisenv` to add/change variables in `.grassX/rc`.
> 
> Alternatively, an environment variable could be another option. It would be more permanent...
> 
> > Why not create it in current database? Disadvantage is that if you use multiple databases, it would eventually be created in each.
> 
> That will be a _big_ disadvantage.
> 
> > Alternatively, we could create a temporary empty location, but that would be probably little more complex, e.g. maybe we would want to hide it from data catalog and when user would switch to other location, it would need to be removed. I think we wouldn't have issues with user importing data into it as long as we clearly inform them in the infobar.
> 
> Only if we can guarantee a clean removal of the temp location. e.g., multiple sessions in the same temp location and exit one session at a time? Or temp location per session?



> 
> 
> > > > @lindakladivova Please make the location of the default database (`$HOME/grassdata`) configurable at least through `$HOME/.grass7/rc` manually. I have my own folder structure. Or maybe, please move it to `$HOME/.grass7` so it's hidden from `ls`.
> > > 
> > > 
> > > @HuidaeCho I like your note about extending the `$HOME/.grass7/rc` file with the default db. But I am not sure about making the `world_latlong_wgs84` location directly in rc. This would mean that a user cannot treat it as a normal db/loc (it would be something like temporary db/loc). I think that the "grassdata" db folder is quite good. And also if the first-time user will see this path `$HOME/.grass7/world_latlong_wgs84/PERMANENT`, I am not sure if it wouldn't be rather confusing.... However, I do not have any really strong opinion about this. It depends, how we perceive the default location. (@landa, @wenzeslaus ?)
> > 
> > 
> > Absolutely, the `default_gisdbase` setting as you mentioned in #1419 will address my concern.
> 
> This doesn't make much sense to me, rc is not a place users should be editing. Why not create it in current database? Disadvantage is that if you use multiple databases, it would eventually be created in each.
> 
> Alternatively, we could create a temporary empty location, but that would be probably little more complex, e.g. maybe we would want to hide it from data catalog and when user would switch to other location, it would need to be removed. I think we wouldn't have issues with user importing data into it as long as we clearly inform them in the infobar.

> > I think that concept of temporary location will be again something confusing for a first-time user. There is still a risk that a user would not read the advice in the infobar... and then he could lose all the data.
The second thing is that if we hide temporary empty location, we destroy the important concept of first infobar which explains data hierarchy...
Or we can have two concept - one for first-time startup, I think this concept of "home/grassdata" default grassdb works fine here and then a different concept of temporary location for other cases...

petrasovaa:
> > > This doesn't make much sense to me, rc is not a place users should be editing.
> > 
> > 
> > Not necessarily directly. Users can use `g.gisenv` to add/change variables in `.grassX/rc`.
> > Alternatively, an environment variable could be another option. It would be more permanent...
> > > Why not create it in current database? Disadvantage is that if you use multiple databases, it would eventually be created in each.
> > 
> > 
> > That will be a _big_ disadvantage.
> > > Alternatively, we could create a temporary empty location, but that would be probably little more complex, e.g. maybe we would want to hide it from data catalog and when user would switch to other location, it would need to be removed. I think we wouldn't have issues with user importing data into it as long as we clearly inform them in the infobar.
> > 
> > 
> > Only if we can guarantee a clean removal of the temp location. e.g., multiple sessions in the same temp location and exit one session at a time? Or temp location per session?
> 
> > > > > @lindakladivova Please make the location of the default database (`$HOME/grassdata`) configurable at least through `$HOME/.grass7/rc` manually. I have my own folder structure. Or maybe, please move it to `$HOME/.grass7` so it's hidden from `ls`.
> > > > 
> > > > 
> > > > @HuidaeCho I like your note about extending the `$HOME/.grass7/rc` file with the default db. But I am not sure about making the `world_latlong_wgs84` location directly in rc. This would mean that a user cannot treat it as a normal db/loc (it would be something like temporary db/loc). I think that the "grassdata" db folder is quite good. And also if the first-time user will see this path `$HOME/.grass7/world_latlong_wgs84/PERMANENT`, I am not sure if it wouldn't be rather confusing.... However, I do not have any really strong opinion about this. It depends, how we perceive the default location. (@landa, @wenzeslaus ?)
> > > 
> > > 
> > > Absolutely, the `default_gisdbase` setting as you mentioned in #1419 will address my concern.
> > 
> > 
> > This doesn't make much sense to me, rc is not a place users should be editing. Why not create it in current database? Disadvantage is that if you use multiple databases, it would eventually be created in each.
> > Alternatively, we could create a temporary empty location, but that would be probably little more complex, e.g. maybe we would want to hide it from data catalog and when user would switch to other location, it would need to be removed. I think we wouldn't have issues with user importing data into it as long as we clearly inform them in the infobar.
> 
> I think that concept of temporary location will be again something confusing for a first-time user. There is still a risk that a user would not read the advice in the infobar... and then he could lose all the data.
> The second thing is that if we hide a temporary empty location, we destroy the important concept of the first infobar which explains data hierarchy...
> Or we can have two concepts - one for the first-time startup, I think this concept of "home/grassdata" default grassdb works 

Yes that's what I meant, the temporary one would be only for these cases, when you can't use the last used mapset.

>fine here and then a different concept of a temporary location for other cases... - but still there will be a risk of losing data which is IMHO very serious.

I wouldn't be worried about that too much. The infobar is hard to miss and even if you ignore it, you don't have any data there, which should tell you something is wrong. GUI could be aware that it is running in a temporary location (e.g. based on the name) so we could use that in some way to additionally warn users when they try to import data etc. I don't have this fully thought through, there may be some associated problems with this, but I think it's worth looking into this direction.

petrasovaa:
> > Alternatively, we could create a temporary empty location, but that would be probably little more complex, e.g. maybe we would want to hide it from data catalog and when user would switch to other location, it would need to be removed. I think we wouldn't have issues with user importing data into it as long as we clearly inform them in the infobar.
> 
> Only if we can guarantee a clean removal of the temp location. e.g., multiple sessions in the same temp location and exit one session at a time? Or temp location per session?

The idea is to have ne

mlennert:
On 9/03/21 15:45, Anna Petrasova wrote:
>     I think that concept of temporary location will be again something
>     confusing for a first-time user. There is still a risk that a user
>     would not read the advice in the infobar... and then he could lose
>     all the data.
>     The second thing is that if we hide a temporary empty location, we
>     destroy the important concept of the first infobar which explains
>     data hierarchy...
>     Or we can have two concepts - one for the first-time startup, I
>     think this concept of "home/grassdata" default grassdb works
> 
> Yes that's what I meant, the temporary one would be only for these 
> cases, when you can't use the last used mapset.

But then this temporary location should be read-only, or ?

petrasovaa:
> But then this temporary location should be read-only, or ?

Theoretically yes, but I would be worried there might be some technical issues with this.

HuidaeCho:
@lindakladivova Please check this minor refactoring in #1428 for potential conflicts.

lindakladivova:
Guys! :-) I think we have the solution which we came up with yesterday on a video call with Anna and Vashek.
We will have two concepts.

1. The first one is the concept of default location (EPSG:4326) with world map and data structure infobar which is strictly related to first-time user experience. If the new user starts GRASS he will be placed into `$HOME/grassdata` with prepared location in WGS84. 

2. The second concept is the temporary location that will be used in other cases when mapset not usable (deleted, owned by different user, locked). 
In this situation, we create a temporary direction in temporary Session dir (my path on Linux is e.g. `computer/tmp/grass7-linwe-4427`). It is exactly the same place where GISRC file for a particular session is created. This temporary dir is only for a given session and after closing the GRASS, this directory is deleted. The location will be called "temporary location". There is a bit of risk that a user could lose their data, but I think that the name "temporary location" is quite clear :-) and also the need for creating a different location will be emphasized in the relevant Infobar.

What do you think?

HuidaeCho:
> 
> 
> Guys! :-) I think we have the solution which we came up with yesterday on a video call with Anna and Vashek.
> We will have two concepts.
> 
>     1. The first one is the concept of default location (EPSG:4326) with world map and data structure infobar which is strictly related to first-time user experience. If the new user starts GRASS he will be placed into `$HOME/grassdata` with prepared location in WGS84.
> 
>     2. The second concept is the temporary location that will be used in other cases when mapset not usable (deleted, owned by different user, locked).
>        In this situation, we create a temporary location in temporary Session dir (my path on Linux is e.g. `computer/tmp/grass7-linwe-4427`). It is exactly the same place where GISRC file for a particular session is created. This temporary dir is only for a given session and after closing the GRASS, this directory is deleted. The location will be called "temporary location". There is a bit of risk that a user could lose their data, but I think that the name "temporary location" is quite clear :-) and also the need for creating a different location will be emphasized in the relevant Infobar.
> 
> 
> What do you think?

That means, if there is no global rc (e.g., `$HOME/.grass7/rc`), this is the first-time run and create a non-temporary world dataset in hardcoded `$HOME/grassdata` database. If there is already global rc, create a temporary world database in e.g., `/var/tmp/grass7-user-XXXX/`. In other words, once you used GRASS, you'll never end up in the `$HOME/grassdata` database unless you decide to use it on a regular basis? And you'll remove the `default_gisdbase` setting from the global rc?

lindakladivova:
> 
> 
> > Guys! :-) I think we have the solution which we came up with yesterday on a video call with Anna and Vashek.
> > We will have two concepts.
> > ```
> > 1. The first one is the concept of default location (EPSG:4326) with world map and data structure infobar which is strictly related to first-time user experience. If the new user starts GRASS he will be placed into `$HOME/grassdata` with prepared location in WGS84.
> > 
> > 2. The second concept is the temporary location that will be used in other cases when mapset not usable (deleted, owned by different user, locked).
> >    In this situation, we create a temporary location in temporary Session dir (my path on Linux is e.g. `computer/tmp/grass7-linwe-4427`). It is exactly the same place where GISRC file for a particular session is created. This temporary dir is only for a given session and after closing the GRASS, this directory is deleted. The location will be called "temporary location". There is a bit of risk that a user could lose their data, but I think that the name "temporary location" is quite clear :-) and also the need for creating a different location will be emphasized in the relevant Infobar.
> > ```
> > 
> > 
> > What do you think?
> 
> That means, if there is no global rc (e.g., `$HOME/.grass7/rc`), this is the first-time run and create a non-temporary world dataset in hardcoded `$HOME/grassdata` database. If there is already global rc, create a temporary world database in e.g., `/var/tmp/grass7-user-XXXX/`. In other words, once you used GRASS, you'll never end up in the `$HOME/grassdata` database unless you decide to use it on a regular basis? And you'll remove the `default_gisdbase` setting from the global rc?

Yes, exactly. And we will also not be bothered by user_1, user_2 mapsets,... because for each session we have a special temporary dir with a special location.

gui/wxpython/datacatalog/catalog.py
```diff
@@ -135,6 +163,13 @@ def OnDownloadLocation(self, event):
         db_node, loc_node, mapset_node = self.tree.GetCurrentDbLocationMapsetNode()
         self.tree.DownloadLocation(db_node)
 
+    def OnSwitchMapset(self, event):
+        """Switch to given mapset"""
+        last_mapset_path = read_gisrc()['LAST_MAPSET_PATH']
```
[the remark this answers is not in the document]
lindakladivova: I will change it, but I think GUI is not under Black yet. I had problems only with grass.py where I applied Black.

gui/wxpython/datacatalog/catalog.py
```diff
@@ -85,6 +104,15 @@ def _layout(self):
...
+        self.infoManager.ShowInvalidMapsetInfo()
+
+    def showDifferentMapsetOwnerInfo(self):
+        self.infoManager.ShowDifferentMapsetOwnerInfo()
+
+    def showLockedMapsetInfo(self):
+        self.infoManager.ShowLockedMapsetInfo(self.OnSwitchMapset)
+
     def showImportDataInfo(self):
         self.infoManager.ShowImportDataInfo(self.OnImportOgrLayers, self.OnImportGdalLayers)
```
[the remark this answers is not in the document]
lindakladivova: I think that the ones with buttons can be separated functions. It is more well-arranged. I let showLockedMapsetInfo as function; showInvalidMapsetInfo, showDifferentMapsetOwnerInfo are not functions anymore.

lindakladivova:
@wenzeslaus There is a problem with Black vs. Flake8. On row 875 in grass.py.  Flake8 says W503 line break before binary operator. But Black wants `and` at the beginning...

lindakladivova:
So, how it works now. :-) It is quite simple. 
We have the default location EPSG:4326 which is used only for the first-time user. In all other cases of non_standard startup, there is a temporary location. 
I also corrected the state when we are in the temporary location and we switch to non-temporary - in this case, our location parameters are written to RC file, so we can start the next session in the last used mapset if usable.

We can have several situations:

**The first first-time user startup is the same as above.**

**Locked mapset - first-time user**
last_used_mapset_locked

**Locked mapset - normal user**
normal_lock

**Invalid mapset**
invalid_mapset

**Mapset owned by a different user**
different_user_mapset

python/grass/grassdb/checks.py
```diff
@@ -169,6 +173,30 @@ def can_start_in_mapset(mapset_path, ignore_lock=False):
...
+    if mapset_path is None:
+        return
+    message = None
+
+    # Check whether mapset is valid
+    if not is_mapset_valid(mapset_path):
+        message = _("invalid")
+    # Check whether mapset is owned by current user
+    elif not is_current_user_mapset_owner(mapset_path):
+        message = _("owned by different user")
```
wenzeslaus, line 193: Here, the `_()` is actually harmful since you are using it for comparison (and even not using `_()` there). All these are just internal constants or flags, so they should not be translated and they should look like ones, so this one can be something like `"different-owner"`. Sure, it can be the whole things with spaces and what not, but that be quite counter-productive. The function doc should also say the reason is indicated by these strings. It is not a message, or at least not meant to be used as one or in one unlike the other functions. Maybe the name should be different, like `reason_id`? I'm not sure.

wenzeslaus (changes requested):
The message does not disappear and the temporary location is not deleted after switching in the "invalid" case. Know issue? Intention?

gui/wxpython/startup/guiutils.py
```diff
@@ -138,6 +140,29 @@ def read_gisrc():
...
+    """
+    Check if a user is a first-time user.
+    (User is in default location and gisrc file has initial settings.)
+    """
+    if "LAST_MAPSET_PATH" in read_gisrc().keys():
+        initial_gisrc = read_gisrc()["LAST_MAPSET_PATH"] == os.path.join(
+            os.getcwd(), "<UNKNOWN>", "<UNKNOWN>"
+        )
+        return initial_gisrc and is_current_mapset_in_default_location()
+    return False
```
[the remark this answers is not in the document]
lindakladivova: Ok. But I am not sure if read_gisrc() function in guiutils is placed meaningfully. Furthermore, as I looked into it correctly, this function has not been in use till now. 
Anyway, I have made the strings.py file with these strings.

python/grass/grassdb/checks.py
lindakladivova: Makes sense. I like` reason_id`.

mlennert:
On 14/03/21 12:11, Linda Kladivova wrote:
> So, how it works now. :-) It is quite simple.
> We have the default location EPSG:4326 which is used only for the 
> first-time user. In all other cases of non_standard startup, there is a 
> temporary location.
> I also corrected the state when we are in the temporary location and we 
> switch to non-temporary - in this case, our location parameters are 
> written to RC file, so we can start the next session in the last used 
> mapset if usable.
> 
> We can have several situations:
> 
> *The first first-time user startup is the same as above.*
> 
> *Locked mapset - first-time user*
> last_used_mapset_locked 
> <https://user-images.githubusercontent.com/49241681/111065397-c1319380-84b9-11eb-9b8e-e287456ff081.PNG>

For new users, a few words about why a mapset is locked and what 
removing .gislock implies might be helpful, no ? Or a link to a place 
that explains this.

gui/wxpython/datacatalog/infomanager.py
```diff
@@ -41,21 +46,77 @@ def ShowDataStructureInfo(self, onCreateLocationHandler):
...
-        ).format(loc=gisenv()['LOCATION_NAME'])
+        ).format(loc=gisenv()["LOCATION_NAME"])
+        self.infoBar.ShowMessage(message, wx.ICON_INFORMATION, buttons)
+
+    def ShowInvalidMapsetInfo(self):
+        """Show info when last used mapset is invalid"""
+        last_used_mapset_path = read_gisrc()["LAST_MAPSET_PATH"]
+        lastdb, lastloc, lastmapset = split_mapset_path(last_used_mapset_path)
+        message = _(
+            "Last used mapset in path '{lastdb}/{lastloc}/{lastmapset}' is invalid."
```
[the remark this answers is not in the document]
lindakladivova: ok, but we have two different situations - mapset invalid and mapset does not exist. What do you think, do we need to separate messages for these cases, or shell I just let one function with not exist case (supposing that an invalid case should not be as common)?

lindakladivova:
> 
> 
> The message does not disappear and the temporary location is not deleted after switching in the "invalid" case. Know issue? Intention?

Good point @wenzeslaus . But I am not sure how to propagate the disappear message event... I will have a look at it in the next commit, not in 72eab77d8bc73dc7e8672deb77484f723a8fbd12

lindakladivova:
> 
> On 14/03/21 12:11, Linda Kladivova wrote: So, how it works now. :-) It is quite simple. We have the default location EPSG:4326 which is used only for the first-time user. In all other cases of non_standard startup, there is a temporary location. I also corrected the state when we are in the temporary location and we switch to non-temporary - in this case, our location parameters are written to RC file, so we can start the next session in the last used mapset if usable. We can have several situations: *The first first-time user startup is the same as above.* *Locked mapset - first-time user* last_used_mapset_locked <https://user-images.githubusercontent.com/49241681/111065397-c1319380-84b9-11eb-9b8e-e287456ff081.PNG>
> For new users, a few words about why a mapset is locked and what removing .gislock implies might be helpful, no ? Or a link to a place that explains this.

I could add a button with more info? ... but I have not found any source that explains it..  have you @mlennert ?

python/grass/grassdb/session.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+
+import os
+import sys
+
+# TODO move SetSessionMapset and GetVersion from guiutils here
+# TODO add read_gisrc and write_gisrcrc and other functions from grass.py
+
+
+def read_gisrc():
```
wenzeslaus, line 19: I hope this function is not needed anymore. See my other comments.

gui/wxpython/datacatalog/infomanager.py
```diff
@@ -41,22 +45,71 @@ def ShowDataStructureInfo(self, onCreateLocationHandler):
...
+            "GRASS GIS has started in a temporary Location {loc}. "
+            "To continue, find or create another location through "
+            "Data Catalog below, or remove .gislock and switch to last used mapset."
+        ).format(
+            mapsetpath=last_used_mapset_path,
+            loc=gisenv()["LOCATION_NAME"],
+        )
+        self.infoBar.ShowMessage(message, wx.ICON_INFORMATION, buttons)
+
+    def _getString(self, reason_id):
```
wenzeslaus, line 95: What about `text_from_reason_id`?

gui/wxpython/datacatalog/catalog.py
```diff
@@ -29,7 +29,11 @@
 
 from grass.pydispatch.signal import Signal
 
-from grass.grassdb.checks import is_current_mapset_in_demolocation
+from grass.grassdb.session import read_gisrc
+from grass.grassdb.manage import split_mapset_path
+from grass.grassdb.checks import (get_reason_mapset_not_usable,
+                                  is_nonstandard_startup,
+                                  is_first_time_user)
```
wenzeslaus, line 36: Since you used the Black style elsewhere...

wenzeslaus (changes requested):

python/grass/grassdb/session.py
```diff
@@ -0,0 +1,47 @@
...
+for details.
+
+.. sectionauthor:: Linda Kladivova <linda.kladivova seznam cz>
+"""
+
+import os
+import sys
+
+# TODO move SetSessionMapset and GetVersion from guiutils here
+# TODO add read_gisrc and write_gisrcrc and other functions from grass.py
```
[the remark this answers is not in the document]
lindakladivova: I think, in future, we would need to hive off more functions from grass.py regarding gisrc and gisrcrc,
But I agree that we do not have to do it now,

python/grass/grassdb/checks.py
```diff
@@ -235,24 +206,24 @@ def can_start_in_mapset(mapset_path, ignore_lock=False):
...
-    Returns 3 if mapset is locked.
+    Returns a reason id as string if there is a reason, otherwise None.
     """
     if mapset_path is None:
         return
     reason_id = None
 
+    # Check whether mapset exists
+    if not os.path.exists(mapset_path):
+        reason_id = "non-existent"
```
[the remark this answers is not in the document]
lindakladivova: None is a bit flat, what is None? I simplified it but I think it is good to have for all reasons var, it is conceptually clearer.

python/grass/grassdb/checks.py
```diff
@@ -169,6 +203,30 @@ def can_start_in_mapset(mapset_path, ignore_lock=False):
     return True
 
 
+def get_reason_mapset_not_usable(mapset_path):
```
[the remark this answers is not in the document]
lindakladivova: Ok, I changed to `get_reason_id_mapset_not_usable ` :-)

python/grass/grassdb/checks.py
```diff
@@ -115,8 +117,40 @@ def get_mapset_owner(mapset_path):
         return None
 
 
-def is_current_mapset_in_demolocation():
-    return gisenv()["LOCATION_NAME"] == "world_latlong_wgs84"
+def is_nonstandard_startup():
```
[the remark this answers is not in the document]
lindakladivova: What about `is_backup_startup`?

python/grass/grassdb/session.py
```diff
@@ -0,0 +1,47 @@
...
+License (>=v2). Read the file COPYING that comes with GRASS
+for details.
+
+.. sectionauthor:: Linda Kladivova <linda.kladivova seznam cz>
+"""
+
+import os
+import sys
+
+# TODO move SetSessionMapset and GetVersion from guiutils here
```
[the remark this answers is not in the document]
lindakladivova: Ok, I would remove them, what do you think about it?

python/grass/grassdb/checks.py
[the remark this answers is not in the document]
lindakladivova: Mmm, we need to catch non-existent reason at the beginning (due to invalid reason). I will let it as it was.

lindakladivova:
@wenzeslaus I was testing it and it works good, I noticed just one thing - when a user switches inside the temporary location, not sure how to solve this case. Do not remove the temporary location in this case? ...

python/grass/grassdb/checks.py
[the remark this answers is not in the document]
wenzeslaus: What I meant is `return "non-existent"` ... `return "other-reason"` ... and `return None` at the end. Same behavior, just different code.

You do that already anyway when `if mapset_path is None:` where you should probably have explicit `return None` and a comment or doc why it behaves like this in this case.

python/grass/grassdb/checks.py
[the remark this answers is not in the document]
wenzeslaus: I'm not sure about the startup/session/mapset/location part, but  `is_fallback_...()` is the right term here I think.

python/grass/grassdb/session.py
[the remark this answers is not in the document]
wenzeslaus: Yes, eventually, grass.py should be one function call or so.

python/grass/grassdb/session.py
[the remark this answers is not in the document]
wenzeslaus: Yes, please.

wenzeslaus:
> when a user switches inside the temporary location - not sure how to solve this case. Do not remove the temporary location in this case? ...

You mean when the user creates another mapset in the tmp location? I tested that and deleting the location is pretty bad in that case. It should be deleted only when it is not current. There is more questions about how it should behave in this case, but not deleting is the start.

python/grass/grassdb/checks.py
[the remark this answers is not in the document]
lindakladivova: I have chosen the `is_fallback_location`. We have already the default location, so it sounds good. :-)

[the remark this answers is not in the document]
lindakladivova: Or session? mmm I do not know.. maybe we need something more general and the session is more general.

python/grass/grassdb/checks.py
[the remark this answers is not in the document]
lindakladivova: Clear. I get it! :-)

lindakladivova:
> 
> 
> > when a user switches inside the temporary location - not sure how to solve this case. Do not remove the temporary location in this case? ...
> 
> You mean when the user creates another mapset in the tmp location? I tested that and deleting the location is pretty bad in that case. It should be deleted only when it is not current. There is more questions about how it should behave in this case, but not deleting is the start.

> 
> 
> > when a user switches inside the temporary location - not sure how to solve this case. Do not remove the temporary location in this case? ...
> 
> You mean when the user creates another mapset in the tmp location? I tested that and deleting the location is pretty bad in that case. It should be deleted only when it is not current. There is more questions about how it should behave in this case, but not deleting is the start.

I think we can figure that out simply - temporary location can be deleted only if it is no longer current. If a user creates a new mapset in the current temporary location, it is his "fight". We can display another warning infobar but I think it's a bit counterproductive... In any case, the important thing here is that we are not allowed to delete the current location.
@wenzeslaus  please test it if it works for you. I think it is good now.

lindakladivova:
> 
> 
> > > when a user switches inside the temporary location - not sure how to solve this case. Do not remove the temporary location in this case? ...
> > 
> > 
> > You mean when the user creates another mapset in the tmp location? I tested that and deleting the location is pretty bad in that case. It should be deleted only when it is not current. There is more questions about how it should behave in this case, but not deleting is the start.
> 
> > > when a user switches inside the temporary location - not sure how to solve this case. Do not remove the temporary location in this case? ...
> > 
> > 
> > You mean when the user creates another mapset in the tmp location? I tested that and deleting the location is pretty bad in that case. It should be deleted only when it is not current. There is more questions about how it should behave in this case, but not deleting is the start.
> 
> I think we can figure that out simply - temporary location can be deleted only if it is no longer current. If a user creates a new mapset in the current temporary location, it is his "fight". We can display another warning infobar but I think it's a bit counterproductive... In any case, the important thing here is that we are not allowed to delete the current location.
> @wenzeslaus please test it if it works for you. I think it is good now.

lindakladivova:
@petrasovaa btw I have noticed that Download Location is not connected to MapsetChanged - when I download location, I am not switched to it, and infobar still remains (thought on another PR).

python/grass/grassdb/checks.py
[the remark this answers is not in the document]
lindakladivova: I would go with `is_fallback_session`. Difficult decision making.

python/grass/grassdb/checks.py
```diff
@@ -169,6 +202,30 @@ def can_start_in_mapset(mapset_path, ignore_lock=False):
...
 
+def get_reason_id_mapset_not_usable(mapset_path):
+    """It finds a reason why mapset is not usable.
+
+    Returns a reason id as a string.
+    If mapset path is None or no reason found, returns None.
+    """
+    if not mapset_path:
+        return None
+
```
wenzeslaus, line 213: Why this? Most likely not needed.

lib/init/grass.py
```diff
@@ -586,6 +586,21 @@ def read_gisrc(filename):
     return kv
 
 
+def write_gisrcrc(gisrcrc, gisrc, skipped_parameter=None):
```
wenzeslaus, line 589: skipped_parameter -> skip_variable

lib/init/grass.py
```diff
@@ -1284,11 +1249,7 @@ def load_gisrc(gisrc, gisrcrc):
 
     :returns: MapsetSettings object
     """
-    mapset_settings = MapsetSettings()
-    kv = read_gisrc(gisrc)
-    mapset_settings.gisdbase = kv.get("GISDBASE")
-    mapset_settings.location = kv.get("LOCATION_NAME")
-    mapset_settings.mapset = kv.get("MAPSET")
+    mapset_settings = getMapsetSettings(gisrc)
     if not mapset_settings.is_valid():
```
wenzeslaus, line 1253: When it is not valid, your function returns None, not an (invalid) settings, so this like fails. The right solution probably depends on how is_valid is defined.

lib/init/grass.py
```diff
@@ -1276,6 +1229,18 @@ def is_valid(self):
         return self.gisdbase and self.location and self.mapset
 
 
+def getMapsetSettings(gisrc):
```
wenzeslaus, line 1232: getMapsetSettings -> get_mapset_settings

lib/init/grass.py
```diff
@@ -2555,39 +2509,87 @@ def main():
...
+            if is_first_time_user():
+                # Ensure default data hiearchy
+                (
+                    default_gisdbase,
+                    default_location,
+                    default_mapset,
+                ) = ga.ensure_default_data_hierarchy()
+
+                if default_mapset is None:
+                    sys.exit("Failed to start GUI, GRASS GIS is not running.")
```
wenzeslaus, line 2543: The message does not fit the problem.

lib/init/grass.py
```diff
@@ -1307,17 +1268,10 @@ def load_gisrc(gisrc, gisrcrc):
...
-    mapset_settings = MapsetSettings()
-    kv = read_gisrc(gisrc)
-    mapset_settings.gisdbase = kv.get("GISDBASE")
-    mapset_settings.location = kv.get("LOCATION_NAME")
-    mapset_settings.mapset = kv.get("MAPSET")
-    if not mapset_settings.is_valid():
-        return False
     return can_start_in_mapset(
         mapset_path=mapset_settings.full_mapset, ignore_lock=ignore_lock
     )
```
wenzeslaus, line 1277: What is the purpose of `can_start_in_gisrc_mapset` since it doesn't use gisrc after your change?

lib/init/grass.py
```diff
@@ -2555,39 +2509,87 @@ def main():
...
+            else:
+                # Create temporary location
+                params.tmp_location = True
+                set_mapset(
+                    gisrc=gisrc,
+                    geofile="XY",
+                    create_new=True,
+                    tmp_location=params.tmp_location,
+                    tmpdir=tmpdir,
+                )
```
wenzeslaus, line 2584: You define the XY part in two places. Ideally restructure this code: 1) The ifs are too long to navigate. 2) The whole "not-usable" block adds too many lines to main(). What about a separate function?

wenzeslaus (changes requested):

python/grass/grassdb/checks.py
```diff
@@ -115,8 +116,40 @@ def get_mapset_owner(mapset_path):
...
-def is_current_mapset_in_demolocation():
-    return gisenv()["LOCATION_NAME"] == "world_latlong_wgs84"
+def is_fallback_session():
+    """Checks if a user encounters a fallback GRASS session.
+
+    Returns True if a user encounters a fallback session.
+    It occurs when a last mapset is not usable and at the same time
+    a user is in a temporary location.
+    """
+    if "LAST_MAPSET_PATH" in gisenv().keys():
```
petrasovaa, line 126: Didn't test, but I believe this can be shortened:
```suggestion
    if "LAST_MAPSET_PATH" in gisenv():
```

gui/wxpython/lmgr/frame.py
```diff
@@ -1122,16 +1122,30 @@ def OnChangeMapset(self, event):
...
+        if fallback_session:
+            gisenv = grass.gisenv()
+            grassdb = os.environ["TMPDIR"]
+            location = cfg.temporary_location
+            if not grassdb == gisenv["GISDBASE"]:
+                # Delete temporary location
+                delete_location(grassdb, location)
+                self._giface.grassdbChanged.emit(
+                    location=location, grassdb=grassdb, action="delete", element="grassdb"
+                )
```
petrasovaa, line 1145: Could we handle this in switch_mapset_interactively directly and not introduce the fallback_session parameter everywhere? Would then be the only problem the dismissing of the infobar? The info bar could connect the mapset changed signal and dismiss itself, but maybe I am missing something (possibly some problem with first time infobars?)

petrasovaa (changes requested):

lib/init/grass.py
lindakladivova: None. I removed this function and used directly `can_start_in_mapset`.

python/grass/grassdb/checks.py
lindakladivova: It does not work like that... I let it as it was.

lib/init/grass.py
lindakladivova: Well, I know that it is hard to navigate.. but there is quite a lot of cases. We have first-time user, then first-time user with fallback session (e.g. second startup), and then normal fallback session...
I tried to simplify it a bit and defined a new function `create_fallback_session`. But I am afraid that "ifs" cannot be more simplified than I did it now. Maybe you will have better idea how to do it.

mlennert:
On 1/04/21 03:22, Vaclav Petras wrote:
>     when a user switches inside the temporary location - not sure how to
>     solve this case. Do not remove the temporary location in this case? ...
> 
> You mean when the user creates another mapset in the tmp location? I 
> tested that and deleting the location is pretty bad in that case. It 
> should be deleted only when it is not current. There is more questions 
> about how it should behave in this case, but not deleting is the start.
> 
I thought the idea was that the tmp location is read-only ?

mlennert:
On 30/03/21 20:11, Linda Kladivova wrote:
>     On 14/03/21 12:11, Linda Kladivova wrote: So, how it works now. :-)
>     It is quite simple. We have the default location EPSG:4326 which is
>     used only for the first-time user. In all other cases of
>     non_standard startup, there is a temporary location. I also
>     corrected the state when we are in the temporary location and we
>     switch to non-temporary - in this case, our location parameters are
>     written to RC file, so we can start the next session in the last
>     used mapset if usable. We can have several situations: /The first
>     first-time user startup is the same as above./ /Locked mapset -
>     first-time user/ last_used_mapset_locked
>     https://user-images.githubusercontent.com/49241681/111065397-c1319380-84b9-11eb-9b8e-e287456ff081.PNG
>     <https://user-images.githubusercontent.com/49241681/111065397-c1319380-84b9-11eb-9b8e-e287456ff081.PNG>
>     For new users, a few words about why a mapset is locked and what
>     removing .gislock implies might be helpful, no ? Or a link to a
>     place that explains this.
> 
> I could add a button with more info? ... but I have not found any source 
> that explains it.. have you @mlennert <https://github.com/mlennert> ?
> 

No, I don't think there is a good source. We should probably create some 
info somewhere (possibly in the GRASS startup manual ?).

gui/wxpython/lmgr/frame.py
lindakladivova: I think that it is okay to dismiss infobar in case of mapset changes. I did it as you suggested. Just one thing - the second infobar for data import is not this case of mapset changed so now it is not being dismissed... This issue could be solved by third ShowImportSuccessfulInfo message in https://github.com/OSGeo/grass/pull/1205.

petrasovaa:
> @petrasovaa btw I have noticed that Download Location is not connected to `self._giface.grassdbChanged` and `self.SwitchMapset` - when I download location, I am not switched to it, and infobar still remains (thought on another PR).

I don't think we wanted to automatically switch, it's not clear to which mapset you would want to switch.

gui/wxpython/startup/guiutils.py
```diff
@@ -715,7 +657,7 @@ def import_file(guiparent, filePath, env):
 
 
 def switch_mapset_interactively(guiparent, giface, dbase, location, mapset,
-                                show_confirmation=False):
+                                fallback_session=False, show_confirmation=False):
```
petrasovaa, line 660: You don't need the parameter, you can just call is_fallback_session right here

lindakladivova:
> 
> 
> > @petrasovaa btw I have noticed that Download Location is not connected to `self._giface.grassdbChanged` and `self.SwitchMapset` - when I download location, I am not switched to it, and infobar still remains (thought on another PR).
> 
> I don't think we wanted to automatically switch, it's not clear to which mapset you would want to switch.
To permanent mapset in downloaded location?

petrasovaa:
> > I don't think we wanted to automatically switch, it's not clear to which mapset you would want to switch.
> 
> To permanent mapset in downloaded location?

Typically not, PERMANENT contains data you don't want to mess with, so for example I would create a new mapset there.

gui/wxpython/datacatalog/infomanager.py
```diff
@@ -31,8 +32,10 @@ def __init__(self, infobar, giface):
 
     def ShowDataStructureInfo(self, onCreateLocationHandler):
         """Show info about the data hierarchy focused on the first-time user"""
-        buttons = [("Create new Location", onCreateLocationHandler),
-                   ("Learn More", self._onLearnMore)]
+        buttons = [
+            ("Create new Location", onCreateLocationHandler),
+            ("Learn More", self._onLearnMore),
```
wenzeslaus, line 37: 1) These two should be translatable.

2) "Learn More" -> "Learn more"

gui/wxpython/datacatalog/catalog.py
```diff
@@ -47,6 +51,8 @@ def __init__(self, parent, giface=None, id=wx.ID_ANY,
 
         Debug.msg(1, "DataCatalog.__init__()")
 
+        delay = 2000
```
wenzeslaus, line 54: The scope of this variable is unnecessary large. Why not to define it right before you need it for the info bars. It would help identifying what delay is that.

gui/wxpython/datacatalog/infomanager.py
```diff
@@ -41,22 +44,66 @@ def ShowDataStructureInfo(self, onCreateLocationHandler):
...
             "Mapsets. You can create new Mapsets for different tasks by right "
             "clicking on the Location name.\n\n"
             "To import data, go to the toolbar above or use the buttons below."
-        ).format(loc=gisenv()['LOCATION_NAME'])
+        ).format(loc=gisenv()["LOCATION_NAME"])
         self.infoBar.ShowMessage(message, wx.ICON_INFORMATION, buttons)
 
+    def ShowFallbackSessionInfo(self, reason_id):
+        """Show info when last used mapset is not usable"""
+        string = self._text_from_reason_id(reason_id)
```
wenzeslaus, line 68: string -> reason

gui/wxpython/datacatalog/tree.py
```diff
@@ -1498,6 +1497,7 @@ def SwitchMapset(self, grassdb, location, mapset, show_confirmation=False):
         """
         Switch to location and mapset interactively.
         """
+
```
wenzeslaus, line 1500: Not many programmers know this, but Python does not require an empty line between every two statements.

lib/init/grass.py
```diff
@@ -2555,39 +2512,71 @@ def main():
...
+                    default_mapset,
+                    default_mapset_path,
+                ) = ensure_default_data_hierarchy()
+
+                if not default_gisdbase:
+                    sys.exit("Failed to start GRASS GUI, no grassdata directory found.")
+                elif not default_location:
+                    sys.exit(
+                        "Failed to start GRASS GUI, no default location to copy in the installation or copying failed."
+                    )
```
wenzeslaus, line 2551: 1) The original messages were passed to fatal. Why not these?

2) The original messages were translatable, so these can be translatable too, no?

3) GRASS GUI -> GRASS GIS

4) "no grassdata directory found": If none is found, a new one is created, isn't it? So some "...and cannot be created" is needed.

python/grass/grassdb/manage.py
```diff
@@ -46,3 +46,10 @@ def rename_mapset(database, location, old_name, new_name):
 def rename_location(database, old_name, new_name):
     """Rename location from *old_name* to *new_name*"""
     os.rename(os.path.join(database, old_name), os.path.join(database, new_name))
+
+
+def split_mapset_path(mapset_path):
+    """Split mapset path to three parts - grassdb, location, mapset"""
+    path, mapset = os.path.split(mapset_path)
+    grassdb, location = os.path.split(path)
+    return grassdb, location, mapset
```
petrasovaa, line 55: I think this should handle the case when you have slash at the end:
```
>>> split_mapset_path('/home/anna/grassdata/ncspm/anna')
('/home/anna/grassdata', 'ncspm', 'anna')
>>> split_mapset_path('/home/anna/grassdata/ncspm/anna/')
('/home/anna/grassdata/ncspm', 'anna', '')
```

```suggestion
def split_mapset_path(mapset_path):
    """Split mapset path to three parts - grassdb, location, mapset"""
    path, mapset = os.path.split(mapset_path.rstrip(os.sep))
    grassdb, location = os.path.split(path)
    return grassdb, location, mapset
```

wenzeslaus (changes requested):

petrasovaa (changes requested):
When testing with --text mode, when last used mapset is in use, it will jump to tmploc instead of failing as before.

lib/init/grass.py
lindakladivova: Changed to:
```
if not default_gisdbase:
  fatal(
      _(
          "Failed to start GRASS GIS, grassdata directory could not be found or created."
      )
  )
elif not default_location:
  fatal(
      _(
          "Failed to start GRASS GIS, no default location to copy in the installation or copying failed."
      )
  )
```

lindakladivova:
> 
> 
> When testing with --text mode, when last used mapset is in use, it will jump to tmploc instead of failing as before.

So we would prefer the failing as before or you think it is correct to jump to tmploc? Maybe, we should have the same behavior everywhere...

gui/wxpython/datacatalog/infomanager.py
lindakladivova: Yes, I think that all of the infobar buttons should be translatable...

wenzeslaus:
> But then this temporary location should be read-only, or ?

Let's discuss this in a separate issue if it seems like needed after merging this PR. Current but read-only would be a new concept we don't have any mechanism for.

wenzeslaus:
> > When testing with --text mode, when last used mapset is in use, it will jump to tmploc instead of failing as before.
> 
> So we would prefer the failing as before or you think it is correct to jump to tmploc? Maybe, we should have the same behavior everywhere...

Just to record here what we discussed elsewhere: There are different expectation for the CLI behavior than for GUI, so it makes sense here to keep the current `--text` behavior, i.e., fail, not create a temporary location. It is easy enough in command line to correct that situation just by running a different command.

gui/wxpython/startup/guiutils.py
```diff
@@ -751,4 +697,16 @@ def switch_mapset_interactively(guiparent, giface, dbase, location, mapset,
...
+                                             location=None,
+                                             mapset=mapset)
+
+    if fallback_session:
+        tmp_dbase = os.environ["TMPDIR"]
+        tmp_loc = cfg.temporary_location
+        if tmp_dbase != gisenv()["GISDBASE"]:
+            # Delete temporary location
+            delete_location(tmp_dbase, tmp_loc)
+            giface.grassdbChanged.emit(
```
wenzeslaus, line 710: Note here in a comment that we pretend that the whole database for deleted because we assume it is not useful after deleting the temporary location.

lindakladivova:
> 
> 
> > > When testing with --text mode, when last used mapset is in use, it will jump to tmploc instead of failing as before.
> > 
> > 
> > So we would prefer the failing as before or you think it is correct to jump to tmploc? Maybe, we should have the same behavior everywhere...
> 
> Just to record here what we discussed elsewhere: There are different expectation for the CLI behavior than for GUI, so it makes sense here to keep the current `--text` behavior, i.e., fail, not create a temporary location. It is easy enough in command line to correct that situation just by running a different command.

Yep, clear. I have corrected the behavior in this situation.

wenzeslaus (approved):
I agree with the code after my fixes to checking behavior. --text with broken mapset now gives the detailed message with reasons as it does with explicit mapset in command line, not just mapset invalid. For consistency, safety, and possibly future all paths now lead to a set_mapset call.

The code needs refactoring more than it did before and first time user gets the WGS84 location even in the command line. I suggest that this and any other issues are left for new issues/PRs.

petrasovaa (approved):

wenzeslaus:
Merged, but there is couple of loose ends which will need to be addressed in subsequent PRs, most importantly what to do with `--text` for first time user and the whole `--gtext` option.

PR #1404, opened 2021-02-25
r.region: Vect_get_map_box() requires level 2

HuidaeCho:


neteler:
Why would this change to "topology required" be needed?

HuidaeCho:
@neteler Without this PR, `Vect_get_map_box()` returns 0, 0, 0, 0 in `box` at line 180 `Vect_get_map_box(&Map, &box);`. The API doc for `Vect_get_map_box()` says that it requires level 2.

HuidaeCho:
Maybe, https://github.com/OSGeo/grass/commit/ae625bf48a603515f60ca2702a6dc0a324916ed4 is not enough?

HuidaeCho:
If "Vect_get_map_box(): recover old behaviors, output box is filled even on level1 (see v.info/level1.c)" really happened, the function would have to return 1, not 0. So that commit itself needs to change either `return 0` to `return 1` if filling the output box really works or change the API doc saying return 0 on level 2 and 1 on level 1 and there will be no failure?

HuidaeCho:
In my testing, Vect_get_map_box() didn't fill the output box.

metzm (approved):
The change is correct.

An addition would be to determine the vector's bbox on the fly if it is not available. See [v.info/level1.c](https://github.com/OSGeo/grass/blob/master/vector/v.info/level1.c)

HuidaeCho:
> The change is correct.
> 
> An addition would be to determine the vector's bbox on the fly if it is not available. See [v.info/level1.c](https://github.com/OSGeo/grass/blob/master/vector/v.info/level1.c)

Does it mean https://github.com/OSGeo/grass/commit/ae625bf48a603515f60ca2702a6dc0a324916ed4 didn't work and was unnecessary? Actually, this commit has two problems as I mentioned above:
1. If it worked as expected, it should return 1 only because that function is supposed to work for both levels 0 and 1. Unless we change the API doc: 0 for successful level 2 return and 1 for successful level 1 return. No failure?
2. Now, we learned that this commit doesn't work as intended, can we safely revert it to avoid an unnecessary call to `Vect_box_copy(Box, &(Plus->box));`? Or does it handle a special case that is not the issue this PR tries to fix?

metzm:
> > The change is correct.
> > An addition would be to determine the vector's bbox on the fly if it is not available. See [v.info/level1.c](https://github.com/OSGeo/grass/blob/master/vector/v.info/level1.c)
> 
> Does it mean [ae625bf](https://github.com/OSGeo/grass/commit/ae625bf48a603515f60ca2702a6dc0a324916ed4) didn't work and was unnecessary? 

This change was indeed strange. The bbox is set, but `Vect_get_map_box()` still returns 0 = error. `Vect_get_map_box()` should not even attempt to access topology information if the level is < 2.

And r.region should use the method of `v.info` to determine the bbox of a vector if topology is not available.

HuidaeCho:
> > > The change is correct.
> > > An addition would be to determine the vector's bbox on the fly if it is not available. See [v.info/level1.c](https://github.com/OSGeo/grass/blob/master/vector/v.info/level1.c)
> > 
> > 
> > Does it mean [ae625bf](https://github.com/OSGeo/grass/commit/ae625bf48a603515f60ca2702a6dc0a324916ed4) didn't work and was unnecessary?
> 
> This change was indeed strange. The bbox is set, but `Vect_get_map_box()` still returns 0 = error. `Vect_get_map_box()` should not even attempt to access topology information if the level is < 2.
> 
> And r.region should use the method of `v.info` to determine the bbox of a vector if topology is not available.

`v.info/level1.c` also populates the number of features, but that's not necessary for `r.gegion`. `Vect_get_map_box1()` is good enough. I think we should move `level_one_info()` to `Vlib/level_one.c` and name it `Vect_get_level1_info()`. Also, `Cats` in this function is never used once filled in.

HuidaeCho:
> > > > The change is correct.
> > > > An addition would be to determine the vector's bbox on the fly if it is not available. See [v.info/level1.c](https://github.com/OSGeo/grass/blob/master/vector/v.info/level1.c)
> > > 
> > > 
> > > Does it mean [ae625bf](https://github.com/OSGeo/grass/commit/ae625bf48a603515f60ca2702a6dc0a324916ed4) didn't work and was unnecessary?
> > 
> > 
> > This change was indeed strange. The bbox is set, but `Vect_get_map_box()` still returns 0 = error. `Vect_get_map_box()` should not even attempt to access topology information if the level is < 2.
> > And r.region should use the method of `v.info` to determine the bbox of a vector if topology is not available.
> 
> `v.info/level1.c` also populates the number of features, but that's not necessary for `r.gegion`. `Vect_get_map_box1()` is good enough. I think we should move `level_one_info()` to `Vlib/level_one.c` and name it `Vect_get_level1_info()`. Also, `Cats` in this function is never used once filled in.

This change is for another PR. For now, I'll merge it.

PR #1406, opened 2021-02-26
Fix compiler warnings, part 12

nilason:
Fixes compiler warnings:

- -Wformat-overflow=

Twelfth part addressing #1247.

- lib/db/dbmi_base
- lib/imagery
- lib/symbol
- lib/vector/Vlib
- d.linegraph
- d.vect
- d.vect.thematic
- i.maxlik
- i.rectify
- ps.map
- r.li.daemon
- r.viewshed
- v.info
- v.lidar.correction
- v.lidar.edgedetection
- v.lidar.growing
- v.lidar.growing
- v.outlier
- v.surf.bspline
- v.to.db

### DIFF
--- a/display/d.linegraph/main.c
+++ b/display/d.linegraph/main.c
@@ -89,7 +89,7 @@ static int cmp(const void *a, const void *b)
 static char *icon_files(void)
 {
     char **list, *ret;
-    char buf[GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
+    char buf[GNAME_MAX + GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
     int i, count;
     size_t len;
     DIR *dir, *dir_i;
@@ -200,9 +200,9 @@ int main(int argc, char **argv)
     float xscale;
     float yscale;
 
-    char txt[1024], xlabel[512];
+    char txt[1024];
     char tic_name[1024];
-    char *name;
+    char *name, *xlabel;
     char color_name[20];
 
     FILE *fopen();
@@ -898,7 +898,7 @@ int main(int argc, char **argv)
     if ((strcmp(title[0]->answer, "") == 0) && (strcmp(tic_name, "") == 0))
         *xlabel = '\0';
     else
-        sprintf(xlabel, "X: %s %s", title[0]->answer, tic_name);
+        G_asprintf(&xlabel, "X: %s %s", title[0]->answer, tic_name);
     text_height = (b - t) * TEXT_HEIGHT;
     text_width = (r - l) * TEXT_WIDTH * 1.5;
     D_text_size(text_width, text_height);
@@ -1010,7 +1010,7 @@ int main(int argc, char **argv)
     if ((strcmp(title[1]->answer, "") == 0) && (strcmp(tic_name, "") == 0))
         *xlabel = '\0';
     else
-        sprintf(xlabel, "Y: %s %s", title[1]->answer, tic_name);
+        G_asprintf(&xlabel, "Y: %s %s", title[1]->answer, tic_name);
     text_height = (b - t) * TEXT_HEIGHT;
     text_width = (r - l) * TEXT_WIDTH * 1.5;
     D_text_size(text_width, text_height);
--- a/display/d.vect.thematic/main.c
+++ b/display/d.vect.thematic/main.c
@@ -647,7 +647,7 @@ int cmp(const void *a, const void *b)
 char *icon_files(void)
 {
     char **list, *ret;
-    char buf[GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
+    char buf[GNAME_MAX + GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
     int i, count;
     size_t len;
     DIR *dir, *dir_i;
--- a/display/d.vect/main.c
+++ b/display/d.vect/main.c
@@ -516,7 +516,7 @@ int cmp(const void *a, const void *b)
 char *icon_files(void)
 {
     char **list, *ret;
-    char buf[GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
+    char buf[GNAME_MAX + GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
     int i, count;
     size_t len;
     DIR *dir, *dir_i;
--- a/imagery/i.maxlik/main.c
+++ b/imagery/i.maxlik/main.c
@@ -141,13 +141,13 @@ int main(int argc, char *argv[])
     Rast_free_cats(&cats);
 
     if (reject_fd > 0) {
-	char title[100];
+        char *title = NULL;
 	char *label[] = { "no data", "0.1%", "0.5%",
 			  "1%", "2%", "5%", "10%",
 			  "20%", "30%", "50%", "70%",
 			  "80%", "90%", "95%", "98%",
 			  "99%", "100%", "bad" };
-	sprintf(title, "Rejection Probability for %s", class_name);
+        G_asprintf(&title, "Rejection Probability for %s", class_name);
 
 	Rast_init_cats(title, &cats);
 	Rast_set_cats_title(title, &cats);
--- a/lib/symbol/read.c
+++ b/lib/symbol/read.c
@@ -236,7 +236,7 @@ SYMBOL *S_read(const char *sname)
 {
     int i, j, k, l;
     FILE *fp;
-    char group[500], name[500], buf[2001];
+    char group[500], name[500], buf[2001], buf2[2048];
     const char *ms;
     char *c;
     double x, y, x2, y2, rad, ang1, ang2;
@@ -357,8 +357,8 @@ SYMBOL *S_read(const char *sname)
 		sscanf(data, "%lf %lf %lf %lf %lf %c", &x, &y, &rad, &ang1,
 		       &ang2, &clock);
 	    if (ret < 5) {
-		sprintf(buf, "Incorrect arc definition: '%s'", buf);
-		return (err(fp, symb, buf));
+		sprintf(buf2, "Incorrect arc definition: '%s'", buf);
+		return (err(fp, symb, buf2));
 	    }
 	    if (ret == 6 && (clock == 'c' || clock == 'C'))
 		i = 1;
@@ -441,8 +441,8 @@ SYMBOL *S_read(const char *sname)
 	    }
 	}
 	else {
-	    sprintf(buf, "Unknown keyword in symbol: '%s'", buf);
-	    return (err(fp, symb, buf));
+	    sprintf(buf2, "Unknown keyword in symbol: '%s'", buf);
+	    return (err(fp, symb, buf2));
 	    break;
 	}
     }
--- a/lib/vector/Vlib/field.c
+++ b/lib/vector/Vlib/field.c
@@ -352,7 +352,8 @@ struct field_info *Vect_default_field_info(struct Map_info *Map,
 					   int field, const char *field_name, int type)
 {
     struct field_info *fi;
-    char buf[GNAME_MAX], buf2[GNAME_MAX];
+    char buf[GNAME_MAX];
+    char *buf2;
     const char *schema;
     dbConnection connection;
 
@@ -418,8 +419,8 @@ struct field_info *Vect_default_field_info(struct Map_info *Map,
     }
     schema = connection.schemaName;
     if (schema && strlen(schema) > 0) {
-	sprintf(buf2, "%s.%s", schema, buf);
-	fi->table = G_store(buf2);
+        G_asprintf(&buf2, "%s.%s", schema, buf);
+        fi->table = buf2;
     }
     else {
 	fi->table = G_store(buf);
--- a/lib/vector/Vlib/map.c
+++ b/lib/vector/Vlib/map.c
@@ -491,7 +491,13 @@ int Vect__delete(const char *map, int is_tmp)
             (strcmp(ent->d_name, "..") == 0))
             continue;
         
-        sprintf(path_buf, "%s/%s", path, ent->d_name);
+        ret = snprintf(path_buf, GPATH_MAX, "%s/%s", path, ent->d_name);
+        if (ret >= GPATH_MAX) {
+            G_warning(_("Filepath '%s/%s' exceeds max length"), path,
+                      ent->d_name);
+            closedir(dir);
+            return -1;
+        }
         G_debug(3, "delete file '%s'", path_buf);
         ret = unlink(path_buf);
         if (ret == -1) {
--- a/raster/r.li/r.li.daemon/daemon.c
+++ b/raster/r.li/r.li.daemon/daemon.c
@@ -110,13 +110,19 @@ int calculateIndex(char *file, rli_func *f,
 	if (doneDir == -1 && errno != EEXIST)
 	    G_fatal_error(_("Cannot create %s directory"), out);
 
-	/* check if ~/.grass8/r.li/output exists */
-	sprintf(out, "%s%s", rlipath, "output");
+	/* check if ~/.grass7/r.li/output exists */
+        if (snprintf(out, GPATH_MAX, "%s%s", rlipath, "output") >= GPATH_MAX)
+            G_fatal_error(_("Filepath '%s%s' exceeds max length"), rlipath,
+                          "output");
 	doneDir = G_mkdir(out);
 	if (doneDir == -1 && errno != EEXIST)
 	    G_fatal_error(_("Cannot create %s directory"), out);
-	sprintf(out, "%s%s%c%s", rlipath, "output", HOST_DIRSEP, output);
-	res = open(out, O_WRONLY | O_CREAT | O_TRUNC, 0644);
+        if (snprintf(out, GPATH_MAX, "%s%s%c%s", rlipath, "output", HOST_DIRSEP,
+                     output) >= GPATH_MAX)
+            G_fatal_error(_("Filepath '%s%s%c%s' exceeds max length"), rlipath,
+                          "output", HOST_DIRSEP, output);
+        if ((res = open(out, O_WRONLY | O_CREAT | O_TRUNC, 0644)) == -1)
+            G_fatal_error(_("Cannot create %s output"), out);
     }
     i = 0;
 
--- a/vector/v.lidar.correction/main.c
+++ b/vector/v.lidar.correction/main.c
@@ -33,7 +33,7 @@ int main(int argc, char *argv[])
     int nsubregion_col, nsubregion_row;
     int subregion = 0, nsubregions = 0;
     const char *dvr, *db, *mapset;
-    char table_name[GNAME_MAX];
+    char table_name[GNAME_MAX + 64];
     char xname[GNAME_MAX], xmapset[GMAPSET_MAX];
     double lambda, mean, stepN, stepE, HighThresh,
 	LowThresh;
--- a/imagery/i.rectify/cp.c
+++ b/imagery/i.rectify/cp.c
@@ -4,7 +4,7 @@
 
 int get_control_points(struct Image_Group *group, int order)
 {
-    char msg[200];
+    char msg[GNAME_MAX + GMAPSET_MAX + 64];
 
     if (!I_get_control_points(group->name, &group->control_points))
 	exit(0);
--- a/include/grass/iostream/ami_stream.h
+++ b/include/grass/iostream/ami_stream.h
@@ -669,7 +669,7 @@ void AMI_STREAM<T>::persist(persistence p)  {
 // if you are careful.
 template<class T>
 char *AMI_STREAM<T>::sprint()  {
-  static char desc[BUFSIZ];
+  static char desc[BUFSIZ + 256];
   sprintf(desc, "[AMI_STREAM %s %ld]", path, (long)stream_len());
   return desc;
 }
--- a/lib/db/dbmi_base/error.c
+++ b/lib/db/dbmi_base/error.c
@@ -101,7 +101,7 @@ void db_protocol_error(void)
 void db_syserror(const char *s)
 {
     char lead[1024];
-    char msg[1024];
+    char msg[2048];
 
 
     err_flag = 0;
@@ -192,7 +192,7 @@ void db_print_error(void)
 	sprintf(lead, "%s: ", who);
 
     if (user_print_function) {
-	char buf[1024];
+        char buf[2048];
 
 	sprintf(buf, "%s%s\n", lead, err_msg);
 	user_print_function(buf);
--- a/ps/ps.map/r_plt.c
+++ b/ps/ps.map/r_plt.c
@@ -8,7 +8,7 @@
 
 int read_point(double e, double n)
 {
-    char buf[1024], symb[1024];
+    char buf[2048], symb[1024];
     int r, g, b;
     int color_R, color_G, color_B;
     int fcolor_R, fcolor_G, fcolor_B;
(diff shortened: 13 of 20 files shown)


display/d.linegraph/main.c
```diff
@@ -89,7 +89,7 @@ static int cmp(const void *a, const void *b)
 static char *icon_files(void)
 {
     char **list, *ret;
-    char buf[GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
+    char buf[GNAME_MAX + GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
```
metzm, line 92:
```suggestion
    char buf[GPATH_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
```
A path is stored in buf, thus GPATH_MAX seems appropriate.

display/d.linegraph/main.c
```diff
@@ -898,7 +898,7 @@ int main(int argc, char **argv)
     if ((strcmp(title[0]->answer, "") == 0) && (strcmp(tic_name, "") == 0))
         *xlabel = '\0';
     else
-        sprintf(xlabel, "X: %s %s", title[0]->answer, tic_name);
+        sprintf(xlabel, "X: %.250s %.250s", title[0]->answer, tic_name);
```
metzm, line 901: How about
```
char *xlabel = NULL:
...
        G_asprintf(xlabel, "X: %s %s", title[0]->answer, tic_name);
```

display/d.vect.thematic/main.c
```diff
@@ -647,7 +647,7 @@ int cmp(const void *a, const void *b)
 char *icon_files(void)
 {
     char **list, *ret;
-    char buf[GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
+    char buf[GNAME_MAX + GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
```
metzm, line 650: Same code as in `d.linegraph`, a path is stored in buf, thus GPATH_MAX seems appropriate.

display/d.vect/main.c
```diff
@@ -516,7 +516,7 @@ int cmp(const void *a, const void *b)
 char *icon_files(void)
 {
     char **list, *ret;
-    char buf[GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
+    char buf[GNAME_MAX + GNAME_MAX], path[GPATH_MAX], path_i[GPATH_MAX];
```
metzm, line 519: Same code as in d.linegraph, a path is stored in buf, thus GPATH_MAX seems appropriate.

imagery/i.maxlik/main.c
```diff
@@ -141,7 +141,7 @@ int main(int argc, char *argv[])
     Rast_free_cats(&cats);
 
     if (reject_fd > 0) {
-	char title[100];
+	char title[GNAME_MAX + 30];
```
metzm, line 144: A name is not stored in title, thus GNAME_MAX seems not appropriate.

How about
```
char *title = NULL;
...
G_asprintf(&title, "Rejection Probability for %s", class_name);
```
?

lib/imagery/list_gp.c
```diff
@@ -36,7 +36,7 @@ int I_list_group(const char *group, const struct Ref *ref, FILE * fd)
     }
     max = 0;
     for (i = 0; i < ref->nfiles; i++) {
-	sprintf(buf, "<%s@%s>", ref->file[i].name, ref->file[i].mapset);
+        sprintf(buf, "<%.38s@%.38s>", ref->file[i].name, ref->file[i].mapset); /* truncates! */
```
metzm, line 39: Use
```
char *buf = NULL;
...
G_asprintf(&buf, "<%s@%s>", ref->file[i].name, ref->file[i].mapset);
```
This is in a for loop, thus `G_free()` would need to be added.

lib/imagery/list_gp.c
```diff
@@ -45,7 +45,7 @@ int I_list_group(const char *group, const struct Ref *ref, FILE * fd)
     fprintf(fd, "-------------\n");
     tot_len = 0;
     for (i = 0; i < ref->nfiles; i++) {
-	sprintf(buf, "<%s@%s>", ref->file[i].name, ref->file[i].mapset);
+        sprintf(buf, "<%.38s@%.38s>", ref->file[i].name, ref->file[i].mapset); /* truncates! */
```
metzm, line 48: As above, `G_asprintf()` + `G_free()` ?

lib/imagery/list_subgp.c
```diff
@@ -93,7 +93,7 @@ int I_list_subgroup(const char *group,
     }
     max = 0;
     for (i = 0; i < ref->nfiles; i++) {
-	sprintf(buf, "<%s@%s>", ref->file[i].name, ref->file[i].mapset);
+        sprintf(buf, "<%.38s@%.38s>", ref->file[i].name, ref->file[i].mapset); /* truncates! */
```
metzm, line 96: As above, `G_asprintf()` + `G_free()` ?

lib/imagery/list_subgp.c
```diff
@@ -105,7 +105,7 @@ int I_list_subgroup(const char *group,
     fprintf(fd, "-------------\n");
     tot_len = 0;
     for (i = 0; i < ref->nfiles; i++) {
-	sprintf(buf, "<%s@%s>", ref->file[i].name, ref->file[i].mapset);
+        sprintf(buf, "<%.38s@%.38s>", ref->file[i].name, ref->file[i].mapset); /* truncates! */
```
metzm, line 108: As above, `G_asprintf()` + `G_free()` ?

lib/symbol/read.c
```diff
@@ -357,8 +357,8 @@ SYMBOL *S_read(const char *sname)
 		sscanf(data, "%lf %lf %lf %lf %lf %c", &x, &y, &rad, &ang1,
 		       &ang2, &clock);
 	    if (ret < 5) {
-		sprintf(buf, "Incorrect arc definition: '%s'", buf);
-		return (err(fp, symb, buf));
+		sprintf(buf2, "Incorrect arc definition: '%s'", buf); //
```
metzm, line 360:
```suggestion
		sprintf(buf2, "Incorrect arc definition: '%s'", buf);
```
Please no `//` comments in C code.

lib/symbol/read.c
```diff
@@ -441,8 +441,8 @@ SYMBOL *S_read(const char *sname)
 	    }
 	}
 	else {
-	    sprintf(buf, "Unknown keyword in symbol: '%s'", buf);
-	    return (err(fp, symb, buf));
+	    sprintf(buf2, "Unknown keyword in symbol: '%s'", buf); //
```
metzm, line 444:
```suggestion
	    sprintf(buf2, "Unknown keyword in symbol: '%s'", buf);
```
Please no `//` comments in C code.

lib/symbol/read.c
metzm: Any idea how to avoid the below warning about "Uncontrolled format string" ?

lib/vector/Vlib/field.c
```diff
@@ -352,7 +352,7 @@ struct field_info *Vect_default_field_info(struct Map_info *Map,
 					   int field, const char *field_name, int type)
 {
     struct field_info *fi;
-    char buf[GNAME_MAX], buf2[GNAME_MAX];
+    char buf[GNAME_MAX], buf2[GNAME_MAX + GNAME_MAX];
```
metzm, line 355:
```
char *buf2 = NULL;
...
G_asprintf(&buf2, ...);
```
?

lib/vector/Vlib/map.c
```diff
@@ -384,7 +384,7 @@ int Vect_delete(const char *map)
 int Vect__delete(const char *map, int is_tmp)
 {
     int ret;
-    char path[GPATH_MAX], path_buf[GPATH_MAX];
+    char path[GPATH_MAX], path_buf[GPATH_MAX + GPATH_MAX];
```
metzm, line 387: `GPATH_MAX` must not be exceeded, thus please leave it at `GPATH_MAX` and add a check that `path, ent->d_name` do not exceed `GPATH_MAX - 2`

raster/r.li/r.li.daemon/daemon.c
```diff
@@ -39,7 +39,7 @@ int calculateIndex(char *file, rli_func *f,
 		   char **parameters, char *raster, char *output)
 {
 
-    char pathSetup[GPATH_MAX], out[GPATH_MAX], parsed;
+    char pathSetup[GPATH_MAX], out[GPATH_MAX + GPATH_MAX], parsed;
```
metzm, line 42: As before, `GPATH_MAX` can not be exceeded, additional tests are needed.

display/d.linegraph/main.c
nilason: > A path is stored in buf, thus GPATH_MAX seems appropriate.

It is not a path, but a concatenated string of two directory names `[dirname]/[dirname]` like e.g. `basic/arrow`, which is exactly maximum `GNAME_MAX + GNAME_MAX`.

display/d.vect.thematic/main.c
nilason: see https://github.com/OSGeo/grass/pull/1406#discussion_r595225478

display/d.vect/main.c
nilason: see https://github.com/OSGeo/grass/pull/1406#discussion_r595225478

lib/imagery/list_gp.c
nilason: The problem with these `I_list_group()` functions of the  `lib/imagery/list_[...].c` files, is that they result in "nicely" formatted strings, fitting in a 80 column. My solution is probably not the best either, as there is no way to know if truncated or not. In addition, there are two loops (per function), one is for calculating width.

Perhaps I can come up with a solution with added ellipsis in case of truncation?

lib/imagery/list_gp.c
nilason: see https://github.com/OSGeo/grass/pull/1406#discussion_r595233775

lib/imagery/list_subgp.c
nilason: see https://github.com/OSGeo/grass/pull/1406#discussion_r595233775

lib/imagery/list_subgp.c
nilason: see https://github.com/OSGeo/grass/pull/1406#discussion_r595233775

lib/vector/Vlib/map.c
nilason: I updated with a suggestion using `snprintf()`, which is C99 feat, but perhaps give a safer and cleaner code.

raster/r.li/r.li.daemon/daemon.c
nilason: see https://github.com/OSGeo/grass/pull/1406#discussion_r595237414

nilason:
> Any idea how to avoid the below warning about "Uncontrolled format string" ?

I found the cause of that problem and addressed it with e673ee7.

I closed the issues (as resolved) where I updated with your suggestions. There remain some open, which need another round.

lib/imagery/list_gp.c
nilason: Also, my present solution is limited by truncating mapname *and* location separately. Instead their combined length should be accounted for.

display/d.linegraph/main.c
metzm: Plus the `/` char, i.e. `char buf[GNAME_MAX + GNAME_MAX + 1]`?

lib/imagery/list_gp.c
metzm: These (what is printed to buf) are names, thus `GNAME_MAX + GNAME_MAX + 1` (+ 1 for the `@`) might work?
Truncating means removing information and creating ambiguities.

lib/vector/Vlib/map.c
metzm: We have `G_snprintf()`

display/d.linegraph/main.c
nilason: The dirnames input has already included a null character each, now one of them will replaced by a `/` .

metzm:
> > Any idea how to avoid the below warning about "Uncontrolled format string" ?
> 
> I found the cause of that problem and addressed it with [e673ee7](https://github.com/OSGeo/grass/commit/e673ee7cc7321f5816454602bf754d3160eeea72). See: [Code Scanning Results](https://github.com/OSGeo/grass/pull/1406/checks?check_run_id=2122589307)

Nice catch, obvious now that I see it, not obvious to find, great!

vector/v.lidar.correction/main.c
```diff
@@ -34,7 +34,7 @@ int main(int argc, char *argv[])
     int subregion = 0, nsubregions = 0;
     const char *dvr, *db, *mapset;
     char table_name[GNAME_MAX];
-    char xname[GNAME_MAX], xmapset[GMAPSET_MAX];
+    char xname[GNAME_MAX - 5], xmapset[GMAPSET_MAX];
```
metzm, line 37: Why reduce the length of xname? Generally, more space is better?

Why reduce the length of xname even more in ` vector/v.lidar.edgedetection/main.c` and ` vector/v.lidar.growing/main.c` ?

`xname` is always used with `G_name_is_fully_qualified()`

nilason: See line eg https://github.com/OSGeo/grass/blob/777c4e672be4b6fb4504083736259fdeb7bc162f/vector/v.lidar.correction/main.c#L189
table_name is equal in size to xname

lib/vector/Vlib/map.c
nilason: Interestingly  `G_snprintf` calls `vsnprintf`, which is C99 feat.

vector/v.lidar.correction/main.c
nilason: Probably better to use `snprintf()` for these too.

display/d.linegraph/main.c
nilason: Just to clarify the comment above.

`#define GNAME_MAX 256`

"The maximum length of a name in a directory is MAXNAMLEN" ([Darwin](https://github.com/apple/darwin-xnu/blob/main/bsd/sys/dirent.h)):
```
#define MAXNAMLEN   255
struct dirent {
    ...
    char d_name[MAXNAMLEN + 1];
}
```
See also [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=dirent/dirent.h;h=1d1fab7e55cdad4d1196e48a35d7712e4e390f92;hb=refs/heads/master).

Thus:
```
[dirname1] + '/' + [dirname2] + '\0'  =  (GNAME_MAX + GNAME_MAX)
    255       1       255         1   =  (   256    +   256    )
```

vector/v.lidar.correction/main.c
nilason: The problem behind the fixes with reduced xname size have all in common addition of extension to the mapname (which can be `GNAME_MAX`) to the db table name (presently also with size of GNAME_MAX). If there is no limitation to the size of database table name (I couldn't find any), then increasing that seems to be the easier solution.

lib/imagery/list_gp.c
nilason: Just pushed suggestion of an implementation that truncates when needed.
Example of its results with various input:
```
-------------------------------------------------------------------------------| <- 80 char
# Long map name:
ifonlythisisalandsatfilewithaveryverylongnameandagainifonlythisisalandsatfilewithaveryverylongnamePERMANENT
<ifonlythisisalandsatfilewithaveryverylongnameandagainifonly...@PERMANENT>    

# Long mapset name:
shortname@IFONLYTHISISALANDSATFILEWITHAVERYVERYLONGNAMEANDAGAINIFONLYTHISISALANDSATFILEWITHAVERYVERYLONGNAME
<shortname@IFONLYTHISISALANDSATFILEWITHAVERYVERYLONGNAMEANDAGAINIFONLY...>    

# Long map and mapset name:
ifonlythisisalandsatfilewithaveryverylongnameandagainifonlythisisalandsatfilewithaveryverylongname@IFONLYTHISISALANDSATFILEWITHAVERYVERYLONGNAMEANDAGAINIFONLYTHISISALANDSATFILEWITHAVERYVERYLONGNAME
<ifonlythisisalandsatfilewithavery...@IFONLYTHISISALANDSATFILEWITHAVER...>    

# No need to truncate (no changed behaviour):
ifonlythisisalandsatfilewithavery@IFONLYTHISISALANDSATFILEWITHAVERY
<ifonlythisisalandsatfilewithavery@IFONLYTHISISALANDSATFILEWITHAVERY>    
-------------------------------------------------------------------------------| <- 80 char
```

vector/v.lidar.correction/main.c
metzm: In this case I would prefer to increase the size of database table names. These database table names often consist of the map name plus some more information and are hopefully not limited in size (to `GNAME_MAX`).

lib/vector/Vlib/map.c
metzm: In this case I prefer `snprintf` over `G_snprintf`. @neteler you being the author of `G_snprintf`, can you remember the reason for the existence of `G_snprintf`?

neteler: Honestly, I don't. But this is the original commit: https://github.com/OSGeo/grass-legacy/commit/e8361476c09903d60f29dbda8e1d7b0fd43d443f#diff-87363de8ff12b50f8adbfd1fb5baefb5ef308b1fc029a16be86c6e688dbf6beb

which contains a "use is discouraged" comment ...

nilason: Comparing the implementation of G_snprintf to glibc and Darwin/FreeBSD, there is one difference that stands out: a fix for Windows. This limitation/bug with Visual Studio has been fixed since its 2015 version at the latest (according to my investigations). I'm actually planning to propose a PR later on to update/drop G_snprintf.

nilason: From deep in the archives: https://lists.osgeo.org/pipermail/grass-dev/2002-April/005142.html

nilason:
Updated with forced push for cleaner commit history.
Changes: 
- added database table buffer size to `GNAME_MAX + 64`.
- using only `snprintf`, may use this as testing ground for eventually updating/removing `G_snprintf`
- closed conversation on using `GNAME_MAX + GNAME_MAX` for cases of concatenate `[dirname]/[dirname]` (please reopen if you disagree with my explanation)

nilason:
Any further thoughts on this PR? There are some (still open) discussions which I personally believe to be resolved.
Would be good to merge/address this for G8.

lib/imagery/list_gp.c
metzm: OK, I agree that such long names are probably rare, truncating in these cases should be harmless.

lib/imagery/list_subgp.c
metzm: OK

metzm (approved):

nilason:
@wenzeslaus It would be preferable to keep the commits of this PR separate, as I did with previous similar PRs (cherry pick to local master and push), as they are not really related (only grouped together to simplify review). Do you have any suggestion as to how to proceed?

lib/imagery/list_subgp.c
```diff
@@ -135,3 +137,38 @@ int I_list_subgroup_simple(const struct Ref *ref, FILE * fd)
...
+    int mapset_length = (int)strlen(mapset);
+
+    if (name_length + mapset_length + 3 < 75) {
+        frmt = "<%s@%s>";
+    }
+    else if (name_length > 35 && mapset_length > 35) {
+        frmt = "<%.33s...@%.32s...>";
+    }
+    else if (name_length > 35) {
+        sprintf(fr, "<\%%.%ds...@%%s>", 68 - mapset_length);
```
wenzeslaus, line 166: Why backslash here? GitHub syntax highlighter thinks it is somehow wrong and marks it red. No backslash should be needed here.

lib/imagery/list_subgp.c
```diff
@@ -135,3 +137,38 @@ int I_list_subgroup_simple(const struct Ref *ref, FILE * fd)
...
+        frmt = fr;
+    }
+    else {
+        sprintf(fr, "<\%%s@%%.%ds...>", 68 - name_length);
```
wenzeslaus, line 170: Same here. Also extra `\`. Or no?

lib/imagery/list_gp.c
```diff
@@ -75,10 +78,45 @@ int I_list_group_simple(const struct Ref *ref, FILE * fd)
...
+ *
+ * Results in a map name in the "<map@mapset>" form.
+ * If necessary truncates relevant part(s) and denotes
+ * with ellipsis, e.g. "<verylongmapname...@mapset>".
+ *
+ * \param[out] buf formatted map name
+ * \param name map name
+ * \param mapset mapset name
+ */
+static void name_fitted(char *buf, const char *name, const char *mapset)
```
wenzeslaus, line 100: Why not making the function shared? If you don't think this should be part of API, make it double underscore `I__...`. If it would be a one liner, but with the current code...

wenzeslaus (changes requested):

wenzeslaus:
> It would be preferable to keep the commits of this PR separate, as I did with previous similar PRs

The first and last commit are so small that they could be one commit if needed. The overflow part in the two middle ones seems related to each other to me, so putting these two together would be okay. However, the i.group name ellipses changes seem to require a separate PR for the i.group commit.

> (cherry pick to local master and push)

Direct pushes to the master branch are no longer possible without changing the settings. I have disabled those to avoid accidental pushes when people (with write access) confuse their remotes. (This is now set for both this and addons repo.)

> as they are not really related (only grouped together to simplify review).

Whatever were the reasons to start with one PR...

> Do you have any suggestion as to how to proceed?

...splitting this into multiple PRs seems to be the most straightforward way of going forward. No special merging procedures needed in that case.

Likely you know about it already, but if you are seeking a simple way for doing PRs, consider GitHub CLI (`gh pr create -f`).

lib/imagery/list_subgp.c
nilason: Honestly don't remember if it's mistake or deliberate. Will look into it.

lib/imagery/list_gp.c
nilason: The function is general enough for lib API. The question is whether it may be of use elsewhere. Anyone know?

lib/imagery/list_subgp.c
nilason: Looking closer at the code, it’s obvious this was deliberate. It creates a format string, which is used in line 173.

I guess GitHubs syntax highlighting means: Are you sure this is what you wanted?

wenzeslaus: I still think the backslash is not needed. `\` is escape of a special character in C string literal, not an escape of _sprintf_ syntax. The right "escape" is doubling the `%` sign which is what you are doing. Try the following code or [run it online](https://onlinegdb.com/180q-2sjh):

```c
#include <stdio.h>
#include <string.h>

int main()
{
    printf("<%%s@%%.%ds...>\n", 100);
    printf("<\%%s@%%.%ds...>\n", 100);
    printf("%d\n", strcmp("%%s", "\%%s"));
    return 0;
}
```

```
<%s@%.100s...>
<%s@%.100s...>
0
```

nilason: You're absolutely right, the backslash shouldn't be there. Now back at my machine, I also did some testing. Funny thing is, it works either way. Removed them all with latest push.

lib/imagery/list_gp.c
nilason: Just updated with the function now as part of lib/imagery API.

nilason:
@wenzeslaus If you are good with latest changes I will commit them through separate PRs.

wenzeslaus:
> If you are good with latest changes I will commit them through separate PRs.

Yes, that sounds great!

lib/imagery/list_gp.c
nilason: Moved this part of the PR to #1818.

PR #1409, opened 2021-02-27
v.db.join: Use case insensitive comparison for columns

wenzeslaus:
Since SQL is case insensitive, adding a new column to a table needs to use case insensitive
comparison with the existing columns.

Additionally, use fully qualified map name to get the 'found in more than one mapset' message
only once.

### DIFF
--- a/scripts/v.db.join/v.db.join.py
+++ b/scripts/v.db.join/v.db.join.py
@@ -6,7 +6,7 @@
 # AUTHOR(S):    Markus Neteler
 #               Converted to Python by Glynn Clements
 # PURPOSE:      Join a table to a map table
-# COPYRIGHT:    (C) 2007-2009 by Markus Neteler and the GRASS Development Team
+# COPYRIGHT:    (C) 2007-2021 by Markus Neteler and the GRASS Development Team
 #
 #               This program is free software under the GNU General Public
 #               License (>=v2). Read the file COPYING that comes with GRASS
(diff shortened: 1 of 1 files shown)


ninsbl (approved):
This is a valid workaround. I still think a more thorough fix would be to use quoted idetifiers (see [Trac 3071](https://trac.osgeo.org/grass/ticket/3071)). Instead of using backend dependent quoting, standard SQL compliant double quotes could be used. That would laso solve quite common issues with keywords, f.ex.. But that is a different issue and would have to be applied all over the place. One may start with modules like this however...

wenzeslaus:
> This is a valid workaround.

I think it is _the solution_ within the current logic, or am I missing something?

> I still think a more thorough fix would be to use quoted idetifiers (see [Trac 3071](https://trac.osgeo.org/grass/ticket/3071)).

Some users might be surprised by the need to exactly type the case of column name in some contexts (parameter is column name) while in some others you would still be free to do whatever you want (parameters like where). However, I think I would be for this change anyway.

wenzeslaus:
> This is a valid workaround. I still think a more thorough fix would be to use quoted idetifiers (see [Trac 3071](https://trac.osgeo.org/grass/ticket/3071)). Instead of using backend dependent quoting, standard SQL compliant double quotes could be used. That would laso solve quite common issues with keywords, f.ex.. But that is a different issue and would have to be applied all over the place. One may start with modules like this however...

Perhaps a CS/CE oriented GSoC topic? Do you want to add the idea to the list?

nilason:

landam:
> Yes, done so

Thanks! :-)

PR #1413, opened 2021-02-28
v.db.update: Use case insensitive comparison for columns

wenzeslaus:
Returned column names are with case as present in the table, but since unqoted column names in SQL are case insensitive,
the constructed UPDATE statement works in any case.

Without this PR, the following fails for column called ABC:
...column=abc where='abc IS NULL'...
although the where part as well as rest of the UPDATE statement would work.


ninsbl (approved):

nilason:

landam:
Not easy to backport (conflicts). @ninsbl What about removing `backported_needed` label?

PR #1421, opened 2021-03-03
r.geomorphon: fix landform category names

infrastation:
This change comes in two commits to make it easier to review. The substantial part is in the first commit, it changes the landform category names. In other words, for `r.geomorphon forms=xxxx` the raster outputs produced before will continue to have the old (inconsistent) category names, and the new raster outputs will start to have the new (consistent) category names. The documentation page change shows the difference. It is difficult for me to tell whether the change is going to cause any backward compatibility issues or not, but in any case it would be best to make it in a major release (i.e. GRASS 8). Please review.

The second commit is a straightforward refactoring.


marisn (approved):

raster/r.geomorphon/r.geomorphon.html
```diff
@@ -149,15 +149,15 @@ <h3>Geomorphon calculation: extraction of terrestrial landforms</h3>
 # verify terrestrial landforms found in DEM
 r.category eu_dem_25m_geomorph
  1  flat
- 2  summit
+ 2  peak
```
neteler, line 152: To avoid confusion, should the second manual page example also be updated? Or defined, that "summit = peak" to keep it understandable?

infrastation:
Thank you for the review, I have updated the second example and the test (which does not pass because of an unrelated issue discovered a while ago, but anyway). Let me know in case anything else catches your eye.

neteler (approved):
Thanks for your efforts!

neteler:
May this PR be merged?

infrastation:
Yes, as far as I can see. It implements the change as described.

infrastation:
Thank you.

PR #1425, opened 2021-03-06
db.in.ogr: Add gdal_config= and gdal_doo=

HuidaeCho:
Useful for AUTODETECT_TYPE=YES for CSV files


neteler (approved):
Didn't test but looks good to me.

PR #1428, opened 2021-03-07
wxGUI: Exit after fatal lock errors

HuidaeCho:


wenzeslaus:
This is just changing/refactoring the code, but not changing behavior, right?

HuidaeCho:
> This is just changing/refactoring the code, but not changing behavior, right?

Refactoring + `sys.exit()` after these two fatal errors from GUI:
1. Concurrent use not allowed. You can force launching GRASS using -f flag.
2. Unable to properly access . Please notify system personnel.

On the same errors, the text mode calls `fatal()` and exits, but not the GUI.

Try this:
1. Open a mapset with `--gui`
2. Open the same mapset again with `--gui`
3. "Concurrent use not allowed" error message
4. Click OK
5. You're in the second session instead of being kicked out, as if you used `-f` flag.

wenzeslaus (approved):
+1 with two notes:

* I assume you fix the formatting issue in the file. Just run Black.
* Merging this will create a merge conflict for #1400.

wenzeslaus:
> > This is just changing/refactoring the code, but not changing behavior, right?
> 
> Refactoring + `sys.exit()` after these two fatal errors from GUI:

Thanks. Now I see it - including the TODO comment written by myself. :-)

HuidaeCho:
> > > This is just changing/refactoring the code, but not changing behavior, right?
> > 
> > 
> > Refactoring + `sys.exit()` after these two fatal errors from GUI:
> 
> Thanks. Now I see it - including the TODO comment written by myself. :-)

Yup!
```bash
$ black grass.py 
reformatted grass.py
All done! ✨ 🍰 ✨
1 file reformatted.
$ git diff
diff --git a/lib/init/grass.py b/lib/init/grass.py
index 80d4d23d9..e50e7cb0c 100755
--- a/lib/init/grass.py
+++ b/lib/init/grass.py
@@ -2624,7 +2624,7 @@ def main():
         lock_mapset(
             mapset_settings.full_mapset,
             user=user,
-            force_gislock_removal=params.force_gislock_removal
+            force_gislock_removal=params.force_gislock_removal,
         )
     except Exception as e:
         msg = e.args[0]
```
That's interesting. A comma after the last argument.

wenzeslaus:
Black enforces trailing comma because that the only way how to make it consistent and it minimizes diffs for subsequent changes when adding a new parameter since only the last addition will be present. However, more importantly, just forget all this an offload it to Black. Just write the code without it and let Black decide what to do. (That applies to more or less everything except for splitting strings.)

PR #1429, opened 2021-03-07
v.colors: Backport #1393 and #1389 (Resolve #1297)

### LINKED ISSUE #1297: [Bug] v.colors cannot properly copy color table

HuidaeCho:
Backport #1393 and #1389 (Resolve #1297)

### DIFF
--- a/vector/v.colors/local_proto.h
+++ b/vector/v.colors/local_proto.h
@@ -9,8 +9,10 @@ void load_colors(struct Colors *, const char *, DCELL, DCELL, int);
 
 /* scan_attr.c */
 int scan_attr(const struct Map_info *, int, const char *, const char *,
-	      const char *, const struct FPRange *, struct Colors *);
-void color_rules_to_cats(dbCatValArray *, int, struct Colors *, struct Colors *);  
+	      const char *, const struct FPRange *, struct Colors *,
+	      struct Colors *);
+void color_rules_to_cats(dbCatValArray *, int, struct Colors *,
+			 struct Colors *);
 
 /* scan_cats.c */
 void scan_cats(const struct Map_info *, int, const char *, const char *,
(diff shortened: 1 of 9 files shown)


cmbarton (approved):
Works great. Thanks again.

Issue #1360, opened 2021-02-18
[Bug] General fixes for three Perl scripts are needed


wenzeslaus:
**Describe the bug**

Perl::Critic (perlcritic) included in Mega-Linter identified the following errors in `display/d.text/test.pl` and `utils/copywrite.pl`.

```
Results of perlcritic linter (version 1.138)
See documentation on https://nvuillam.github.io/mega-linter/descriptors/perl_perlcritic/
-----------------------------------------------

❌ [ERROR] display/d.text/test.pl
    Code before strictures are enabled at line 2, column 1.  See page 429 of PBP.  (Severity: 5)
    Bareword file handle opened at line 52, column 1.  See pages 202,204 of PBP.  (Severity: 5)
    Two-argument "open" used at line 52, column 1.  See page 207 of PBP.  (Severity: 5)

✅ [SUCCESS] general/g.parser/test.pl
❌ [ERROR] utils/copywrite.pl
    Code before strictures are enabled at line 28, column 1.  See page 429 of PBP.  (Severity: 5)
    Bareword file handle opened at line 76, column 32.  See pages 202,204 of PBP.  (Severity: 5)
    Two-argument "open" used at line 76, column 32.  See page 207 of PBP.  (Severity: 5)
    Bareword file handle opened at line 98, column 2.  See pages 202,204 of PBP.  (Severity: 5)
    Two-argument "open" used at line 98, column 2.  See page 207 of PBP.  (Severity: 5)
    Expression form of "eval" at line 146, column 12.  See page 161 of PBP.  (Severity: 5)
    Bareword file handle opened at line 152, column 5.  See pages 202,204 of PBP.  (Severity: 5)
    Two-argument "open" used at line 152, column 5.  See page 207 of PBP.  (Severity: 5)
```

Running perlcritic manually reveals additional issues with file `vector/v.clean/test/v.rand.poly`.

```
vector/v.clean/test/v.rand.poly: Code before strictures are enabled at line 5, column 1.  See page 429 of PBP.  (Severity: 5)
vector/v.clean/test/v.rand.poly: Bareword file handle opened at line 47, column 1.  See pages 202,204 of PBP.  (Severity: 5)
vector/v.clean/test/v.rand.poly: Two-argument "open" used at line 47, column 1.  See page 207 of PBP.  (Severity: 5)
```

**To Reproduce**

```
perlcritic .
```

**Expected behavior**
No errors reported.

[...]

PR #1431, opened 2021-03-08
perlcritic: fix perlcritic errors

ocsmit:
Converts one perl script to Python and fixes errors identified by perlcritic in two other scripts.

Addresses  #1360


display/d.text/test.py
```diff
@@ -0,0 +1,69 @@
...
+          "purple")
+# fmt: on
+
+
+def rc(*args):
+    print(f".X {args[0]}\n.Y {args[1]}")
+
+
+def xy(*args):
+    print(f".X {args[0]}\n.Y {args[1]}")
```
wenzeslaus, line 21: I had to add `{args[0] / 100}` for something to show up. The original code has `%`. I'm not sure how it should work. It does not work for me. I did this to see it:

```
# in a GRASS session
d.mon wx0
python display/d.text/test.py > test.txt
d.erase
d.text input=test.txt
```

Whether it works for you or not, I think the ` / 100` is needed here.

wenzeslaus (changes requested):
Thank you. This looks good. I have couple of suggestions, but it depends how far you want to go with this. Let me know and we can merge this soon (even as is if needed).

display/d.text/test.py
```diff
@@ -0,0 +1,69 @@
...
+
+size(2)
+rotate(0)
+font("romans")
+color("gray")
+rc(1, 1)
+
+with open(__file__) as f:
+    src = "\n.L 1\n.L 0\n".join(f.read().split("\n")[1:])
+    f.close()
```
HuidaeCho, line 67: No need to close file pointers from inside `with` (https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files).

display/d.text/test.py
```diff
@@ -0,0 +1,69 @@
...
+
+print(re.sub('(".*?")', "\n.C red\n\\g<0>\n.C gray\n", src))
```
HuidaeCho, line 69: This would be a more direct translation. Note the missing `".L 0\n" +` and use of double quotes for consistency.
```python
with open(__file__) as f:
    src = f.read()

print(".L 0\n" +
      re.sub("(\".*?\")", "\n.C red\n\\g<0>\n.C gray\n",
             re.sub("\n", "\n.L 1\n.L 0\n", src)))
```

HuidaeCho (changes requested):

ocsmit:
> 
> 
> Thank you. This looks good. I have couple of suggestions, but it depends how far you want to go with this. Let me know and we can merge this soon (even as is if needed).

I certainly don't mind, and would even prefer to take this further and make it as good as it can be. I will address yours and @HuidaeCho's changes/comments and implement them!

display/d.text/test.py
ocsmit: This one is interesting. It shows up for me the same without either dividing by 100 or without %, which is why I had missed the % being left out.

I was testing with the following, not that it should matter too much.
```
d.text >  d.mon start=wx0
./test.py | d.text
```

display/d.text/test.py
```diff
@@ -63,7 +67,14 @@ def text(*args):
...
-print(re.sub('(".*?")', "\n.C red\n\\g<0>\n.C gray\n", src))
+    src = f.read()
+    "\n.L 1\n.L 0\n".join(f.read().split("\n")[1:])
+
+print(
+    ".L 0\n"
+    + re.sub(
+        '(".*?")',
+        "\n.C red\n,\\g<0>\n.C gray\n",
+        re.sub("\n", "\n.L 1\n.L 0\n", re.sub("(?m)^#.*\n?", "", src)),
```
ocsmit, line 78: Added regular expression to remove commented lines in rendered output.

display/d.text/test.py
wenzeslaus: I see the rotated colorful font names only when I do `/ 100` not when I use `%` in the code. The result I see is not supper nice either way (see below), but the current code does not give me the colorful font names. Are you getting nicer outputs @HuidaeCho @ocsmit?

Screenshot from 2021-03-11 22-53-05

display/d.text/test.py
```diff
@@ -0,0 +1,69 @@
...
+
+# fmt: off
+fonts = ("cyrilc", "gothgbt", "gothgrt", "gothitt", "greekc", "greekcs",
+         "greekp", "greeks", "italicc", "italiccs", "italict", "romanc",
+         "romancs", "romand", "romans", "romant", "scriptc", "scripts",
+         "cyrilc", "gothgbtlu")
+colors = ("red", "orange", "yellow", "green", "blue", "indigo", "violet",
+          "black", "gray", "brown", "magenta", "aqua", "grey", "cyan",
+          "purple")
+# fmt: on
```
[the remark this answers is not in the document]
wenzeslaus: So if I understand #1441 correctly, within this PR there is no way to start the text in the upper left corner, right?

display/d.text/test.py
HuidaeCho: It's supposed to look like this:
image

```bash
d.mon start=png
./test.pl | d.text
d.mon stop=png
```

The reason why you cannot see the rotating font names with `%` is the wx0 monitor and GUI start `at=50,50` (#1441). In fact, the behavior of wx0 and GUI is inconsistent with the other monitors. Starting from the top-left corner used to be and should be the default.

display/d.text/test.py
```diff
@@ -0,0 +1,80 @@
...
+rc(1, 1)
+
+with open(__file__) as f:
+    src = f.read()
+    "\n.L 1\n.L 0\n".join(f.read().split("\n")[1:])
+
+print(
+    ".L 0\n"
+    + re.sub(
+        '(".*?")',
```
HuidaeCho, line 76: This is actually interesting. Black complained about single-quote indexing (#1425), but it leaves regular single-quote strings. It even reformats escaped double quotes inside double quotes. Another confusing lesson from Black.

ocsmit: Yes, I was going to leave a comment pointing this out  since you addressed the want for double quotations. Glad you noticed it.

display/d.text/test.py
ocsmit: I managed to get something closer to the original perl script while compensating for the newer start point with 
```python
def xy(x, y):
    print(f".X {x - 50}%\n.Y {y + 25}%")
```
image

Obviously it is still not the same as the original but its closer to the intended output while appearing much more condensed.

ocsmit: Using the following
```bash 
d.mon start=wx3                    
./test.py | d.text at=1,100
```
I get 
image
without changing any of the code.

display/d.text/test.py
wenzeslaus: Ad quoting, Black abandons(x) the practice of single quotes for indexes/values and double quotes for user strings/output. Instead, it uses the classic/original reasoning for single and double quotes - to limit the need for escaping when quotes are in the strings when, e.g., mixing languages. This has more clear rules compared to the other practice. The point with using Black is: you just do something reasonable and let Black do its thing automatically, then you check the checks in a PR that you didn't forget to run Black. No figuring out what is the proper/project/file/Black style. When everything works as it should (in terms of Black versions, for example), it is not you trying to get the Black formatting right, it is Black taking care of the formatting for you.

x Actually, it is the default settings we decided to use. It has legacy option for project which don't want to switch.

display/d.text/test.py
HuidaeCho: > without changing any of the code.

The code was meant to test `%`, so it should include `%`.

display/d.text/test.py
```diff
@@ -0,0 +1,80 @@
...
+
+size(2)
+rotate(0)
+font("romans")
+color("gray")
+rc(1, 1)
+
+with open(__file__) as f:
+    src = f.read()
+    "\n.L 1\n.L 0\n".join(f.read().split("\n")[1:])
```
HuidaeCho, line 71: Please remove this line. It is done in line 78 again.

display/d.text/test.py
ocsmit: >without changing any of the code.

This was in reference to c714586c8dc596826e9bf8783d697a8345f406bf, which has percents added. I should have been more specific or created a different comment.

display/d.text/test.py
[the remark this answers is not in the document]
HuidaeCho: If you set `at=0,100` (overriding wx's default `at=50,50`), you can start from the upper-left corner. But other monitors always start from there by default. Using `line=` without `at=` should also start from the upper-left corner by default. Unfortunately, at this point, `line=` and `at=` are mutually exclusive and the wx monitor and GUI (default `at=50,50`) behave different from the other monitors (no implicit at=).

HuidaeCho (approved):

display/d.text/test.py
wenzeslaus: Beautiful! Well, not really, but it does what it is supposed to do :-)

wenzeslaus (approved):

wenzeslaus:
Merged, thank you @ocsmit !

ocsmit:
> Merged, thank you @ocsmit !

Absolutely, my pleasure!

wenzeslaus:
Thanks to this, I now enabled Perl check with Super-Linter in #1493 and the possible future addition of Mega-Linter will be smoother.

PR #1433, opened 2021-03-09
parser: Show error for option file

wenzeslaus:
Show the actual error using errno in G_open_option_file function.
Influences g.list, v.external.out, v.in.ascii.

Users won't have to wonder why the file cannot be created or opened.

strerror(errno) is used in lib/gis. Only POSIX fopen requires errno to be set,
but even on Windows errno should be used to obtain specific error information
on error when using fopen according to docs.

### DIFF
--- a/lib/gis/parser.c
+++ b/lib/gis/parser.c
@@ -75,6 +75,7 @@
  * \author Soeren Gebbert added Dec. 2009 WPS process_description document
  */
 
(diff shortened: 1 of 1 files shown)


metzm:
More informative error messages are always an improvement. A candidate for backport?

nilason:
This is tagged for backport to 7.8.6.

PR #1446, opened 2021-03-11
parser: Accept #-space-% to allow writing PEP8-compliant Python code

wenzeslaus:
Accept '# %' in parser script header definition because '#%' is against PEP8
(each line of a block comment starts with a # and a single space).
Only minimal changes to achive this support are included.

'#%' is still supported fully supported and remains in documentation.
The full switch to '# %' is left for v8.

Partial backport of 635224115c5ee9c29a9356210c3b03e5511ef6dc (#1287).


neteler (approved):
Tested with some G79 scripts, seems to work fine!

PR #1447, opened 2021-03-11
Number of bands relevant only to strds

landam:
This PR fixes issue reported in #1342 

Steps to reproduce the bug:

```
t.create A01 type=str3ds title=x d=y
```

fails with

```
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/scripts/t.create", line 94, in <module>
    main()
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/scripts/t.create", line 87, in main
ERROR: Unable to execute transaction:
...
sqlite3.OperationalError: table str3ds_metadata has no column named number_of_bands
```

By applying this PR its possible to create all stds:

```
t.create A02 type=str3ds title=x d=y
t.create A02 type=strds title=x d=y
t.create A02 type=stvds title=x d=y
```

Property `number of bands` is relevant only to strds:

```
t.info A02
...
 | Number of registered bands:. None
...
```

### DIFF
--- a/python/grass/temporal/metadata.py
+++ b/python/grass/temporal/metadata.py
@@ -242,28 +242,29 @@ class RasterMetadataBase(SQLDatabaseInterface):
     min = property(fget=get_min, fset=set_min)
     max = property(fget=get_max, fset=set_max)
 
-    def print_info(self):
-        """Print information about this class in human readable style"""
-        #      0123456789012345678901234567890
-        print(" | Datatype:................... " + str(self.get_datatype()))
-        print(" | Number of columns:.......... " + str(self.get_cols()))
-        print(" | Number of rows:............. " + str(self.get_rows()))
-        print(" | Number of cells:............ " + str(self.get_number_of_cells()))
-        print(" | North-South resolution:..... " + str(self.get_nsres()))
-        print(" | East-west resolution:....... " + str(self.get_ewres()))
-        print(" | Minimum value:.............. " + str(self.get_min()))
-        print(" | Maximum value:.............. " + str(self.get_max()))
+    def _print_info_body(self, shell=False):
+        """Print information about this class (body part).
 
-    def print_shell_info(self):
-        """Print information about this class in shell style"""
-        print("datatype=" + str(self.get_datatype()))
-        print("cols=" + str(self.get_cols()))
-        print("rows=" + str(self.get_rows()))
-        print("number_of_cells=" + str(self.get_number_of_cells()))
-        print("nsres=" + str(self.get_nsres()))
-        print("ewres=" + str(self.get_ewres()))
-        print("min=" + str(self.get_min()))
-        print("max=" + str(self.get_max()))
+        :param bool shell: True for human readable style otherwise shell style
+        """
+        if shell:
+            print("datatype=" + str(self.get_datatype()))
+            print("cols=" + str(self.get_cols()))
+            print("rows=" + str(self.get_rows()))
+            print("number_of_cells=" + str(self.get_number_of_cells()))
+            print("nsres=" + str(self.get_nsres()))
+            print("ewres=" + str(self.get_ewres()))
+            print("min=" + str(self.get_min()))
+            print("max=" + str(self.get_max()))
+        else:
+            print(" | Datatype:................... " + str(self.get_datatype()))
+            print(" | Number of columns:.......... " + str(self.get_cols()))
+            print(" | Number of rows:............. " + str(self.get_rows()))
+            print(" | Number of cells:............ " + str(self.get_number_of_cells()))
+            print(" | North-South resolution:..... " + str(self.get_nsres()))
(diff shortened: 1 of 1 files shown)


veroandreo (approved):

python/grass/temporal/metadata.py
```diff
@@ -1261,15 +1249,9 @@ def print_info(self):
...
         print(" | Aggregation type:........... " + str(self.get_aggregation_type()))
-        print(" | Number of registered bands:. " + str(self.get_number_of_bands()))
-
-        STDSMetadataBase.print_info(self)
 
     def print_shell_info(self):
         """Print information about this class in shell style"""
-        print("aggregation_type=" + str(self.get_aggregation_type()))
-        print("number_of_bands=" + str(self.get_number_of_bands()))
-        STDSMetadataBase.print_shell_info(self)
```
wenzeslaus, line 1272: The way `STDSMetadataBase.print_shell_info(self)` is now called is highly unusual. Please document the reasoning in the code.

wenzeslaus (changes requested):
This looks good and it brings the tests above 90%. Perhaps a test for for these combinations is missing in the first place.

Do you want to go ahead and change `--min-success 80` to `90`? (The two Ubuntu tests are at 91% and 92% in this PR.)

python/grass/temporal/metadata.py
landam: The way how `STDSMetadataBase.print_shell_info(self)` is called haven't changed. I just moved call from `RasterMetadataBase` to `RasterMetadata` and `Raster3DMetadata` in order to keep printing variables in consistent order.

@wenzeslaus Please explain what you mean by "unusual".

wenzeslaus: As far as I can tell, before, a function from a parent class was called (by a direct child). Now, classes call functions of the their grandparents. I think that's unusual (and [possibly also considered a bad practice](https://stackoverflow.com/a/43943861/574907)).

wenzeslaus:
Just to be clear about the impact: This fixes 11 tests according to a rough comparison between current master branch and this PR.

```diff
  Numbers for Ubuntu 18.04
- From them 213 files (87%) were successful and 33 files (13%) failed.
+ From them 224 files (91%) were successful and 22 files (9%) failed.
```

python/grass/temporal/metadata.py
landam: Well, not everything what is mentioned in stackoverflow I would write in the stone. @wenzeslaus Anyway, I took your valid point into consideration.

* `STDSMetadataBase.print_info()` renamed to `print_info_tail()`
* the code changed in order to use `super()` syntax

See https://github.com/OSGeo/grass/pull/1447/commits/57fe78fb867abf865fba2d31b8d029a720788f08

wenzeslaus (approved):

python/grass/temporal/metadata.py
wenzeslaus: > not everything what is mentioned in stackoverflow I would write in the stone

So I though about it again and you can show it just on the standard inheritance example like Animal-Mammal-Dog or Fruit-Apple-Gala. Instances of Dog are Mammals and because Mammals are Animals dogs are Animals (but Dogs are not Animals apart from being Mammals - it is not multiple inheritance). What is relevant here is that, Mammals must be fully Animals, so if there is a `print_attributes()` function in Animal for printing attributes, `print_attributes()` of Mammal is expected to give you attributes from Animal (at least) and (possibly) Mammal (if there are any extra). It is expected because Mammal is an Animal. In this context, Dog should not combine attributes from Animal and attributes from Mammal in order to get some "full Mammal list", the Mammal's attributes contain the Animal attributes because Mammal is an Animal.

> STDSMetadataBase.print_info() renamed to print_info_tail()

I'm afraid that the Animal-Mammal-Dog inheritance example reveals that although this avoids the somewhat evident issue of calling parent and grandparent, it still retains the less apparent issue of not keeping inheritance semantics of children classes being also valid parent classes. Namely, both `STDSRasterMetadataBase` (Mammal) and `STRDSMetadata` (Dog) have `print_info()`, but `STDSRasterMetadataBase.print_info()` does not print the "tail" part while the `STRDSMetadata.print_info()` does. That's like the Mammal not being an Animal and only when we get to Dog, we can see the Animal (`STDSMetadataBase`) there. 

In this class hierarchy, starting with what you did so far, I think the right solution would be having `print_tail` and also `print_body` (and `print_head` or whatever is needed). Each class would define those and there would be default implementation for `print_info` which would be defined as printing all and overwritten when needed.

Perhaps a simpler way to avoid this is to say there is difference between the `*Base` ("abstract") classes and the "final"/"concrete" classes. The base classes define only the building block functions and it is responsibility of the concrete classes to define `print_info`. (In C++, `print_info` would be pure virtual.) So, maybe you could just rename `print_info` in `STDSRasterMetadataBase`.

> the code changed in order to use super() syntax

That's super!

PR #1454, opened 2021-03-12
TGIS DB v3 backward compatible with v2

landam:
Currently TGIS DB created in GRASS <7.9 (version 2) is not supported by GRASS 7.9 (version 3). 

```
> t.info v2
ERROR: Unsupported temporal database: version mismatch.
The format of your actual temporal database is not supported any more.
Please create a backup of your temporal database to avoid lossing data.
SOLUTION: Run t.upgrade command installed from GRASS Addons in order to
upgrade your temporal database.
Supported temporal database version is: 2
Current temporal database info:
DBMI interface:..... sqlite3
Temporal database:.. /home/martin/grassdata/tgis/PERMANENT/tgis/sqlite.db
```

This PR introduces first step for backward compatibility:

```
t.info v2
WARNING: Temporal database version mismatch detected.
         Please create a backup of your temporal database to avoid lossing
         data.
         SOLUTION: Run t.upgrade command installed from GRASS Addons in
         order to upgrade your temporal database.
         Supported temporal database version is: 3
         Current temporal database info:
         DBMI interface:..... sqlite3
         Temporal database:..
         /home/martin/grassdata/tgis/PERMANENT/tgis/sqlite.db
 +-------------------- Space Time Raster Dataset -----------------------------+
 |                                                                            |
...
 | Number of registered maps:.. 1
 |
 | Title:
 | x
 | Description:
 | y
 | Command history:
 | # 2021-03-12 22:19:52 
 | t.create v2 tit="x" d="y"
 | # 2021-03-12 22:23:53 
 | t.register input="v2" maps="rast" start="2020-01-01"
 |     --o
 | 
 +----------------------------------------------------------------------------+
``` 

Note that it's a draft which need to be tested.


python/grass/temporal/core.py
```diff
@@ -374,13 +374,29 @@ def get_tgis_version():
 
 
 def get_tgis_db_version():
-    """Get the version number of the temporal framework
-    :returns: The version number of the temporal framework as string
+    """Get the supported version number of the temporal database
```
wenzeslaus, line 377: See the other "supported" comment, e.g., grass/temporal/metadata.py seems to support 2 and 3. Also, if you want to add an adjective, you should probably drop the word number.

python/grass/temporal/metadata.py
```diff
@@ -1134,7 +1135,8 @@ def __init__(
         self.D["ewres_min"] = None
         self.D["ewres_max"] = None
         self.D["aggregation_type"] = aggregation_type
-        self.D["number_of_bands"] = None
+        if get_tgis_db_version_from_metadata() > 2:
+            self.D["number_of_bands"] = None
```
wenzeslaus, line 1139: Awesome! I miss the context to understand how much of these you plan/need to add. In this particular case, it seems that it would not be a big deal to simply report number of bands as None (or zero or whatever this should translates to). It might be matter of taste and the impact on the user is questionable, but in my view, the library does not have to behave as the older version. In my view, number of bands is zero in 8.0.0 with data from 7.8. Where it should fail in this context is attempt to register bands and maybe in some cases also when explicitly accessing band information.

python/grass/temporal/core.py
```diff
@@ -363,8 +363,8 @@ def get_raise_on_error():
 
 
 def get_tgis_version():
-    """Get the version number of the temporal framework
-    :returns: The version number of the temporal framework as string
+    """Get the supported version number of the temporal framework
+    :returns: The version number of the temporal framework as integer
```
wenzeslaus, line 367: I must say I don't understand the role of tgis_version and how it relates to tgis_db_version, specifically why it is stored in the database, so I can't comment much on this part.

wenzeslaus (changes requested):

python/grass/temporal/core.py
landam: https://github.com/OSGeo/grass/blob/master/python/grass/temporal/core.py#L116
https://github.com/OSGeo/grass/blob/master/python/grass/temporal/core.py#L365

python/grass/temporal/metadata.py
veroandreo: > Awesome! I miss the context to understand how much of these you plan/need to add. In this particular case, it seems that it would not be a big deal to simply report number of bands as None (or zero or whatever this should translates to). It might be matter of taste and the impact on the user is questionable, but in my view, the library does not have to behave as the older version. In my view, number of bands is zero in 8.0.0 with data from 7.8. 

I think it can also be 0 with data from 8.0.0 if we do not register bands... 

> Where it should fail in this context is attempt to register bands and maybe in some cases also when explicitly accessing band information.

So, a warning would be needed then only if a user tries to register or read bands in an old STRDS (i.e. version 2) while in GRASS 8.0. But s/he could still read info from the old STRDS and do operations with it? 

Honestly, I think it's really easy and safe to do a `t.upgrade` in all mapsets with STDS if using grass-dev or once 8.0.0 is released... So, as long as #1435 removes the errors when registering raster maps without band references, I do not have a strong opinion about backward compatibility.

python/grass/temporal/core.py
```diff
@@ -697,8 +713,6 @@ def init(raise_fatal_error=False, skip_db_version_check=False):
             db_exists = True
 
     backup_howto = _(
-        "The format of your actual temporal database is not "
-        "supported any more.\n"
         "Please create a backup of your temporal database "
         "to avoid lossing data.\nSOLUTION: "
```
HuidaeCho, line 717: `losing data`

neteler, line 717:
```suggestion
        "to avoid losing data.\nSOLUTION: "
```

python/grass/temporal/core.py
```diff
@@ -759,22 +775,32 @@ def init(raise_fatal_error=False, skip_db_version_check=False):
...
                 )
-            if "tgis_db_version" in entry and entry[1] != str(get_tgis_db_version()):
+
+        # temporal database version check
+        tgis_db_version_meta = get_tgis_db_version_from_metadata(metadata)
+        if tgis_db_version_meta != tgis_db_version:
+            message = _(
+                "Temporal database version mismatch detected."
+                "\n {backup}Supported temporal database version"
+                " is: {tdb}\nCurrent temporal database info:"
```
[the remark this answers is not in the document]
landam: Done in 0838a7fdc. 

```
WARNING: Temporal database version mismatch detected.
         Please create a backup of your temporal database to avoid losing
         data.
         SOLUTION: Run t.upgrade command in order to upgrade your temporal
         database.
         Supported temporal database version is: 3
         Your existing temporal database version: 2
         Current temporal database info:
         DBMI interface:..... sqlite3
         Temporal database:..
         /home/martin/grassdata/tgis/PERMANENT/tgis/sqlite.db
```

python/grass/temporal/core.py
```diff
@@ -759,22 +775,32 @@ def init(raise_fatal_error=False, skip_db_version_check=False):
...
+                "{info}".format(
```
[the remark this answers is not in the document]
landam: Hopefully solved by 0838a7fdc

python/grass/temporal/core.py
landam: Done by 4d0294a10

python/grass/temporal/metadata.py
landam: I would keep the `None` value, it's default for other values, see https://github.com/landam/grass/blob/tgisdb3_backward_compatibility/python/grass/temporal/metadata.py#L1129 and https://github.com/landam/grass/blob/tgisdb3_backward_compatibility/python/grass/temporal/metadata.py#L1145. Actually the `None` value is retured by `get_number_of_bands` (https://github.com/landam/grass/blob/tgisdb3_backward_compatibility/python/grass/temporal/metadata.py#L1234).

landam:
Current message:

```
WARNING: Temporal database version mismatch detected.
         Please create a backup of your temporal database to avoid losing
         data.
         Run t.upgrade command to upgrade your temporal database.
         Supported temporal database version is: 3
         Your existing temporal database version: 2
         Current temporal database info:
         DBMI interface:..... sqlite3
         Temporal database:..
         /home/martin/grassdata/tgis/PERMANENT/tgis/sqlite.db
```

landam:
Attempt to modify strds still fails

```
t.register input=test maps=x1 start=2020-01-01
...
(SELECT id FROM raster_map_register_147b8e68ee454031a9d993dc5cf31194)
) WHERE id = 'test@PERMANENT';
UPDATE st
ERROR: no such column: semantic_label
````

due to https://github.com/landam/grass/blob/tgisdb3_backward_compatibility/lib/temporal/SQL/update_strds_metadata_template.sql#L45

@marisn Do we want to support modification of strds created by GRASS v7 in v8?

**Comment by landam** [comment]
@marisn We could either split files:

* `update_strds_metadata_template.sql` (common)
* `update_strds_metadata_template_v3.sql` (version 3+)

or to use inline modification as applied in 481f681 (but more robust). What do you think?

File: python/grass/temporal/abstract_space_time_dataset.py [hunk 7]
```diff
@@ -2579,6 +2579,14 @@ def update_from_registered_maps(self, dbif=None):
         sql = sql.replace("SPACETIME_REGISTER_TABLE", stds_register_table)
         sql = sql.replace("SPACETIME_ID", self.base.get_id())
 
+        # check version
+        idx = sql.find("-- VERSION")
+        if idx > -1:
+            from .core import get_tgis_db_version_from_metadata
+            version = int(sql[idx+len("-- VERSION ")])
+            if get_tgis_db_version_from_metadata() < version:
+                sql = sql[:idx] # remove version-relevant code
```
wenzeslaus, line 2588: This looks good. But can you please document somewhere (here or in the SQL file) what is the intended behavior and/or how to use it? ...e.g., in case we need to add one more thing like this.

wenzeslaus (approved):
Thanks for addressing what I raised earlier, esp. the important message-related things. I like how it looks like now, but I can review again if needed.

python/grass/temporal/abstract_space_time_dataset.py
landam: Please consider these changes as a draft. We have to decide whether:

1) keep all SQL statements in one file and introduce VERSION specific code by comments (as used in the draft - but by robust and well-documented code) or
2) we split version specific SQL statements into separated files, eg.

* `update_strds_metadata_template.sql` (all versions)
* `update_strds_metadata_template_v3.sql` (version 3+)

The second options is probably more clear and robust. @wenzeslaus What do you think?

wenzeslaus: +1 for the filenames. Having a convention for filenames is also more expected behavior/code than special comments in this case.

veroandreo (commented):
I like the wording in the new message. 

I'm still somehow confused though. AFAIU, if there's a version mismatch (supported=3, current=2) the user needs to run `t.upgrade`, is that what we call backward compatibility? I'd have assumed compatibility meant some background update of versions transparent to the user.

Also, instead of `number_of_bands` shouldn't we use something with `semantic_label` for consistency?

landam:
> I'm still somehow confused though. AFAIU, if there's a version mismatch (supported=3, current=2) the user needs to run `t.upgrade`, is that what we call backward compatibility? I'd have assumed compatibility meant some background update of versions transparent to the user.

By backward compatibility we mean that there is no need to run `t.upgrade` if you are not interested about v3 new features (like `semantic_labels`).
 
> Also, instead of `number_of_bands` shouldn't we use something with `semantic_label` for consistency?

Probably yes, but it's for separated PR. @marisn Any opinion?

marisn:
> > Also, instead of `number_of_bands` shouldn't we use something with `semantic_label` for consistency?
> 
> Probably yes, but it's for separated PR. @marisn Any opinion?

I'm not that familiar with TGIS internals to comment if it makes sense or not. We must keep in mind is that imagery group members are called bands. i.bands.library has band references as a subclass/a kind of semantic labels and thus having bands, band references is just fine when dealing with remote sensing data.

python/grass/temporal/abstract_space_time_dataset.py
landam: Agree, I will update this PR accordigly.

veroandreo:
> > > Also, instead of `number_of_bands` shouldn't we use something with `semantic_label` for consistency?
> > 
> > 
> > Probably yes, but it's for separated PR. @marisn Any opinion?
> 
> I'm not that familiar with TGIS internals to comment if it makes sense or not. We must keep in mind is that imagery group members are called bands. i.bands.library has band references as a subclass/a kind of semantic labels and thus having bands, band references is just fine when dealing with remote sensing data.

Agreed, but time series of raster data (STRDS) can consist of any type of raster maps, be them satellite imagery or not... this was the whole point of changing band reference to something more general

python/grass/temporal/abstract_space_time_dataset.py
landam: Done 4a2939757.

landam:
> Agreed, but time series of raster data (STRDS) can consist of any type of raster maps, be them satellite imagery or not... this was the whole point of changing band reference to something more general

@marisn @veroandreo So what about changing

```
 | Number of registered bands:. 1
 | Band names:................. S2_1
```

to

```
 | Number of reg. semantic labels:. 1
 | Semantic labels:................. S2_1
```

I can prepare separated PR if you agree.

marisn:
> > Agreed, but time series of raster data (STRDS) can consist of any type of raster maps, be them satellite imagery or not... this was the whole point of changing band reference to something more general
> 
> @marisn @veroandreo So what about changing
> 
> ```
>  | Number of registered bands:. 1
>  | Band names:................. S2_1
> ```
> 
> to
> 
> ```
>  | Number of reg. semantic labels:. 1
>  | Semantic labels:................. S2_1
> ```
> 
> I can prepare separated PR if you agree.

Fine for me.

landam:
If no objection I will merge this PR in the next days.

veroandreo:
> The format of your actual temporal database is not supported any more.



> Quick test: create strds and register 1st Sentinel band
> 
> 1. In GRASS 8 `t.info` reports:
> 
> ```
>  | Number of registered bands:. 1
>  | Band names:................. S2_1
>  | Number of registered maps:.. 1
> ```
> 
> 2. Attempt to run `t.info` on the same strds in GRASS 7.8.7dev fails with:
> 
> ```
> ERROR: Unsupported temporal database: version mismatch.
> The format of your actual temporal database is not supported any more.

Since this would be a downgrade, I'd remove the "any more" part which indicates that you have something old not supported. Instead, in this case, users will have a new format and try to read it in an "old" grass.

> Solution: You need to export it by restoring the GRASS GIS version used for
> creating this DB. From there, create a backup of your temporal database to
> avoid the loss of your temporal data.
> Notes: Use t.rast.export and t.vect.export to make a backup of your
> existing space time datasets.To safe the timestamps of your existing maps

Replace "safe" by save

> and space time datasets, use t.rast.list, t.vect.list and t.rast3d.list.
> You can register the existing time stamped maps easily if you export
> columns=id,start_time,end_time into text files and use t.register to
> register them again in new created space time datasets (t.create). After
> the backup remove the existing temporal database, a new one will be created
> automatically.
> Supported temporal database version is: 2
> Current temporal database info:
> DBMI interface:..... sqlite3
> Temporal database:..
> /home/martin/grassdata/tgis_sentinel/tgis3/tgis/sqlite.db
> ```
> 
> @metzm Should be this message modified (in separated PR) to promote `t.downgrade` in GRASS 7.8?

Sorry to step in here, but, yes, I'd promote t.downgrade as the main and most straightforward solution. Indeed, I'd shrink the message above as the case for upgrade.
 
> 3. Create new strds in GRASS 7.8
> 
> ```
> | Number of registered maps:.. 1
> ```
> 
> 4. Open strds created by GRASS 7.8 in GRASS 8:
> 
> ```
> WARNING: Temporal database version mismatch detected.
>          Please create a backup of your temporal database to avoid losing
>          data.
>          Run t.upgrade command to upgrade your temporal database.
>          Supported temporal database version is: 3
>          Your existing temporal database version: 2
>          Current temporal database info:
>          DBMI interface:..... sqlite3
>          Temporal database:..
>          /home/martin/grassdata/tgis_sentinel/tgis2/tgis/sqlite.db
> ...
>  | Number of registered bands:. None
>  | Band names:................. None
>  | Number of registered maps:.. 1
> ```
> 
> Ready for review and merge if approved.

landam:
> Sorry to step in here, but, yes, I'd promote t.downgrade as the main and most straightforward solution. Indeed, I'd shrink the message above as the case for upgrade.

See https://github.com/OSGeo/grass/pull/2002

PR #1456, opened 2021-03-13
wxGUI/mapwin: check if 'BufferedMapWindow' instance exist before calling '_runUpdateMap()' method

tmszi:
**To Reproduce**

1. Launch wxGUI
2. Display some raster elevation map e.g. `d.rast elevation`
3. Switch from 2D view -> 3D view
4. Try to switch to another LOCATION MAPSET via Data catalog
5. See error message

**Error message**

```
  File "/usr/lib64/grass79/gui/wxpython/core/gthread.py",
line 137, in OnDone

event.ondone(event)
  File "/usr/lib64/grass79/gui/wxpython/mapwin/buffered.py",
line 906, in _onUpdateMap

self._runUpdateMap()
  File "/usr/lib64/grass79/gui/wxpython/mapwin/buffered.py",
line 911, in _runUpdateMap

self._updateM(self.render, self.renderVector)
  File "/usr/lib64/grass79/gui/wxpython/mapwin/buffered.py",
line 928, in _updateM

self.Map.ChangeMapSize(self.GetClientSize())
RuntimeError
:
wrapped C/C++ object of type BufferedMapWindow has been
deleted
```

### DIFF
--- a/gui/wxpython/mapwin/buffered.py
+++ b/gui/wxpython/mapwin/buffered.py
@@ -902,7 +902,7 @@ class BufferedMapWindow(MapWindowBase, Window):
                 return
 
     def _onUpdateMap(self, event):
-        if self.timerRunId == event.pid:
+        if self and self.timerRunId == event.pid:
             self._runUpdateMap()
 
     def _runUpdateMap(self):


petrasovaa (commented):
I assume this happens only when you switch mapset immediately after switching to 3D view while the rendering is not finished?

I would add a comment with explanation there because testing 'if self' looks rather weird...

tmszi:
> I assume this happens only when you switch mapset immediately after switching to 3D view while the rendering is not finished?

I just normally waited for the map to render in 3D, and only then did I try to switch to another LOCATION's MAPSET.

petrasovaa (approved):
I don't get this bug, but this fix is harmless in any case.

PR #1459, opened 2021-03-13
lib/gis: add test for invalid value to parser

metzm:
If an option expects a number (int or double), but the answer is a string, the parser reports an error about a missing value. This PR adds a test for invalid values to the parser with corresponding error message. I find the current error message complaining about a missing value confusing when a value has been provided (thus not missing), therefore I would like to backport this change to G78.

Simple test in nc_spm_08: `r.grow in=lakes out=lakes_grow radius=twenty`

### DIFF
--- a/lib/gis/parser.c
+++ b/lib/gis/parser.c
@@ -65,7 +65,7 @@
  *    out of range.  The acceptable range (or list) will be printed.
  *
  * Overview table: <a href="parser_standard_options.html">Parser standard options</a>
- * 
+ *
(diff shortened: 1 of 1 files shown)


neteler:
Yes, to report:

```
r.grow in=lakes out=lakes_grow radius=twenty
ERROR: Invalid value <twenty> for parameter <radius>
```

thanks to this PR is much better.

neteler (approved):

metzm:
Backported to G78

PR #1460, opened 2021-03-14
libproj: added alias for ITRF92

neteler:
adds "International_Terrestrial_Reference_Frame_1992" as an alias for ITRF92

Ref: https://lists.osgeo.org/pipermail/grass-user/2021-March/082282.html

### DIFF
--- a/lib/proj/convert.c
+++ b/lib/proj/convert.c
@@ -1066,6 +1066,8 @@ static const char *papszDatumEquiv[] = {
     "Deutsches_Hauptdreiecksnetz",
     "South_American_1969",
     "South_American_Datum_1969",
+    "International_Terrestrial_Reference_Frame_1992",
+    "ITRF92",
     "ITRF_1992",
     "ITRF92",
     NULL


veroandreo (commented):
is the duplication intended?

neteler:
It is not a duplication. From the description:

_The list below is only for files that use a non-standard name for a datum that is already supported in GRASS. The number of entries must be even; they are all in pairs._

So far the "ITRF_1992" alias is resolved as "ITRF92".

This PR adds that the long version "International_Terrestrial_Reference_Frame_1992" is also resolved as "ITRF92".

veroandreo:
> It is not a duplication. From the description:
> 
> _The list below is only for files that use a non-standard name for a datum that is already supported in GRASS. The number of entries must be even; they are all in pairs._
> 
> So far the "ITRF_1992" alias is resolved as "ITRF92".
> 
> This PR adds that the long version "International_Terrestrial_Reference_Frame_1992" is also resolved as "ITRF92".
> 
> 

Ah! It's a paired list, got it! Thanks for the explanation! Dismiss the noise please :)

Issue #1480, opened 2021-03-24
[Bug] r.colors: -g (log) and -a (absolute log) don't work as expected


HuidaeCho:
**Describe the bug**
`r.colors -g` clears the map and `r.colors -a` produces different colors from `r.colors -g` for non-negative rasters.

**To Reproduce**

```bash
v.to.rast input=censusblk use=attr attcol=MEDIAN_AGE output=censusblk_median_age
r.colors map=censusblk_median_age col=plasma
r.colors map=censusblk_median_age col=plasma -g
r.colors map=censusblk_median_age col=plasma -a
```

**Expected behavior**

Same colors for raster cell, log(raster cell), and log(abs(raster cell)).

**Screenshots**

```bash
v.to.rast input=censusblk use=attr attcol=MEDIAN_AGE output=censusblk_median_age
r.colors map=censusblk_median_age col=plasma
```
![image](https://user-images.githubusercontent.com/7456117/112320920-da8aca80-8c85-11eb-852d-431cf0935751.png)
```bash
r.colors map=censusblk_median_age col=plasma -g
```
![image](https://user-images.githubusercontent.com/7456117/112321154-1887ee80-8c86-11eb-9450-1e0e9792fded.png)

```bash
r.colors map=censusblk_median_age col=plasma -a
```
![image](https://user-images.githubusercontent.com/7456117/112321059-fa21f300-8c85-11eb-995d-3e4ebce11441.png)

**System description (please complete the following information):**

- Operating System: Linux
- GRASS GIS version: master

**Additional context**

The minimum of `censusblk_median_age` is 0, so `log(min)` is `-inf`
https://github.com/OSGeo/grass/blob/647162b017cfa70c05f7274f7eb0f4a0bfda5892/lib/raster/color_xform.c#L197
and `lx` is `-nan`.
https://github.com/OSGeo/grass/blob/647162b017cfa70c05f7274f7eb0f4a0bfda5892/lib/raster/color_xform.c#L219
This is why its colr file is empty with just header, I think:
```
% 0 -1
nv:255
*:255
```
```

**Comment by HuidaeCho** [comment]
```bash
r.mapcalc ex='tmp=if(censusblk_median_age==0,1,censusblk_median_age)'
r.colors map=tmp col=plasma -g
```
![image](https://user-images.githubusercontent.com/7456117/112325147-d19bf800-8c89-11eb-97ae-a9e5b1299bd0.png)

HuidaeCho:
[...]

PR #1481, opened 2021-03-24
Rast_log_colors: Use a sane default color for non-positive rasters and assume a positive min cell value

HuidaeCho:
An attempt to address #1480.

* If raster is positive, everything is good.
* If max cell value is non-positive (<= 0), use the max color only.
* If min cell value is non-positive (<= 0), set color for min up to `GRASS_EPSILON` to the default color and assume a new positive min cell value using `GRASS_EPSILON` (currently, 1e-15).

A potential issue with this assumption: If cell values are between 0 and 1e-15, they'll get assigned the default color.

### DIFF
--- a/lib/raster/color_xform.c
+++ b/lib/raster/color_xform.c
@@ -188,7 +188,6 @@ void Rast_log_colors(struct Colors *dst, struct Colors *src, int samples)
     DCELL min, max;
     double lmin, lmax;
     int red, grn, blu;
-    int use_lmin_eps;
     DCELL prev;
     int i;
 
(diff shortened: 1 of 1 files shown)


HuidaeCho:
If 0 < max <= GRASS_EPSILON? Just default color.. Maybe, the max color from the rule would be better.

HuidaeCho:
Still, ideally, it would be better to find actually positive min and use that instead of GRASS_EPSILON.

HuidaeCho:
Probably, setting anything below the positive min cell value (or GRASS_EPSILON currently) to the min color rather than default would be better when the default "default color" is white? Or would white be better to let the user know that the raster has non-positive cells (current PR)?

HuidaeCho:
I think GRASS_EPSILON was too small. I changed `eps` to 1 (default for absolute log). Now,
```bash
v.to.rast input=censusblk output=censusblk_median_age use=attr attribute_column=MEDIAN_AGE
r.colors map=censusblk_median_age col=plasma -g
r.colors map=censusblk_median_age col=plasma -a
```
produces the same output.
image

cmbarton:
I’m traveling and won’t be able to check this until next week. Maybe see if @nilason, @petrasovaa, or @hmitaso can test.

HuidaeCho:
Maybe, to be consistent with #1507, for max <= 0, fatal error rather than using the max color only?

cmbarton:
Is there something to test here or has it been merged into master/7.8?

HuidaeCho:
> 
> 
> Is there something to test here or has it been merged into master/7.8?

No, it's not merged yet. The main test should be with rasters with min=0. Without this PR, you'll see #1480.

HuidaeCho:
This PR reduces the color space if there are non-positive cell values while #1937 utilizes the full space by shifting cell values. Closing it in favor of #1937.

PR #1485, opened 2021-03-28
wxGUI: one row for lmgr toolbar after adding 3D view tab

lindakladivova:
Before:
chyba

After:
dobre


wenzeslaus (approved):

wenzeslaus:
Works for me. Thanks.

Just for the record: A more radical change related to this (for some other PR) is removing the 3D three-button toolbar. Possible fixes include creating another toolbar within the 3D view tab and moving these three buttons to 3D view pane itself in some other way than a toolbar.

wenzeslaus:
PS: See how I modified the commit message for a7d1f47329ad0d952b36321fe3e9681697cca990.

Issue #1482, opened 2021-03-24
[Feat] d.legend: allow log scale when min value 0 like with population maps


neteler:
**Is your feature request related to a problem? Please describe.**

When plotting population density maps (e.g. Worldpop), `d.legend` does not allow log scale plotting when the minimum value = 0.
While it mathematically makes sense it does less so from a user point of view since a copy of the map is needed with being 0.000xxx filtered out.

**Describe the solution you'd like**
A flag or behaviour as probably implemented in #1481 to simplify user's life.

**Example**

```
grass79 ~/grassdata/nc_spm_08_grass7/user1/
g.region raster=elevation res=750 -ap

# download Worldpop USA map
wget ftp://ftp.worldpop.org.uk/GIS/Population/Global_2000_2020_1km_UNadj/2020/USA/usa_ppp_2020_1km_Aggregated_UNadj.tif

r.import input=usa_ppp_2020_1km_Aggregated_UNadj.tif output=usa_ppp_2020_1km_Aggregated_UNadj extent=region resolution=value resolution_value=750
r.colors usa_ppp_2020_1km_Aggregated_UNadj color=population_dens

r.info -r usa_ppp_2020_1km_Aggregated_UNadj 
min=0
max=4805.798

d.mon wx0
d.rast usa_ppp_2020_1km_Aggregated_UNadj
d.legend -l raster=usa_ppp_2020_1km_Aggregated_UNadj
ERROR: Range [0.000, 4805.798] out of the logarithm domain.

```

HuidaeCho:
@neteler Do you know an **efficient** way to retrieve the positive min value from a raster? Using GRASS_EPSILON in #1481 stretches the color space too extremely. I want to do this:
1. Set negative and 0 to the default color (default default is white)
2. Use the positive min and max for log colors.

neteler:
> @neteler Do you know an **efficient** way to retrieve the positive min value from a raster? 

Probably @metzm knows...

But isn't for this case only relevant to change 0.00000 to e.g. 0.0000000001? (not sure how to write it properly, needs to meet the C precision conventions)

HuidaeCho:
> > @neteler Do you know an **efficient** way to retrieve the positive min value from a raster?
> 
> Probably @metzm knows...
> 
> But isn't for this case only relevant to change 0.00000 to e.g. 0.0000000001? (not sure how to write it properly, needs to meet the C precision conventions)

[...]

PR #1507, opened 2021-04-06
d.legend: handle 0 for log (-l)

HuidaeCho:
Attempt to address #1482

### DIFF
--- a/display/d.legend/draw.c
+++ b/display/d.legend/draw.c
@@ -44,7 +44,7 @@ void draw(const char *map_name, int maptype, int color, int thin, int lines,
     struct Range range;
     struct FPRange fprange, render_range;
     CELL min_ind, max_ind;
-    DCELL dmin, dmax, val;
+    DCELL dmin, dmax, val, eps = 1.0;
     CELL min_colr, max_colr;
     DCELL min_dcolr, max_dcolr;
     int x0, x1, y0, y1, xyTemp;
@@ -306,13 +306,6 @@ void draw(const char *map_name, int maptype, int color, int thin, int lines,
             }
         }
 
-        /* In case of log. scale raster doesn't contain negative or zero values */
-        if (log_sc)
-            if ((dmin<=0) || (dmax<=0))
-                G_fatal_error(_("Range [%.3f, %.3f] out of the logarithm domain."),
-                              dmin, dmax);
-
-
         if (use_catlist) {
             for (i = 0; i < catlistCount; i++) {
                 if ((catlist[i] < dmin) || (catlist[i] > dmax)) {
@@ -375,7 +368,12 @@ void draw(const char *map_name, int maptype, int color, int thin, int lines,
             for (k = 0; k < lleg; k++) {
                 if (log_sc) { /* logarithmic scale */
                     num = k / lleg;
-                    val = dmin * pow(dmax/dmin, num);
+		    if (dmax <= 0)
+			val = dmax;
+		    else if (dmin <= 0)
+			val = eps * pow(dmax/eps, num);
+		    else
+			val = dmin * pow(dmax/dmin, num);
                     D_d_color(val, &colors);
                     if (!flip) {
                         if (horiz)
@@ -508,8 +506,12 @@ void draw(const char *map_name, int maptype, int color, int thin, int lines,
                         buff[0] = 0;    /* no text */
                     else {
                         if (log_sc) {
-                            num = log10(dmax) - k * ((log10(dmax) - log10(dmin)) / (steps - 1));
-                            val = pow(10,num);
+			    if (dmax <= 0)
+				val = dmax;
+			    else {
+				num = log10(dmax) - k * ((log10(dmax) - log10(dmin > 0 ? dmin : eps)) / (steps - 1));
+				val = pow(10,num);
+			    }
                         }
                         else{
                         if (!flip)
@@ -536,7 +538,10 @@ void draw(const char *map_name, int maptype, int color, int thin, int lines,
                         D_get_text_box(buff, &bb, &bt, &bl, &br);
                         if (!horiz) {
                             if (log_sc) {
-                                coef = (log10(val) - log10(dmin)) / (log10(dmax) - log10(dmin));
+				if (dmax <= 0)
+				    coef = 1;
+				else
+				    coef = (log10(val) - log10(dmin > 0 ? dmin : eps)) / (log10(dmax) - log10(dmin > 0 ? dmin : eps));
                                 if (flip)
                             D_pos_abs(x1 + label_indent,
                                               y1 - coef * lleg + (bb - bt) / 2);
@@ -565,7 +570,10 @@ void draw(const char *map_name, int maptype, int color, int thin, int lines,
                         }
                         else {
                             if (log_sc) {
-                                coef = (log10(val) - log10(dmin)) / (log10(dmax) - log10(dmin));
+				if (dmax <= 0)
+				    coef = 1;
+				else
+				    coef = (log10(val) - log10(dmin > 0 ? dmin : eps)) / (log10(dmax) - log10(dmin > 0 ? dmin : eps));
                                 if (flip)
                                     D_pos_abs(x1 - coef * wleg - ((br - bl) / 2),
                                       y1 + label_indent + txsiz);
@@ -625,7 +633,10 @@ void draw(const char *map_name, int maptype, int color, int thin, int lines,
                 }
 
                 if (log_sc) {
-                    coef = (log10(tick_values[i]) - log10(dmin)) / (log10(dmax) - log10(dmin));
+		    if (dmax <= 0)
+			coef = 1;
+		    else
+			coef = (log10(tick_values[i]) - log10(dmin > 0 ? dmin : eps)) / (log10(dmax) - log10(dmin > 0 ? dmin : eps));
                 }
                 else
                 coef = (tick_values[i] - dmin) / ((dmax - dmin) * 1.0);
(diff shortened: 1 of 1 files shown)


HuidaeCho:
TODO: This module needs `indent` after this PR is merged.

HuidaeCho:
@neteler Please test it with negative, non-positive, negative to positive, and 0 to positive rasters. It should affect positive rasters, but just in case test them too. It still doesn't support absolute logarithm.

Noticed that the flag name (`-l`) is different from `-g` from `r.colors`, but `-l` makes more sense.

neteler:
Thanks a lot!
For testing, I applied the PR to G78 and got this compilation warning (just FYI):

```
grass78_git/display/d.legend]$ make
gcc  -O3 -march=native -fdiagnostics-color   -I/home/mundialis/software/grass78_git/dist.x86_64-pc-linux-gnu/include -I/home/mundialis/software/grass78_git/dist.x86_64-pc-linux-gnu/include    -DPACKAGE=\""grassmods"\"   -I/home/mundialis/software/grass78_git/dist.x86_64-pc-linux-gnu/include -I/home/mundialis/software/grass78_git/dist.x86_64-pc-linux-gnu/include -DRELDIR=\"display/d.legend\" -o OBJ.x86_64-pc-linux-gnu/draw.o -c draw.c
draw.c: In function ‘draw’:
draw.c:273:40: warning: ‘%d’ directive writing between 1 and 11 bytes into a region of size 4 [-Wformat-overflow=]
  273 |                 sprintf(DispFormat, "%%%dd", (int)(log10(fabs(maxCat))) + 1);
      |                                        ^~
draw.c:273:37: note: directive argument in the range [-2147483647, 2147483647]
  273 |                 sprintf(DispFormat, "%%%dd", (int)(log10(fabs(maxCat))) + 1);
      |                                     ^~~~~~~
In file included from /usr/include/stdio.h:866,
                 from /home/mundialis/software/grass78_git/dist.x86_64-pc-linux-gnu/include/grass/gis.h:24,
                 from draw.c:14:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:38:10: note: ‘__builtin___sprintf_chk’ output between 4 and 14 bytes into a destination of size 5
   38 |   return __builtin___sprintf_chk (__s, __USE_FORTIFY_LEVEL - 1,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   39 |       __bos (__s), __fmt, __va_arg_pack ());
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Here my user test:
- not more min=0 related error, great
- the maximum shown is not that reported thru stats (see below, but I didn't manage to reproduce it!):

```
GRASS 7.8.6dev (latlong_wgs84):~ > r.info -r worldpop_2020_1km_aggregated_UNadj
min=0
max=52820.77
GRASS 7.8.6dev (latlong_wgs84):~ > r.univar worldpop_2020_1km_aggregated_UNadj -g | grep min
min=0

d.legend -s -d -l raster=worldpop_2020_1km_aggregated_UNadj@worldpop_south_america title=worldpop_2020_1km_aggregated_UNadj font=Vera
```

image

HuidaeCho:
>     * the maximum shown is not that reported thru stats (see below, but I didn't manage to reproduce it!):

@neteler I couldn't reproduce it. Can you share your data (worldpop_2000_1km_aggregated_UNadj)?

HuidaeCho:
> draw.c:273:40: warning: ‘%d’ directive writing between 1 and 11 bytes into a region of size 4 [-Wformat-overflow=]
>   273 |                 sprintf(DispFormat, "%%%dd", (int)(log10(fabs(maxCat))) + 1);
>       |                                        ^~
> draw.c:273:37: note: directive argument in the range [-2147483647, 2147483647]
>   273 |                 sprintf(DispFormat, "%%%dd", (int)(log10(fabs(maxCat))) + 1);
>       |                                     ^~~~~~~

I think you use GCC >= 7 (https://gcc.gnu.org/gcc-7/changes.html). I'm stuck with GCC 5 :-(. The warning is because DispFormat is not big enough. Buffer overflow can actually happen if maxCat is *very* large such that log10(maxCat)+1 >= 100. That is, maxCat >= 10^99, which looks pretty safe... One solution would be `DispFormat[7]` (two more bytes) and
```c
sprintf(DispFormat, "%%%dd", (char)(log10(fabs(maxCat))) + 1);
```
`char` ranges from -128 to 127 so it can take up four bytes max plus `%`, `d`, and `\0`.

Now, if maxCat is very small and log10(fabs(maxCat)) becomes < -1, text will be left-aligned *unintentionally*.

In general, I think this module needs a major refactoring including the above warnings, uninitialized variables, indentation, breaking long code, etc.

neteler:
> > ```
> > * the maximum shown is not that reported thru stats (see below, but I didn't manage to reproduce it!):
> > ```
> 
> @neteler I couldn't reproduce it. Can you share your data (worldpop_2000_1km_aggregated_UNadj)?

(btw: gcc version 10.2.1 here)

My use case is this one:

https://github.com/OSGeo/grass/issues/1482#issue-839905515

but I'll test again to see if it happens again + (try to) provide synthetic test data generated with r.random.* etc.

neteler:
> @neteler Please test it with negative, non-positive, negative to positive, and 0 to positive rasters. It shouldn't affect positive rasters, but just in case test them too. It still doesn't support absolute logarithm.

OK, I have set up some test cases (see below). Hope they are somehow meaningful...

```
# NC sample dataset
g.region -p n=228500 s=215000 w=630000 e=645000 res=100 -p
d.mon wx0
```

## Negative values
```
# negative, DCELL
r.surf.random out=random_negative min=-100 max=-1
r.univar random_negative
r.info -r random_negative
min=-99.9995498657227
max=-1.01180768013

d.rast random_negative
d.legend raster=random_negative@user1 title=random_negative
d.legend -l raster=random_negative@user1 title="Log random_negative"
```

image

## Negative to positive
```
# negative to positive, DCELL
r.surf.random out=random_negative_positive min=-100 max=100
r.univar random_negative_positive
r.info -r random_negative_positive
min=-99.9983367919922
max=99.9910354614258

d.rast random_negative_positive
d.legend raster=random_negative_positive@user1 title=random_negative_positive
d.legend -l raster=random_negative_positive@user1 title="Log random_negative_positive"
```

image

## 0 to positive
```
# 0 to positive rasters, DCELL
r.surf.random out=tmp min=0 max=100
r.mapcalc "random_positive = if( tmp < 1.0, 0.0, tmp)"
g.remove raster name=tmp -f
r.univar random_positive
r.info -r random_positive
min=0
max=99.9885711669922

d.rast random_positive
d.legend raster=random_positive@user1 title=random_positive
d.legend -l raster=random_positive@user1 title="Log random_positive"
 ```

image

**Comment by veroandreo** [comment]
These are my tests with and without this PR: 

- Without the PR: 

1) positive values starting from 0.something (lots of small values some pretty high)


2) negative and positive values - only legend without `-l` can be drawn, for `-l`, I get the reported error
```
Failed to run command 'd.legend -b -l raster=prueba_grib@pruebas title="With -l flag" at=5,50,47,50'. Details:
GRASS_INFO_ERROR(207604,1): Range [-69.550, 42.850] out of the logarithm domain.
GRASS_INFO_END(207604,1)     
```

- With this PR:

1) looks good - same as before the PR


2) legend with `-l` does not look good, only positive values shown, though I no longer get the error


I wonder if it really make sense to try to plot a logarithmic scale with negative data... For the case of value zero, can't we just add 1 to get 0 as a result?

Also, while no error is reported for negative data, I believe it's worthy to raise a warning at least

**Comment by HuidaeCho** [comment]
Let me summarize what this PR does:
* For a positive raster, everything is good. Output should be exactly the same as without this PR.
* For a 0-to-positive or negative-to-positive raster, use 1 as the min color
  * It can be a problem when the raster's max is smaller than 1. This is why it would be great to know the true positive min value.
  * We also skip all raster colors between the min and 1.
* For a negative raster, use the max color only.
  * Actually, this behavior can be wrong most of the case unless the raster is colorized using `r.colors -g`, which does use the max color only, but there is no smart way to know if this is the case.
  * Probably, we should throw a fatal error in this case because the legend won't match the display most of the case. Commented as XXX in the PR.

**Comment by HuidaeCho** [comment]
> > > @neteler Please test it with negative, non-positive, negative to positive, and 0 to positive rasters. It shouldn't affect positive rasters, but just in case test them too. It still doesn't support absolute logarithm.
> > 
> > 
> > OK, I have set up some test cases (see below). Hope they are somehow meaningful...
> 
> I find the legends pretty odd, tbh, even the one for all positive numbers... I would expect colors to be more equally distributed as in my case 1 examples. Maybe the distribution of values plays a role?

I think your example 1 is a good case for a logarithmic legend. Data looks logarithmically distributed than linearly. The log legend looks good to me.

**Comment by HuidaeCho** [comment]
Now, fatal error on a non-positive raster and warnings on min <= 0.

**Comment by neteler** [comment]
This PR looks good, i.e., it improves the current situation. Merge it?

File: display/d.legend/draw.c [hunk 1]
```diff
@@ -323,13 +323,20 @@ void draw(const char *map_name, int maptype, int color, int thin, int lines,
...
+                G_fatal_error(_("No positive cell values found; "
+                                "logarithmic legend cannot be displayed"));
+            if (dmin > 0)
+                eps_min = dmin;
+            else {
+                G_warning(_("Non-positive cell values found; "
+                            "discarding colors for cells <= %g and "
+                            "assuming a positive min cell value of %g"),
+                          eps, eps);
+                eps_min = eps;
```
petrasovaa, line 337: I wonder if you couldn't just use `dmin = eps_min` here, which would avoid all the changes below.

display/d.legend/draw.c
```diff
@@ -395,7 +402,16 @@ void draw(const char *map_name, int maptype, int color, int thin, int lines,
...
+                    /* XXX: for log scale, if the entire raster is not
+                     * positive, use the max color only; actually, this is
+                     * incorrect because negative values can have different
+                     * colors on the monitor; maybe, we should throw a fatal
+                     * error in this case because using the max color only for
+                     * displaying a raster and its legend is different and the
+                     * legend must always match the display; well... unless the
+                     * raster is colorized using r.colors -g, but how do we
+                     * know that? */
+                    val = dmax <= 0 ? dmax : eps_min * pow(dmax / eps_min, num);
```
petrasovaa, line 414: All these comments are somewhat confusing and perhaps not updated after introducing the fatal error for rasters with all negative values. I would suggest to summarize the reasoning and perhaps move it somewhere close to the fatal error call.

Given the fatal error in the beginning, these expressions don't need to test for the negative cases and can be simplified. This applies to all.most of the changes below.

petrasovaa (changes requested):
I think this implementation is improvement, but the PR needs to be cleaned up, see my other comments.

echoix:
Solved conflicts

PR #1509, opened 2021-04-07
pythonlib: Enable unused var warning

wenzeslaus:
Ignore whole temporal subpackage.
Remove unused in imaging including some Python 2 support (and fix a typo).


wenzeslaus:
I'm wondering if the following is safe to delete or if it should be used in function `grass.script.start_command`:

```
./script/core.py:468:9: F841 local variable 'encoding' is assigned to but never used
        encoding = kwargs.pop("encoding")
        ^
1     F841 local variable 'encoding' is assigned to but never used
```

wenzeslaus:
The encoding issue in start_command is now ignored directly in the source code and reported as #1521.

PR #1512, opened 2021-04-07
Git: Un-ignore locale directory

wenzeslaus:
The whole locale dir was ignored since 4c874e8110df823ed575fec7172a0817a6e5a7db.
Most of the dir is source code, so ignoring just the scriptstrings subdir.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 *.lock
 *.pyc
 OBJ.*
-locale/*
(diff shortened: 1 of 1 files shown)


wenzeslaus:
Requires #1513 to not fail the Black check.

HuidaeCho (approved):
PO files are already tracked. Looks good to me.

PR #1513, opened 2021-04-07
Apply Black to locale

wenzeslaus:
This applies Black 20.8b1 to po file stats script.
Uses: black --target-version=py36


wenzeslaus:
Will be checked in CI only after #1512.

PR #1525, opened 2021-04-13
pythonlib: Use more specific language in CalledModuleError

wenzeslaus:
The error message comming from CalledModuleError was not sufficient for users to find
the error output of a module. This now explicitly says to look above traceback (case
for user scripts) while mentioning the more general case of testing framework and
(Jupyter) notebooks with default (GRASS) settings where stderr shows in a different
place than the traceback.

The module name is now included only if provided and if not already present avoiding
the `None r.univar` and `r.univar r.univar` cases. Backticks are used to mark the code,
although a) the code formatting is still unresolved (e.g., a list will render as just
Python list) and b) backticks are not common in GRASS (but fit perfectly this context
in general).

This also rewrites the code with format, constructs the message without plus, and
extends documentation (in-code and docstrings).

### DIFF
--- a/python/grass/exceptions/__init__.py
+++ b/python/grass/exceptions/__init__.py
@@ -58,25 +58,43 @@ class Usage(Exception):
 class CalledModuleError(subprocess.CalledProcessError):
     """Raised when a called module ends with error (non-zero return code)
 
-    :param module: module name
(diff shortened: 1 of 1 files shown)


ninsbl (approved):
Makes sense. Good improvement.

PR #1526, opened 2021-04-13
lib/init: set HISTSIZE to empty string for unlimited size

nilason:
Value "-1" only works for bash 4.3 and later, whereas built-in bash on Mac is frozen at 3.2.57.


nilason:
Should add that HISTSIZE set to "-1" with build-in bash on Mac is the same as "0" – quite the opposite to unlimited. I tested the empty string with bash 5.1.4 and it seems to work.

nilason:
Any thoughts on this, @wenzeslaus, @cwhite911 ?

susi (approved):

wenzeslaus (approved):
This fits with what I have seen. Originally, I did not consider old Bash an issue, but it is.

PR #1527, opened 2021-04-15
Apply Black to wxGUI

wenzeslaus:
This applies Black 20.8b1 to wxGUI and GUI scripts in gui dir.
Uses Black with settings from pyproject.toml file, i.e., enables the gui subdir.
Ignores lmgr/layertree.py because of a bug in Black.
Fixed formatting checks enabled in Flake8.


wenzeslaus:
As with the other large changes, this will likely conflict with other open PR. However, no need to despair. Based on the previous experience, the fix is usually trivial. You have couple of options:

1) Just merge updated upstream/master into your branch and hope for the best. (You can rebase, too, but don't do it when you have multiple commits otherwise you will have to resolve the conflicts for each commit you did.)
2) For some reason, it seems to me applying Black to your branch before the merge is somehow cleaner (you need to commit before merge). If nothing else, you are resolving conflicts between two Black compliant code bases. (Git works in the way that files which are same will just merge without conflicts.)
3) Apply Black to your branch. Than create a new branch from an updated master branch and use something like [Meld](https://stackoverflow.com/questions/2006032/view-differences-of-branches-with-meld) to get changes from your old branch to your new branch. You can force push this new branch to your origin (fork) branch of the old name which will replace it in the pull request.

For options 1 and 2, the conflicts will be only the lines your touched. For a short change, anything will work. For more changes, so Git conflict resolution support (like e.g., [VS Code/VSCodium](https://stackoverflow.com/questions/38216541/visual-studio-code-how-to-resolve-merge-conflicts-with-git) has) in editor comes handy.

In any case, commit your current changes before you do anything.

PR #1535, opened 2021-04-17
wxGUI: Enable all Flake8 warnings

wenzeslaus:
* Enable remaning Flake8 warnings globally.
* More precise (narrow) pattern for scripts and temporal dirs.
* Previously ignored gui/scripts directory is now checked.
* Simplify d.rast3d.
* Fix some hash-hash comments and long lines.
* Ignore remaining warnings using per file ignores.
* Break down some directory ignore into true per file ignores for easier fixing
  and greater current checks.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -30,6 +30,8 @@ per-file-ignores =
     doc/gui/wxpython/example/dialogs.py: F401
     doc/gui/wxpython/example/g.gui.example.py: E501
     locale/grass_po_stats.py: E122, E128, E231, E401, E501, E722, E741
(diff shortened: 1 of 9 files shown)


neteler:
This is outstanding, thanks for your hard work, @wenzeslaus! 
Worth a news entry on the website..

wenzeslaus:
Thanks for checking this out @neteler. News would be great, but with highlighting the need for resolving the list(s) of temporary ignores for specific files. Great for new and potential contributors ("good first issue"). The grass pkg (python/grass) still has some warnings enabled globally.

PR #1539, opened 2021-04-19
gunittest: Deduplicate output decoding code

wenzeslaus:
* Resolves code duplication between stdout and stderr decoding.
* Removes while, use of indices, and control variable.
* Documents the operation (although not the motivation).
* Adds direct bytes.decode call with replace as a fallback.
* Uses UnicodeError instead of the bare except (original motivation for this change).


ninsbl (approved):
Looks good.

PR #1543, opened 2021-04-20
pythonlib: Reduce usage of bare except in script

wenzeslaus:
* Finer error reporting in the parser() function.
* Only ValueError errors in the key-value parsing function.
* Only specific errors in the del_temp_region() function.
* Generic Exception used instead of bare except (satisifies Flake8, although not Pylint).


ninsbl (approved):
Looks good as well.

PR #1544, opened 2021-04-20
pygrass: Remove usage of bare except

wenzeslaus:
* Plain try-finally is how re-raising is expressed according to the doc:
  'If there is a saved exception it is re-raised at the end of the finally clause.'
* datetime.strptime() produces ValueError.
* Only ImportError for import statement.


ninsbl (approved):
Looks good. (putting the "finally" statement on my what-I-learned-today list...)

PR #1546, opened 2021-04-21
pythonlib: Remove star imports

wenzeslaus:
Replaces star import (from x import *) by regular imports (from x import y).

Enables the relevant Flake8 warnings.

Adds the warning to per-file ignores for init files where we expect and allow
using star imports (unfortunatelly some init files contain actual code suffers
from these warning being disabled, but that's a different issue).


wenzeslaus:
@lrntct Can you please review the temporal imports in this PR? There are test and they seem to be running fine (same number of passing test files), so totally optional, just for extra safety.

PR #1549, opened 2021-04-22
pythonlib: Enable bare except warning

wenzeslaus:
* Enable Flake8 warning for bare except globally.
* Fix files with only ImportError missing or bare except-raise with finally
  (where finally is sufficient).
* Replace temporal directory ignores by true per-file ignores.


wenzeslaus:
Depends on #1543 and #1544.

Issue #1555, opened 2021-04-28
[Bug] PyGRASS: Confusing TypeError on VectorTopo.open.


Viech:
**Describe the bug**
I can read `archsites` of `spearfish_grass70data_0` with `VectorTopo` just fine but when I use `v.select` to filter points not in the computational region, opening the resulting vector map with `VectorTopo` raises `TypeError: Driver is not supported yet, pleas use: sqlite or pg`. I suspect that the filtered map is missing some important metadata (I'm happy to receive a hint on how to make it work) but in any case **the raised `TypeError` is confusing**, as I did not specify any database driver. Additional error handling seems necessary to tell the user exactly why the map can't be loaded.

**To Reproduce**
```python
#!/usr/bin/env python3

from grass.pygrass.modules import Module
from grass.pygrass.vector import VectorTopo

try:
    a = VectorTopo("archsites", mode="r")
    a.open()  # works

    Module("v.in.region", output="test_region", quiet=True)
    Module(
        "v.select",
        ainput="archsites",
        binput="test_region",
        operator="overlap",
        output="test_filtered",
        quiet=True,
    )

    b = VectorTopo("test_filtered", mode="r", quiet=True)
    b.open()  # confusing error
finally:
    Module(
        "g.remove",
        type="vector",
        name=("test_region", "test_filtered"),
        flags="f",
        quiet=True,
    )
```
Output:
```
Traceback (most recent call last):
  File "…/test.py", line 21, in <module>
    b.open()  # misleading error
  File "/opt/grass/etc/python/grass/pygrass/vector/abstract.py", line 393, in open
    self.table = self.dblinks.by_layer(layer).table()
  File "/opt/grass/etc/python/grass/pygrass/vector/table.py", line 792, in table
    return Table(self.table_name, self.connection(), self.key)
  File "/opt/grass/etc/python/grass/pygrass/vector/table.py", line 775, in connection
    raise TypeError(str_err)
TypeError: Driver is not supported yet, pleas use: sqlite or pg
```

**Expected behavior**
Opening the map should either work, if possible, or raise an exception that makes it clear why it would not work.

**Screenshots**
–

**System description (please complete the following information):**

- Operating System: Linux 5.11.16-arch1-1
- GRASS GIS version: 7.8.5

```shell
% g.version -rge
version=7.8.5
date=2020
revision=4e3f393
build_date=2021-04-27
build_platform=x86_64-pc-linux-gnu
build_off_t_size=8
[...]

PR #1559, opened 2021-04-30
PyGRASS: Improve exception handling in Info.open (#1555)

Viech:


python/grass/pygrass/vector/abstract.py
```diff
@@ -413,14 +413,33 @@ def open(
...
             self.n_lines = 0
         else:
-            self.layer = self.dblinks.by_layer(layer).layer
-            self.table = self.dblinks.by_layer(layer).table()
+            layer_db_link = self.dblinks.by_layer(layer)
+            if not layer_db_link:
+                raise LookupError(
+                    "There appears to be no database link for layer %d of <%s>."
+                    % (layer, self.name)
+                )
```
Viech, line 427: Previously, this would error because it would try to access `None.layer`, so I kept the general behavior of producing an exception. However, if there are zero links (case above) then it is OK to just not load the table, so this may still be inconsistent behavior.

python/grass/pygrass/vector/abstract.py
```diff
@@ -413,14 +413,33 @@ def open(
...
+            if layer_db_link.layer != layer:
+                raise RuntimeError(
+                    "The databse link for layer %d of <%s> references layer %d."
+                    % (layer, self.name, layer_db_link.layer)
+                )
```
Viech, line 432: This is a new exception; the previous code would just set `self.layer = layer_db_link.layer`. I don't know if there is a scenario where one would want to allow such a layer number mismatch.

python/grass/pygrass/vector/abstract.py
```diff
@@ -413,14 +413,33 @@ def open(
...
+            self.layer = layer
+            try:
+                self.table = layer_db_link.table()
+            except Exception as error:
+                raise RuntimeError(
+                    "Loading the attribute table for layer %d of <%s> failed: %s"
+                    % (layer, self.name, error)
+                )
```
Viech, line 440: In a hypothetical future where you require a recent Python version, this exception should not copy the error message but just `raise … from error`.

wenzeslaus: The code on master branch is Python >=3.6, so if you don't want this backported to 7.8 or you prepare another PR for the 7.8 branch, you can do that already.

Viech: Good to hear, I was judging from the `from __future__ import print_function` statement I saw. :wink:
I'll update the commit accordingly.

wenzeslaus: Right, another PR needed there.

ninsbl:

python/grass/pygrass/vector/table.py
```diff
@@ -714,9 +714,11 @@ def _get_driver(self):
         return decode(self.c_fieldinfo.contents.driver)
 
     def _set_driver(self, driver):
-        if driver not in ("sqlite", "pg"):
-            str_err = "Driver not supported, use: %s." % ", ".join(DRIVERS)
-            raise TypeError(str_err)
+        if driver not in DRIVERS:
+            raise NotImplementedError(
```
wenzeslaus, line 718: From the Python doc, it actually seems to me that ValueError is more appropriate. NotImplementedError is related to abstract and unfinished class and it does not mention missing parts of user code. Also, will I get this if I pass `non-sense-drive-name`? ValueError seems quite appropriate then. Unless there is a good reason for NotImplementedError, I think this needs to be ValueError.

> [exception ValueError](https://docs.python.org/3/library/exceptions.html#ValueError)
> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

> [exception NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError)
> This exception is derived from [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError). In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

wenzeslaus (changes requested):
After consulting Python documentation, I found one more issue here.

The rest of the code seems to be ready to be merged.

I merged main branch here to resolve conflict with new code on main (only a context line changed).

python/grass/pygrass/vector/table.py
Viech: If a driver is supported by GRASS but not by PyGRASS, then `NotImplementedError` is appropriate ("your value is valid in theory but we haven't implemented this feature"). I will raise it for `ogr` and `dbf` only. Otherwise, indeed, `ValueError` would be correct. One could make a case for raising further a `TypeError` for non-string arguments but I think this would be over-engineered and possibly unpythonic.

(As far as I know abstract base methods are supposed to just `pass`; an exception for them is raised by Python.)

wenzeslaus: Do you have some examples or best practices which show `NotImplementedError` should be used here? The use case in Python doc for `NotImplementedError` mentions two one use cases and that is abstract methods and not yet implemented methods, although, as these two SO/SE posts mention (https://stackoverflow.com/a/44316063/574907, https://softwareengineering.stackexchange.com/a/231398/221122), it does not explicitly exclude other use cases, the focus seems to be on methods (and both posts also state that in the context of todos). I think the `pass` you are thinking about applies to defining methods using the *abc* package.

I searched for "NotImplementedError + backend" to get some examples. In Celery, `NotImplementedError` is really used for methods which are not implemented. Not necessarily todo, but not implemented methods which are there to satisfy interface.

It seems that [usb/backend/libusb1](https://github.com/pyusb/pyusb/blob/master/usb/backend/libusb1.py#L600) uses that outside of context of methods, although the actual context is hard to tell for me.

On the other hand, [pyca/cryptography](https://github.com/pyca/cryptography/commit/3225358d1581fc5041663895aa33898db20b0a5e) had `NotImplementedError` there and then they changed it to ValueError.

I like the last change with two different errors and I agree with not raising type error. Given the above, I think it should be two value errors with different messages. However, I'm willing to merge it as is. Let me know what you think about it.

wenzeslaus:
All right, this can be merged as is. (CI is still running now.)

@Viech Two questions:

1) You are using `%` formatting, but in this context `.format()` or f-string would be more appropriate with current Python. Any reason for that?

2) Your error messages are not translatable (not marked with `_(...)`, `.format()` or `%` are then needed, not f-string). Is that intentional? Python does not translate its error messages, but in GRASS we generally do, although for programming-only messages it is permissible not to translate (most of the new grass.jupyter does not have translated messages now). It would be good to discuss this on grass-dev and arrive at some best practice.

PR #1567, opened 2021-05-06
v.hull: use standard C boolean type

### LINKED ISSUE #1563: [Bug] v.hull fails to compile in GRASS 7.8.5
**Describe the bug**
When compiling GRASS 7.8.4 against GDAL 3.3.0 the v.hull module fails.

**To Reproduce**
```
GRASS GIS 7.8.5 exported compilation log
Started compilation: Mon 03 May 2021 10:51:18 PM MST
--
Errors in:
/tmp/makepkg/grass/src/grass-7.8.5/vector/v.hull
```

and the error from the v.hull folder in the src package gives:

```
grass/src/grass-7.8.5/vector/v.hull % make
gcc -D_FORTIFY_SOURCE=2 -march=native -O2 -pipe -fstack-protector-strong -fno-plt   -I/tmp/makepkg/grass/src/grass-7.8.5/dist.x86_64-pc-linux-gnu/include -I/tmp/makepkg/grass/src/grass-7.8.5/dist.x86_64-pc-linux-gnu/include   -I/usr/include -I/usr/include -DPACKAGE=\""grassmods"\"   -I/tmp/makepkg/grass/src/grass-7.8.5/dist.x86_64-pc-linux-gnu/include -I/tmp/makepkg/grass/src/grass-7.8.5/dist.x86_64-pc-linux-gnu/include -DRELDIR=\"vector/v.hull\" -o OBJ.x86_64-pc-linux-gnu/chull.o -c chull.c
In file included from /usr/include/ogr_api.h:47,
                 from /tmp/makepkg/grass/src/grass-7.8.5/dist.x86_64-pc-linux-gnu/include/grass/vect/dig_structs.h:27,
                 from /tmp/makepkg/grass/src/grass-7.8.5/dist.x86_64-pc-linux-gnu/include/grass/vect/digit.h:3,
                 from /tmp/makepkg/grass/src/grass-7.8.5/dist.x86_64-pc-linux-gnu/include/grass/vector.h:4,
                 from chull.c:27:
chull.c:34:19: error: expected ‘;’, identifier or ‘(’ before ‘_Bool’
   34 | { BFALSE, BTRUE } bool;
      |                   ^~~~
make: *** [../../include/Make/Compile.make:32: OBJ.x86_64-pc-linux-gnu/chull.o] Error 1
```

- Operating System: Arch Linux (5.11.6 kernel)
- GRASS GIS 7.8.4
- GDAL 3.3.0

**Additional context**
Downgrading GDAL to 3.2.2 lets GRASS 7.8.4 compile.

nilason:
GDAL 3.3.0 introduces the use of standard C bool type. Redefining bool locally, as in `vector/v.hull/chull.c` causes failure. Since GRASS now adhere to C11 standard this PR addresses this issue by replacing the local bool with the standard C bool.

Fixes #1563

### DIFF
--- a/include/grass/gis.h
+++ b/include/grass/gis.h
@@ -23,6 +23,8 @@
 /* System include files */
 #include <stdio.h>
 #include <stdarg.h>
+#include <stdbool.h>
+
 
 /* Grass and local include files */
 #include <grass/config.h>
@@ -51,13 +53,14 @@ static const char *GRASS_copyright __attribute__ ((unused))
 #define G_gisinit(pgm) G__gisinit(GIS_H_VERSION, (pgm))
 #define G_no_gisinit() G__no_gisinit(GIS_H_VERSION)
 
-/* Define TRUE and FALSE for boolean comparisons */
+/* For boolean values and comparisons use the C99 type 'bool' with values 'true' */
+/* and 'false' For historical reasons 'TRUE' and 'FALSE' are still valid.        */
 #ifndef TRUE
-#define TRUE 1
+#define TRUE true
 #endif
 
 #ifndef FALSE
-#define FALSE 0
+#define FALSE false
 #endif
 
 #if (defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64) || (__APPLE__ && __LP64__)
--- a/lib/lidar/lidar.h
+++ b/lib/lidar/lidar.h
@@ -62,9 +62,6 @@
     /* INTERPOLATOR */
 #define P_BILINEAR 		1
 #define P_BICUBIC 		0
-    /* Boolean definitions */
-#define TRUE 			1
-#define FALSE 			0
 
 /*----------------------------------------------------------------------------------------------------------*/
     /*STRUCTS DECLARATION */
--- a/raster/r.param.scale/param.h
+++ b/raster/r.param.scale/param.h
@@ -18,8 +18,6 @@
 				/* 'blank' edge around raster.          */
 #define MAX_WSIZE 499		/* Maximum dimensions of window.        */
 				/* Some useful labels.                  */
-#define TRUE 1
-#define FALSE 0
 
 #define RAD2DEG M_R2D
 #define DEG2RAD M_D2R
--- a/raster/r.surf.idw/main.h
+++ b/raster/r.surf.idw/main.h
@@ -1,8 +1,6 @@
 #include <grass/raster.h>
 
 #define         SHORT           short
-#define         TRUE    1
-#define         FALSE   0
 
 #define MELEMENT        struct Melement
 MELEMENT {
--- a/raster3d/r3.showdspf/Ball.c
+++ b/raster3d/r3.showdspf/Ball.c
@@ -7,8 +7,6 @@
 #include "Ball.h"
 #include "BallMath.h"
 #include <stdio.h>
-#define TRUE 1
-#define FALSE 0
 
 HMatrix mId = { {1, 0, 0, 0}
 , {0, 1, 0, 0}
--- a/raster3d/r3.showdspf/togif.c
+++ b/raster3d/r3.showdspf/togif.c
@@ -324,9 +324,6 @@ static int ditherrow(unsigned short *r, unsigned short *g, unsigned short *b,
  *
  *****************************************************************************/
 
-#define TRUE 1
-#define FALSE 0
-
 
 /************************** BumpPixel() ********************************/
 /*
--- a/vector/v.hull/chull.c
+++ b/vector/v.hull/chull.c
@@ -22,16 +22,13 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <math.h>
+#include <stdbool.h>
 
 #include <grass/gis.h>
 #include <grass/vector.h>
 #include <grass/glocale.h>
 
 #include "globals.h"
-
-/*Define Boolean type */
-typedef enum
-{ BFALSE, BTRUE } bool;
 
 /* Define vertex indices. */
 #define X   0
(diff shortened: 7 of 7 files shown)


nilason:
If there are no objections, I will soon merge and backport this. This is needed for successful compilation with GDAL 3.3.

metzm:
A much easier solution could be to use `TRUE|FALSE` from https://github.com/OSGeo/grass/blob/master/include/grass/gis.h#L54, no need to `#include <stdbool.h>` or to define boolean type in `v.hull`. Many other GRASS modules use GRASS defined `TRUE|FALSE` without problems.

nilason:
> A much easier solution could be to use `TRUE|FALSE` from https://github.com/OSGeo/grass/blob/master/include/grass/gis.h#L54, no need to `#include <stdbool.h>` or to define boolean type in `v.hull`. Many other GRASS modules use GRASS defined `TRUE|FALSE` without problems.

I did consider using GRASS’s TRUE/FALSE https://github.com/OSGeo/grass/issues/1563#issuecomment-832251652, but in the end sided with bool, which gives incomparably clearer code. I see no reason not to use the standard bool, in fact I’d welcome its use more often in GRASS in general. Perhaps by include stdbool.h in gis.h.

metzm:
> I did consider using GRASS’s TRUE/FALSE [#1563 (comment)](https://github.com/OSGeo/grass/issues/1563#issuecomment-832251652), but in the end sided with bool, which gives incomparably clearer code. I see no reason not to use the standard bool, in fact I’d welcome its use more often in GRASS in general. Perhaps by include stdbool.h in gis.h.

Using GRASS's TRUE/FALSE, you don't need to `#include <stdbool.h>` and only replace true with TRUE and false with FALSE in this PR. Looks like pretty clear code in `v.hull`. What exactly is unclear in your opinion?

If you really want to `#include <stdbool.h>`, the correct place would be in `gis.h`. What would be the benefits?

nilason:
> > I did consider using GRASS’s TRUE/FALSE [#1563 (comment)](https://github.com/OSGeo/grass/issues/1563#issuecomment-832251652), but in the end sided with bool, which gives incomparably clearer code. I see no reason not to use the standard bool, in fact I’d welcome its use more often in GRASS in general. Perhaps by include stdbool.h in gis.h.
> 
> Using GRASS's TRUE/FALSE, you don't need to `#include <stdbool.h>` and only replace true with TRUE and false with FALSE in this PR. Looks like pretty clear code in `v.hull`. What exactly is unclear in your opinion?
> 
> If you really want to `#include <stdbool.h>`, the correct place would be in `gis.h`. What would be the benefits?

Starting with GDAL 3.3 `stdbool.h` **is** included (in ogr_api.h), which causes the issue at hand in the first place. Therefore all `bool` types declared in `chull.c` must also be redeclared as `int`'s if we go by your suggestion. Precisely this is less clear code in my opinion. (I agree v.hull is has clear code and this PR is suggesting to keep it that way, dealing with the conflict with GDAL.) Compare e.g.:

```c
bool AddOne(tVertex p);

struct tVertexStructure {
  double v[3];
  int vnum;
  tEdge duplicate;  /* pointer to incident cone edge (or NULL) */
  bool onhull;      /* T iff point on hull. */
  bool mark;        /* T iff point already processed. */
  tVertex next, prev;
};

# vs

int AddOne(tVertex p);

struct tVertexStructure {
  double v[3];
  int vnum;
  tEdge duplicate;  /* pointer to incident cone edge (or NULL) */
  int onhull;       /* T iff point on hull. */
  int mark;         /* T iff point already processed. */
  tVertex next, prev;
};
```

In my mind it's quite obvious that using `bool` over `int` is making the intent of the code much clearer. `Int` is now used in so many various ways: as ordinary integer value, return codes, bit flags and also boolean value. Often only the documentation can describe an Int's intentional use. Using bool makes these cases clear reading the code alone.

I did some testing with putting this in `gis.h`:
```c
....

/* System include files */
#include <stdio.h>
#include <stdarg.h>
#include <stdbool.h>

...

/* Define TRUE and FALSE for boolean comparisons */
#ifndef TRUE
#define TRUE true
#endif

#ifndef FALSE
#define FALSE false
#endif

...
```

then there is no need to make the include in `chull.c`, and this and all other code works without any further modification. (Note: the change to `#define TRUE true` is not really necessary as true is defined as `1` anyway and similarly for FALSE.) But with the advantage of enabling possible future modifications/additions like:
`int G_is_gisbase(const char *path)`
to 
`bool G_is_gisbase(const char *path)`
or
`static void log_error(const char *msg, int fatal)`
to
`static void log_error(const char *msg, bool fatal)`

I agree including stdbool.h in gis.h would be the better way to go. However, I can't see *any* benefits of using ints over bools for boolean values.

(see [stdbool.h specs](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdbool.h.html) for reference)

nilason:
I have now updated this PR, to include `stdbool.h` in `gis.h` for global availability of C99's `bool` type and values `true` and `false`. The old definitions of `TRUE` and `FALSE` remain and are fully working as before; I changed their definitions to `true` and `false` (which are the same as the `1` and `0` before), to make their equivalency unambiguous.

metzm:
> I have now updated this PR, to include `stdbool.h` in `gis.h` for global availability of C99's `bool` type and values `true` and `false`. The old definitions of `TRUE` and `FALSE` remain and are fully working as before; I changed their definitions to `true` and `false` (which are the same as the `1` and `0` before), to make their equivalency unambiguous.

Thanks, fine with me. We could now clean up code and use `bool` types where appropriate (there are many places where int or char are used as booleans).

nilason:
> Thanks, fine with me. We could now clean up code and use bool types where appropriate (there are many places where int or char are used as booleans).

@metzm Thank you, as always, for your invaluable input!

Merged and backported.

Issue #1420, opened 2021-03-03
[Bug] r.in.poly fails trying to process stdin


infrastation:
**Describe the bug**
I use `r.in.poly` quite a bit and it works well, less one minor inconvenience. Specifically, it takes to save the ASCII data to a temporary file before the plugin can take it as input. This is because internally it processes the input in several passes and calls `G_fseek()` before each pass. Which is an error for the standard input file descriptor.

**To Reproduce**
```text
GRASS UK_ALOS_reprojected/PERMANENT:~ > r.in.poly --help
Creates raster maps from ASCII polygon/line/point data files.

Usage:
 r.in.poly input=name output=name [title=phrase] [type=string]
   [null=value] [rows=value] [--overwrite] [--help] [--verbose]
   [--quiet] [--ui]

Parameters:
   input   Name of input file; or "-" to read from stdin
  output   Name for output raster map
   title   Title for resultant raster map
    type   Type of raster map to be created
           options: CELL,FCELL,DCELL
           default: CELL
    null   Integer representing NULL value data cell
    rows   Number of rows to hold in memory
           default: 4096
GRASS UK_ALOS_reprojected/PERMANENT:~ > r.in.poly input=- output=test
ERROR: Unable to seek: Illegal seek
GRASS UK_ALOS_reprojected/PERMANENT:~ > 
```

**Expected behavior**
Either the notions of stdin should be removed from the code and the parser props (trivial), or the internal logic should be changed to read the input file exactly once without rewinding (may be complicated).

**Additional context**
From the git history it looks like this bug is not new. It reproduces with GRASS 7.4.0 and with the current master branch.

PR #1569, opened 2021-05-09
r.in.poly: do not try to use stdin as input

infrastation:


neteler:
fixes #1420

metzm:
A better solution would be to write stdin input to a temp file, then open this temp file and proceed as before. See `v.in.ascii` as an example.

infrastation:
Thank you for making this point. When a program input explicitly presents itself as a pipe (not a file), using a temporary file to emulate the pipe goes against the principle of least surprise. This exact "workaround" in other programs in the past had unexpectedly caused me magnetic disk bottlenecks or excess SSD wear or running out of tmpfs space and breaking things. The issue becomes more apparent when the amount of data starts to be in gigabytes. Although this plugin usually processes small files, perhaps it would be better to keep as many plugins on the safe side as possible. Does it make sense?

infrastation:
Rebased on the latest master branch.

infrastation:
Rebased on the latest master branch.

PR #1580, opened 2021-05-16
wxGUI/gui_core: fix trigger 'wx.EVT_CHOICE' event if you set widget selection programatically 

tmszi:
**To Reproduce**

1. Launch Graphical Modeler
2. From the Graphical Modeler toolbar launch 'Add command (GRASS module) to model' tool dialog
3. Try to type 'db.' (without quotation marks)
4. Choose first option from the Choice widget options 'db.columns'
5. TextCtrl widgets 'Command' and 'Label' value is not updated automatically after you select module from Choice widget

Another example:
3. Try type exactly 'db.tables' (without quotation marks)

gmodeler_add_gg_module_dialog_choice_widget

**Expected behavior:**

gmodeler_add_gg_module_dialog_choice_widget_exp

### DIFF
--- a/gui/wxpython/gmodeler/dialogs.py
+++ b/gui/wxpython/gmodeler/dialogs.py
@@ -214,9 +214,9 @@ class ModelSearchDialog(wx.Dialog):
             parent=self.panel, model=menuModel.GetModel(), showTip=True
         )
         self.search.moduleSelected.connect(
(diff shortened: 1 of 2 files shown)


petrasovaa (approved):
Testes, seems to work. I think ideally we would replace this whole search widget with something similar to the Modules tab in lmgr.

PR #1596, opened 2021-05-23
wxGUI/datacatalog: fix setting output vector/raster format

tmszi:
**To Reproduce**

1. Launch wxGUI
2.  On the Data tab (page) choose from toolbar 'Select another import option' tool and invoke menu and choose 'Link external data' -> 'Set raster output format [r.external.out]'
3. Choose Directory as Output type
4. Choose e.g. GeoTIFF format 
5. Set Directory path e.g.  /tmp
6.  Hit Ok button
7. You get error message

**Error message**

wxgui_datacatalog_set_output_raster_format_error_message

**Additional info**
1. 'Set raster output format [r.external.out]' dialog internally use `v.external.out` command instead of `r.external.out` and params args are incorrect
2. Mapping output formats between dialog wx.Choice widget items (which use long format name) and command `r.external.out` or `v.external.out` (which use format abbreaviation) is incorrect . Simple e.g. 'Set raster output format [r.external.out]'  dialog wx.Choice format type widget use for GeoTIFF format string 'GeoTIFF', but internall command `r.external.out` as format param argument  accept for GeoTIFF format string arg 'GTiff' (check all format options `r.external.out`), the same is true for 'Set vector output format [v.external.out]' dialog.

### DIFF
--- a/gui/wxpython/core/utils.py
+++ b/gui/wxpython/core/utils.py
@@ -19,6 +19,7 @@ import glob
 import shlex
 import re
 import inspect
+import operator
 import six
 
 from grass.script import core as grass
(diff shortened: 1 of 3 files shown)


petrasovaa (approved):
Thanks, looks good to me, I recommend giving it little bit of time before backport, just to make sure it doesn't impact things outside of r/v.external.out.

PR #1597, opened 2021-05-23
GRASS GIS 8.0 changes

neteler:
This PR ~~is meant for testing (do not merge into master yet!) as it~~ contains needed changes for a GRASS GIS 8.0 ~~.0alpha release~~.
~~It may be used for local testing by applying this PR to the latest GRASS GIS-core in the master branch.~~

_Update 28 May 2021:_
This PR may become the basis for the changes in the master branch to GRASS GIS 8.0. 
A  8.0.0alpha release branch may be forked after merge of this PR.

Please verify thoroughly.

(note that some CI tests will fail as they expect `grass79` and not `grass80` as the name of startup script; in addition, there is no `grass8` branch yet in addons)

To locally apply and try it:

1. checkout GRASS GIS master from GitHub, `cd` into it
2. if you already have it, `make distclean`
2. `wget https://github.com/OSGeo/grass/pull/1597.diff`
3. `patch -p1 < 1597.diff`
4. `configure ... ; make`
5. start it from `bin.x86_64-pc-linux-gnu/grass80` (the name of architecture in the path may differ)


NEWS
```diff
@@ -1,7 +1,9 @@
 NEWS
 
-GRASS GIS 7
+GRASS GIS 8
+ o https://trac.osgeo.org/grass/wiki/Grass7/NewFeatures80
```
hellik, line 4: Grass7 => Grass8?

REQUIREMENTS.html
```diff
@@ -55,7 +55,7 @@ <h3>General requirements:</h3>
 <a href="https://gdal.org">https://gdal.org</a>
 </li>
 
-<li><b>Python &gt;= 2.7 (recommended &gt;= 3.5)</b> (for temporal framework, scripts, wxGUI, and ctypes interface)<br>
+<li><b>Python &gt;= 3.5</b> (for temporal framework, scripts, wxGUI, and ctypes interface)<br>
```
hellik, line 58: what is the minimum Python 3.x? really 3.5?

db/db.login/db.login.html
```diff
@@ -15,8 +15,8 @@ <h2>NOTE</h2>
 file in the user account, specifically
 
 <ul>
-<li> in the 'home' directory, i.e. <tt>$HOME/.grass7/dblogin</tt> (Unix-like systems)</li>
-<li> <tt>%APPDATA%/GRASS7/dblogin</tt> (MS-Windows)</li>
+<li> in the 'home' directory, i.e. <tt>$HOME/.grass8/dblogin</tt> (Unix-like systems)</li>
+<li> <tt>%APPDATA%\GRASS8\dblogin</tt> (MS-Windows)</li>
```
hellik, line 19: wouldn't it be

`C:\Users\youruser\AppData\Roaming\GRASS7`

?

hellik: in this case `%APPDATA%\Roaming\GRASS8`

doc/grass_database.html
```diff
@@ -222,18 +222,18 @@ <h3>Starting GRASS GIS using command line</h3>
 in a given Mapset with only command line interface:
 
 <div class="code"><pre>
-grass79 --text ~/grassdata/mylocation/mymapset
+grass80 --text ~/grassdata/mylocation/mymapset
```
hellik, line 225: that's starting a text session in *nix. should we add an example how to do it in windows?

NEWS
neteler: Good catch, fixed in 8078cc676df18473fc3aceee6b9f7a78822b46e9

REQUIREMENTS.html
neteler: That's still to be defined... 

See RFC 8: Language Standards Support: Python (Draft), https://trac.osgeo.org/grass/wiki/RFC/8_LanguageStandardsSupport

db/db.login/db.login.html
neteler: ok, fixed in 9ee56ed3c

doc/grass_database.html
neteler: Yes, please donate an example (here is fine)

python/grass/script/setup.py
```diff
@@ -24,20 +24,20 @@
...
     # needs to be edited by the user
-    grass7bin = 'grass79'
+    grass8bin = 'grass80'
     if sys.platform.startswith('win'):
         # MS Windows
-        grass7bin = r'C:\OSGeo4W\bin\grass79.bat'
+        grass8bin = r'C:\OSGeo4W\bin\grass80.bat'
         # uncomment when using standalone WinGRASS installer
-        # grass7bin = r'C:\Program Files (x86)\GRASS GIS 7.9.0\grass79.bat'
+        # grass8bin = r'C:\Program Files (x86)\GRASS GIS 8.0.0\grass80.bat'
```
hellik, line 32: do we provide still a 32 bit standalone installer C:\Program Files (x86)? @landam

hellik (changes requested):
left mostly windows related comments

doc/grass_database.html
hellik: grass80 --text D:\grassdata\nc_spm_08_grass7\user1

doc/grass_database.html
```diff
@@ -222,18 +222,18 @@ <h3>Starting GRASS GIS using command line</h3>
...
 </pre></div>
 
 And the following will create the given Location with projection given
 by the EPSG code and it will start the default interface
 (GUI or command line):
 
 <div class="code"><pre>
-grass79 -c EPSG:5514:3 ~/grassdata/mylocation
+grass80 -c EPSG:5514:3 ~/grassdata/mylocation
```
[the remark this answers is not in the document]
hellik: grass80 -c EPSG:31287 D:\grassdata\mylocation_epsg31287

gui/wxpython/docs/wxGUI.html
```diff
@@ -655,7 +655,7 @@ <h3>Starting the GUI from command line</h3>
 </pre></div>
 
 Alternatively, it may be defined in the main configuration file
-(<tt>$HOME/.grass7/rc</tt> on GNU/Linux and macOS, <tt>$APPDATA\GRASS7\rc</tt>
+(<tt>$HOME/.grass8/rc</tt> on GNU/Linux and macOS, <tt>$APPDATA\GRASS8\rc</tt>
```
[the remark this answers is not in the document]
neteler: fixed in 7704a7a8f

lib/init/variables.html
```diff
@@ -97,8 +97,8 @@ <h2>List of selected (GRASS related) shell environment variables</h2>
     for local GRASS modules (normally installed as GRASS Addons
     by <tt>g.extension</tt> module) extra to standard
     distribution. The default on GNU/Linux
-    is <tt>$HOME/.grass7/addons</tt>, on MS
-    Windows <tt>$APPDATA\GRASS7\addons</tt>.</dd>
+    is <tt>$HOME/.grass8/addons</tt>, on MS
+    Windows <tt>$APPDATA\GRASS8\addons</tt>.</dd>
```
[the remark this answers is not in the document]
neteler: fixed in 7704a7a8f

doc/grass_database.html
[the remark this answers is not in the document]
neteler: done (modified) in d83447807

doc/grass_database.html
neteler: done in d83447807

gui/wxpython/docs/wxGUI.toolboxes.html
```diff
@@ -18,7 +18,7 @@ <h2>DESCRIPTION</h2>
 Toolboxes are configured through two XML files (<tt>main_menu.xml</tt> and
 <tt>toolboxes.xml</tt>) located in your user home
 GRASS directory, subdirectory <tt>toolboxes</tt>
- (<tt>$HOME/.grass7/toolboxes/</tt> on UNIX).
+ (<tt>$HOME/.grass8/toolboxes/</tt> on UNIX).
```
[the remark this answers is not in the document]
neteler: yes, pls donate an example

wenzeslaus:
> To locally apply and try it:
>  ...
>     3. `wget https://github.com/OSGeo/grass/pull/1597.diff`
>     4. `patch -p1 < 1597.diff`

Or `gh pr checkout 1597` following the _Open with_ button right from the PR title. (Or manually, add the neteler fork repo as another remote to your local repo and get the grass_gis_800_alpha from there.)

doc/grass_database.html
```diff
@@ -222,18 +222,26 @@ <h3>Starting GRASS GIS using command line</h3>
 in a given Mapset with only command line interface:
 
 <div class="code"><pre>
-grass79 --text ~/grassdata/mylocation/mymapset
+# Linux, Mac, *BSD, ...:
+grass80 --text ~/grassdata/nc_spm_08_grass7/user1
+
+# Windows
+grass80 --text D:\grassdata\nc_spm_08_grass7\user1
 </pre></div>
```
wenzeslaus, line 230: Do you think that someone who has an issue figuring out the need for a Windows path from the unix example (as before) will parse through the `#` comments in the example? Either delete the Windows example or make it a separate paragraph.

INSTALL
```diff
@@ -151,16 +151,16 @@ See also CODE OPTIMIZATION below.
 
 After compilation, the resulting code is stored in the directory
     ./dist.$ARCH
-and the scripts (grass79, ...) in
+and the scripts (grass80, ...) in
```
wenzeslaus, line 154: What about dropping the 80 part at least from the documentation? On a reasonable system, it should install as `grass` anyway. I also noticed I usually write `ls` and not `ls830`, `gdalinfo`, not `gdalinfo32`... Seems like less work for the releases, too.

INSTALL
```diff
@@ -189,7 +189,7 @@ GitHub, you have to perform a few steps. In general:
 
 In detail:
 
-    cd /where/your/grass7sourcecode/lives/
+    cd /where/your/grass8sourcecode/lives/
```
wenzeslaus, line 192: This should definitively say `grass-source-code` without the number.

REQUIREMENTS.html
wenzeslaus: Since, as the RFC says, _the oldest still-alive Python version is 3.6, to be retired in December 2021_, I don't see any reason to have anything lower than 3.6 here.

db/db.login/db.login.html
```diff
@@ -15,8 +15,8 @@ <h2>NOTE</h2>
 file in the user account, specifically
 
 <ul>
-<li> in the 'home' directory, i.e. <tt>$HOME/.grass7/dblogin</tt> (Unix-like systems)</li>
-<li> <tt>%APPDATA%/GRASS7/dblogin</tt> (MS-Windows)</li>
+<li> in the 'home' directory, i.e. <tt>$HOME/.grass8/dblogin</tt> (Unix-like systems)</li>
+<li> <tt>%APPDATA%\Roaming\GRASS8\dblogin</tt> (MS-Windows)</li>
```
wenzeslaus, line 19: Doesn't `/` just work on Windows [these days](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.altdirectoryseparatorchar?view=net-5.0), @hellik?

wenzeslaus:
> This PR is meant for testing (do not merge into master!) as it contains needed changes for a GRASS GIS 8.0.0alpha release.

I don't understand where this will go. If it will be released, it should go to the main repo. A branch would make sense to me.

wenzeslaus:
> (note that some CI tests will fail as they expect `grass79` and not `grass80` as the name of startup script)

Can we install just `grass` to avoid this?

doc/grass_database.html
veroandreo: What do you mean? In my experience, students want to see it written, they do not want to figure it out themselves...

lib/init/variables.html
```diff
@@ -42,18 +42,18 @@ <h2>Setting shell environment variables</h2>
...
-<tt>$HOME/.grass7/cshrc</tt></li>
+<tt>$HOME/.grass8/cshrc</tt></li>
 </ul>
 
 
 <h2>Setting GRASS gisenv variables</h2>
 
 Use <em><a href="g.gisenv.html">g.gisenv</a></em> within GRASS. This permanently
-predefines GRASS variables in the <tt>$HOME/.grass7/rc</tt> file after the
+predefines GRASS variables in the <tt>$HOME/.grass8/rc</tt> file after the
```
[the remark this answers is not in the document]
veroandreo: would you please provide some, @hellik ?

python/grass/script/core.py
```diff
@@ -347,7 +347,7 @@ def make_command(
                         "To run the module <%s> add underscore at the end"
                         " of the option <%s> to avoid conflict with Python"
                         " keywords. Underscore at the beginning is"
-                        " depreciated in GRASS GIS 7.0 and will be removed"
+                        " depreciated in GRASS GIS 7.0 and has been removed"
```
veroandreo, line 350:
```suggestion
                        " deprecated in GRASS GIS 7.0 and has been removed"
```

db/db.login/db.login.html
HuidaeCho: / works on Windows (https://answers.microsoft.com/en-us/windows/forum/windows_7-files/when-did-windows-start-accepting-forward-slash-as/aa033741-0fc6-43e3-ab6e-165020794126), but I don't think it's a good idea to use / in general (https://bytes.com/topic/python/answers/23123-when-did-windows-start-accepting-forward-slash-path-separator), not in this specific case.
```dos
cd /users # works fine
mkdir \dev
cd /dev # doesn't work because /d is an option
```

INSTALL
neteler: Generally fine for me but then we may also need to actually generate a `grass` script at compile time.

INSTALL
neteler: fixed in 5bd452e4e

REQUIREMENTS.html
neteler: Updated to Python 3.6 in dcfb6c548

python/grass/script/core.py
neteler: fixed here and elsewhere in b7c892155

INSTALL
neteler: Yes, fine for me.

> > This PR is meant for testing (do not merge into master!) as it contains needed changes for a GRASS GIS 8.0.0alpha release.
> 
> I don't understand where this will go.

As I understand there are still different opinions (https://lists.osgeo.org/pipermail/grass-dev/2021-May/095136.html and related) on when to create a branch.

>  If it will be released, it should go to the main repo. A branch would make sense to me.

We could asap, as I don't see much point in keeping "7.9" forever:
- merge this PR into the main branch (changing VERSION from "8.0.0alpha" to "8.0.dev")
- then create the "releasebranch_8_0" branch off the main branch (changing VERSION therein to "8.0.0alpha")
- cherry-pick future relevant changes from main branch to "releasebranch_8_0" as we have done with other release branches

neteler:
> > (note that some CI tests will fail as they expect `grass79` and not `grass80` as the name of startup script)
> 
> Can we install just `grass` to avoid this?

Would be practical, however, I routinely use the stable and the dev versions in parallel. But that could be solved by a link...

doc/grass_database.html
wenzeslaus: My question is how large is the overlap between the groups: Group 1 can replace unix path by Windows path easily. Group 2 is not distracted or confused by the `#` comments. I'm guessing pretty small. In other words, if the reader is familiar with unix/Bash syntax, the Windows piece is not needed. And if a unix path is a stumbling block, so is a Bash comment.

wenzeslaus:
> > > (note that some CI tests will fail as they expect `grass79` and not `grass80` as the name of startup script)
> > 
> > 
> > Can we install just `grass` to avoid this?
> 
> Would be practical, however, I routinely use the stable and the dev versions in parallel. But that could be solved by a link...

Let's not make it hard everywhere and for the users because developers need to have multiple versions.

wenzeslaus:
> We could asap, as I don't see much point in keeping "7.9" forever:
>     * merge this PR into the main branch (changing VERSION from "8.0.0alpha" to "8.0.dev")
>     * then create the "releasebranch_8_0" branch off the main branch (changing VERSION therein to "8.0.0alpha")
>     * cherry-pick future relevant changes from main branch to "releasebranch_8_0" as we have done with other release branches

Or you could do the 8.0dev part, but branch out only to create the alpha and then abandon that branch and create the 8.0 release branch later avoiding any backports between the alpha and the more mature release branching.

In any case, the "odd versions are dev and never released" system seems to break here, but after creating 8.0 release branch, master branch would need to be 8.1.

neteler:
> Let's not make it hard everywhere and for the users because developers need to have multiple versions.

Ok, what do you suggest? I didn't get it.

wenzeslaus:
> > Let's not make it hard everywhere and for the users because developers need to have multiple versions.
> 
> Ok, what do you suggest? I didn't get it.

That installing `grass` is the expected behavior and it should be the default.

doc/grass_database.html
hellik: in my experience, many manual examples are too *nix biased and exclude _average_ windows user by not giving literal working windows examples. I agree here with @veroandreo.

neteler:
> > > Let's not make it hard everywhere and for the users because developers need to have multiple versions.
> That installing `grass` is the expected behavior and it should be the default.

Ok, I have changed it (82e8429e8a6c20a88275720447147b6c66a74dfe), it naturally implied other places to be changed.

lib/init/variables.html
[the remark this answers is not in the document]
hellik: it's again in  `%APPDATA%\Roaming\GRASS8\` :

`%APPDATA%\Roaming\GRASS8\rc`

neteler:
> > (note that some CI tests will fail as they expect `grass79` and not `grass80` as the name of startup script)
> 
> Can we install just `grass` to avoid this?

Done (82e8429) - now it would be great to have the CI accept `grass` as well...

lib/init/variables.html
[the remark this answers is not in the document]
neteler: added in efe90a9f7

db/db.login/db.login.html
hellik: > cd /data  # doesn't work because /d is an option

AFAIK it's 

`cd data`

`cd \data` changes here to an existing directory in the toplevel d: , altough I've started e.g. in D:\temp\tm18> where the data subfolder exists.

```
D:\temp\tm18>cd /data
Das System kann den angegebenen Pfad nicht finden.
```

hellik: > 
> 
> Doesn't `/` just work on Windows [these days](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.altdirectoryseparatorchar?view=net-5.0), @hellik?

the interesting thing is - and almost nobody knows it ;-)  - it worked since the beginning of MS windows

```
D:\>cd D:\temp\tm18\data
D:\temp\tm18\data>
```

```
D:\>cd D:/temp/tm18/data
D:\temp\tm18\data>
```

though the MS windows standard, everywhere promoted, is:

`D:\temp\tm18\data`

and you can't be sure that it works in all circumstances of myriads of different varieties of batch files, python scripts, R scripts in windows .....

INSTALL
```diff
@@ -212,9 +212,9 @@ the main source code directory or locally with
     "INST_NOW=y make"
 
 You may want to define an alias for this:
-    alias gmake7='INST_NOW=y make'
+    alias gmake8='INST_NOW=y make'
 Then simply compile/install the current module with
-    gmake7
+    gmake8
```
landam, line 217: What about simply `gmake`?

neteler: Deal: 98388f565

db/db.login/db.login.html
wenzeslaus: I wonder if there are any circumstances where it is still an issue except for that `/d` issue. Python scripts should never have explicit slash or backslash in them. The best practices for R I have seen argued for forward slashes on all platforms (saying to make them work on macOS and Linux, so they actually seemed written from Windows point of view).

wenzeslaus (commented):
The `grass79` -> `grass` seems like the best part of this PR!

neteler:
> > Done ([82e8429](https://github.com/OSGeo/grass/commit/82e8429e8a6c20a88275720447147b6c66a74dfe)) - now it would be great to have the CI accept `grass` as well...
> 
> You can just change it together with the rest of the files. Your find and replace should work.

Ok, done in d885e67d2

neteler:
> > You can just change it together with the rest of the files. Your find and replace should work.
> 
> Ok, done in [d885e67](https://github.com/OSGeo/grass/commit/d885e67d2206cb9727d8897ba44990b5400c487b)

Something is still missing concerning the CI (I wonder where...):

Ubuntu log:

```
mkdir -p -m 755 /home/runner/install//grass80
mkdir -p -m 755 /home/runner/install//bin
tar cBCf /home/runner/work/grass/grass/dist.x86_64-pc-linux-gnu - . | tar xBCf /home/runner/install//grass80 - 2>/dev/null
rm /home/runner/install//grass80/grass80.tmp
rm: cannot remove '/home/runner/install//grass80/grass80.tmp': No such file or directory
```

wenzeslaus:
It looks like a problem with the `make install` step. Seem to be coming from `include/Make/Install.make:101: -rm $(DESTDIR)$(INST_DIR)/$(GRASS_NAME).tmp`. Does it work locally? (Can't investigate the GRASS_NAME piece now.)

doc/grass_database.html
wenzeslaus: > ...many manual examples are too *nix biased...

I think what would help is saying that it is a Bash or whatever example so that it would be clear that commands or syntax may be different (on Windows) and that it is not "standard GRASS" but rather "standard Bash/unix/...".

Otherwise, I think we need Python examples in manual pages more than Windows examples anyway.

> ...exclude average windows user...

Similarly to the "dropping terminal" PR, I continue being puzzled about who are these _average Windows users_ attached to their command lines. Isn't that a stereotype about Linux users, not Windows ones? :-)

neteler:
> It looks like a problem with the `make install` step. 

Found it: fcb524dfc

INSTALL
ninsbl: > Generally fine for me but then we may also need to actually generate a `grass` script at compile time.

A version independend startup script is nice if this does not interfere with installing different GRASS versions in parallel.
So we might need to keep `grass78` and `grass80` anyway? Just to make sure different version in parallel are considered...

wenzeslaus:
It seems like you can re-conceptualize this PR to be a 7-to-8 transition, merge it to master branch, and do the alpha release changes elsewhere. Thoughts?

neteler:
Getting pretty close to "CI passing", yet this one I cannot identify in the code/workflow:

```
Run .github/workflows/test_simple.bat 'C:\OSGeo4W64\opt\grass\grass.bat'

D:\a\grass\grass>set grass=C:\OSGeo4W64\opt\grass\grass.bat 

D:\a\grass\grass>call C:\OSGeo4W64\opt\grass\grass.bat --tmp-location EPSG:4326 --exec g.region res=0.1 -p 
(null): can't open file 'C:\OSGEO4~1\opt\grass\etc\grass80.py': [Errno 2] No such file or directory
Press any key to continue . . . 
```

nilason:
> It seems like you can re-conceptualize this PR to be a 7-to-8 transition, merge it to master branch, and do the alpha release changes elsewhere. Thoughts?

A lot of the changes, certainly the ones related to `grass79->grass` and `grass8` in docs need to be merged into master before any branching. I would like to suggest to merge this PR (when approved) and give it some days/a week for testing the grassXX->grass (👍 BTW) changes in particular. Then tag a 8.0.0.alpha1 on master for release. It's not unlikely there will pop up issues quite immediately after testing in a wider group, these issues can hopefully quickly be addressed and applied directly to master. When these immediate issues are ironed out (I'm thinking of a time frame of about 2-4 weeks) a 8.0 branch may be made, perhaps starting with a beta version or - even better if feasable - a release candidate.

nilason:
> Getting pretty close to "CI passing", yet this one I cannot identify in the code/workflow:
> 
> ```
> Run .github/workflows/test_simple.bat 'C:\OSGeo4W64\opt\grass\grass.bat'
> 
> D:\a\grass\grass>set grass=C:\OSGeo4W64\opt\grass\grass.bat 
> 
> D:\a\grass\grass>call C:\OSGeo4W64\opt\grass\grass.bat --tmp-location EPSG:4326 --exec g.region res=0.1 -p 
> (null): can't open file 'C:\OSGEO4~1\opt\grass\etc\grass80.py': [Errno 2] No such file or directory
> Press any key to continue . . . 
> Error: Process completed with exit code 1.
> ```

Could it be:
https://github.com/OSGeo/grass/blob/2048e101650b587d641dbe0d7ec561f84ec1194b/mswindows/osgeo4w/grass.bat.tmpl#L18

neteler:
> > D:\a\grass\grass>call C:\OSGeo4W64\opt\grass\grass.bat --tmp-location EPSG:4326 --exec g.region res=0.1 -p 
> > (null): can't open file 'C:\OSGEO4~1\opt\grass\etc\grass80.py': [Errno 2] No such file or directory
> Could it be:
> 
> https://github.com/OSGeo/grass/blob/2048e101650b587d641dbe0d7ec561f84ec1194b/mswindows/osgeo4w/grass.bat.tmpl#L18

Ah, thanks. Changed in ec8e141a0

nilason:
> > > D:\a\grass\grass>call C:\OSGeo4W64\opt\grass\grass.bat --tmp-location EPSG:4326 --exec g.region res=0.1 -p
> > > (null): can't open file 'C:\OSGEO4~1\opt\grass\etc\grass80.py': [Errno 2] No such file or directory
> > > Could it be:
> > 
> > 
> > https://github.com/OSGeo/grass/blob/2048e101650b587d641dbe0d7ec561f84ec1194b/mswindows/osgeo4w/grass.bat.tmpl#L18
> 
> Ah, thanks. Changed in [ec8e141](https://github.com/OSGeo/grass/commit/ec8e141a04d6130f99a64fc4582a58fcfaef9f6e)

I did notice `POSTFIX` is used in several places and files in similar way. Remain to be see CIs turnout, but may be tested with `export POSTFIX=""`, or evaluate case by case.

nilason:
I see it fails again:
```
D:\a\grass\grass>call C:\OSGeo4W64\opt\grass\grass.bat --tmp-location EPSG:4326 --exec g.region res=0.1 -p 
Starting GRASS GIS...
  File "C:\OSGEO4~1\opt\grass\etc\grass.py", line 2721, in <module>
    main()
  File "C:\OSGEO4~1\opt\grass\etc\grass.py", line 2595, in main
    tmpdir=tmpdir,
  File "C:\OSGEO4~1\opt\grass\etc\grass.py", line 1035, in set_mapset
    from grass.grassdb.checks import (
ModuleNotFoundError: No module named 'grass.grassdb'; 'grass' is not a package
Press any key to continue . . . 
```
my guess it's related to `POSTFIX` (looking for the module at a path including non-existing `grass80`) as I noted above. Maybe you could try set it to an empty string.

neteler:
> I see it fails again:
...
> my guess it's related to `POSTFIX` (looking for the module at a path including non-existing `grass80`) as I noted above. Maybe you could try set it to an empty string.

Ok, I try that now: 1fe7743ab

mswindows/osgeo4w/package.sh
```diff
@@ -111,7 +111,8 @@ read MINOR <&3
 read PATCH <&3
 
 export VERSION=${MAJOR}.${MINOR}.${PATCH}
-export POSTFIX=${MAJOR}${MINOR}
+#export POSTFIX=${MAJOR}${MINOR}
```
hellik, line 114: we have this logic with ${MAJOR}${MINOR} that we're able to build/compile e.g. 7.6, 7.8, 7.9, 8.0 etc in the same OSGeo4W build environment side by side; IIRC also used later for compiling the winGRASS addons for different versions.

so if we decide not to use ${MAJOR}${MINOR}, then maybe OSGeo4W build environment has to be re-organized.

@landam what are your suggestions?

wenzeslaus:
>  `ModuleNotFoundError: No module named 'grass.grassdb'; 'grass' is not a package`

My guess: The directory with grass.py is on Python (import) path (`sys.path`). Likely because it is the current directory. Python picks grass.py instead of the grass package. Startup ends in flames.

neteler:
> It seems like you can re-conceptualize this PR to be a 7-to-8 transition, merge it to master branch, and do the alpha release changes elsewhere. Thoughts?

Meanwhile I also believe that this PR is the preparation of the master branch for G8. So.
- a 8.0.0alpha branch shall just be a "throw away/ignore later" branch, created after merging of this PR
- the real release_branch_8_0 can then be created at a later stage, after gaining some experience with the "throw away" 8.0.0alpha branch.

neteler:
> > `ModuleNotFoundError: No module named 'grass.grassdb'; 'grass' is not a package`
> 
> My guess: The directory with grass.py is on Python (import) path (`sys.path`). _Likely because it is the current directory._ Python picks grass.py instead of the grass package. Startup ends in flames.

I see. Shall I revert the parts concerning Windows at this point? I happily accept suggestions how to deal with this.

neteler:
> > `ModuleNotFoundError: No module named 'grass.grassdb'; 'grass' is not a package`
> 
> My guess: The directory with grass.py is on Python (import) path (`sys.path`). _Likely because it is the current directory._ Python picks grass.py instead of the grass package. Startup ends in flames.

Does anyone have an idea how to deal with this collision? It's the remaining showstopper to merge this PR.

neteler:
> Shall I revert the parts concerning Windows at this point? 

Ok, I have reverted all Windows related changes for now.
The only Windows related change is "grass79" -> "grass80".

landam:
> The only Windows related change is "grass79" -> "grass80" (ignoring any documentation related issues which may be addressed later/in a separate PR).

The Windows issue could be solved in separated PR. Lets move on with important PR.

neteler:
> The Windows issues can be solved in separated PR. Lets move on with this important PR.

With the CI hack in ea31d27 also OSGeo4W builds again. The CI only fails because of the not (yet) existing `grass8` branch in the addons repo.

I suggest to merge this PR.

nilason (approved):
Shouldn't the following lines:

https://github.com/OSGeo/grass/blob/e5379bbd7e534071eae392bf416865fdbf109f01/configure.in#L244

https://github.com/OSGeo/grass/blob/e5379bbd7e534071eae392bf416865fdbf109f01/configure#L1476

be changed to:
`INSTDIR='${prefix}'"/grass"`
?

Or is this needed for Windows build?

Else, fine for me.

neteler:
> Shouldn't the following lines:
...
> be changed to:
> `INSTDIR='${prefix}'"/grass"`
> ?
> 
> Or is this needed for Windows build?

I'd keep it for now and move on.

nilason:
> > Shouldn't the following lines:
> > ...
> > be changed to:
> > `INSTDIR='${prefix}'"/grass"`
> > ?
> > Or is this needed for Windows build?
> 
> I'd keep it for now and move on.

I needed that change on Mac at least, compilation (or was it install?) failed without it. But that can be fixed with other issues which will likely pop up after merge.

nilason:
> > > Shouldn't the following lines:
> > > ...
> > > be changed to:
> > > `INSTDIR='${prefix}'"/grass"`
> > > ?
> > > Or is this needed for Windows build?
> > 
> > 
> > I'd keep it for now and move on.
> 
> I needed that change on Mac at least, compilation (or was it install?) failed without it. But that can be fixed with other issues which will likely pop up after merge.

I now tested again, working fine. Please ignore my previous comment.

HuidaeCho:
Dropping the version from the startup script broke my cross-compilation cron jobs. Have we decided to not use `${MAJOR}${MINOR}`? What was wrong with it?

neteler:
> Dropping the version from the startup script broke my cross-compilation cron jobs. Have we decided to not use `${MAJOR}${MINOR}`? What was wrong with it?

It has been kept for Windows, and dropped elsewhere. Maybe @wenzeslaus and @nilason can comment here?

HuidaeCho:
> It has been kept for Windows, and dropped elsewhere.

And why is this inconsistency? Quoting @ninsbl

> A version independend startup script is nice if this does not interfere with installing different GRASS versions in parallel. So we might need to keep grass78 and grass80 anyway? Just to make sure different version in parallel are considered...

wenzeslaus:
> Have we decided to not use `${MAJOR}${MINOR}`? What was wrong with it?

Quoting my initial comment above:
 
> What about dropping the 80 part at least from the documentation? On a reasonable system, it should install as `grass` anyway. I also noticed I usually write `ls` and not `ls830`, `gdalinfo`, not `gdalinfo32`... Seems like less work for the releases, too.

In other words, nobody types `vim81` or `ogr2ogr321`. The only exception I can think of is Python where you may have `python`, `python3`, `python3.8`, `python2`, `python2.7`, but that's an extreme case. Moreover, it is a packaging issue more than a basic install issue.

Additionally, the current version in the name is misleading and would require a change anyway. Q: What the `80` in `grass80` means? A: Perhaps that it was not updated since the 80s.

> > It has been kept for Windows, and dropped elsewhere.
> 
> And why is this inconsistency?

Some Python import issue needs to be resolved to fix the Windows install, but that's not a reason to compromise Linux or macOS. The expectations for executables differ anyway between these platforms.

> > A version independend startup script is nice if this does not interfere with installing different GRASS versions in parallel. So we might need to keep grass78 and grass80 anyway? Just to make sure different version in parallel are considered...

Quoting earlier comments:

> > > > (note that some CI tests will fail as they expect `grass79` and not `grass80` as the name of startup script)
> > > 
> > > Can we install just `grass` to avoid this?
> > 
> > Would be practical, however, I routinely use the stable and the dev versions in parallel. But that could be solved by a link...
> 
> Let's not make it hard everywhere and for the users because developers need to have multiple versions.

In other words, the default result you get should be `grass` because that's expected. If someone is managing different versions, they likely have the skills to deal with that accordingly. There is many ways how parallel versions of one software are managed (symlinks, PATH modifications, Docker, VMs, Python virtual environment, ...), so it is not only about adding version number to the name. This is not to say that GRASS cannot help this in some way, but the default behavior should be that you get a basic `grass` command in one way or another without any extra steps.

Scripts, CI, etc. should generally just require change of the installed version (Git reference, package version, Docker image, ...) and the rest should work out of the box (assuming no changes in the API), i.e., without the need to go over the files and modify all references to the executable or other things to update the version. This should happen without the need to parametrize the version number and the need to pass it through all the files.

PR #1598, opened 2021-05-27
wxGUI: easier switching between locations (GRASS GIS 8)

lindakladivova:
When switching to a map layer in a different location this message will appear:

Map layer in different location_004

### DIFF
--- a/gui/wxpython/datacatalog/tree.py
+++ b/gui/wxpython/datacatalog/tree.py
@@ -990,10 +990,10 @@ class DataCatalogTree(TreeView):
                             selected_mapset.data["name"],
                             selected_loc.data["name"],
                         ),
-                        caption=_("Map layer in different location"),
+                        caption=_("Map layer in a different location"),
                         style=wx.YES_NO | wx.NO_DEFAULT | wx.ICON_QUESTION,
                     )
-                    dlg.SetYesNoLabels("S&witch now", "S&tay in current mapset")
+                    dlg.SetYesNoLabels("S&witch now", "S&tay in current location")
                     if dlg.ShowModal() == wx.ID_YES:
                         if can_switch_mapset_interactive(
                             self,
--- a/gui/wxpython/datacatalog/tree.py
+++ b/gui/wxpython/datacatalog/tree.py
@@ -982,13 +982,14 @@ class DataCatalogTree(TreeView):
                     dlg = wx.MessageDialog(
                         parent=self,
                         message=_(
-                            "Map <{0}@{1}> is not in the current location. "
-                            "Do you want to switch to the location '{2}'? \n\n"
-                            "All current Map Displays will be closed."
+                            "Map <{map_name}@{map_mapset}> is not in the current location. "
+                            "If you switch to the location <{map_location}> all current "
+                            "Map Displays will be closed.\n\n"
+                            "Do you want to switch anyway?"
                         ).format(
-                            selected_layer.data["name"],
-                            selected_mapset.data["name"],
-                            selected_loc.data["name"],
+                            map_name=selected_layer.data["name"],
+                            map_mapset=selected_mapset.data["name"],
+                            map_location=selected_loc.data["name"],
                         ),
                         caption=_("Map layer in a different location"),
                         style=wx.YES_NO | wx.NO_DEFAULT | wx.ICON_QUESTION,
--- a/gui/wxpython/datacatalog/tree.py
+++ b/gui/wxpython/datacatalog/tree.py
@@ -983,7 +983,8 @@ class DataCatalogTree(TreeView):
                         parent=self,
                         message=_(
                             "Map <{map_name}@{map_mapset}> is not in the current location. "
-                            "If you switch to the location <{map_location}> all current "
+                            "To be able to display it you need to switch to <{map_location}> "
+                            "location. Note that if you switch there all current "
                             "Map Displays will be closed.\n\n"
                             "Do you want to switch anyway?"
                         ).format(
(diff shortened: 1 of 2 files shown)


neteler (approved):
This is a great solution! Thanks so much @lindakladivova

gui/wxpython/datacatalog/tree.py
```diff
@@ -981,17 +982,31 @@ def OnDoubleClick(self, node):
                     dlg = wx.MessageDialog(
                         parent=self,
                         message=_(
-                            "Map <{0}@{1}> is not in the current location"
-                            " and therefore cannot be displayed."
-                            "\n\n"
-                            "To display this map switch to mapset <{1}> first."
+                            "Map <{0}@{1}> is not in the current location. "
+                            "Do you want to switch to the location '{2}'? "
+                            "Your currently displayed layers will disappear."
```
wenzeslaus, line 987: The reason I used the word "disappear" in the original issue is that it expresses well the confusion a user may have when the layers are unloaded, i.e., something unexpected happened. Furthermore, "disappear" in this context may sound as "underlying data will be deleted". I don't have another suggestion at this point, but Perhaps mentioning again the reason for it in connection with the change in wording can bring some understanding.

wenzeslaus (changes requested):

gui/wxpython/datacatalog/tree.py
```diff
@@ -1838,30 +1860,30 @@ def SwitchMapset(self, grassdb, location, mapset, show_confirmation=False):
         """
         Switch to location and mapset interactively.
         """
-        if can_switch_mapset_interactive(self, grassdb, location, mapset):
```
petrasovaa, line 1841: What is the reason for moving this if out of the function?

lindakladivova: The reason is that we display a layer only if a switching is successful. I had another idea how to solve this issue - make a optional argument display_layer for SwitchMapset method. But finally I realized that SwitchMapset is something not much related to displaying layer.
Now code also seems to me a bit clearer because SwitchMapset really switch in all cases.
Making a display_layer argument would be much shorter in code but I think not much meaningful in terms of logic.
Or we could return True or False from SwitchMapset but I am not sure If this function should return something, It is more of procedure. What do you think @petrasovaa ?

petrasovaa: I see. I would be in favor of returning true/false since we need to use can_switch_mapset_interactive anyway every time we want to switch it, so it may be better have it in the SwitchMapset method.

gui/wxpython/datacatalog/tree.py
petrasovaa: Maybe something along the lines "All current Map Displays will be closed"?

lindakladivova: And put it on the next line? Now it looks as follows: switch

gui/wxpython/datacatalog/tree.py
```diff
@@ -981,17 +982,25 @@ def OnDoubleClick(self, node):
                     dlg = wx.MessageDialog(
                         parent=self,
                         message=_(
-                            "Map <{0}@{1}> is not in the current location"
-                            " and therefore cannot be displayed."
-                            "\n\n"
-                            "To display this map switch to mapset <{1}> first."
+                            "Map <{0}@{1}> is not in the current location. "
+                            "Do you want to switch to the location '{2}'? \n\n"
```
petrasovaa, line 986: Just a small change for consistency, it should be <location>, not 'location'.

wenzeslaus, line 986: `{0} {1} {2}` should be `{map_name} {map_mapset} {map_location}`?

wenzeslaus (changes requested):

gui/wxpython/datacatalog/tree.py
```diff
@@ -981,17 +982,25 @@ def OnDoubleClick(self, node):
...
+                            "All current Map Displays will be closed."
```
wenzeslaus, line 987: This looks like "will be closed in any case". I think the new paragraph needs to be between "...the current location" and "All current Map...".

wenzeslaus (changes requested):

gui/wxpython/datacatalog/tree.py
lindakladivova: Ok, then I would suggest this formulation:
easier_switch

veroandreo: That message sounds a bit intimidating to me... Can we say something like: 

```
Map <bla> is not in the current location. To be able to display it you need to switch to <blabla> location. Note that if you switch there all current Map Displays will be closed. 

Do you want to switch anyway?
```

Too long? My intention is to "explain" in short that to display a map from a different location the user needs to switch there.

lindakladivova: I like your idea more Vero. It is true that it is quite long but we need to fully explain it. @petrasovaa do you agree with that?

nilason: > That message sounds a bit intimidating to me... Can we say something like:
> 
> ```
> Map <bla> is not in the current location. To be able to display it you need to switch to <blabla> location. 
> Note that if you switch there all current Map Displays will be closed. 
> 
> Do you want to switch anyway?
> ```
> 
> Too long? My intention is to "explain" in short that to display a map from a different location the user needs to switch there.

I think @veroandreo 's formulation is a bit more appealing. However, I think the button labels are too wordy. What about `Cancel` (as default) and `Switch`?

lindakladivova: switch

PR #1600, opened 2021-05-27
PyGRASS: MultiModule: return MultiModule object

ninsbl:
This PR should make MultiModule work again in the PrallelModuleQueue.

See:
https://lists.osgeo.org/pipermail/grass-dev/2021-May/095177.html
and
https://github.com/OSGeo/grass/blob/master/temporal/t.rast.neighbors/testsuite/test_neighbors.py

### DIFF
--- a/python/grass/pygrass/modules/interface/module.py
+++ b/python/grass/pygrass/modules/interface/module.py
@@ -996,7 +996,7 @@ class MultiModule(object):
             for module in self.module_list:
                 module.finish_ = True
                 module.run()
-            return None
+            return self
         else:
             if self.set_temp_region is True:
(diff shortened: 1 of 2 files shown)


ninsbl:
Hm, seems the return is not the issue.
The CI-tests fail and on my development box the tests succeed also without returning a Module object.
Python version there is 3.6.9. CI-python version is 3.8. However, on my other box where the test fails too, the Python version is 3.6.9 too. So that is probably not the issue either...

ninsbl:
Closing this as invalid, as the issue is solved in master already. See discussion in the mailinglist referenced above.

PR #1603, opened 2021-05-30
Added Binder setup files

chaedri:
I'm just opening this pull request draft to get the ball rolling for my Google Summer of Code project. As part of the GRASS and Jupyter Notebook project, we discussed adding setup files to run binder.


binder/postBuild
```diff
@@ -0,0 +1,43 @@
...
+
+# exit on error
+set -e
+
+# use code dir for code
+mkdir code
+cd code
+
+# get the source code
+git clone https://github.com/OSGeo/grass.git
```
ninsbl, line 11: This seems to allway build the official master branch. So, it wil not work for testing PRs, right?

I tried to find some info on how to solve this but did not come across a solution that is known to work, but you could try if it is possible in the postbuild script to:

1 fetch teh url of the current repo:
`git config --get remote.origin.url`

and
2. use that to clone the code from which GRASS is supposed to be compiled...

But maybe @wenzeslaus has some more clear ideas about this...

ninsbl (commented):
Looks good to me.
Probably best to include download of sample data into the notebook if needed...
Then the commented code at the bottom of the postBuild script could be removed...

binder/postBuild
wenzeslaus: Just removing all the `cd`s, `mkdir`s and clone should be enough. Binder actually puts the repo content to what is the current directory when the Docker container starts. This makes it simply use always the current code whatever is the the current one, master branch in upstream, branch on a fork, ...

You can test that by going to Binder, the URL is (obtained by filling out mybinder.org):

<https://mybinder.org/v2/gh/chaedri/grass/binder>

@ninsbl I'm not sure what you are aiming at, but if I understand correctly, if you can do `git config --get remote.origin.url`, you probably have the repo clone already, so you don't need to do it at all.

binder/postBuild
```diff
@@ -0,0 +1,43 @@
...
+# create link to build
+ln -s $HOME/code/grass/bin.*/grass* $HOME/.local/bin/grass
+
+# OPTIONAL
+# download a sample dataset
+# mkdir -p data/grassdata \
+#  && curl -SL https://grass.osgeo.org/sampledata/north_carolina/nc_basic_spm_grass7.zip > nc_basic_spm_grass7.zip \
+#  && unzip -qq nc_basic_spm_grass7.zip \
+#  && mv nc_basic_spm_grass7 data/grassdata \
+#  && rm nc_basic_spm_grass7.zip
```
wenzeslaus, line 43: As GitHub Changes view says, no new line at the end of file.

binder/requirements.txt
```diff
@@ -0,0 +1,2 @@
+numpy
```
wenzeslaus, line 1: Move the pil and ply dependencies here from the apt file. Delete numpy there. Test if the packages are there, e.g., with (in the notebook):

```
!grass --tmp-location XY --exec python -c "import matplotlib"
```

wenzeslaus (changes requested):
Thank you. I think this will be very helpful.

wenzeslaus:
> Probably best to include download of sample data into the notebook if needed...
> Then the commented code at the bottom of the postBuild script could be removed...

Or download the data always? We don't know how this will be used, so maybe the download on demand in the notebook is best now. Definitively remove the comment.

wenzeslaus:
This is for a separate PR, but I think a GitHub Action should generate a Binder button aka Binder badge (which is just a URL and an image) and post it as a comment to each new PR so that you don't have to guess or construct the Binder URL yourself. That's my idea how this should be eventually used. The URL can also point to a specific notebook so you don't need to go through the directory structure to find it. The button/badge for this PR with notebooks and JupyterLab instead of a plain notebook:

[Binder](https://mybinder.org/v2/gh/chaedri/grass/binder?urlpath=lab/tree/doc%2Fnotebooks%2Fexample_notebook.ipynb)

```
 _____________________________
| Boootton. How cool is that? |
 -----------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
```

Suggestions about the content of the notebook are welcome (there can be also more than one), but that's also for another PR. Empty one is fine with me now as well as skipping the dataset download.

binder/postBuild
ninsbl: @wenzeslaus , thanks great, I was not aware of how this is handled in binder. I just saw the line that clones the official repo which whould not result in the behavior we discused. So, if compiling from current directory works, that is great and please disregard my suggestion of `git config`...

ninsbl:
> Or download the data always? We don't know how this will be used, so maybe the download on demand in the notebook is best now. Definitively remove the comment.

Yes, I think both ways of handling that would be good. My thought was to have a blank setup and add (sample) data - including location / mapset setup - in the respective example notebook... As allways there are pros and cons and I am fine with either solution...

ninsbl:
> This is for a separate PR, but I think a GitHub Action should generate a Binder button aka Binder badge (which is just a URL and an image) and post it as a comment to each new PR so that you don't have to guess or construct the Binder URL yourself. 

Sounds like a very good idea.

binder/requirements.txt
chaedri: When I try to move the pil and ply dependencies, I get an error when I try to build binder:

'''
ERROR: Could not find a version that satisfies the requirement pil
ERROR: No matching distribution found for pil
'''

From my googling, it seems like PIL has to be installed first. Should it remain in the apt file?

binder/postBuild
```diff
@@ -0,0 +1,27 @@
...
+    --with-opengl-libs=/usr/include/GL \
+    --with-freetype=yes --with-freetype-includes="/usr/include/freetype2/" \
+    --with-sqlite=yes
+make
+
+# put command on path
+# ensure the user specific bin dir exists (already on path)
+mkdir -p $HOME/.local/bin/
+# create link to build
+ln -s $HOME/code/grass/bin.*/grass* $HOME/.local/bin/grass
```
ninsbl, line 27:
```suggestion
ln -s $HOME/grass/bin.*/grass* $HOME/.local/bin/grass
```

binder/requirements.txt
wenzeslaus: What is in the apt file goes to system Python and is not available in the notebook or GRASS when I test in the notebook using the above (which is using conda in this default Binder setup). You can try to confirm it. If that holds, apt file is not useful for Python packages. And python and python-dev should not be there either (I'm not even sure how python-dev got there, maybe to compile numpy or something).

What you import as pil, is installed with its proper name PIL. The requirements file goes to this repo: https://pypi.org/project/PIL/. So try PIL.

binder/postBuild
wenzeslaus: @chaedri to test this use `!grass --version` in the notebook or `grass --version` in a shell you get in the JupyterLab.

ninsbl:
> This is for a separate PR, but I think a GitHub Action should generate a Binder button aka Binder badge

So something like along this line I guess:
https://github.com/jupyterhub/repo2docker-action#use-github-actions-to-cache-the-build-for-binderhub

binder/postBuild
wenzeslaus: The first `grass` part is still one dir extra here even in the suggestion.

wenzeslaus (changes requested):
I think we can merge after you accept and test the changes.

BTW, perhaps obvious, but I used the button/link to test this and come up with the suggestion.

ninsbl (approved):
Looks good to me.

wenzeslaus (approved):
The `grass` (or `!grass`) command runs in Binder which is what I saw as the goal for this PR.

wenzeslaus:
Awesome! Now all PRs can be tested in Binder. One just needs to figure out the URL either by filling out the form at <https://mybinder.org> or by filling out this template:

```
https://mybinder.org/v2/gh/<PR author name>/grass/<branch name>
```

PR #1608, opened 2021-05-31
scripts/r.plane/r.plane.py:fixed bare except

ShubhamSwati:
r.plane/r.plane.py: E722: fixed bare except


scripts/.flake8
```diff
@@ -29,10 +29,10 @@ per-file-ignores =
...
     r.fillnulls/r.fillnulls.py: E722
     d.rast.edit/d.rast.edit.py: E722
-    g.manual/g.manual.py: E722
-    v.to.lines/v.to.lines.py: E722
-    v.in.wfs/v.in.wfs.py: E722
-    r.plane/r.plane.py: E722
+    
+    
+    
+
```
wenzeslaus, line 35: Just delete the lines. No need to keep them.

wenzeslaus (changes requested):

scripts/.flake8
ShubhamSwati: Now I have deleted these 4 lines from scripts/.flake8. Kindly review.

wenzeslaus (approved):
Thank you. This works!

FYI, `except Exception:` is not good enough for more strict checks from Pylint, but it is good enough for Flake8 and that's all we were aiming for now.

PR #1612, opened 2021-06-02
g.extension: workaround to use grass7 addon repo

neteler:
In order to avoid that `g.extension` fails to find extensions after the merge of #1597 here a temporary workaround of using the `grass7/` subdirectory of the GRASS GIS addons repository.

Once the new addons repository structure is implemented (see also https://github.com/OSGeo/grass-addons/issues/528) then this PR shall be reverted/updated.


neteler:
Yay!!
image

neteler:
Ops I already merged it... wrong button.

PR #1621, opened 2021-06-05
wxGUI: Single-Window GUI: arrangement of basic widgets

lindakladivova:
This PR deals with the base settings for the full screen layout with main dockable panes. 
Map Displays as well as 3D view are not solved in this PR - we just need to prepare the base structure here.

There are two main things that we need to figure out:

- Any content "not dockable" pane is necessary in terms of implementation. So first we need to decide whether to employ Data Catalog as the content pane.  The advantage is that all other panes would be dockable then. :-) And also after changes made last year the Data Catalog is kind of the center point of GRASS so it could make sense to have it as the content  pane. The con is that visually it might look a bit strange. There the layout I suggest so far in this PR (using wx.AUI):

tmploc-PERMANENT - GRASS GIS_010

- Second important thing is to decide which library to use. We have two options: the traditional **wx.AUI** C++ library and the newer **wx.lib.agw.aui** python library. I tried to compared them a bit. There is a bit strange thing about wx.AUI -> MinimizeButton method does not work to me although according to https://wxpython.org/Phoenix/docs/html/wx.aui.AuiPaneInfo.html#wx.aui.AuiPaneInfo.MinimizeButton documentation, it should work. :-( The second thing I have come up with so far is the missing guide docking styles such as nice aui.AUI_MGR_WHIDBEY_DOCKING_GUIDES style.
Both of these functions work for wx.lib.agw.aui.

But there may be (and probably are) more things that are not implemented for wx.AUI. But I think we do not have to make the decision about the library right now.

### DIFF
--- a/gui/wxpython/main_window/frame.py
+++ b/gui/wxpython/main_window/frame.py
@@ -2158,16 +2158,6 @@ class GMFrame(wx.Frame):
         """Key pressed"""
         kc = event.GetKeyCode()
 
-        #        if event.ControlDown():
-        #            if kc == wx.WXK_TAB:
-        #                # switch layer list / command output
-        #                if self.notebook.GetSelection() == self.notebook.GetPageIndexByName(
-        #                    "layers"
-        #                ):
-        #                    self.notebook.SetSelectionByName("output")
-        #                else:
-        #                    self.notebook.SetSelectionByName("layers")
-
         try:
             kc = chr(kc)
         except ValueError:
--- a/gui/wxpython/main_window/frame.py
+++ b/gui/wxpython/main_window/frame.py
@@ -92,7 +92,7 @@ class GMFrame(wx.Frame):
         id=wx.ID_ANY,
         title=None,
         workspace=None,
-        size=wx.Display(0).GetGeometry().GetSize(),
+        size=globalvar.GM_WINDOW_SIZE,
         style=wx.DEFAULT_FRAME_STYLE,
         **kwargs,
     ):
@@ -150,7 +150,7 @@ class GMFrame(wx.Frame):
         self.dialogs["atm"] = list()
 
         # set pane sizes according to the full screen size of the primary monitor
-        size = wx.Display(0).GetGeometry().GetSize()
+        size = wx.Display().GetGeometry().GetSize()
         self.PANE_BEST_SIZE = tuple(t / 5 for t in size)
         self.PANE_MIN_SIZE = tuple(t / 10 for t in size)
 
--- a/gui/wxpython/main_window/frame.py
+++ b/gui/wxpython/main_window/frame.py
@@ -25,8 +25,13 @@ import platform
 import re
 
 from core import globalvar
+#try:
+#    from agw import aui
+#except ImportError:
+#    import wx.lib.agw.aui as aui
+
 import wx
-import wx.aui
+import wx.aui as aui
 
 try:
     import wx.lib.agw.flatnotebook as FN
@@ -45,7 +50,7 @@ from core.utils import SetAddOnPath, GetLayerNameFromCmd, command2ltype, get_she
 from gui_core.preferences import MapsetAccess, PreferencesDialog
 from lmgr.layertree import LayerTree, LMIcons
 from lmgr.menudata import LayerManagerMenuData, LayerManagerModuleTree
-from gui_core.widgets import GNotebook, FormNotebook
+from gui_core.widgets import GNotebook
 from core.gconsole import GConsole, EVT_IGNORED_CMD_RUN
 from core.giface import Notification
 from gui_core.goutput import GConsoleWindow, GC_PROMPT
@@ -87,7 +92,7 @@ class GMFrame(wx.Frame):
         id=wx.ID_ANY,
         title=None,
         workspace=None,
-        size=globalvar.GM_WINDOW_SIZE,
+        size=wx.DisplaySize(),
         style=wx.DEFAULT_FRAME_STYLE,
         **kwargs,
     ):
@@ -136,7 +141,7 @@ class GMFrame(wx.Frame):
         self._menuTreeBuilder = LayerManagerMenuData(message_handler=add_menu_error)
         # the search tree and command console
         self._moduleTreeBuilder = LayerManagerModuleTree(message_handler=add_menu_error)
-        self._auimgr = wx.aui.AuiManager(self)
+        self._auimgr = aui.AuiManager(self)
 
         # list of open dialogs
         self.dialogs = dict()
@@ -144,83 +149,15 @@ class GMFrame(wx.Frame):
         self.dialogs["nvizPreferences"] = None
         self.dialogs["atm"] = list()
 
-        # create widgets
-        self._createMenuBar()
-        self.statusbar = self.CreateStatusBar(number=1)
-        self.notebook = self._createNotebook()
-        self._createDataCatalog(self.notebook)
-        self._createDisplay(self.notebook)
-        self._createSearchModule(self.notebook)
-        self._createConsole(self.notebook)
-        self._createPythonShell(self.notebook)
-        self._addPagesToNotebook()
-        self.toolbars = {
-            "workspace": LMWorkspaceToolbar(parent=self),
-            "tools": LMToolsToolbar(parent=self),
-            "misc": LMMiscToolbar(parent=self),
-            "nviz": LMNvizToolbar(parent=self),
-        }
-        self._toolbarsData = {
-            "workspace": (
-                "toolbarWorkspace",  # name
-                _("Workspace Toolbar"),  # caption
-                1,
-                0,
-            ),  # row, position
-            "tools": ("toolbarTools", _("Tools Toolbar"), 1, 1),
-            "misc": ("toolbarMisc", _("Misc Toolbar"), 1, 2),
-            "nviz": ("toolbarNviz", _("3D view Toolbar"), 1, 3),
-        }
-        toolbarsList = ("workspace", "tools", "misc", "nviz")
-        for toolbar in toolbarsList:
-            name, caption, row, position = self._toolbarsData[toolbar]
-            self._auimgr.AddPane(
-                self.toolbars[toolbar],
-                wx.aui.AuiPaneInfo()
-                .Name(name)
-                .Caption(caption)
-                .ToolbarPane()
-                .Top()
-                .Row(row)
-                .Position(position)
-                .LeftDockable(False)
-                .RightDockable(False)
-                .BottomDockable(False)
-                .TopDockable(True)
-                .CloseButton(False)
-                .Layer(2)
-                .BestSize((self.toolbars[toolbar].GetBestSize())),
-            )
-
-        self._auimgr.GetPane("toolbarNviz").Hide()
-        # bindings
-        self.Bind(wx.EVT_CLOSE, self.OnCloseWindowOrExit)
-        self.Bind(wx.EVT_KEY_DOWN, self.OnKeyDown)
+        # create widgets and build panes
+        self.CreateMenuBar()
+        self.CreateStatusBar(number=1)
+        self.BuildPanes()
+        self.BuildEvents()
 
         self._giface.mapCreated.connect(self.OnMapCreated)
         self._giface.updateMap.connect(self._updateCurrentMap)
         self._giface.currentMapsetChanged.connect(self.OnMapsetChanged)
-
-        # minimal frame size
-        self.SetMinSize(globalvar.GM_WINDOW_MIN_SIZE)
-
-        # AUI stuff
-        self._auimgr.AddPane(
-            self.notebook,
-            wx.aui.AuiPaneInfo()
-            .Left()
-            .CentrePane()
-            .BestSize((-1, -1))
-            .Dockable(False)
-            .CloseButton(False)
-            .DestroyOnClose(True)
-            .Row(1)
-            .Layer(0),
-        )
-
-        self._auimgr.Update()
-
-        wx.CallAfter(self.notebook.SetSelectionByName, "catalog")
 
         # use default window layout ?
         if UserSettings.Get(group="general", key="defWindowPos", subkey="enabled"):
@@ -238,6 +175,7 @@ class GMFrame(wx.Frame):
 
         self.Layout()
         self.Show()
+        self.Maximize(True)
 
         # load workspace file if requested
         if workspace:
@@ -247,7 +185,7 @@ class GMFrame(wx.Frame):
             # start default initial display
             self.NewDisplay(show=False)
 
-        # show map display widnow
+        # show map display window
         # -> OnSize() -> UpdateMap()
         for mapdisp in self.GetMapDisplay(onlyCurrent=False):
             mapdisp.Show()
@@ -291,7 +229,7 @@ class GMFrame(wx.Frame):
                 )
             )
 
-    def _createMenuBar(self):
+    def CreateMenuBar(self):
         """Creates menu bar"""
         self.menubar = GMenu(
             parent=self, model=self._menuTreeBuilder.GetModel(separators=True)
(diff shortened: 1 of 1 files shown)


gui/wxpython/main_window/frame.py
```diff
@@ -136,92 +141,24 @@ def show_menu_errors(messages):
...
-
-        self._auimgr.GetPane("toolbarNviz").Hide()
-        # bindings
-        self.Bind(wx.EVT_CLOSE, self.OnCloseWindowOrExit)
-        self.Bind(wx.EVT_KEY_DOWN, self.OnKeyDown)
+        # create widgets and build panes
+        self.CreateMenuBar()
+        self.CreateStatusBar(number=1)
+        self.BuildPanes()
+        self.BuildEvents()
```
petrasovaa, line 156: change to BindEvents

petrasovaa:
> 
> * Second important thing is to decide which library to use. We have two options: the traditional **wx.AUI** C++ library and the newer **wx.lib.agw.aui** python library. I tried to compared them a bit. There is a bit strange thing about wx.AUI -> MinimizeButton 

I am leaning now towards the agw one since the wxWidgets one seems to be quite limited for our use case. For example we would like to have the option to group Modules, Data and Display into one side notebook widget, which you can do only with agw.

veroandreo:
> This PR deals with the base settings for the full screen layout with main dockable panes.
> Map Displays as well as 3D view are not solved in this PR - we just need to prepare the base structure here.
> 
> There are two main things that we need to figure out:
> 
>     * Any content "not dockable" pane is necessary in terms of implementation. So first we need to decide whether to employ Data Catalog as the content pane.  The advantage is that all other panes would be dockable then. :-) And also after changes made last year the Data Catalog is kind of the center point of GRASS so it could make sense to have it as the content  pane. The con is that visually it might look a bit strange. There the layout I suggest so far in this PR (using wx.AUI):
> 
> 
> tmploc-PERMANENT - GRASS GIS_010

This looks really cool! Where will the *layer* tab go? Is it possible to put it below the data catalog similar to QGIS?

lindakladivova:
> 
> 
> > This PR deals with the base settings for the full screen layout with main dockable panes.
> > Map Displays as well as 3D view are not solved in this PR - we just need to prepare the base structure here.
> > There are two main things that we need to figure out:
> > ```
> > * Any content "not dockable" pane is necessary in terms of implementation. So first we need to decide whether to employ Data Catalog as the content pane.  The advantage is that all other panes would be dockable then. :-) And also after changes made last year the Data Catalog is kind of the center point of GRASS so it could make sense to have it as the content  pane. The con is that visually it might look a bit strange. There the layout I suggest so far in this PR (using wx.AUI):
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > tmploc-PERMANENT - GRASS GIS_010
> 
> This looks really cool! Where will the _layer_ tab go? Is it possible to put it below the data catalog similar to QGIS?

Yes, we can make it as I suggested here: (1) .
But I cannot decide which panel should take the role of the Content pane (non-dockable pane). I think it would be nice to have the Single-Window GUI where we can reach the Multi-Window GUI as well (only by pulling out Map Displays). In (1) we cannot do that because the notebook containing all Map Displays have the role of the Content pane.

So we need either **use Data Catalog as the Content pane** (as this PR suggests) - then all Map Displays and other panes will be dockable, or we could **make the first opened Map Display as the Content pane** and all other Map Displays will be dockable - this means, however, allowing the user to change it, e.g. suddenly a user wants to place the first Map Display on the second monitor, so he/she changes some settings and the first Map Display become dockable while the second Map Display becomes the Content pane.

I would add that people are already used to have some data organization and layer management on the left side and tools on the right side, so I would keep the arrangement in (1). And also having the Map Display in the center looks impressive. :-) Any suggestions about that?

Which idea regarding the Content pane you like more @veroandreo? Do you have any other ideas?

(1) https://trac.osgeo.org/grass/attachment/wiki/wxGUIDevelopment/SingleWindow/single_window_layout.png

veroandreo:
> > This looks really cool! Where will the _layer_ tab go? Is it possible to put it below the data catalog similar to QGIS?
> 
> Yes, we can make it as I suggested here: (1) .

Yes, that one I like, so user can have control of layer properties

> But I cannot decide which panel should take the role of the Content pane (non-dockable pane). I think it would be nice to have the Single-Window GUI where we can reach the Multi-Window GUI as well (only by pulling out Map Displays). In (1) we cannot do that because the notebook containing all Map Displays has the role of the Content pane.

Ok, I see. Indeed, it'd be nice to be able to switch to multiple window by pulling the maps, seems intuitive. However, just to confirm, would having the data catalog as content pane as in the fig above, still allow to place the layer manager below the data catalog as in (1)? If yes, that's perfect. I think placing the layer manager below data catalog is kinda natural, so the user has full control of layers, all in the same side of the screen (this is what I prefer, at least)

> So we need to either use **the Data Catalog as the Content pane** (as this PR suggests) - then all Map Displays will be dockable, or we could use **the first opened Map Display as the Content pane** and all other Map Displays except the first one will be dockable - this means, however, allowing the user to change it, e.g. suddenly a user wants to place the first Map Display on the second monitor, so he/she changes some settings and the first Map Display become dockable while the second Map Display becomes the Content pane.

So using the map display as content pane would not allow to go back to multiple window layout as we have now, because one of them would always stay as content pane, right?
 
> I would add that people are already used to have some data organization and layer management on the left side and tools on the right side, so I would keep the arrangement in (1). And also having the Map Display in the center looks impressive. :-) Any suggestions about that?
> 
> Which idea regarding the Content pane you like more @veroandreo? Do you have any other ideas?

I'm with you there! I love the layout in (1) ! This is my preferred one. In my personal use, indeed, I'd remove all those dockable panes on the right and only keep data catalog + layer manager (as in 1) on the left, and map display(s) over the rest of the screen :)

> (1) https://trac.osgeo.org/grass/attachment/wiki/wxGUIDevelopment/SingleWindow/single_window_layout.png

mlennert:
Am 08.06.2021 14:39 schrieb Veronica Andreo:
>>> This looks really cool! Where will the _layer_ tab go? Is it
>>> possible to put it below the data catalog similar to QGIS?
>> 
>> Yes, we can make it as I suggested here: (1) .
> 
> Yes, that one I like, so user can have control of layer properties
> 
>> But I cannot decide which panel should take the role of the Content
>> pane (non-dockable pane). I think it would be nice to have the
>> Single-Window GUI where we can reach the Multi-Window GUI as well
>> (only by pulling out Map Displays). In (1) we cannot do that because
>> the notebook containing all Map Displays has the role of the Content
>> pane.
> 
> Ok, I see. Indeed, it'd be nice to be able to switch to multiple
> window by pulling the maps, seems intuitive. However, just to confirm,
> would having the data catalog as content pane as in the fig above,
> still allow to place the layer manager below the data catalog as in
> (1)? If yes, that's perfect. I think placing the layer manager below
> data catalog is kinda natural, so the user has full control of layers,
> all in the same side of the screen (this is what I prefer, at least)
> 
>> So we need to either use THE DATA CATALOG AS THE CONTENT PANE (as
>> this PR suggests) - then all Map Displays will be dockable, or we
>> could use THE FIRST OPENED MAP DISPLAY AS THE CONTENT PANE and all
>> other Map Displays except the first one will be dockable - this
>> means, however, allowing the user to change it, e.g. suddenly a user
>> wants to place the first Map Display on the second monitor, so
>> he/she changes some settings and the first Map Display become
>> dockable while the second Map Display becomes the Content pane.
> 
> So using the map display as content pane would not allow to go back to
> multiple window layout as we have now, because one of them would
> always stay as content pane, right?
> 
>> I would add that people are already used to have some data
>> organization and layer management on the left side and tools on the
>> right side, so I would keep the arrangement in (1). And also having
>> the Map Display in the center looks impressive. :-) Any suggestions
>> about that?
>> 
>> Which idea regarding the Content pane you like more @veroandreo [1]?
>> Do you have any other ideas?
> 
> I'm with you there! I love the layout in (1) ! This is my preferred
> one. In my personal use, indeed, I'd remove all those dockable panes
> on the right and only keep data catalog + layer manager (as in 1) on
> the left, and map display(s) over the rest of the screen :)

+1 also from me for Data catalog as content pane.

lindakladivova:
> > > This looks really cool! Where will the _layer_ tab go? Is it possible to put it below the data catalog similar to QGIS?
> > 
> > 
> > Yes, we can make it as I suggested here: (1) .
> 
> Yes, that one I like, so user can have control of layer properties
> 
> > But I cannot decide which panel should take the role of the Content pane (non-dockable pane). I think it would be nice to have the Single-Window GUI where we can reach the Multi-Window GUI as well (only by pulling out Map Displays). In (1) we cannot do that because the notebook containing all Map Displays has the role of the Content pane.
> 
> Ok, I see. Indeed, it'd be nice to be able to switch to multiple window by pulling the maps, seems intuitive. However, just to confirm, would having the data catalog as content pane as in the fig above, still allow to place the layer manager below the data catalog as in (1)? If yes, that's perfect. I think placing the layer manager below data catalog is kinda natural, so the user has full control of layers, all in the same side of the screen (this is what I prefer, at least)
> 
> > So we need to either use **the Data Catalog as the Content pane** (as this PR suggests) - then all Map Displays will be dockable, or we could use **the first opened Map Display as the Content pane** and all other Map Displays except the first one will be dockable - this means, however, allowing the user to change it, e.g. suddenly a user wants to place the first Map Display on the second monitor, so he/she changes some settings and the first Map Display become dockable while the second Map Display becomes the Content pane.
> 
> So using the map display as content pane would not allow to go back to multiple window layout as we have now, because one of them would always stay as content pane, right?
> 
> > I would add that people are already used to have some data organization and layer management on the left side and tools on the right side, so I would keep the arrangement in (1). And also having the Map Display in the center looks impressive. :-) Any suggestions about that?
> > Which idea regarding the Content pane you like more @veroandreo? Do you have any other ideas?
> 
> I'm with you there! I love the layout in (1) ! This is my preferred one. In my personal use, indeed, I'd remove all those dockable panes on the right and only keep data catalog + layer manager (as in 1) on the left, and map display(s) over the rest of the screen :)
> 
> > (1) https://trac.osgeo.org/grass/attachment/wiki/wxGUIDevelopment/SingleWindow/single_window_layout.png

lindakladivova:
I first thought that we could employ the Data Catalog as the content pane but it is not possible since we need the Data Catalog to occupy only the small part of the window but the Best Size settings do not work for the center pane. :-( It means that the Center pane always adapts and extends according to other widgets. That would not be good:

tmploc-PERMANENT - GRASS GIS_012

In yesterday's video call, we decided to use the Map Display as the Content pane instead. We can either follow the first layout proposal where Map Displays are in notebooks. There is a nice functionality that the notebook can be split so that we can see several Map Displays side by side as I suggested in the mock-up...:
split_displays

Or we would have one center Map Display and each extra Map Display could be launched as a separate window as we know  from the current solution - here we could utilize the second monitor.

So in this PR I proposed the basic skeleton and the Map Display content pane is left blank:
world_latlong_wgs84-PERMANENT - GRASS GIS_014
 For purpose of Map Display we need the Map Display in the form of wx.Panel (not only wx.Frame as now) and it is the task for the next PR.

veroandreo:
> I first thought that we could employ the Data Catalog as the content pane but it is not possible since we need the Data Catalog to occupy only the small part of the window but the Best Size settings do not work for the center pane. :-( It means that the Center pane always adapts and extends according to other widgets. That would not be good:
> 
> tmploc-PERMANENT - GRASS GIS_012
> 
> In yesterday's video call, we decided to use the Map Display as the Content pane instead. We can either follow the first layout proposal where Map Displays are in notebooks. There is a nice functionality that the notebook can be split so that we can see several Map Displays side by side as I suggested in the mock-up...:
> split_displays

Will we be able to collapse several map displays as tabs too as the tabs in the display pane in the mock-up? (I lack the right vocabulary, sorry)

gui/wxpython/main_window/frame.py
```diff
@@ -136,92 +141,28 @@ def show_menu_errors(messages):
...
-                .LeftDockable(False)
-                .RightDockable(False)
-                .BottomDockable(False)
-                .TopDockable(True)
-                .CloseButton(False)
-                .Layer(2)
-                .BestSize((self.toolbars[toolbar].GetBestSize())),
-            )
+        # set pane sizes according to the full screen size
+        self.PANE_BEST_SIZE = tuple(t / 5 for t in wx.DisplaySize())
```
petrasovaa, line 153: There is a note warning about using this function in the [documentation](https://wxpython.org/Phoenix/docs/html/wx.functions.html#wx.DisplaySize), so I suggest to follow it.

lindakladivova:
> 
> 
> > I first thought that we could employ the Data Catalog as the content pane but it is not possible since we need the Data Catalog to occupy only the small part of the window but the Best Size settings do not work for the center pane. :-( It means that the Center pane always adapts and extends according to other widgets. That would not be good:
> > tmploc-PERMANENT - GRASS GIS_012
> > In yesterday's video call, we decided to use the Map Display as the Content pane instead. We can either follow the first layout proposal where Map Displays are in notebooks. There is a nice functionality that the notebook can be split so that we can see several Map Displays side by side as I suggested in the mock-up...:
> > split_displays
> 
> Will we be able to collapse several map displays as tabs too as the tabs in the display pane in the mock-up? (I lack the right vocabulary, sorry)

Well, we can have several Map Displays as tabs in the notebook:
single_window_layout

but! :-) we can also split the notebook in order to see MapDisplays side by side (the previous picture).
But as the notebook is the Content pane, we can pull out neither the whole notebook nor one of the "side by side" Map Displays on the second monitor.

Have I answered your question Vero?  :-) I am not sure if I get it right.

gui/wxpython/main_window/frame.py
lindakladivova: I am not sure if [0] really means primary, it is not written anywhere... if not it could be possible to determine which monitor it is according to GetPosition, but unfortunately, I do not have the second monitor to try it out.

lindakladivova: mmm.. actually, neither on them work (neither wx.DisplaySize nor wx.Display(0).GetGeometry().Get(Size)... I need to have self.Maximize(True) to extend GRASS on the whole screen. I will have a look on it again tomorrow.

veroandreo:
> Well, we can have several Map Displays as tabs in the notebook:
> single_window_layout
> 
> but! :-) we can also split the notebook in order to see MapDisplays side by side (the previous picture).
> But as the notebook is the Content pane, we can pull out neither the whole notebook nor one of the "side by side" Map Displays on the second monitor.
> 
> Have I answered your question Vero? :-) I am not sure if I get it right.

Yes! That's exactly what I meant. IMHO, It looks beautiful! :-)
I understand the "limitation" though of not being able of pulling the the maps to a second monitor

gui/wxpython/main_window/frame.py
```diff
@@ -87,7 +92,7 @@ def __init__(
         id=wx.ID_ANY,
         title=None,
         workspace=None,
-        size=globalvar.GM_WINDOW_SIZE,
+        size=wx.Display(0).GetGeometry().GetSize(),
```
petrasovaa, line 95: This should be `wx.Display().GetGeometry()` which gives you the wx.Rect of the display the application was started on.

But `wx.DisplaySize()` gives me (3520, 1080) when I have 2 displays.

petrasovaa: Also you shouldn't need this here since you Maximize the window later, no?

lindakladivova: Ok. I already get it. :-)  We do not have to solve different sizes of 2 displays at all.

lindakladivova:
@petrasovaa it works for me, could we merge it? If some problem occurs we can deal with that later. This PR is just the initial simple sketch.

PR #1624, opened 2021-06-07
Dockerfiles: remove workaround for grass-session

neteler:
Partial revert of #1623


wenzeslaus:
...anyway, I think the ultimate fix is to build usable images from GRASS source code without a 3rd party library which is not an official dependency. Either GRASS needs grass-session to work and then it should be part of the source code or GRASS should be fixed to work without it.

neteler:
Agreed to make it part of core. Discussion started in https://lists.osgeo.org/pipermail/grass-dev/2021-June/095211.html

PR #1629, opened 2021-06-09
Session initiations functions for Jupyter Notebooks

chaedri:
This PR contains several functions that simplify and shorten the launch of GRASS in Jupyter Notebooks, written as part of my Google Summer of Code project (https://trac.osgeo.org/grass/wiki/GSoC/2021/JupyterAndGRASS). It also contains a Jupyter Notebook (which launches in Binder) to demonstrate usage of the functions.


python/grass/jupyter/init-funcs.py
```diff
@@ -0,0 +1,23 @@
+import os
+import grass.script as gs
+import grass.script.setup as gsetup
+
+
+def init(gisbase, grassdata_filepath, location, mapset):
```
wenzeslaus, line 6: Suggestions for a different name are welcome, but `init` is okay.

As of now, the GISBASE environmental variable needs to be defined before `import grass.script as gs`, so we don't need ask user to pass it here as we can just take it from the environment here and remove the `gisbase` parameter. This could be done in `grass.script.setup.init` too, but let's do it here now since we are sure what is our context here.

I'm not sure what to do with the db/loc/mapset trio, we had discussions in the past about passing just path to mapset versus the three variables. This might be the occasion to try that in a prominent part of the API.

wenzeslaus (changes requested):
Thanks for submitting this. I think it will be good to get this to a working state even with some prototype API and merge it (based on the timing, we may need to consider how it will fit into 8.0 release - currently, the things merged to master will become part of 8.0 release).

Anyway, this will change massively, but having something working in the code may be helpful for testing and developing ideas, so that's what I think should be the aim in this PR.

python/grass/jupyter/init_funcs.py
```diff
@@ -0,0 +1,61 @@
...
+    # We want functions to raise exceptions and see standard output of
+    # the modules in the notebook.
+    gs.set_raise_on_error(True)
+    gs.set_capture_stderr(True)
+
+    # Allow overwrite of existing maps
+    os.environ["GRASS_OVERWRITE"] = "1"
+
+
+def init(grassdata_filepath, location, mapset):
```
wenzeslaus, line 23: `grassdata_filepath` is too long and `filepath` part is not great. You can try:

```
init(path, location=None, mapse=None):
```

`path` will be path to mapset or path to database if location and mapset were provided. To follow the new v8 GUI behavior of using PERMANENT mapset automatically, you could even allow providing only `path` and `location`, but not mapset and also if only `path` is provided and points to an existing location (`grass.grassdb` is your friend), assume mapset=PERMANENT.

python/grass/jupyter/__init__.py
```diff
@@ -0,0 +1,3 @@
+from __future__ import absolute_import
+
+from .init_funcs import *
```
wenzeslaus, line 3: To make the grass.jupyter subpackage really work, you will need to add it to makefiles, see the `Makefile` file in `grass/app` or `grass/grassdb` and also add the name to `Makefile` in the parent directory. It should be just copy-pasting, no real Makefile writing is needed, but let me know.

doc/notebooks/example_notebook.ipynb
```diff
@@ -10,9 +10,13 @@
...
    "pygments_lexer": "ipython3",
-   "version": 3
+   "version": "3.8.3"
   },
-  "orig_nbformat": 2
+  "orig_nbformat": 2,
+  "kernelspec": {
+   "name": "python383jvsc74a57bd07e402764773cbeaa2d2d6f3815f29bb5da9599569c66a0a6d6a02c360a55847f",
+   "display_name": "Python 3.8.3 64-bit ('base': conda)"
+  }
```
wenzeslaus, line 19: You can manually edit the file to revert this change. Something like `git checkout master -- doc/notebooks/example_notebook.ipynb` should bring the old state too if you want to go the Git way, but commit all your changes before, I'm wring the command without checking it.

doc/notebooks/GSoC2021_GRASS_Jupyter_Integration.ipynb
```diff
@@ -0,0 +1,90 @@
...
+    "gj.init(\"$HOME/Documents/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
+    "\n",
+    "# Set Display Settings\n",
+    "gj.display_settings(font=\"romans\") # Default font is \"sans\", let's change it to Romans just for fun!"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "name": "python383jvsc74a57bd07e402764773cbeaa2d2d6f3815f29bb5da9599569c66a0a6d6a02c360a55847f",
```
wenzeslaus, line 72: I have seen issues before with conda specifying its complex kernel name and then the notebook failing elsewhere because the very specific kernel is not available. I'm not sure if it is the case here or not, but we have to check this time to time and will need to modify these by hand if needed. Jupyter Notebooks are unfortunately not as great with Git as one would wish.

doc/notebooks/GSoC2021_GRASS_Jupyter_Integration.ipynb
```diff
@@ -0,0 +1,90 @@
+{
```
wenzeslaus, line 1: The name can be just `jupyter_integration` or `jupyter` or some other simple name.

doc/notebooks/.ipynb_checkpoints/GSoC2021_GRASS_Jupyter_Integration-checkpoint.ipynb
```diff
@@ -0,0 +1,103 @@
+{
```
wenzeslaus, line 1: Delete this file and then if you feel like you can manage some more Git today, find the `.gitignore` in the source code root directory and add `.ipynb_checkpoints` directory there.

wenzeslaus (changes requested):

python/grass/jupyter/init_funcs.py
chaedri: I'm not quite sure what this would look like (or that I understand correctly). Maybe like this:
`def init(path, location=None):
.
.
.
gsetup.init(os.environ["GISBASE"], path, location, mapset="PERMANENT")`

Is that close to what you were thinking? If location is None, I think this would cause an error?

doc/notebooks/example_notebook.ipynb
chaedri: I think I just need to delete this file completely since I'm working on it in the binder-button PR.

.gitignore
```diff
@@ -41,3 +41,5 @@ mswindows/osgeo4w/package/
 mswindows/osgeo4w/setup*.hint
 mswindows/osgeo4w/vc/
 daag*.s
+doc/notebooks/.ipynb_checkpoints/
+
```
wenzeslaus, line 45: Unnecessary change plus it ends an empty line to the end of the file (there is already a line ending at the end of the file, so you are adding a second one).

wenzeslaus (changes requested):
Poor doc/notebooks/example_notebook.ipynb, by deleting the file, you are asking for it to be deleted in the main code. Practically, this will cause a merge conflict with the PR where you are modifying it.

doc/notebooks/example_notebook.ipynb
wenzeslaus: Making it same as on the master branch is what "removes" it from the PR. Deleting in PR is really deleting.

python/grass/jupyter/init_funcs.py
wenzeslaus: Let's just leave this for another PR, so that you get the basic things working quickly and can use them. It's not just about setting some combination of defaults, but rather about running series of checks to figure out what the user meant.

wenzeslaus:
I think the data is in `/home/data/grassdata/` not `/home/grassdata/`, so it does not work in Binder.

```
access: No such file or directory
ERROR: LOCATION </home/grassdata/nc_basic_spm_grass7> not available
```

We can revisit the question of where the data should be, but that's for another PR.

Let's keep the policy of un-computed notebooks in the source code. We can also revisit this later.

Try-in-Binder link:

https://mybinder.org/v2/gh/chaedri/grass/init-funcs?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb

chaedri:
> I think the data is in `/home/data/grassdata/` not `/home/grassdata/`, so it does not work in Binder.
> 
> ```
> access: No such file or directory
> ERROR: LOCATION </home/grassdata/nc_basic_spm_grass7> not available
> ```
> 

I still get an error with this and I'm having trouble figuring out why. Working on that today...

python/grass/jupyter/.ipynb_checkpoints/init_funcs-checkpoint.py
```diff
@@ -0,0 +1,61 @@
+import os
```
wenzeslaus, line 1: When this file got here? Delete it.

doc/notebooks/example_notebook.ipynb
```diff
@@ -1,28 +0,0 @@
-{
```
wenzeslaus, line 1: You really need to do the `git checkout` command to get the original version of this file or copy-paste the content from the _File changes_ here and place it to the empty file. Committing an empty file says I want this file to be empty.

python/grass/jupyter/init_funcs.py
```diff
@@ -0,0 +1,61 @@
...
+    # We want functions to raise exceptions and see standard output of
+    # the modules in the notebook.
+    gs.set_raise_on_error(True)
+    gs.set_capture_stderr(True)
+
+    # Allow overwrite of existing maps
+    os.environ["GRASS_OVERWRITE"] = "1"
+
+
+def init(path, location=None, mapset=None):
```
wenzeslaus, line 23: If you are not going to handle the more fancy parameters here, delete the default values. Both location and mapset are mandatory form grass.script.setup.init, so they need to be mandatory here to when you are just passing them through. The name `path` can stay as is. It makes sense even in this context.

python/grass/jupyter/init_funcs.py
```diff
@@ -0,0 +1,61 @@
+import os
```
wenzeslaus, line 1: Let's rename this file to `setup`. `funcs` that's kind of clear as usually it is functions or functionality. `init.py` as file name would be little confusing in Python due to `__init__.py` file. Two words are fine especially for internal modules, but I don't have a better idea now, `setup` is okay, already used for the `init` function you are wrapping, so let's go with that.

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -79,7 +79,7 @@
    "outputs": [],
    "source": [
     "# Start GRASS Session\n",
-    "gj.init(\"/home/data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
+    "gj.init(\"../../data/grassdata", \"nc_basic_spm_grass7\", \"user1\")\n",
```
wenzeslaus: The red in the `+` line marks the error in the JSON syntax which is what is causing the following message when you open the notebook:

```
File Load Error for jupyter_integration.ipynb
Unreadable Notebook: /home/jovyan/doc/notebooks/jupyter_integration.ipynb NotJSONError('Notebook does not appear to be JSON: \'{\\n "cells": [\\n {\\n "cell_type": "m...')
```

The fix is putting back the `\` for the double quote after `grassdata` which should cause the red-marked `\` to be interpreted correctly.

wenzeslaus (changes requested):

python/grass/jupyter/init_funcs.py
chaedri: When I change the name to setup.funcs.py, I get an error in the Jupyter Notebook:

`ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-3-9dda28f7b25c> in <module>
      1 import grass.script as gs
----> 2 import grass.jupyter as gj

~/dist.x86_64-pc-linux-gnu/etc/python/grass/jupyter/__init__.py in <module>
      1 from __future__ import absolute_import
      2 
----> 3 from .setup.funcs import *

ModuleNotFoundError: No module named 'grass.jupyter.setup'`

Is there a better/correct way to call setup.funcs that doesn't result in this error? Or should be named something different?

doc/notebooks/example_notebook.ipynb
wenzeslaus: I don't see anything expect the last character (last two actually), but there must be something different. Maybe the line endings. Try to get the original file from Git:

```
git checkout upstream/master -- .../the/file/path/file/name.py
```

Some other things than upstream/master may work too. Ideally, _without_ updating to latest master branch first.

python/grass/jupyter/setup.funcs.py
```diff
@@ -0,0 +1,61 @@
...
+import grass.script as gs
+import grass.script.setup as gsetup
+
+
+def _set_notebook_defaults():
+    """
+    This function sets several GRASS environment variables that are
+    important for GRASS to run smoothly in Jupyter. A complete list and
+    description of environment variables can be found here:
+    https://grass.osgeo.org/grass78/manuals/variables.html
```
wenzeslaus, line 12: The whole second sentence can be deleted. Not really relevant for the user. Next programmer will hopefully be able to find that page without specific link. The link itself is adding complexity, it contains `78`, so it needs to be changed for every minor version - in fact it is not matching this version. Alternative would be to link `variables.html` in the source code, but I don't think that's needed in the first place.

python/grass/jupyter/setup.funcs.py
```diff
@@ -0,0 +1,61 @@
+import os
```
wenzeslaus, line 1: Run Pylint, so it tells you you should put a docstring here describing what is in this file (one sentence, it may appear in online documentation, aimed at users of the functions in the file).

Also at the very begging, please add file header with license:

https://trac.osgeo.org/grass/wiki/Submitting/Python
https://trac.osgeo.org/grass/wiki/Submitting/C

I'm little torn between using `#` comments and a docsting for that statement, but given that the docsting is meant for user documentation, `#` comments seem more appropriate for the header information which is aiming at maintainers/authors.

python/grass/jupyter/setup.funcs.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # Allow overwrite of existing maps
+    os.environ["GRASS_OVERWRITE"] = "1"
+
+
+def init(path, location, mapset):
+    """
+    This function initiates a GRASS session and sets GRASS
+    environment variables.
+
```
wenzeslaus, line 27: This might be a good place to mention that it will be overwriting the existing maps of the same name, i.e., as it happens with Python variables - ideally in some way which is 1) not confusing Python users and 2) still warns GRASS users that the default is different than what GRASS usually does. ... or you can put a `# TODO:` comment above the function to get to this later.

wenzeslaus (changes requested):
What I meant to say in the previous round is that `funcs` in the name is not adding any information. The dot in the name is asking for trouble. Just `setup.py` is sufficient.

python/grass/jupyter/init_funcs.py
wenzeslaus: Try to parse my comment again and stop the first sentence right with the first period-space :-)

chaedri: Oh my gosh - I see it now. Thank you.

python/grass/jupyter/setup.funcs.py
chaedri: I wrote a short sentence to that effect. I'm not sure if it's the clearest - I can put a `#TODO:` comment to clarify later too?

wenzeslaus: Good enough for now.

wenzeslaus:
@chaedri You can also try to use the _Resolve conflicts_ button here in the PR on GitHub. There, just keep whatever is marked as master branch, i.e., the long content coming from from the other PR and deleting all what is on this branch.

python/grass/jupyter/__init__.py
```diff
@@ -0,0 +1,3 @@
+from __future__ import absolute_import
```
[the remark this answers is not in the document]
wenzeslaus: I meant to say: not needed! Search for `from __future__ import absolute_import` e.g., https://portingguide.readthedocs.io/en/latest/imports.html#absolute-imports

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -0,0 +1,155 @@
...
...Ho0qNA+ONzNGg0Tt4Jqebuuac3CXELUGsirfOHWG7vcbeaEjsE0hexbZpTP8KfgofwVP4MyBTGmMLuIbbrmsYSK6Z9t/DFdzBlYQtjRV6q11zhN97Bt/BU01Qsk++Eo4yKqlsBYLSZfaDQo2igKPqetssaoXmtF6pope12kSCyZOOdwIFDJiY1Rz+RmycoiGc78QCeu8NJd7gWRVTjZGUGeApPIH/AX4X3vxmw/9IjkfvgG2T7eEdvBeDcfHuI7ZQPZh8j4kYrwwM8JX4+P8BBzx0NtE2xXs4ktfUlKc06ifO7Mwj2lsLrCQjtoysxvvdTLOd/F57xNrPrzXc5K30gIIBhYO0IQeVTnuyUmpIrywjVrU9MwFPuDG42Ly4OdqNQxtbDuq6EAeO91w0iw906W8tzme2yCZlkmVIQJyKJy4JJvTvtZCC9LzCvnKmMaJtcOyZnWYGO6g+EgdJPM5MJS9h2BEiIVqOvuc0skws5uMHH/3kU4mG/WlTMtqratQ0T7IyQsuPT+6c4lKk3tByVg9cqv6xIc8Wc7Sus4HVI3e7uU9kPxJD9p6r6t7UM/7YhhYk2XFklYCJMFs1Qdql0seIoTwQmOz8vZMbqsoldzJR80Y8HlbnssVYfGrQTnXHPfFYagnhouoJqwBnFz01LF1jtpskZzT+fCVwJFJCrlhl12z8lcq3zwT2tILoQagJnbezX6J8MFD5NvBEJN4PdiLetHAnqLd+7CDTloYe9IgBA6fR7/kJXq1an6AlbXBrO5mqtC92tZtWTr6VvkeAd533tfvYNC+rkrKiEQK5q/4G2ZykEEFVqktd5SwPaq6ZE2RiZAumm2FBn3/Pq8ifD3wuRRYTz/r38LvwFj5UqPPf2gcaCe4a3so1tgen5fpF1qkqwbSFamzwWQFK0L+/hWsR+EMn3tCKDv163nF5wRJcxKPCa/n3DG/hFp7ACV7DPXwjbtpJ7jnJpjYdsx1cJH/i/0iDb34oQPzYjkfvgO24knfkIZ6zCd9jbsksyMow305TmFmkiPriNwDRl3ADdxJEPznEXeXI7dOAVewhQ2ujBv81mHHH2ZVbU2dfGu5kGN1Lha6zmthvYKYQE6X6YN0grpAzrtdu1EG7CHMbC5PlW9e8t2wVCLHWEmhiBSMxlFSztVHtDDMlWJ90JQTvI4qZYWQx7a2zf3IZGWcmizB283m
```
wenzeslaus, line 95: Let's follow the policy of having un-computed notebooks in this repo. The point is that you can compute them easily, right? They are different from the standard documentation (where we have images) and from the notebooks linked from the website which should really look as nice as possible, so they include images.

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -0,0 +1,155 @@
...
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Start GRASS Session\n",
+    "gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
+    "\n",
+    "# Set Display Settings\n",
+    "gj.display_settings(font=\"romans\", driver=\"cairo\") # Default font is \"sans\", let's change it to Romans just for fun!"
```
wenzeslaus, line 85: I don't think `font="romans"` does what you expect. It uses the default font in GRASS display drivers which is not high quality. Perhaps `font="serif"` would give a good result. However, `grass.jupyter` is about reasonable defaults, so why to set anything at all in this basic example and not go with `gj.display_settings()`? This API and example will change, but this comment applies to the new API as well.

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -0,0 +1,155 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Start GRASS Session\n",
+    "gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
+    "\n",
+    "# Set Display Settings\n",
```
wenzeslaus, line 84: Given my comment for the line below, this would be _Set default display settings_.

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -0,0 +1,155 @@
...
+      "        path - path to grass databases\n",
+      "        location - name of GRASS location\n",
+      "        mapset - name of mapset within location\n",
+      "    \n"
+     ]
+    }
+   ],
+   "source": [
+    "# Let's check our import by printing the __doc__ for init()\n",
+    "print(gj.init.__doc__)"
```
wenzeslaus, line 72: I like this. `help(gj.init)` is perhaps more elegant. You can also consider more Jupyter-ish `gj.init?`.

wenzeslaus (changes requested):
Almost there!

wenzeslaus:
This is now available for general testing and it will be merged soon. Notably, the API will be changed significantly, but getting feedback about that now will be helpful in the next phase.

No-install no-download (some-wait) try-in-binder link:

https://mybinder.org/v2/gh/chaedri/grass/init-funcs?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -0,0 +1,113 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Start GRASS Session\n",
+    "gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
+    "\n",
+    "# Set default display settings\n",
+    "gj.display_settings() # Default font is \"sans\", let's change it to Romans just for fun!"
```
wenzeslaus, line 69: No comment needed here anymore after the last change.

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -0,0 +1,113 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "gisbase = subprocess.check_output([\"grass\", \"--config\", \"path\"], text=True).strip()\n",
+    "os.environ['GISBASE'] = gisbase\n",
```
wenzeslaus, line 35: Since you are doing changes anyway, what about changing all single quotes in the notebook to double quotes to follow Black? Looks much worse in the raw JSON, but it will be better when viewing notebook. No need to wonder if there is a difference or what is the style used.

python/grass/jupyter/__init__.py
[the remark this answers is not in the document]
wenzeslaus: This still needs to be removed. We don't keep Python 2 compatibility anymore.

wenzeslaus (changes requested):

doc/notebooks/jupyter_integration.ipynb
wenzeslaus: Use double quotes, not single quotes.

wenzeslaus (changes requested):
The quotes are consistent, +1, but it is the other kind. Then, I think we are done.

chaedri:
Oops - I'm reading your comments above correctly now. They are double quotes now.

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -78,22 +78,21 @@
...
    ]
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
-   "language": "python",
-   "name": "python3"
+   "name": "python3",
+   "display_name": "Python 3.8.3 64-bit ('base': conda)"
```
wenzeslaus: I'm not sure how to handle those in the future, but right now, the best practice seem to keep here something simple without the conda complexities (if for no other reason, it is meant to be used also without conda), so revert this piece.

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -105,7 +104,10 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.5"
+   "version": "3.8.3"
+  },
+  "interpreter": {
+   "hash": "7e402764773cbeaa2d2d6f3815f29bb5da9599569c66a0a6d6a02c360a55847f"
```
wenzeslaus: As above, revert this part.

wenzeslaus (changes requested):

wenzeslaus (approved):
Great. Works great for me with the [Binder link](https://mybinder.org/v2/gh/chaedri/grass/init-funcs?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb). The CI fail is unrelated.

PR #1634, opened 2021-06-11
r.univar: Add parallel support

aaronsms:
This PR implements parallelization for all r.univar options except for when "extended (stats)" flag is set to true. This is because there are dynamic allocation and sorting involved, which is trickier to implement, but this could be a work for the future.

Checklists before merging:
- [x] code review
- [x] CI passes
- [x] run tests without OpenMP (runs in CI)
- [x] rebase to main
- [x] run valgrind


raster/r.univar/testsuite/benchmark_r_univar.py
```diff
@@ -0,0 +1,59 @@
+"""Benchmarking of r.univar
```
wenzeslaus, line 1: test_r_univar.py should not be deleted. This benchmark_r_univar.py file is extra.

wenzeslaus (commented):
Just some initial things:

Fails on Ubuntu 18.04, but not 20.04 in CI, too old version of OpenMP?

```
2021-06-12T13:19:51.6016965Z r.univar_main.c: In function ‘process_raster_threaded’:
2021-06-12T13:19:51.6019743Z r.univar_main.c:500:11: error: ‘value_sz’ is predetermined ‘shared’ for ‘shared’
2021-06-12T13:19:51.6020739Z      shared(stats, fd, fdz, raster_row, zoneraster_row, n, sum, sumsq, sum_abs, min, max, size, region, \
2021-06-12T13:19:51.6021406Z            ^
2021-06-12T13:19:51.6022224Z r.univar_main.c:500:11: error: ‘map_type’ is predetermined ‘shared’ for ‘shared’
2021-06-12T13:19:51.6025019Z r.univar_main.c:500:11: error: ‘n_zones’ is predetermined ‘shared’ for ‘shared’
2021-06-12T13:19:51.6027487Z r.univar_main.c:500:11: error: ‘cols’ is predetermined ‘shared’ for ‘shared’
2021-06-12T13:19:51.6028607Z r.univar_main.c:500:11: error: ‘rows’ is predetermined ‘shared’ for ‘shared’
```

raster/r.univar/r.univar_main.c
```diff
@@ -52,6 +57,14 @@ void set_params()
 	_("Percentile to calculate (requires extended statistics flag)");
     param.percentile->guisection = _("Extended");
     
+    param.threads = G_define_option();
+    param.threads->key = "nprocs";
```
ninsbl, line 61: Should we add a "nproc" standard option for the parser (like e.g G_OPT_M_NPROCS)?
`nprocs` is used in several Python scripts as well. So having a standard option could secure a harmonized way to handle such an option...

HuidaeCho: @ninsbl That's a good idea.

HuidaeCho: Please check #1644.

raster/r.univar/testsuite/benchmark_r_univar.py
HuidaeCho: Fails on my Linux machine too with the same message. What is the min version of GCC that this parallel code supports?

HuidaeCho: * https://stackoverflow.com/a/13770822/16079666
* https://stackoverflow.com/a/13203075/16079666

HuidaeCho: FYI, gcc 5.5.0

HuidaeCho: OK, all those variables are `const` and removing them from `shared` worked. I think `const`s are shared by default.

raster/r.univar/r.univar_main.c
aaronsms: Right, I'll modify the code using the standard option.

raster/r.univar/testsuite/benchmark_r_univar.py
aaronsms: I think I must've accidentally deleted the file when separating the scripts to respective directories, will fix it.

raster/r.univar/r.univar_main.c
```diff
@@ -202,7 +213,24 @@ int main(int argc, char *argv[])
...
 
+    sscanf(param.threads->answer, "%d", &threads);
+    if (threads < 1)
+    {
+      G_warning(_("<%d> is not valid number of threads. Number of threads will be set to <%d>"),
+      threads, abs(threads));
+      threads = abs(threads);
+    }
+#if defined(_OPENMP)
+    if (!param.extended->answer) {
```
marisn, line 224: If I understand it right, reason for not using parallel version when extended is set, is lack of parallelization in calculation of extended statistics. If it is so, just add a comment here.
```suggestion
    /* Calculation of extended statistics is not parallelized yet */
    if (!param.extended->answer) {
```

raster/r.univar/r.univar_main.c
```diff
@@ -379,7 +408,215 @@ process_raster(univar_stat * stats, int fd, int fdz, const struct Cell_head *reg
...
+		if (n_zones)
+		    zptr++;
+		continue;
+	    }
+
+	    val = ((map_type == DCELL_TYPE) ? *((DCELL *) ptr)
+			  : (map_type == FCELL_TYPE) ? *((FCELL *) ptr)
+			  : *((CELL *) ptr));
+
+	    sum[id][zone] += val;
```
marisn, line 542: Would it be worth to go here with Kahan–Babuska algorithm?

marisn (commented):
My OpenMP knowledge is too weak to judge if this is the right approach :-(

HuidaeCho:
I have [i5-7300U](https://ark.intel.com/content/www/us/en/ark/products/97472/intel-core-i5-7300u-processor-3m-cache-up-to-3-50-ghz.html) with 2 cores and 4 threads on my laptop. What does this module do with `nprocs` > 4? The benchmark script runs fine with up to 12 threads, but I thought my CPU can only do up to 4 threads... I might be wrong. interestingly, more threads didn't *always* mean faster. See my results below. Is it possible to determine the number of max threads in the code and print a warning if `nprocs` is greater than that? Also, please consider implementing a fallback terminal size when redirecting benchmarking outputs; `os.get_terminal_size()` raises an exception `OSError: [Errno 25] Inappropriate ioctl for device` on `python3 benchmark_r_univar.py >& benchmark_r_univar.log`.

```
r.univar map=elevation,elevation,elevation,elevation,elevation,elevation,elevation,elevation,elevation,elevation percentile=90.0 nprocs=1 separator=pipe -g

Benchmark with 1 thread(s)...
Result - 0.1885438919067383s

Benchmark with 2 thread(s)...
Result - 0.1510328769683838s

Benchmark with 3 thread(s)...
Result - 0.18701410293579102s

Benchmark with 4 thread(s)...
Result - 0.2988687038421631s

Benchmark with 5 thread(s)...
Result - 0.14902639389038086s

Benchmark with 6 thread(s)...
Result - 0.16875758171081542s

Benchmark with 7 thread(s)...
Result - 0.16923255920410157s

Benchmark with 8 thread(s)...
Result - 0.20600790977478028s

Benchmark with 9 thread(s)...
Result - 0.1692127227783203s

Benchmark with 10 thread(s)...
Result - 0.13888721466064452s

Benchmark with 11 thread(s)...
Result - 0.16875271797180175s

Benchmark with 12 thread(s)...
Result - 0.17763447761535645s

r.univar map=elevation,elevation,elevation,elevation,elevation,elevation,elevation,elevation,elevation,elevation zones=basin_50K percentile=90.0 nprocs=1 separator=pipe -g

Benchmark with 1 thread(s)...
Result - 0.27266292572021483s

Benchmark with 2 thread(s)...
Result - 0.17659420967102052s

Benchmark with 3 thread(s)...
Result - 0.2019331455230713s

Benchmark with 4 thread(s)...
Result - 0.42590484619140623s

Benchmark with 5 thread(s)...
Result - 0.2039196491241455s

Benchmark with 6 thread(s)...
Result - 0.1847921848297119s

Benchmark with 7 thread(s)...
Result - 0.1929023265838623s

Benchmark with 8 thread(s)...
Result - 0.22778925895690919s

Benchmark with 9 thread(s)...
Result - 0.3162210464477539s

Benchmark with 10 thread(s)...
Result - 0.2332921504974365s

Benchmark with 11 thread(s)...
Result - 0.2835871696472168s

Benchmark with 12 thread(s)...
Result - 0.24240994453430176s
```

wenzeslaus:
I tested with 4 cores - 8 threads processor and made some additions to the benchmark script. Here is a test up to nprocs=16, 10 runs, and two additional scenarios. The data is still too small I think, so that's still a todo.

Screenshot from 2021-06-19 15-57-35

wenzeslaus:
> What does this module do with `nprocs` > 4? ... Is it possible to determine the number of max threads in the code and print a warning if `nprocs` is greater than that?

I don't think that needs a warning. You need to explicitly say you want n threads and likely you know the number cores/threads on the machine you are using or a look into process manager or specs can tell you. So, warning is really no necessary since likely that's what you meant. Additionally, what will be the result (improvement or degradation) depends on the specific setup, so the warning anyway cannot claim it will be worse.

Well, now, automatically detecting number of cores with nprocs=auto or by default that's a different story!

HuidaeCho:
> I don't think that needs a warning. You need to explicitly say you want n threads and likely you know the number cores/threads on the machine

I know I have 4 threads, but I don't know what it's doing with 12 threads when I only have 4. What does it even mean? It needs to be explained at least.

HuidaeCho:
> I tested with 4 cores - 8 threads processor and made some additions to the benchmark script. Here is a test up to nprocs=16, 10 runs, and two additional scenarios. The data is still too small I think, so that's still a todo.
> 
> Screenshot from 2021-06-19 15-57-35

That looks nice. Interesting it's not happening on my machine.

HuidaeCho:
`opm_get_num_threads`?

raster/r.univar/r.univar_main.c
aaronsms: I think we can add this, but it is not the original behavior before this parallelization, perhaps we can do this in the next PR?

petrasovaa:
r.univar on a desktop with 28 cores, slope map, 16832104560 cells:
results
Any idea what is going on?

HuidaeCho:
@aaronsms Please check this first module. That jump at 14 processes (half of the 28 cores) is interesting. @petrasovaa What is the name of the CPU? Does it have 28 cores with 56 threads or 14 cores with 28 threads (usually 2 threads per core)?

HuidaeCho:
OK, in the weekly meeting with @aaronsms, @petrasovaa confirmed that it has 14 physical cores and 28 threads. Maybe, it's related to this fact.

raster/r.univar/r.univar_main.c
```diff
@@ -379,7 +409,215 @@ process_raster(univar_stat * stats, int fd, int fdz, const struct Cell_head *reg
...
+    double **min = G_malloc(threads * sizeof(double*));
+    double **max = G_malloc(threads * sizeof(double*));
+    size_t **size = G_malloc(threads * sizeof(double*));
+
+    for (int i = 0; i < threads; i++) {
+        n[i] = G_calloc(n_alloc, sizeof(size_t));
+        sum[i] = G_calloc(n_alloc, sizeof(double));
+        sumsq[i] = G_calloc(n_alloc, sizeof(double));
+        sum_abs[i] = G_calloc(n_alloc, sizeof(double));
+        size[i] = G_calloc((n_alloc), sizeof(size_t));
```
marisn, line 479: There is no need for round brackets around n_alloc.

marisn (changes requested):
I could not run the benchmark as it ended with OOM in r.surf.fractal with 100000000 cells in a region. Is there something that can be done to adapt to the RAM size of machine instead of failing with OOM and thus losing all results?

raster/r.univar/r.univar_main.c
aaronsms: Alright, will fix that.

wenzeslaus:
> Is there something that can be done to adapt to the RAM size of machine instead of failing with OOM and thus losing all results?

Unlike for the test, which should just run everywhere (although skipped in some cases), for benchmarks, we don't have a any portability conceptualized. For example, the grass.benchmark package helps you write a benchmark, but does not tell you how to write it (i.e., you can write a benchmark without using grass.benchmark and that's perfectly fine). So far, we were adapting the benchmark scripts to the test we wanted to do. Suggestions welcome.

As for this particular case (OOM), do you envision the Python code to do some heuristics for memory requirements of r.surf.fractal versus your size of your RAM and do benchmarks according to that?

marisn:
> 
> As for this particular case (OOM), do you envision the Python code to do some heuristics for memory requirements of r.surf.fractal versus your size of your RAM and do benchmarks according to that?

Heuristics would be good, but might need too much work. Probably the easiest solution would be to add try: except: around r.surf.fractal calls to not fail if OOM situation is encountered.

petrasovaa:
> r.univar on a desktop with 28 cores, slope map, 16832104560 cells:
> results
> Any idea what is going on?

Repeated my benchmark with latest code, this looks much better!
results

aaronsms:
> > r.univar on a desktop with 28 cores, slope map, 16832104560 cells:
> > results
> > Any idea what is going on?
> 
> Repeated my benchmark with latest code, this looks much better!
> results

@petrasovaa yea I suspect the issues previously was due to memory bandwidth issues or false sharing due to cache inefficiencies. So I made some effort to refactor such that threads now are less likely to share the same cache for accessing variables, to achieve higher cache hits. So I won't no longer include the issues regarding worsening performance with threads overloading. I believe we need to check this for other modules as well.

raster/r.univar/benchmark/benchmark_r_univar.py
```diff
@@ -0,0 +1,46 @@
+"""Benchmarking of r.univar
+
+@author Aaron Saw Min Sern
+"""
+
+from grass.pygrass.modules import Module
+from subprocess import DEVNULL
+
+import grass.benchmark as bm
```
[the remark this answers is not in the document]
aaronsms: yeap I have rebased it.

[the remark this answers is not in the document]
aaronsms: yeap I have rebased it on master. now I have the script to use r.random.surface to generate if encounter OOM for r.surf.fractal.

petrasovaa:
With extended statistics (with nprocs=1) valgrind is getting mad:
```
==90673== Invalid write of size 2
==90673==    at 0x4842B33: memmove (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==90673==    by 0x10C225: process_raster._omp_fn.0 (r.univar_main.c:441)
==90673==    by 0x4A988E5: GOMP_parallel (in /usr/lib/x86_64-linux-gnu/libgomp.so.1.0.0)
==90673==    by 0x10B6F8: process_raster (r.univar_main.c:342)
==90673==    by 0x10B1FC: main (r.univar_main.c:240)
==90673==  Address 0xee42980 is 0 bytes after a block of size 36,000 alloc'd
==90673==    at 0x483DFAF: realloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==90673==    by 0x4899F82: G__realloc (alloc.c:126)
==90673==    by 0x10C344: process_raster._omp_fn.0 (r.univar_main.c:431)
==90673==    by 0x4A988E5: GOMP_parallel (in /usr/lib/x86_64-linux-gnu/libgomp.so.1.0.0)
==90673==    by 0x10B6F8: process_raster (r.univar_main.c:342)
==90673==    by 0x10B1FC: main (r.univar_main.c:240)
```

petrasovaa (approved):

PR #1461, opened 2021-03-14
Python scripts: remove UTF-8 coding line

neteler:
UTF-8 is the default with Python 3, no need for the coding line.

Ref: https://github.com/OSGeo/grass/pull/1449#discussion_r593748122

PR #1643, opened 2021-06-13
i.tasscap: utf-8 encoding for non-ASCII chars

HuidaeCho:
This PR addresses:
```bash
$ xgettext i.tasscap.py 
xgettext: Non-ASCII comment at or before i.tasscap.py:49.
          Please specify the source encoding through --from-code or through a comment
          as specified in http://www.python.org/peps/pep-0263.html.
```


nilason:
Why not use `--from-code=UTF-8`? I see no reason to start re-adding this coding line for no good reason, py3-files are by default (interpreted as) utf-8.

nilason:
In case you it slipped your attention see #1461.

neteler:
> Why not use `--from-code=UTF-8`? I see no reason to start re-adding this coding line for no good reason, py3-files are by default (interpreted as) utf-8.

At time we have a compilation issue in the snapshot compilation log on grass.osgeo.org (Debian):
See: https://grass.osgeo.org/grass80/binary/linux/snapshot/build.log.txt

```
xgettext: Non-ASCII comment at or before ../scripts/i.tasscap/i.tasscap.py:49.
          Please specify the source encoding through --from-code or
through a comment
          as specified in http://www.python.org/peps/pep-0263.html.
```

Indeed, in line 49 there is a French accent in "Académie". I also thought is was ok with UTF-8 but apparently not entirely.

HuidaeCho:
> Indeed, in line 49 there is a French accent in "Académie". I also thought is was ok with UTF-8 but apparently not entirely.

Weird. I think adding from-code to xgettext should be the right fix at this point.

HuidaeCho:
One more message:
https://github.com/OSGeo/grass/blob/55079f3bdbe36a2d26fa690d0b708cf2398aac24/scripts/r.plane/r.plane.py#L6
```
xgettext: Non-ASCII comment at or before ../scripts/r.plane/r.plane.py:6.
          Please specify the source encoding through --from-code or through a comment
          as specified in http://www.python.org/peps/pep-0263.html.
```

HuidaeCho:
Closed by #2061.

PR #1644, opened 2021-06-13
Add a standard option for nprocs= (G_OPT_M_NPROCS)

HuidaeCho:
This PR implements the standard G_OPT_M_NPROCS option for the number of threads for parallel computing. See https://github.com/OSGeo/grass/pull/1634#discussion_r650575434.

### DIFF
--- a/include/grass/gis.h
+++ b/include/grass/gis.h
@@ -99,12 +99,12 @@ static const char *GRASS_copyright __attribute__ ((unused))
 #define U_DEGREES	8
 #define U_USFEET	9
 /* Temporal units from the datetime library */
-#define U_YEARS         DATETIME_YEAR   
-#define U_MONTHS        DATETIME_MONTH  
-#define U_DAYS          DATETIME_DAY    
-#define U_HOURS         DATETIME_HOUR   
-#define U_MINUTES       DATETIME_MINUTE 
-#define U_SECONDS       DATETIME_SECOND 
+#define U_YEARS         DATETIME_YEAR
+#define U_MONTHS        DATETIME_MONTH
+#define U_DAYS          DATETIME_DAY
+#define U_HOURS         DATETIME_HOUR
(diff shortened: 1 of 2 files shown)


ninsbl (approved):
Thanks for following up so swiftly. One question: should it be backported? 
It is a new feature but very little invasive.
With regards to usage in AddOns backporting may be an advantage to safeguard AddOn-compatibility across versions...

HuidaeCho:
> Thanks for following up so swiftly. One question: should it be backported?
> It is a new feature but very little invasive.
> With regards to usage in AddOns backporting may be an advantage to safeguard AddOn-compatibility across versions...

I think it makes sense to backport it.

HuidaeCho:
Backport PR in #1648

wenzeslaus:
Here, many tests fail which don't fail on master. They fail on master in the commit subsequent to this one (there is only one at this point) and they fail in an PR which merged the current master (so it fails the same way in 4 different instances and tracks lead here).

Also, the code says "Values also need to be added to general/g.parser/standard_option.c".

HuidaeCho:
Do only tests fail? The core compiles fine on my machine. Let me add it to standard_option.c.

HuidaeCho:
How can I reproduce it?

wenzeslaus:
> Do only tests fail?

You can see that when you click the red x ❌. In the overview, it will mark the failing jobs, when you click it, it will go to the end of the failing step in that job. Tests fail slow (i.e., they don't fail fast) so you may need to scroll up to see which test(s) are failing.

> The core compiles fine on my machine.

Many things need to run to see if they work. That's why we have the tests, so hooray for the tests! :tada: We need more of them!

> How can I reproduce it?

Pick what you need from: https://grass.osgeo.org/grass80/manuals/libpython/gunittest_running_tests.html

PR #1645, opened 2021-06-14
Fix source code URL when addon path starts with root path

HuidaeCho:
Fix https://github.com/OSGeo/grass-addons/issues/546

* Core path: `/home/user/usr/grass/grass`
* Addon path: `/home/user/usr/grass/grass-addons`

Addon path starts with core path, but that doesn't mean I'm inside the core.


neteler:
Honestly, hard to judge for me. But if it works then fine. I just wonder why the issue didn't pop up earlier...?

HuidaeCho:
I think the issue was already there. Just noticed it now.

PR #1646, opened 2021-06-15
wxGUI: refactoring: build display status bar based on wx.StatusBar widget

lindakladivova:
This PR redesigning the form of display map status bar implementation.
It was originally created directly using CreateStatusBar function, which is, however, directly connected to the wx.Frame widget. 
For further implementation of the SingleLayout, we need to replace MapFrame widget by MapPanel widget. 
MapPanel widget does not support CreateStatusBar function.

Reimplemented status bar in a Map Display for a better context:
Map Display 1_024

Original status bar:
Map Display 1_023

### DIFF
--- a/gui/wxpython/mapdisp/frame.py
+++ b/gui/wxpython/mapdisp/frame.py
@@ -223,7 +223,8 @@ class MapFrame(SingleMapFrame):
             .CloseButton(False)
             .DestroyOnClose(True)
             .ToolbarPane()
-            .Dockable(False),
+            .Dockable(False)
+            .Gripper(False),
         )
         self._mgr.Update()
 
@@ -282,7 +283,6 @@ class MapFrame(SingleMapFrame):
         statusbarpanel = wx.Panel(self, size=wx.DefaultSize, style=wx.SUNKEN_BORDER)
         statusbar = wx.StatusBar(statusbarpanel, id=wx.ID_ANY)
         statusbar.SetFieldsCount(4)
-        statusbar.SetMinHeight(30)
         statusbar.SetStatusWidths([-5, -2, -1, -1])
 
         container = wx.BoxSizer(wx.VERTICAL)
--- a/gui/wxpython/mapdisp/frame.py
+++ b/gui/wxpython/mapdisp/frame.py
@@ -312,17 +312,12 @@ class MapFrame(SingleMapFrame):
 
     def ShowStatusbar(self, show):
         """Show/hide statusbar and associated pane"""
-        self.statusbarManager.Show(show)
-        if show:
-            self._mgr.GetPane("statusbar").Show()
-        else:
-            self._mgr.GetPane("statusbar").Hide()
-
+        self._mgr.GetPane("statusbar").Show(show)
         self._mgr.Update()
 
     def IsStatusbarShown(self):
         """Check if statusbar is shown"""
-        return self.statusbarManager.IsShown()
+        return self._mgr.GetPane("statusbar").IsShown()
 
     def SetStatusText(self, *args):
         """Overide wx.StatusBar method"""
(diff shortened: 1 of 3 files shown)


gui/wxpython/mapdisp/frame.py
```diff
@@ -266,9 +280,16 @@ def CreateStatusbar(self):
         )
 
         # create statusbar and its manager
-        statusbar = self.CreateStatusBar(number=4, style=0)
-        statusbar.SetMinHeight(24)
+        statusbarpanel = wx.Panel(self, size=wx.DefaultSize, style=wx.SUNKEN_BORDER)
```
petrasovaa, line 283: Get rid of the panel, just work with the statusbar directly.

gui/wxpython/mapdisp/frame.py
```diff
@@ -212,6 +212,20 @@ def __init__(
...
+            statusbarpanel,
+            wx.aui.AuiPaneInfo()
+            .Bottom()
+            .MinSize(30, 30)
+            .Fixed()
+            .Name("statusbar")
+            .CloseButton(False)
+            .DestroyOnClose(True)
+            .ToolbarPane()
+            .Dockable(False)
```
petrasovaa, line 226: add PaneBorder(False)

gui/wxpython/mapdisp/frame.py
```diff
@@ -453,7 +475,7 @@ def AddNviz(self):
         self._giface.WriteCmdLog(
             _("Starting 3D view mode..."), notification=Notification.HIGHLIGHT
         )
-        self.SetStatusText(_("Please wait, loading data..."), 0)
+        self.statusbar.SetStatusText(_("Please wait, loading data..."), 0)
```
petrasovaa, line 478: I suggest to override SetStatusText function (meaning create your own where you call self.statusbar.SetStatusText)

petrasovaa (changes requested):
When you right click on map display and select Hide statusbar, it's not working because the pane itself is not hidden. Currently the hiding/showing is in SbManager, but it may be better to keep SbManager for managing the content and the hiding/showing could be done by the frame directly (it knows about the pane unlike the SbManager). So create ShowStatusbar and IsStatusbarShown functions in MapFrame that show/hide the pane.

gui/wxpython/mapdisp/frame.py
```diff
@@ -293,6 +308,25 @@ def CreateStatusbar(self):
...
+            self._mgr.GetPane("statusbar").Show()
+        else:
+            self._mgr.GetPane("statusbar").Hide()
+
+        self._mgr.Update()
+
+    def IsStatusbarShown(self):
+        """Check if statusbar is shown"""
+        return self.statusbarManager.IsShown()
+
```
petrasovaa, line 326: It's enough to show/hide the pane. Also I would remove the unused show functions from sbManager, but check if they are really unused.

```suggestion
    def ShowStatusbar(self, show):
        """Show/hide pane associated with statusbar"""
        self._mgr.GetPane("statusbar").Show(show)
        self._mgr.Update()

    def IsStatusbarShown(self):
        """Check if statusbar is shown"""
        return self._mgr.GetPane("statusbar").IsShown()

```

petrasovaa (changes requested):

petrasovaa:
> @petrasovaa Not sure why there are these slant stripes now:
> Výstřižek
> 
> I first thought it is something related to the pane resizing but it looks more like related to the whole frame... (it disappears when Map Window is maximized). I noticed we have the same stripes in the bottom of the mainframe window as well. Do you have any suggestion about that?
> 
I noticed it but I am not sure either. It's probably ok.

gui/wxpython/mapdisp/frame.py
lindakladivova: I think it works fine now.

petrasovaa (approved):

PR #1652, opened 2021-06-16
g.parser: Add G_OPT_M_NPROCS

HuidaeCho:
Address #1644?


wenzeslaus (approved):
Yes, that's it. Right place and the tests are running. ✔️

PR #597, opened 2020-05-07
r.neighbors: add option for exponential weighting

ninsbl:
r.neighbors has already a nice option to weigh cells within the neighborhood using a mathematical function (2D gaussian distribution).

This PR adds an option for exponential weighting, that has applications e.g. in spatial ecology. This is a rather simple implementation with some copy and paste from the gaussian function.

It would probably be preferable to have a more generic approach that allows also for other distribution functions (see e.g. en.wikipedia.org/wiki/List_of_probability_distributions) to be added without increasing the number of options too much?

The PR also addresses #593 (but that can be easily separated out from this PR if preferable)

PR #1654, opened 2021-06-17
r.neighbors: Add parallel support

aaronsms:
- Update Makefile
- Refactor inputs to use pointers
- Refactor to use segment libraries
- Add parallel region
- Add parallel test cases

All input and output variables are transformed into pointers to support parallel execution. There is no change in the underlying algorithm except of the introduction of segmented temporary files to allow for random access write. So the extra memory usage will be scaled according to the number of threads used plus a constant overhead from the usage of segmented files. I think we should introduce a memory parameter to allow users to control the how much memory to be used for the segment. Right now, it is just a fixed constant, which is to be changed.

This PR contains an addition of a test file which contains the parallel version of the original sequential test cases. Upon addition of benchmark script, this will be combined into the original test file to avoid code duplication. On my machine, there is some speedup of ~2x on 4 threads, though this is not accurate without the benchmark.

I am thinking to refactor my code to encapsulate the usage of Segments.


ninsbl:
Nice!
Could this be coordinated with: #597
Sorry for not having committed before you started on r.neighbors...

aaronsms:
> Nice!
> Could this be coordinated with: #597
> Sorry for not having committed before you started on r.neighbors...

No worries, yeap will incorporate the changes once #597 is merged.

aaronsms:
- [ ] Fix CentOS build fail due to C98 incompatibility (will take note of this in the future)

test_pygrass_modules_interface_doctests.py was failing, I am guessing this has to do with the newly added test cases. 20.04 build fail is incidental, due to connection issues to repository.

HuidaeCho:
@aaronsms Please consider using the standard option `G_OPT_M_NPROCS` introduced in #1644.

aaronsms:
> @aaronsms Please consider using the standard option `G_OPT_M_NPROCS` introduced in #1644.

Noted! Will use that for future modules.

aaronsms:
I will close this PR in favor of #1724.

PR #1662, opened 2021-06-19
libpython: Make parallel region test somewhat more robust

wenzeslaus:
Test introduced in #638 uses raster maps to determine if the subprocess was
successfully executed. However, in GitHub Actions this fails randomly
with a random number of rasters missing.

This PR makes the test wait for couple seconds before counting the rasters
in case the file system needs to catch up. If there is an actual error,
waiting won't fix it, so the test will still do its job.

If running and rerunning the tests here is successful, it suggests
that it is a good fix at least in the sense of accomodating
the given environment.

Alternative fix for the test would be to replace Bash by Python
which would allow leaving out the raster creation check (which is
replacing return code check here) and provide more control.

On one hand, we want to make the test run, on the other,
we want to have the actual code robust, but maybe writing 50
rasters in parallel in a small VM in GitHub Actions is not
a use case we need to cover in code and thus modifying the test is enough.

### DIFF
--- a/python/grass/script/testsuite/test_parallel_temp_region.sh
+++ b/python/grass/script/testsuite/test_parallel_temp_region.sh
@@ -56,6 +56,9 @@ fi
 MAP_PARALLEL_PATTERN="${MAP_PARALLEL}_*_size_*_nesting_*"
 MAP_NEST_PATTERN="${MAP_NEST}_*_size_*_nesting_*"
 
+# After each wait, sleep for additional n seconds.
+SLEEP_AFTER_WAIT=5
+
 # Remove maps at exit.
 cleanup () {
     EXIT_CODE=$?
@@ -99,6 +102,7 @@ do
 done
 
 wait
+sleep $SLEEP_AFTER_WAIT
 
 EXPECTED=$NUM_PARALLELS
 NUM=$(g.list type=raster pattern="${MAP_PARALLEL_PATTERN}" mapset=. | wc -l)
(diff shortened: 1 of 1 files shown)


wenzeslaus:
With 5s wait, still:

```
Pure parallel test: Got 49 but expected 50 maps
```

wenzeslaus:
Update and follow up from #1272:

> > @marisn: I can not reproduce the issue locally on Ubuntu 20.10.
> 
> @wenzeslaus: And I don't expect you would. My guess is it is file system related. I executed that test on my machine many times when working on #638.

Now I think, it is actually much simpler than my initial guess which was influenced too much by considering that the return code may be zero (which is why the map presence is tested in the first place). The conditions are very similar to the temporal tests which are failing due to creating common directories in a mapset, so this is the most likely cause here. I'm trying to address that in #1681.

The reason we were not able to reproduce this locally is that these test are using an empty mapset, but just repeating test locally usually means that you create all these sub-directories in the first run even if you use an empty mapset, so no issue will arise when running the test again.

wenzeslaus:
The two observed errors (in various CI runs; now actually visible) are creation of raster directories (at least `cell`) and a host-specific `.tmp/nodeid` temporary directory. Both of these should be addressed in #1681, so closing this, although some better error handling may still be a good idea for this test.

```
ERROR: Unable to make mapset element cell (/home/runner/nc_spm_full_v2alpha2/__python_grass_script_test_parallel_temp_region/cell): File exists
ERROR: Unable to make mapset element .tmp/fv-az180-858 (/home/runner/nc_spm_full_v2alpha2/__python_grass_script_test_parallel_temp_region/.tmp): File exists
```

PR #1663, opened 2021-06-19
r.proj: Add test cases

aaronsms:
- [x] Add test cases for r.proj


raster/r.proj/testsuite/test_r_proj.py
```diff
@@ -0,0 +1,302 @@
...
+    """
+
+    Used dataset: nc_spm_08_grass7
+
+    Test cases:
+    test_same_resolution: Test output with region using the same resolution.
+    test_different_resolution: Test output with region using a different resolution.
+    """
+
+    location = "nc_spm_08_grass7"
```
aaronsms, line 15: Currently, r.proj requires a location parameter, and although the workflow uses nc_spm dataset, they have different names in different context. Changing the hardcoded location to "nc_spm_full_v2alpha2" would probably solve the build issues. But is there a better way to handle this?

wenzeslaus: This is where the testing framework design doesn't work that well. Another location with data is needed, but this test uses the NC SPM as another, so I don't know, perhaps it assumes it is not in NC SPM which is not true when the tests are automatically executed.

Maybe it is good enough test if the same location is used as both current and the other (I'm not sure how r.proj behaves in that case, e.g., if it tries to skip some the steps and just copy). If this is the case, getting the current location name would solve the naming issue.

The more advanced solution is to create a new location (e.g. NC SPF - EPSG:2264) and make it the current one. That's likely better done in a separate environment (copy of os.environ). See, e.g., code in `r.import` or `v.import` or the [create_environment function doc](https://grass.osgeo.org/grass79/manuals/libpython/script.html#script.core.create_environment). You will need to leave out the `use_temp_region` part as it currently does not support passing of the `env` parameter.

raster/r.proj/testsuite/test_r_proj.py
```diff
@@ -0,0 +1,302 @@
...
+
+    def test_different_resolution(self):
+        self.runModule("g.region", raster="elevation", res=25)
+        for method, reference in self.references["different"].items():
+            output_file = "elevation_diff_{}".format(method)
+            self.to_clear.append(output_file)
+            self.runModule(
+                "r.proj",
+                input="elevation",
+                location=self.location,
```
wenzeslaus, line 274: I think you will need to add `mapset="PERMANENT"`.

raster/r.proj/testsuite/test_r_proj.py
```diff
@@ -0,0 +1,302 @@
...
+            cls.runModule(
+                "g.remove", flags="f", type="raster", name=",".join(cls.to_clear)
+            )
+
+    def test_same_resolution(self):
+        self.runModule("g.region", raster="elevation")
+        for method, reference in self.references["same"].items():
+            output_file = "elevation_same_{}".format(method)
+            self.to_clear.append(output_file)
+            self.runModule(
```
wenzeslaus, line 237: `assertModule` is better in this context as this tells the framework that this is the module you are testing. It may improve the error messages you (don't) get in the CI.

raster/r.proj/testsuite/test_r_proj.py
aaronsms: Thanks! The pointer to the doc for creating new location is really useful. The current behavior of r.proj should perform the same computation (so it's not a direct copy) but show a warning that we are using the same location. Since for now it's more important for me to check the integrity of computation after applying parallelization, I will leave doing the advanced step for the future.

raster/r.proj/testsuite/test_r_proj.py
aaronsms: Noted!

aaronsms:
18.04 uses PROJ 4 variant, and 20.04 uses PROJ 5+. r.proj has different behavior for both of them (thus the failing test). Is this to be expected?

neteler:
(re-run flake8)

raster/r.proj/testsuite/test_r_proj.py
```diff
@@ -0,0 +1,305 @@
...
+    def test_different_resolution(self):
+        self.runModule("g.region", raster="elevation", res=25)
+        for method, reference in self.references["different"].items():
+            output_file = "elevation_diff_{}".format(method)
+            self.to_clear.append(output_file)
+            self.assertModule(
+                "r.proj",
+                input="elevation",
+                location=self.location,
+                mapset=self.mapset,
```
neteler, line 277: So, shouldn't this be `mapset="PERMANENT"`? (see comments above)

PR #1668, opened 2021-06-21
Non-interactive display

chaedri:
This is a set of functions to display non-interactive maps in Jupyter Notebooks, as proposed in the Google Summer of Code project, [Improved Integration of GRASS and Jupyter Notebooks](https://trac.osgeo.org/grass/wiki/GSoC/2021/JupyterAndGRASS). You can test the functionality by launching the repository in Binder here:

https://mybinder.org/v2/gh/chaedri/grass/non-interactive-display?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb


python/grass/jupyter/display.py
```diff
@@ -0,0 +1,101 @@
...
+#            in Jupyter Notebooks
+#
+# COPYRIGHT: (C) 2021 Caitlin Haedrich, and by the GRASS Development Team
+#
+#           This program is free software under the GNU General Public
+#           License (>=v2). Read the file COPYING that comes with GRASS
+#           for details.
+
+import os
+import folium
```
wenzeslaus, line 16: Probably won't be part of this PR at the end and not a big deal in Binder, but in general, this will need to be a lazy import, only in the function which needs it unless you make the "interactive" part a separate module and class which may or may not happen in the future. As in the case of `from IPython.display import Image`.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,101 @@
...
+import grass.script as gs
+
+class grassRenderer:
```
wenzeslaus, line 19: Classes should be CamelCase, just run Pylint to get all these complaints sooner :-)

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,101 @@
...
+        Adds a raster to the display
+        """
+        gs.run_command("r.colors", map=raster, color=color)
+        gs.run_command("d.rast", map=raster)
+
+
+    def d_vect(self, vector, fill="grey", color="black", width=2):
+        """
+        Adds a vector to the display
+        """
```
wenzeslaus, line 45: This is a common style for these short docstrings:

```suggestion
        """Adds a vector to the display"""
```

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -53,7 +53,8 @@
    "outputs": [],
    "source": [
     "# Let's check our import by printing docstring for init()\n",
-    "gj.init?"
+    "gj.init?\n",
+    "gj.grassRenderer?"
```
wenzeslaus, line 57: Only the last line seems to really "counts" here, so you need decide or do two different ones. I would say only one is needed or push the second one later in the nb.

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -75,17 +76,35 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Let's display the DTM of our sample area to ensure all's working\n",
-    "\n",
+    "# Let's see what is in the example database\n",
+    "print(gs.read_command(\"g.list\", type=\"all\"))"
```
wenzeslaus, line 80: I would like to see the map first and the list only later, but that might be just me.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,101 @@
...
+import folium
+import grass.script as gs
+
+class grassRenderer:
+    """
+    The grassRenderer class creates and displays GRASS maps in Jupyter
+    Notebooks. 
+    """
+
+    def __init__(self, width=600, height=400, text_size=12):
```
wenzeslaus, line 25: Add env parameter and when set, use it instead of os.environ and also pass it to the other commands.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def d_rast(self, raster, color="greys"):
+        """
+        Adds a raster to the display
+        """
+        gs.run_command("r.colors", map=raster, color=color)
+        gs.run_command("d.rast", map=raster)
+
+
+    def d_vect(self, vector, fill="grey", color="black", width=2):
```
wenzeslaus, line 42: The defaults are probably better left to d.vect itself unless we have good reasons to have them here (different values, need for documenting them right here).

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,101 @@
...
+        os.environ["GRASS_RENDER_HEIGHT"] = str(height)
+        os.environ["GRASS_TEXT_SIZE"] = str(text_size)
+        gs.run_command("d.erase")
+
+
+    def d_rast(self, raster, color="greys"):
+        """
+        Adds a raster to the display
+        """
+        gs.run_command("r.colors", map=raster, color=color)
```
wenzeslaus, line 38: Since this actually modified the raster, this may not be what is expected from the display function. Unfortunately, it also means that you need to do it ahead of time and can't set it here. I'm interested what everybody thinks about this.

wenzeslaus (changes requested):
This looks good! The difference to the `example_notebook.ipynb` is huge, but we need resolve couple of things before this is a full replacement. Please, specify in the PR description what you now see as a goal for this PR (you can change that any time later as we discuss things further).

python/grass/jupyter/display.py
veroandreo: I tested the notebook, looks nice ;)
Indeed, it seems odd to me to be forced to set a color when d.* modules should just read the defined color table or use the default, setting a color should be optional IMHO

doc/notebooks/jupyter_integration.ipynb
chaedri: > I had it the other way because, if you weren't familiar with the spm dataset, seeing the contents would be a pre-requisite to knowing how to modify/play with the code below.

wenzeslaus: Right, but I would say it is not a prerequisite to run the code. You can then executed the next cell to see what else is in the dataset and then go back and play. The other consideration is showing images versus text first. What do you want to see? This of course depends on the target audience and purpose of the notebook which I'm not sure if we decided. Promotion of the Jupyter integration is probably a good goal at this point for this notebook.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,48 @@
...
+
+import os
+import grass.script as gs
+
+
+class GrassRenderer:
+    """
+    The grassRenderer class creates and displays GRASS maps in Jupyter
+    Notebooks.
+    """
```
wenzeslaus, line 22: GrassRenderer and ideally one line including the triple quotes.

python/grass/jupyter/display.py
wenzeslaus: Also, if you just use `**kwargs`, you right away support all what d.vect supports.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,48 @@
...
+        gs.run_command("d.rast", map=raster)
+
+    def d_vect(self, vector, fill="grey", color="black", width=2):
+        """Adds a vector to the display"""
+        gs.run_command(
+            "d.vect", map=vector, fill_color=fill, color=color, width=str(width)
+        )
+
+    def show(self):
+        """Displays a png image of the map (non-interactive)"""
```
wenzeslaus, line 45: Usually PNG I think.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,48 @@
...
+        from IPython.display import Image
```
wenzeslaus, line 46: Please not the reason for lazy import: We don't want Jupyter to be an import-time dependency.

wenzeslaus (changes requested):

python/grass/jupyter/display.py
ninsbl: For listing of data also:
https://grass.osgeo.org/grass79/manuals/libpython/script.html?highlight=list#script.core.list_strings
could be used...

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -75,17 +75,35 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
     "# Set computational region to the study area.\n",
-    "gs.parse_command(\"g.region\", raster=\"elevation\", flags=\"pg\")\n",
+    "gs.run_command(\"g.region\", raster=\"elevation\", flags=\"pg\")"
```
ninsbl, line 89: Should be sufficient to use either p or g flag.

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -75,17 +75,35 @@
...
+   "execution_count": null,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "# Demonstration of grassRenderer for non-interactive map display\n",
+    "spm_map = gj.GrassRenderer(height=600)\n",
+    "\n",
+    "spm_map.d_rast(raster=\"elevation\", color=\"grey\")\n",
```
ninsbl, line 103: Would `spm_map.raster` be for new users more intuitive than `spm_map.d_rast` (or `spm_map.vector` instead of `spm_map.d_vect`)...?

python/grass/jupyter/display.py
ninsbl: Agreed with Vero here...

doc/notebooks/jupyter_integration.ipynb
ninsbl: For listing of data also:
https://grass.osgeo.org/grass79/manuals/libpython/script.html?highlight=list#script.core.list_strings
could be used...

chaedri: That's true, the list takes up a lot of space and maybe the more likely workflow is to run the whole notebook then go back and modify. I'm convinced - I'll switch it. 

@ninsbl That's a cool function! I'll try it out.

python/grass/jupyter/display.py
chaedri: I'm not 100% sure what this would look like. Is this what you're thinking?

`def __init__(self, env=None, ... ):`

What would go in the function then? I imagine and if-else statement?

doc/notebooks/jupyter_integration.ipynb
chaedri: This is a great question. I think it would be easier for new users but it also goes against the GRASS convention and might be less intuitive for experienced GRASS users... I'd be curious to hear what others think. I can write it either way.

python/grass/jupyter/display.py
wenzeslaus: Exactly. If not env then os.environ. Look to grass.script for more examples, but here, it's just that.

wenzeslaus: In my previous comment, I was thinking about storing (the reference to) it as `self._env`. However, to make this work in parallel (so that two of these things can run in parallel) we need to store a copy. See:

https://github.com/ncsu-geoforall-lab/gis714-2021-tangible-landscape/blob/main/website/render_activities.py#L49

doc/notebooks/jupyter_integration.ipynb
wenzeslaus: With run_command, nothing is printed (as expected), so the `-p` is useless. Either -g and parse_command or nothing. I think nothing is fine here. For reading region to Python we have a better function anyway (`region`).

binder/requirements.txt
```diff
@@ -2,3 +2,4 @@ numpy
 matplotlib
 Pillow
 ply
+folium
```
wenzeslaus, line 5: Just add that with the rest of folium support.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,41 @@
...
+
+class GrassRenderer:
+    """The grassRenderer class creates and displays GRASS maps in
+    Jupyter Notebooks."""
+
+    def __init__(self, width=600, height=400, text_size=12):
+        """Initiates an instance of the GrassRenderer class."""
+        os.environ["GRASS_RENDER_WIDTH"] = str(width)
+        os.environ["GRASS_RENDER_HEIGHT"] = str(height)
+        os.environ["GRASS_TEXT_SIZE"] = str(text_size)
```
wenzeslaus, line 27: There is no reason to keep display_settings function around. Instance of GrassRenderer should be sufficient. Delete it and move things here as in:

https://github.com/ncsu-geoforall-lab/gis714-2021-tangible-landscape/blob/main/website/render_activities.py#L47

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,41 @@
...
+        # gs.run_command("r.colors", map=raster, color=color)
+        gs.run_command("d.rast", map=raster, **kwargs)
+
+    def d_vect(self, vector, **kwargs):
+        """Adds a vector to the display"""
+        gs.run_command("d.vect", map=vector, **kwargs)
+
+    def show(self):
+        """Displays a PNG image of the map (non-interactive)"""
+        return Image("map.png")
```
wenzeslaus, line 41: At least make the filename for the image modifiable as in 

https://github.com/ncsu-geoforall-lab/gis714-2021-tangible-landscape/blob/main/website/render_activities.py#L50

Using `"map.png"` when not provided is fine for now, but eventually (i.e., in a subsequent PR), the automatic option should be probably some temporary file somewhere.

wenzeslaus (changes requested):
Here is couple more things about handling of the rendering environment. Basically, it's: see GrassRenderer here:

https://github.com/ncsu-geoforall-lab/gis714-2021-tangible-landscape/blob/main/website/render_activities.py#L41

doc/notebooks/jupyter_integration.ipynb
wenzeslaus: I also think this is a great question. Even the name `GrassRenderer` is not necessary set. `spm_map` was little strange at first, but I think it is actually okay.

In a [GrassRenderer](https://github.com/ncsu-geoforall-lab/gis714-2021-tangible-landscape/blob/main/website/render_activities.py#L41) class I wrote for a different purpose, I had only a `run` method which just executed whatever command (like `run_command` does) since I wanted to pass any `d.*` commands which I was reading and [running programmatically](https://github.com/ncsu-geoforall-lab/gis714-2021-tangible-landscape/blob/main/website/render_activities.py#L233). I guess you could try to enforce that the command name must start with `d` so that it is not misused, but I'm not convinced it is needed.

`d_rast`, `d_vect` is a sure middle ground, but it requires either all modules to be mirrored or implementing the "resolve a missing attribute" magic as in PyGRASS module shortcuts to support all commands. It would be harder to implement what I did above using just this because you need to translate what you are reading to function calls, but it is doable in Python and perhaps a fringe use case in Jupyter.

`vector`, `raster` may work well as a hand-picked subset in addition to a more general interface or if you conceptualize the whole thing as a "display" (`display = GrassDisplay()`) where d.vect translates to `display.vector()`. Yet another approach is to have these as a common API with the folium-based renderer in parallel with the `d_vect` style.

python/grass/jupyter/display.py
wenzeslaus: Perhaps the issue is not as much "being forced to set a color" (defaults can be changed) as that the expectation is that you set the color at this point. And perhaps that the reason @chaedri wrote it initially this way. Like for vectors. Not Jupyter specific, but it is more visible here.

I have never tried that, but even with the current system, you could implement ad-hoc/temporary color tables for rasters, I think. You create a temporary mapset, set color table for raster of interest there (raster color tables are per-mapset), display  the raster from that mapset, so the mapset-specific color table kicks in. I'm not suggesting @chaedri should do this :-)

python/grass/jupyter/display.py
wenzeslaus: Typo: I meant "note" the reason. But I'm fine with Jupyter as an import-time dependency in grass.jupyter. Makes sense.

doc/notebooks/jupyter_integration.ipynb
ninsbl: In the introduction it is probably worth mentioning where to find more information on what one can do with GRASS GIS APIs, like:
* Manuals in general: https://grass.osgeo.org/grass-stable/manuals/
* Python API: https://grass.osgeo.org/grass-stable/manuals/libpython/
* Module manuals: https://grass.osgeo.org/grass-stable/manuals/full_index.html (or https://grass.osgeo.org/grass78/manuals/graphical_index.html)

In addition to listing data, also modules could be listed (at the end):
`print(gs.read_command("g.search.modules", flags=g"))`

doc/notebooks/jupyter_integration.ipynb
ninsbl: What about module shortcuts here?
https://grass.osgeo.org/grass79/manuals/libpython/pygrass.modules.html?highlight=shortcut#module-pygrass.modules.shortcuts

If there is value in making the API as much as possible compatibel with e.g. `folium` that should probably go into consideration for the non-interactive API as well?

I have to admitt that I am not familiar with folium. But for passing vector data to folium, that we discussed earlier, [viter](https://grass.osgeo.org/grass78/manuals/libpython/pygrass.vector.html?highlight=viter#pygrass.vector.VectorTopo.viter) from pygrass maybe an option too? There you could request only geometries within the map window of folium with the [find]() method.

doc/notebooks/jupyter_integration.ipynb
chaedri: @ninsbl I was thinking this might be better placed in the example notebook, since this one is more focused in Jupyter-GRASS integration. So, I made a new PR with those changes here: https://github.com/OSGeo/grass/pull/1686

You can also see the changes in binder here:
https://mybinder.org/v2/gh/chaedri/grass/example_notebook_update?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fexample_notebook.ipynb

I'm undecided on whether listing all the modules is helpful - perhaps since the manual link is there, the list is unnecessary?

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,60 @@
...
+            os.environ["GRASS_RENDER_HEIGHT"] = str(height)
+            os.environ["GRASS_TEXT_SIZE"] = str(text_size)
+            os.environ["GRASS_RENDER_IMMEDIATE"] = "cairo"
+            os.environ["GRASS_RENDER_FILE"] = filename
+            os.environ["GRASS_RENDER_FILE_READ"] = "TRUE"
+            self._legend_file = Path(filename).with_suffix(".grass_vector_legend")
+            os.environ["GRASS_LEGEND_FILE"] = str(self._legend_file)
+            self._env = os.environ.copy()
+        else:
+            self._env = os.environ.copy()
```
wenzeslaus, line 40: env is unused except for the if statement. `self._env = env`?

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,60 @@
...
+            self._env["GRASS_RENDER_WIDTH"] = str(width)
+            self._env["GRASS_RENDER_HEIGHT"] = str(height)
+            self._env["GRASS_TEXT_SIZE"] = str(text_size)
+            self._legend_file = Path(filename).with_suffix(".grass_vector_legend")
+            self._env["GRASS_LEGEND_FILE"] = str(self._legend_file)
```
wenzeslaus, line 45: Code duplication and missing variables.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,60 @@
...
+
+        gs.run_command("d.erase")
```
wenzeslaus, line 47: You need to pass the environment as in:

https://github.com/ncsu-geoforall-lab/gis714-2021-tangible-landscape/blob/main/activities/activity_haedrich.py#L33

python/grass/jupyter/display.py
wenzeslaus: Now you added filename as a parameter, so you should reflect that here.

wenzeslaus (changes requested):

python/grass/jupyter/display.py
chaedri: Ok, I took a swing at this in my latest commit. Let me know if that's in line with what you were thinking.

doc/notebooks/jupyter_integration.ipynb
wenzeslaus: Yep, module shortcuts is what I meant by the "resolve a missing attribute" magic.

I suggest to:

1. Add `run(name, **kwargs)` now to:
    a. Support any command right away.
    b. Have a single place where the actual subprocess call is handled (i.e., use this method from the other methods).
2. Go with `d_rast` and `d_vect` or `raster` and `vector` (whatever seems better to you @chaedri) in this PR to have a nice interface.
3. Later, in another PR, adding more (all) `d_vect`-like methods later either explicitly defined, dynamically generated (`self.d_vect = some_locally_defined_function`), or through custom, simpler version of module shortcuts. I'm saying `d_vect`-like in a broad sense: they could be done like `display.vect` similarly to what I mentioned above. However, I think closer to module names, i.e., `d_vect` makes more sense here as it will clearly translate to all display module names.
4. Leave the compatibility with interactive display for later (again another PR) when the interactive display is somewhat finished. I think switching between displays just by changing the class name but not the rest of your code would be nice.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(
+        self, env=None, width=600, height=400, filename="map.png", text_size=12
+    ):
+        """Initiates an instance of the GrassRenderer class."""
+
+        if env:
+            self._env = env
+        if not env:
+            self._env = os.environ.copy()
```
wenzeslaus, line 32: Pylint may give you this complaint too: the second condition is negation of the first one which is in this context best expressed as `else:`.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,58 @@
...
+
+        self._env["GRASS_RENDER_WIDTH"] = str(width)
+        self._env["GRASS_RENDER_HEIGHT"] = str(height)
+        self._env["GRASS_TEXT_SIZE"] = str(text_size)
+        self._env["GRASS_RENDER_IMMEDIATE"] = "cairo"
+        self._env["GRASS_RENDER_FILE"] = filename
```
wenzeslaus, line 38: Let's get ahead of the game and support `Path` here by converting this to `str` as we do with the numbers. This is how it behaves:

```
> str(Path("/home"))
'/home'
```

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,58 @@
...
+        self._env["GRASS_RENDER_FILE_READ"] = "TRUE"
+        
+        self._legend_file = Path(filename).with_suffix(".grass_vector_legend")
+        self._env["GRASS_LEGEND_FILE"] = str(self._legend_file)
+
+        self._filename = filename
+
+        gs.run_command("d.erase", env = self._env)
```
wenzeslaus, line 46: Use `self.run()` here too.

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def run(self, module,  **kwargs):
+        """Run modules from "d." GRASS library"""
+        # Check module is from display library then run
+        if module[0]== "d":
+            gs.run_command(module, env=self._env, **kwargs)
+        else:
+            print("Module must be from GRASS display library. It must begin with letter \'d\'.")
```
wenzeslaus, line 54: See "raising exceptions" and "reporting errors in Python".

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,58 @@
...
+    """The grassRenderer class creates and displays GRASS maps in
+    Jupyter Notebooks."""
+
+    def __init__(
+        self, env=None, width=600, height=400, filename="map.png", text_size=12
+    ):
+        """Initiates an instance of the GrassRenderer class."""
+
+        if env:
+            self._env = env
```
wenzeslaus, line 30: Make a copy here too. Easier to document. More predictable for the user. It also makes the environment private, so more control in the implementation.

wenzeslaus (changes requested):

python/grass/jupyter/display.py
chaedri: So, the line would read: `self._env["GRASS_RENDER_FILE"] = str(Path(filename))` ?

wenzeslaus: No, no need to construct Path. The point is to handle the case when user passes filename as an Path object. Try the behavior in an interactive console. The point is to handle these two or three cases:

```python
file = Path("/base/dir") / more_dirs / added_here
renderer1 = GrassRenderer(filename=file)
renderer2 = GrassRenderer(filename="map2.png")
renderer3 = GrassRenderer(filename=os.path.join("base/dir", "map2.png"))
```

python/grass/jupyter/display.py
```diff
@@ -0,0 +1,58 @@
...
+
+        self.run("d.erase", env=self._env)
+
+    def run(self, module, **kwargs):
+        """Run modules from "d." GRASS library"""
+        # Check module is from display library then run
+        if module[0] == "d":
+            gs.run_command(module, env=self._env, **kwargs)
+        else:
+            raise Exception("Module must begin with letter 'd'.")
```
wenzeslaus, line 54: See what is the most fitting exception type for this problem, probably some value or attribute error whatever is the actual name.

wenzeslaus (changes requested):

python/grass/jupyter/display.py
chaedri: I couldn't get any of those to produce an error. But, I think I understand: line 38 should read `self._env["GRASS_RENDER_FILE"] = str(filename)`.

wenzeslaus: Right, no error, there is just no point in making the filename a Path object if it is not (well, at least for this line, we need Path later).

wenzeslaus (approved):
This looks great and it is ready to be tested!

wenzeslaus:
Try this PR live at:

https://mybinder.org/v2/gh/chaedri/grass/non-interactive-display?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb

PR #1670, opened 2021-06-21
libpython: Add helper library for benchmarking

wenzeslaus:
An experimental collection of simple functions to help benchmarking
and reduce code duplication between benchmarks.

The current design ideas are be minimalist, pragmatic with no API promisses.
The tools are meant for developers tracking the latest development version.

Co-authored-by: Aaron Saw Min Sern (thread/nprocs benchmarking function)

### DIFF
--- a/python/grass/benchmark/plots.py
+++ b/python/grass/benchmark/plots.py
@@ -14,6 +14,29 @@
 """Plotting functionality for benchmark results"""
 
 
+def get_pyplot(to_file):
+    """Get pyplot from matplotlib
+
+    Lazy import to easily run code importing this function on limited installations.
+    Only actual call to this function requires matplotlib.
+
+    The *to_file* parameter can be set to True to avoid tkinter dependency
+    if the interactive show method is not needed.
+    """
+    import matplotlib  # pylint: disable=import-outside-toplevel
+
+    if to_file:
+        backend = "agg"
+    else:
+        backend = None
+    if backend:
+        matplotlib.use(backend)
+
+    import matplotlib.pyplot as plt  # pylint: disable=import-outside-toplevel
+
+    return plt
+
+
 def nprocs_plot(results, filename=None):
     """Plot results from a multiple nprocs (thread) benchmarks.
 
@@ -31,10 +54,7 @@ def nprocs_plot(results, filename=None):
     Each result can come with a different list of nprocs, i.e., benchmarks
     which used different values for nprocs can be combined in one plot.
     """
-    # Lazy import to easily run code importing this function on limited installations.
-    # Only actual call to this function requires matplotlib.
-    import matplotlib.pyplot as plt  # pylint: disable=import-outside-toplevel
-
+    plt = get_pyplot(to_file=bool(filename))
     axes = plt.gca()
 
     x_ticks = set()  # gather x values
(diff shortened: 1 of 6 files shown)


wenzeslaus:
@aaronsms I would be interested in what you think (I can't put you on review directly).

wenzeslaus:
This is addressing some of the needs from @aaronsms's GSoC 2021, but see also [[GRASS-dev] measuring cpu time for each module and estimating modules scalability](https://lists.osgeo.org/pipermail/grass-dev/2014-August/070343.html) which was _not_ taken into account at this point.

python/grass/benchmark/plots.py
```diff
@@ -0,0 +1,49 @@
...
+    import matplotlib.pyplot as plt
+
+    for result in results:
+        fig, ax = plt.subplots()
+        ax.invert_xaxis()
+        plt.plot(result.resolutions, result.avg_times, label=result.label)
+        mins = [min(i) for i in result.all_times]
+        maxes = [max(i) for i in result.all_times]
+        plt.fill_between(result.resolutions, mins, maxes, color="gray", alpha=0.3)
+    plt.legend()
```
aaronsms, line 45: For resolution_plot, this is generating different figures for each plot, then the `plt.legend` needs to be within the loop to display the label for each plot.

python/grass/benchmark/plots.py
```diff
@@ -0,0 +1,49 @@
...
+    if filename:
+        plt.savefig(filename)
```
aaronsms, line 47: Also, only the last figure/plot will be saved in file.

aaronsms:
I think this is nice to add, I have tried using this for r.benchmark and r.univar. For the `resolution_plot`, do you think it's better to have the results be fitted into the same plot just like `nprocs_plot`, even if the resolutions points are slightly different, because users can always choose call `resolutions_plot` multiple times if they want separate graphs. Having them sharing a same graph allows us to see how it scales against bigger inputs, for example;
image

python/grass/benchmark/plots.py
HuidaeCho: I think he meant:
```python
plt.gca().invert_xaxis()
for result in results:
    plt.plot(result.resolutions, result.avg_times, label=result.label)
    mins = [min(i) for i in result.all_times]
    maxes = [max(i) for i in result.all_times]
    plt.fill_between(result.resolutions, mins, maxes, color="gray", alpha=0.3)
plt.legend()
```

HuidaeCho (changes requested):

HuidaeCho:
> For the `resolution_plot`, do you think it's better to have the results be fitted into the same plot just like `nprocs_plot`

I think it would be more useful to plot them all in the same plot just like `nprocs_plot`.

python/grass/benchmark/plots.py
wenzeslaus: Right, modified to make only one plot.

python/grass/benchmark/plots.py
```diff
@@ -0,0 +1,49 @@
...
+#
+#            This program is free software under the GNU General Public
+#            License (>=v2). Read the file COPYING that comes with GRASS
+#            for details.
+
+
+"""Plotting functionality for benchmark results"""
+
+
+def nprocs_plot(results, filename=None):
```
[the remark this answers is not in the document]
wenzeslaus: All functions are now documented.

wenzeslaus:
> > For the `resolution_plot`, do you think it's better to have the results be fitted into the same plot just like `nprocs_plot`
> 
> I think it would be more useful to plot them all in the same plot just like `nprocs_plot`.

Agreed. Done.

wenzeslaus:
This now has documentation and test. Additional benchmarking, outputs, plots are left for future PRs.

HuidaeCho (approved):

wenzeslaus:
I'm not sure if the test will run in the CI, but I did two rounds of dealing with the dependencies, so hopefully it will. I can't confirm that now, due to addon installation failing which fails the tests in CI.

I merged this anyway so it is available for @aaronsms and anybody who will be testing his work.

PR #1677, opened 2021-06-24
libgis: Recover when directory already exists

wenzeslaus:
When another process creates the directory between the access test
and mkdir call, mkdir ends with an error even though the desired
state now exists.

This chanages the error handling so that on mkdir error,
the access test is still performed and only when it fails
the mkdir errno is used for fatal error. The access errno
is used when mkdir succeeds but access fails.

This follows the behavior before 3c374600e3cc7d91c542cab4411fdf40f31ce931
where only the second access failure causes an error, but not a failed
mkdir call in betwee the access calls.

This error handling is also possibly more aligned with the documentation
of the higher level function from API G_make_mapset_element() which
should 'do nothing if it is found'. However, the documentation is
not written with parallel processes in mind, so it only states that
the function 'can be called even if the element already exists'
and does not specify what if the element comes to existence during
the function execution. However, I would say that since existence
is a valid pre-condition and it does not matter how the post-condition,
i.e., existing directory, was achieved. Hence finding out that some
other process created it is fine.

If my understanding is correct, this approach should work well for
directories in mapset such as cell, fcell, or vector, for example
we have calls such as `G_make_mapset_element("fcell");`.

However, it seems to me that in several places this is used for to create
nested directories such as grid3/some_raster3d_name. This is is place
where you want the function to fail when the some_raster3d_name directory
exists, because that's definitevely a race condition unlike creating grid3
which may be just two processes peacefully co-existing.
To accomode this scenario the new error handling is not good
and the original one works. What would be needed is a better API which
does not use vague element in the function name, but instead distinguishes
between element group directories (such as cell) and specific elements
(sub-directory for a specific vector map).
If this is the correct approach, this PR is currently experimental.

### DIFF
--- a/lib/gis/mapset_msc.c
+++ b/lib/gis/mapset_msc.c
@@ -85,6 +85,7 @@ int make_mapset_element(const char *p_path, const char *p_element)
     while (1) {
 	if (*element == '/' || *element == 0) {
 	    *p = 0;
+            char *msg = NULL;
             if (access(path, 0) != 0) {
                 /* Assuming that directory does not exist. */
                 if (G_mkdir(path) != 0)
--- a/lib/gis/mapset_msc.c
+++ b/lib/gis/mapset_msc.c
@@ -85,14 +85,24 @@ int make_mapset_element(const char *p_path, const char *p_element)
     while (1) {
 	if (*element == '/' || *element == 0) {
 	    *p = 0;
-	    if (access(path, 0) != 0) { /* directory not yet created */
-		if (G_mkdir(path) != 0)
-		    G_fatal_error(_("Unable to make mapset element %s (%s): %s"),
-				  p_element, path, strerror(errno));
-	    }
-	    if (access(path, 0) != 0)  /* directory not accessible */
-		G_fatal_error(_("Unable to access mapset element %s (%s): %s"),
-			      p_element, path, strerror(errno));
+            if (access(path, 0) != 0) {
+                /* Assuming that directory does not exist. */
+                if (G_mkdir(path) != 0)
+                    msg = G_store(strerror(errno));
+            }
+            if (access(path, 0) != 0) {
+                /* Directory is not accessible even after attempt to create it. */
+                if (msg) {
+                    /* Error already happened when mkdir. */
+                    G_fatal_error(_("Unable to make mapset element %s (%s): %s"),
+                                  p_element, path, strerror(errno));
+                }
+                else {
+                    /* Access error is not related to mkdir. */
+                    G_fatal_error(_("Unable to access mapset element %s (%s): %s"),
+                                  p_element, path, strerror(errno));
+                }
+            }
 	    if (*element == 0)
 		return 1;
 	}


wenzeslaus:
Among other things, this aims to fix some of the temporal modules which you can sometimes see failing in the CI:

```
Running ./temporal/t.rast.gapfill/testsuite/test_gapfill.py...

ERROR: Unable to make mapset element fcell (/home/runner/nc_spm_full_v2alpha2/__temporal_t_rast_gapfill_test_gapfill/fcell): File exists
```

HuidaeCho:
> some of the temporal modules which you can sometimes see failing in the CI

Is it happening because the modules run concurrently? I believe many, if not all, parts of GRASS are not safe for parallel processes.

HuidaeCho:
> instead distinguishes
> between element group directories (such as cell) and specific elements
> (sub-directory for a specific vector map).

Do you mean create individual API functions for different elements, or one API function for groups and another for elements in a group? This PR reverts back to the 2009 behavior, but it looks harmless to me... I tried to think of any race conditions that could cause any problems during parallel proceses (very common now? Or can it even happen in a non-parallel environment?). Probably, @neteler still remembers why he changed the behavior. I guess he only meant to add an error message on mkdir failure?

neteler:
> Probably, @neteler still remembers why he changed the behavior. I guess he only meant to add an error message on mkdir failure?

I believe to have found the related discussion in grass-dev:

- https://lists.osgeo.org/pipermail/grass-dev/2009-March/042741.html
    - ... (see thread)
        - https://lists.osgeo.org/pipermail/grass-dev/2009-March/042774.html

HuidaeCho:
> I believe to have found the related discussion in grass-dev:
> 
>     * https://lists.osgeo.org/pipermail/grass-dev/2009-March/042741.html
>       
>       * ... (see thread)
>         
>         * https://lists.osgeo.org/pipermail/grass-dev/2009-March/042774.html

Thanks for the links, @neteler. AFAICT, I cannot see any direct intention of changing the behavior. It was a side effect of printing the error message on mkdir failure. His reason for the mkdir failure was the number of directories, not parallel processing.

wenzeslaus:
@HuidaeCho @neteler Thanks for the input.

See #1677 for an alternative approach. Instead of changing the behavior, it adds the behavior from this PR and uses that based on the context.

HuidaeCho:
> See #1677 for an alternative approach.

You mean #1681.

wenzeslaus:
> > 
> 
> Is it happening because the modules run concurrently?

To be clear, the tests in CI do not run in parallel. However, inside the modules, other modules do run in parallel.

> I believe many, if not all, parts of GRASS are not safe for parallel processes.

Many parts are working quite well in parallel if used properly. It would be also different if we speak about C libraries, Python, or user scripts. Sometimes just little more care is need to make things work smoothly which is the case also here.

> Do you mean create individual API functions for different elements, or one API function for groups and another for elements in a group (can it address race conditions BTW?)?

Right and I think yes, it can. 

> I tried to think of any race conditions that could cause any problems during parallel proceses (very common now? Or can it even happen in a non-parallel environment?).

Just having GUI makes things potentially parallel, so yes, not necessarily explicitly parallel and common.

wenzeslaus:
Right, I mean #1681.

HuidaeCho:
> Right, I mean #1681.

#1681 looks better.

wenzeslaus:
For the record, this should fix the errors coming from temporal modules:

```
Running ./temporal/t.rast.gapfill/testsuite/test_gapfill.py...

ERROR: Unable to make mapset element fcell (/home/runner/nc_spm_full_v2alpha2/__temporal_t_rast_gapfill_test_gapfill/fcell): File exists
r.series.interp input=a_1@__temporal_t_rast_gapfill_test_gapfill,a_2@__temporal_t_rast_gapfill_test_gapfill datapos=0,1 samplingpos=0.3333333333333333,0.6666666666666666 method=linear output=test_6_1,test_6_2 --q
...
```

And it also should fix the parallel test of temporary region (where many raster maps are created at once):

```
Running ./python/grass/script/testsuite/test_parallel_temp_region.sh...
Pure parallel test: Got 48 but expected 50 maps
FAILED ./python/grass/script/testsuite/test_parallel_temp_region.sh
```

marisn (approved):
Good catch!

wenzeslaus:
@marisn Thanks checking this PR. The PR #1681 replaces this. It uses the same function, but the race recovery is optional and the two different contexts are distinguished in the API. My intention is to close this PR and merge #1681 instead.

wenzeslaus:
See #1681 instead.

PR #1684, opened 2021-06-25
Jupyter: Interactive display

chaedri:
This PR contains interactive functions for displaying GRASS maps in Jupyter Notebooks using folium. It is part of an ongoing Google Summer of Code project, Improved Integration of GRASS and Jupyter Notebooks. You can find more information [here](https://trac.osgeo.org/grass/wiki/GSoC/2021/JupyterAndGRASS). 

The functions can also be tried in this Jupyter Notebook:

https://mybinder.org/v2/gh/chaedri/grass/interactive-display?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb

### DIFF
--- a/binder/requirements.txt
+++ b/binder/requirements.txt
@@ -2,3 +2,4 @@ numpy
 matplotlib
 Pillow
 ply
+folium
--- a/doc/notebooks/jupyter_integration.ipynb
+++ b/doc/notebooks/jupyter_integration.ipynb
@@ -21,8 +21,7 @@
    "source": [
     "import os\n",
     "import subprocess\n",
-    "import sys\n",
-    "from IPython.display import Image"
+    "import sys"
    ]
   },
   {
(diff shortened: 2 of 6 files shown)


chaedri:
This is just a draft (it doesn't allow the display of vectors at all yet!) but I wanted to open a PR so we could discuss and I could have some feedback. One of the issues I'm working on is figuring out the best way to pass raster/vector data between GRASS and folium. In this first draft, a PNG of the raster is written by GRASS then overlayed on a map in folium. But, I think folium only takes certain projections (WGS84), so I reproject the bounding box of the region and stretch the PNG into the box when it's overlayed. Thoughts and feedback welcome.

wenzeslaus (changes requested):
Leaving aside the issue of reprojection for now, the image size does not correspond to the region. The region is fitted into the image, but the image contains whitespace on the sides which is outside of the computational region (in other context, it could be used e.g. for legend). However, to georeference the image, it needs to actually exactly fit the region, so the ratio needs to be the same. See for example of the this procedure:

https://github.com/ncsu-geoforall-lab/grass-web-publishing

You can test the basic fit with the base map using:

```
gs.run_command("v.to.rast", input="roadsmajor", output="roadsmajor", use="val")
spm_map.d_rast("roadsmajor")
spm_map.show_interactively(opacity=0.7)
```

chaedri:
These changes are in PR #1710.

PR #1686, opened 2021-06-26
Added additional GRASS info to example_notebook

chaedri:
This PR contains a few changes proposed in #1668 that I think would be better placed here. I'm working on the example notebook and other Jupyter Notebooks as part of the Google Summer of Code project, [Improved Integration of GRASS GIS and Jupyter Notebooks[(https://trac.osgeo.org/grass/wiki/GSoC/2021/JupyterAndGRASS).

You can see and test this updated notebook here:
https://mybinder.org/v2/gh/chaedri/grass/example_notebook_update?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fexample_notebook.ipynb

### DIFF
--- a/doc/notebooks/example_notebook.ipynb
+++ b/doc/notebooks/example_notebook.ipynb
@@ -1,23 +1,8 @@
 {
- "metadata": {
-  "language_info": {
-   "codemirror_mode": {
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
-   "name": "python",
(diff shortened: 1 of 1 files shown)


wenzeslaus (approved):
This is good and I think a better option than having this in the notebooks with the new Jupyter-oriented API.

I'm not 100% happy with the output, esp. from g.list, but the `list_*` are not better. Either hard to read or rasters and vectors mixed or both. Leaving out the print actually helps with with some combinations. Suggestions welcome.

PR #1687, opened 2021-06-27
g.filename: Create dir in mapset only when asked

wenzeslaus:
The previous behavior is creating the corresponding directory in the current
mapset anytime the module is executed (the intention was anytime the path is
in the current mapset according to the doc).

Creating the directory makes sense given the intended use of the module
(notably, the module is used only in one v6 addon as of now).
Given that there is a distinction between creation of an element and file
(even when file is a directory), having the functionality in a module
allows for treating the element case in a special way (see #1681).

However, given the primary function of this module (constructing a path regardless of
its existence), writing operation is unexpected, so the new default is not writing
anything and creation needs to be explicitly requested using a new -c flag
which causes fatal error when the constructed path is not in the current mapset
(asking for creation of something outside of the current mapset is considered a usage error).


wenzeslaus:
Launch binder to test this:

https://mybinder.org/v2/gh/wenzeslaus/grass/no-mkdir-in-g_filename?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fexample_notebook.ipynb

PR #1650, opened 2021-06-16
t.remove: changed behaviour of flags to remove stds and connected maps

lucadelu:
Implemented behaviour suggested in https://trac.osgeo.org/grass/ticket/3362#comment:1

PR #1695, opened 2021-06-30
t.remove: new implementation of flags

lucadelu:
new pull request from previous one #1650


veroandreo (approved):

veroandreo:
> There are some `temporal` tests failing on CI. Could they be caused by changes with this PR?

Right! I didn't check that. Most likely yes, if they use t.remove, since the flag behavior changes with this PR

lucadelu:
@nilason thanks to catch that. Looking previous log only one error seems related to my changes. 
However I updated all the tests replacing `flags="rf"` with the new `flas="df"` that has the same meaning of previous

PR #1678, opened 2021-06-24
g.extension: draft to address new addon-repo structure

ninsbl:
See also: https://github.com/OSGeo/grass-addons/issues/528
This is a draft and may have to be adjusted once the new repo structure is settled...

PR #1700, opened 2021-07-01
g.extension: get branch from version

ninsbl:
Replaces #1678
Once we have a grass8 branch in grass-addons, the hardcoded version 7 should be removed from g.extension...

### DIFF
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -77,7 +77,6 @@
 # % description: Specific branch to fetch addon from (only used when fetching from git)
 # % required: no
 # % multiple: no
-# % answer: main
 # %end
 
 # %flag
@@ -232,6 +231,21 @@ def urlopen(url, *args, **kwargs):
     """
     request = urlrequest.Request(url, headers=HEADERS)
     return urlrequest.urlopen(request, *args, **kwargs)
+
+
+def get_branches(github_api_url, version_only=False):
+    """Get ordered list of branch names in repo using github API"""
+    req = urlrequest.urlopen(github_api_url)
+    content = json.loads(req.read())
+    branches = [repo_branch["name"] for repo_branch in content]
+    if version_only:
+        branches = [
+            version_branch
+            for version_branch in branches
+            if version_branch.startswith("grass")
+        ]
+    branches.sort()
+    return branches
 
 
 def download_addons_paths_file(url, response_format, *args, **kwargs):
@@ -1612,9 +1626,9 @@ def install_extension_std_platforms(name, source, url, branch):
     """Install extension on standard platforms"""
     gisbase = os.getenv("GISBASE")
     # TODO: workaround, https://github.com/OSGeo/grass-addons/issues/528
-    source_url = "https://github.com/OSGeo/grass-addons/tree/grass{}/src".format(
-        version[0]
-    )
+    # source_url = "https://github.com/OSGeo/grass-addons/tree/grass{}/src".format(
+    #     version[0]
+    # )
 
     # to hide non-error messages from subprocesses
     if grass.verbosity() <= 2:
@@ -1696,7 +1710,7 @@ def install_extension_std_platforms(name, source, url, branch):
         "SCRIPTDIR=%s" % dirs["script"],
         "STRINGDIR=%s" % dirs["string"],
         "ETC=%s" % os.path.join(dirs["etc"]),
-        "SOURCE_URL=%s" % source_url,
+        "SOURCE_URL=%s" % url,
     ]
 
     install_cmd = [
@@ -2304,12 +2318,18 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
     # Handle URL for the offical repo
     if not url and name:
         module_class = get_module_class_name(name)
+        repo_branches = get_branches(
+            "https://api.github.com/repos/OSGeo/grass-addons/branches", True
+        )
+        # Define branch to fetch from (latest or current version)
+        version_branch = "grass{}".format(version[0])
+        if version_branch not in repo_branches:
+            version_branch = repo_branches[-1]
+
         # Set URL for the given GRASS version
         git_url = (
             "https://github.com/OSGeo/grass-addons/branches/"
-            "grass{version}/src/{module_class}/{module_name}".format(
-                version=version[0], module_class=module_class, module_name=name
-            )
+            f"{version_branch}/src/{module_class}/{name}"
         )
         return "official", git_url
 
@@ -2319,7 +2339,19 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
 
         # note: 'trunk' is required to make URL usable for 'svn export' call
         if branch is None:
-            svn_reference = "branches/grass{}".format(version)
+            repo_branches = get_branches(
+                url.rstrip("/").replace("github.com/", "api.github.com/repos/")
+                + "/branches",
+                True,
+            )
+            # Define branch to fetch from (latest or current version)
+            if not repo_branches:
+                svn_reference = "trunk"
+            else:
+                version_branch = "grass{}".format(version[0])
+                if version_branch not in repo_branches:
+                    version_branch = repo_branches[-1]
+                svn_reference = "branches/{}".format(version_branch)
         elif branch in ["master", "main"]:
             svn_reference = "trunk"
         else:
@@ -2499,8 +2531,11 @@ if __name__ == "__main__":
     version = grass_version["version"].split(".")
     # TODO: update temporary workaround of using grass7 subdir of addon-repo, see
     #       https://github.com/OSGeo/grass-addons/issues/528
-    version[0] = 7
-    version[1] = 9
+
+    # if version[0] > 7:
+    #     version[0] = 7
+    #     version[1] = 9
+
     build_platform = grass_version["build_platform"].split("-", 1)[0]
 
     sys.exit(main())
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -1612,7 +1612,7 @@ def install_extension_std_platforms(name, source, url, branch):
     """Install extension on standard platforms"""
     gisbase = os.getenv("GISBASE")
     # TODO: workaround, https://github.com/OSGeo/grass-addons/issues/528
-    source_url = "https://github.com/OSGeo/grass-addons/tree/master/grass7/"
+    source_url = "https://github.com/OSGeo/grass-addons/tree/grass{}/src".format(version[0])
 
     # to hide non-error messages from subprocesses
     if grass.verbosity() <= 2:
@@ -2303,8 +2303,8 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
         module_class = get_module_class_name(name)
         # note: 'trunk' is required to make URL usable for 'svn export' call
         git_url = (
-            "https://github.com/OSGeo/grass-addons/trunk/"
-            "grass{version}/{module_class}/{module_name}".format(
+            "https://github.com/OSGeo/grass-addons/branches/"
+            "grass{version}/src/{module_class}/{module_name}".format(
                 version=version[0], module_class=module_class, module_name=name
             )
         )
@@ -2319,16 +2319,15 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
         module_class = get_module_class_name(name)
 
         # note: 'trunk' is required to make URL usable for 'svn export' call
-        if branch in ["master", "main"]:
-            svn_reference = "trunk"
+        if branch is None or branch in ["master", "main"]:
+            svn_reference = "branches/grass{}".format(version)
         else:
             svn_reference = "branches/{}".format(branch)
 
         git_url = (
             "{url}/{branch}/"
-            "grass{version}/{module_class}/{module_name}".format(
+            "src/{module_class}/{module_name}".format(
                 url=url,
-                version=version[0],
                 module_class=module_class,
                 module_name=name,
                 branch=svn_reference,
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -248,6 +248,28 @@ def get_github_branches(github_api_url, version_only=False):
     return branches
 
 
+def get_default_branch(full_url):
+    """Get default branch for repo (currently only implemented for github API and gitlab API"""
+    if "github.com" in full_url:
+        organization, repository = full_url.split("github.com/")[1].split("/")[0:2]
+        req = urlrequest.urlopen(f"https://api.github.com/orgs/{organization}/repos")
+        content = json.loads(req.read())
+        default_branch = [
+            repo["default_branch"] for repo in content if repo["name"] == repository
+        ][0]
+    elif "gitlab.com" in full_url:
+        organization, repository = full_url.split("gitlab.com/")[1].split("/")[0:2]
+        req = urlrequest.urlopen(
+            f"https://gitlab.com/api/v4/projects/{organization}%2F{repository}"
+        )
+        content = json.loads(req.read())
+        default_branch = content["default_branch"]
+
+    else:
+        default_branch = "main"
+    return default_branch
+
+
 def download_addons_paths_file(url, response_format, *args, **kwargs):
     """Generates JSON file containing the download URLs of the official
     Addons
@@ -1567,7 +1589,7 @@ def download_source_code(
             response = urlopen(url)
         except URLError:
             # Try download add-on from 'master' branch if default "main" fails
-            if branch is None or branch == "":
+            if not branch:
                 try:
                     url = url.replace("main", "master")
                     gscript.message(
@@ -2232,9 +2254,7 @@ def resolve_known_host_service(url, name, branch):
         else:
             actual_start = ""
         if "branch" in match["url_end"]:
-            suffix = match["url_end"].format(
-                name=name, branch=branch if branch else "main"
-            )
+            suffix = match["url_end"].format(name=name, branch=get_default_branch(url))
         else:
             suffix = match["url_end"].format(name=name)
         url = "{prefix}{base}{suffix}".format(
@@ -2319,7 +2339,8 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
     if not url and name:
         module_class = get_module_class_name(name)
         repo_branches = get_github_branches(
-            "https://api.github.com/repos/OSGeo/grass-addons/branches", True
+            "https://api.github.com/repos/OSGeo/grass-addons/branches",
+            version_only=True,
         )
         # Define branch to fetch from (latest or current version)
         version_branch = "grass{}".format(version[0])
@@ -2342,7 +2363,7 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
             repo_branches = get_github_branches(
                 url.rstrip("/").replace("github.com/", "api.github.com/repos/")
                 + "/branches",
-                True,
+                version_only=True,
             )
             # Define branch to fetch from (latest or current version)
             if not repo_branches:
@@ -2419,10 +2440,11 @@ def get_addons_paths(gg_addons_base_dir):
     and their paths (mkhmtl.py tool)
     """
     get_addons_paths.json_file = "addons_paths.json"
-
-    url = (
-        "https://api.github.com/repos/OSGeo/grass-addons/git/trees/" "main?recursive=1"
+    addons_branch = get_github_branches(
+        "https://github.com/OSGeo/grass-addons", version_only=True
     )
+
+    url = f"https://api.github.com/repos/OSGeo/grass-addons/git/trees/{addons_branch}?recursive=1"
 
     response = download_addons_paths_file(
         url=url,
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -233,8 +233,12 @@ def urlopen(url, *args, **kwargs):
     return urlrequest.urlopen(request, *args, **kwargs)
 
 
-def get_github_branches(github_api_url, version_only=False):
-    """Get ordered list of branch names in repo using github API"""
+def get_github_branches(
+    github_api_url="https://api.github.com/repos/OSGeo/grass-addons/branches",
+    version_only=True,
+):
+    """Get ordered list of branch names in repo using github API
+    For the official repo we assume that at least one version branch is present"""
     req = urlrequest.urlopen(github_api_url)
     content = json.loads(req.read())
     branches = [repo_branch["name"] for repo_branch in content]
@@ -249,16 +253,23 @@ def get_github_branches(github_api_url, version_only=False):
 
 
 def get_default_branch(full_url):
-    """Get default branch for repo (currently only implemented for github API and gitlab API"""
-    if "github.com" in full_url:
-        organization, repository = full_url.split("github.com/")[1].split("/")[0:2]
-        print(organization, repository)
-        req = urlrequest.urlopen(f"https://api.github.com/repos/{organization}/{repository}")
+    """Get default branch for repo (currently only implemented for github API and gitlab API)"""
+    if "/github.com/" in full_url:
+        organization, repository = re.split(r"/github.com/", full_url)[1].split("/")[
+            0:2
+        ]
+        req = urlrequest.urlopen(
+            f"https://api.github.com/repos/{organization}/{repository}"
+        )
         content = json.loads(req.read())
         default_branch = content["default_branch"]
-    elif "gitlab.com" in full_url:
-        organization, repository = full_url.split("gitlab.com/")[1].split("/")[0:2]
-        req = urlrequest.urlopen(f"https://gitlab.com/api/v4/projects/{organization}%2F{repository}")
+    elif "/gitlab.com/" in full_url:
+        organization, repository = re.split(r"/gitlab.com/", full_url)[1].split("/")[
+            0:2
+        ]
+        req = urlrequest.urlopen(
+            f"https://gitlab.com/api/v4/projects/{organization}%2F{repository}"
+        )
         content = json.loads(req.read())
         default_branch = content["default_branch"]
 
@@ -2173,7 +2184,13 @@ def resolve_xmlurl_prefix(url, source=None):
     gscript.debug("resolve_xmlurl_prefix(url={0}, source={1})".format(url, source))
     if source == "official":
         # use pregenerated modules XML file
-        url = "https://grass.osgeo.org/addons/grass%s/" % version[0]
+        repo_branches = get_github_branches()
+        # Define branch to fetch from (latest or current version)
+        version_branch = "grass{}".format(version[0])
+        if version_branch not in repo_branches:
+            version_branch = repo_branches[-1]
+
+        url = "https://grass.osgeo.org/addons/{}/".format(version_branch)
     # else try to get extensions XMl from SVN repository (provided URL)
     # the exact action depends on subsequent code (somewhere)
 
@@ -2251,7 +2268,9 @@ def resolve_known_host_service(url, name, branch):
         else:
             actual_start = ""
         if "branch" in match["url_end"]:
-            suffix = match["url_end"].format(name=name, branch=banch if branch else get_default_branch(url))
+            suffix = match["url_end"].format(
+                name=name, branch=branch if branch else get_default_branch(url)
+            )
         else:
             suffix = match["url_end"].format(name=name)
         url = "{prefix}{base}{suffix}".format(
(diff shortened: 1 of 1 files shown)


ninsbl:
It would be nice if we could remove the hardcoded version here:
https://github.com/OSGeo/grass/blob/3d86470027eb390248d1c49e5a4b43a6d2b6e1c7/scripts/g.extension/g.extension.py#L2508-L2511

Then we do not have to change that in master later on and it probably causes less trouble with addons on windows where we have pre-compiled addons going back to version 7.4...

But that requires a branch "grass8" in the addon repo for `g.extension` to work with GRASS 8 (or current master)...

scripts/g.extension/g.extension.py
```diff
@@ -2319,20 +2321,16 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
         module_class = get_module_class_name(name)
 
         # note: 'trunk' is required to make URL usable for 'svn export' call
-        if branch in ["master", "main"]:
-            svn_reference = "trunk"
+        if branch is None or branch in ["master", "main"]:
+            svn_reference = "branches/grass{}".format(version)
```
wenzeslaus, line 2325: This is confusing. None (default) may be expected to resolve to the default branch which is represented by `trunk` in Subversion interface, although perhaps default resulting in the version branch makes sense in this context. However, explicit master or main should resolve to branch/master and branch/main unless there is a clear reason to do it otherwise.

I though -o is somewhat "behave as with the official repository" which is broader than its defintion "url refers to a fork of the official extension repository". Also depends on what people think fork is. Overall, this might be limiting when you let's say create an institutional repo which should work as the official repository, but you support only one version.

scripts/g.extension/g.extension.py
```diff
@@ -2319,20 +2321,16 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
...
-        git_url = (
-            "{url}/{branch}/"
-            "grass{version}/{module_class}/{module_name}".format(
-                url=url,
-                version=version[0],
-                module_class=module_class,
-                module_name=name,
-                branch=svn_reference,
-            )
+        git_url = "{url}/{branch}/" "src/{module_class}/{module_name}".format(
```
wenzeslaus, line 2329: Delete the `" "` part in the middle.

wenzeslaus (changes requested):

wenzeslaus:
> But that requires a branch "grass8" in the addon repo for `g.extension` to work with GRASS 8 (or current master)...

To maximize freshness of addons for v7 and minimize the backporting needs, it would be best to have a fix in code for v8 and create grass8 branch only close to the release (perhaps we are already?). The v8 specific fix can be later removed or rewritten for v9 because the idea now is that we just do the same grassX to grassY branch switch for v8->v9 as we will do for v7->v8.

ninsbl:
> > But that requires a branch "grass8" in the addon repo for `g.extension` to work with GRASS 8 (or current master)...
> 
> To maximize freshness of addons for v7 and minimize the backporting needs, it would be best to have a fix in code for v8 and create grass8 branch only close to the release (perhaps we are already?). The v8 specific fix can be later removed or rewritten for v9 because the idea now is that we just do the same grassX to grassY branch switch for v8->v9 as we will do for v7->v8.

In 1b39a61 I tried to solve this by listing branches via github API (should be OK, if addons are supposed to be fetched from online repo).
The logic is to pick the branch of the current version if present in the repo (and if no branch is specified explicitly), otherwise take the latest branch. That should work for grass8 and 9 and hardcoded versions are not necessary, unless you would want to fetch addons for e.g. GRASS 7.8. from grass8 branch by default...

scripts/g.extension/g.extension.py
```diff
@@ -2299,46 +2317,54 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
...
     if not url and name:
         module_class = get_module_class_name(name)
-        # note: 'trunk' is required to make URL usable for 'svn export' call
+        repo_branches = get_branches(
+            "https://api.github.com/repos/OSGeo/grass-addons/branches", True
+        )
+        # Define branch to fetch from (latest or current version)
+        version_branch = "grass{}".format(version[0])
+        if version_branch not in repo_branches:
+            version_branch = repo_branches[-1]
```
wenzeslaus, line 2329: That's nice. Querying GitHub API seems to be worth it to have this.

scripts/g.extension/g.extension.py
```diff
@@ -2299,46 +2317,54 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
...
+            # Define branch to fetch from (latest or current version)
+            if not repo_branches:
+                svn_reference = "trunk"
+            else:
+                version_branch = "grass{}".format(version[0])
+                if version_branch not in repo_branches:
+                    version_branch = repo_branches[-1]
+                svn_reference = "branches/{}".format(version_branch)
+        elif branch in ["master", "main"]:
             svn_reference = "trunk"
```
wenzeslaus, line 2358: master and main are really just names for branches which are branches like any other branch. As far as I understand, in this GitHub Git-Subversion bridge, trunk refers to the default branch. It is not a nickname for master or main, so there is no reason to translate master and main to trunk. What needs to resolve to trunk is the state when no branch is provided, i.e., you are asking for the default branch which is what trunk refers to.

https://docs.github.com/en/github/importing-your-projects-to-github/working-with-subversion-on-github/support-for-subversion-clients
https://docs.github.com/en/github/administering-a-repository/managing-branches-in-your-repository/changing-the-default-branch

scripts/g.extension/g.extension.py
```diff
@@ -1403,8 +1417,8 @@ def download_source_code_official_github(url, name, outdev, directory=None):
     """
     if not directory:
         directory = os.path.join(os.getcwd, name)
-    classchar = name.split(".", 1)[0]
-    moduleclass = expand_module_class_name(classchar)
+    # classchar = name.split(".", 1)[0]
+    # moduleclass = expand_module_class_name(classchar)
```
wenzeslaus, line 1421: I'm not sure what you mean by these commented out lines. Variables seems unused. Just delete?

scripts/g.extension/g.extension.py
```diff
@@ -1555,7 +1569,7 @@ def download_source_code(
             response = urlopen(url)
         except URLError:
             # Try download add-on from 'master' branch if default "main" fails
-            if branch == "main":
+            if branch is None or branch == "":
```
wenzeslaus, line 1572: Can be written as `if branch:` but I think the logic does not make sense. See my comments about trunk. Here perhaps result of having main as the default for the option.

scripts/g.extension/g.extension.py
```diff
@@ -2218,7 +2234,9 @@ def resolve_known_host_service(url, name, branch):
         else:
             actual_start = ""
         if "branch" in match["url_end"]:
-            suffix = match["url_end"].format(name=name, branch=branch)
+            suffix = match["url_end"].format(
+                name=name, branch=branch if branch else "main"
```
wenzeslaus, line 2238: Perhaps a comment here about the assumptions. This resolves to a zip, so branch name is needed and there is no easy default. Getting default branch name for different services would be more work, so here I would say "main" is close enough, esp. since it is unrelated to the goal of this PR.

neteler:
(for an easier backporting, `g.extension`in the G78 branch needs to receive a `black` update first)

scripts/g.extension/g.extension.py
```diff
@@ -234,6 +233,21 @@ def urlopen(url, *args, **kwargs):
     return urlrequest.urlopen(request, *args, **kwargs)
 
 
+def get_branches(github_api_url, version_only=False):
```
wenzeslaus, line 236: Maybe get_github_branches?

wenzeslaus (changes requested):

scripts/g.extension/g.extension.py
ninsbl: Yes. Sounds good.

scripts/g.extension/g.extension.py
ninsbl: OK

scripts/g.extension/g.extension.py
ninsbl: Yes. If we get branches, we can also get the default branch. For github, it could be done like this:

```
def get_github_default_branch(organization="OSGeo", repository="grass-addons"):
    """Get default branch for github repo using github API"""
    req = urlrequest.urlopen(f"https://api.github.com/orgs/{organization}/repos")
    content = json.loads(req.read())
    default_branch = [repo["default_branch"] for repo in content if repo["name"] == repository][0]
    return default_branch
``` 
That function could be included / merged with get_branches function.

But for the official repo, that is probably not necessary (as we use the version branches) and as you say, `trunk` refers to the default branch anyway, so as long as we use svn export that should be OK. For all other repos user specific repos, users can specify a branch. But there we would need a solution for bitbucket, gitlab ... too...

scripts/g.extension/g.extension.py
ninsbl: Yes, see also comment above.

scripts/g.extension/g.extension.py
ninsbl: See: ca14583

ninsbl:
Thanks for your feedback!

scripts/g.extension/g.extension.py
wenzeslaus: I agree that it is not needed for the official repo. For other repos, I think it would be fine to get the default branch for GitHub and for others, at least for now, either require branch to be set or use master or main as a (internally set) default value. I think that is still reasonable behavior which can documented clearly enough. However, I'm fine with requiring users to set branch explicitly too, but not having to do that is closer to Git and pip behavior.

`get_github_default_branch` looks like a separate function from _get branches_.

scripts/g.extension/g.extension.py
wenzeslaus: Can you be more specific on how you addressed this? The condition is still unnecessary complicated and I don't understand how it fits to the code which follows. Perhaps update the comment.

scripts/g.extension/g.extension.py
```diff
@@ -1612,7 +1624,9 @@ def install_extension_std_platforms(name, source, url, branch):
     """Install extension on standard platforms"""
     gisbase = os.getenv("GISBASE")
     # TODO: workaround, https://github.com/OSGeo/grass-addons/issues/528
-    source_url = "https://github.com/OSGeo/grass-addons/tree/master/grass7/"
+    # source_url = "https://github.com/OSGeo/grass-addons/tree/grass{}/src".format(
+    #     version[0]
+    # )
```
wenzeslaus, line 1629: Delete including the comment?

scripts/g.extension/g.extension.py
```diff
@@ -2299,46 +2315,47 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
...
         module_class = get_module_class_name(name)
 
         # note: 'trunk' is required to make URL usable for 'svn export' call
-        if branch in ["master", "main"]:
-            svn_reference = "trunk"
+        if branch is None:
+            repo_branches = get_github_branches(
+                url.rstrip("/").replace("github.com/", "api.github.com/repos/")
+                + "/branches",
+                True,
```
wenzeslaus, line 2345: Use keyword argument for this argument for readability.

wenzeslaus (changes requested):

scripts/g.extension/g.extension.py
```diff
@@ -234,6 +233,40 @@ def urlopen(url, *args, **kwargs):
...
+            for version_branch in branches
+            if version_branch.startswith("grass")
+        ]
+    branches.sort()
+    return branches
+
+
+def get_default_branch(full_url):
+    """Get default branch for repo (currently only implemented for github API and gitlab API"""
+    if "github.com" in full_url:
```
wenzeslaus, line 253: Although CodeQL thinks this is a security-related context (`Incomplete URL substring sanitization`), it is right about this being not 100% robust. I'm afraid adding `://.../` is still similar which would mean using regexp. I think the split does _not_ suffer from the same issue, so maybe doing the split on first `://` but perhaps there is no reason to avoid regexp here.

scripts/g.extension/g.extension.py
```diff
@@ -1555,7 +1586,7 @@ def download_source_code(
             response = urlopen(url)
         except URLError:
             # Try download add-on from 'master' branch if default "main" fails
-            if branch == "main":
+            if not branch:
```
wenzeslaus, line 1589: If the branch is coming from `get_default_branch`, checking main makes more sense to me here, but I can't figure out how branch is passed in the code. `resolve_known_host_service` seem to not use the parameter... Perhaps a topic for another PR.

scripts/g.extension/g.extension.py
```diff
@@ -2218,7 +2251,7 @@ def resolve_known_host_service(url, name, branch):
         else:
             actual_start = ""
         if "branch" in match["url_end"]:
-            suffix = match["url_end"].format(name=name, branch=branch)
+            suffix = match["url_end"].format(name=name, branch=get_default_branch(url))
```
wenzeslaus, line 2254: This is what seems to me like always using default, never user-provided branch name.

scripts/g.extension/g.extension.py
```diff
@@ -2299,46 +2332,48 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
...
         module_class = get_module_class_name(name)
-        # note: 'trunk' is required to make URL usable for 'svn export' call
+        repo_branches = get_github_branches(
+            "https://api.github.com/repos/OSGeo/grass-addons/branches",
+            version_only=True,
+        )
+        # Define branch to fetch from (latest or current version)
+        version_branch = "grass{}".format(version[0])
+        if version_branch not in repo_branches:
+            version_branch = repo_branches[-1]
```
wenzeslaus, line 2345: Just add a comment here that we assume the official repo always has at least one version branch. May be useful if this code gets reused for something else.

wenzeslaus (changes requested):

scripts/g.extension/g.extension.py
ninsbl: yes, right. Thanks for catching this.

scripts/g.extension/g.extension.py
ninsbl: Yes. I think, in general, refactoring g.extension could be a very good task for the next community sprint (including improved git support and other issues and feaure requests that are open...)

scripts/g.extension/g.extension.py
ninsbl: Not sure I understood correctly. But I tried to address it by using `re` and adding `/` to the host...
Hope that solves it...

wenzeslaus: That's better, but still not perfect, although apparently CodeQL missises that. The new regexp split is doing the same as the original one since there is no regular expression there. The plain split is actually fine as is and you could use the same trick to get the domain. That's what I meant.

But there is actually a better solution:

```python
>>> from urllib.parse import urlparse  # Fine as is in Python 3
>>> urlparse("https://github.com/OSGeo/grass-addons/")
ParseResult(scheme='https', netloc='github.com', path='/OSGeo/grass-addons/', params='', query='', fragment='')
>>> url_parts = urlparse("https://github.com/OSGeo/grass-addons/")
>>> url_parts.netloc
'github.com'
>>> url_parts.netloc == "github.com"
True
```

You can parse it once and then reuse it in both tests and for both split calls.

scripts/g.extension/g.extension.py
```diff
@@ -234,6 +233,40 @@ def urlopen(url, *args, **kwargs):
...
+        print(organization, repository)
+        req = urlrequest.urlopen(f"https://api.github.com/repos/{organization}/{repository}")
+        content = json.loads(req.read())
+        default_branch = content["default_branch"]
+    elif "gitlab.com" in full_url:
+        organization, repository = full_url.split("gitlab.com/")[1].split("/")[0:2]
+        req = urlrequest.urlopen(f"https://gitlab.com/api/v4/projects/{organization}%2F{repository}")
+        content = json.loads(req.read())
+        default_branch = content["default_branch"]
+
```
wenzeslaus, line 275: if you don't mind, please delete this empty line.

scripts/g.extension/g.extension.py
wenzeslaus: I already started to put some useful functions (originally taken from g.extension) to the library, so we are on a good track. The GitHub/GitLab API stuff could eventually go there too (later).

wenzeslaus (changes requested):

wenzeslaus:
The tests ends with _rate limit exceeded_ from get_github_branches:

```
WARNING: The download of the json file with add-ons paths from the github
         server wasn't successful, rate limit exceeded. The previous
         downloaded json file will be used if exists.
  File "/home/runner/install/grass80/scripts/g.extension", line 2565, in <module>
    sys.exit(main())
  File "/home/runner/install/grass80/scripts/g.extension", line 2540, in main
    name=options["extension"], url=original_url, branch=branch, fork=flags["o"]
  File "/home/runner/install/grass80/scripts/g.extension", line 2357, in resolve_source_code
    repo_branches = get_github_branches()
  File "/home/runner/install/grass80/scripts/g.extension", line 242, in get_github_branches
    req = urlrequest.urlopen(github_api_url)
...
urllib.error.HTTPError: HTTP Error 403: rate limit exceeded
```

wenzeslaus:
If we merge #1715 first, we can get tests running even here (after merge/rebase) and perhaps gain a greater freedom in what can we try or merge here.

scripts/g.extension/g.extension.py
```diff
@@ -2299,46 +2351,45 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
...
     """
+    # Handle URL for the offical repo
     if not url and name:
         module_class = get_module_class_name(name)
-        # note: 'trunk' is required to make URL usable for 'svn export' call
+        repo_branches = get_github_branches()
+        # Define branch to fetch from (latest or current version)
+        version_branch = "grass{}".format(version[0])
+        if version_branch not in repo_branches:
+            version_branch = repo_branches[-1]
```
wenzeslaus, line 2361: Given that the API has rate limits, you can consider abandoning the idea of this ideal test and go with something like:

```suggestion
        # Hard-coded value which needs to be updated when new official addon branch is created.
        addons_available_up_to_version = 7  # perhaps as a global ADDONS_AVAILABLE...
        use_version = int(version[0])
        if use_version > addons_available_up_to_version:
            # Fallback to the highest available version.
            use_version = addons_available_up_to_version
        version_branch = f"grass{use_version}"
```

scripts/g.extension/g.extension.py
```diff
@@ -2140,7 +2184,13 @@ def resolve_xmlurl_prefix(url, source=None):
...
     if source == "official":
         # use pregenerated modules XML file
-        url = "https://grass.osgeo.org/addons/grass%s/" % version[0]
+        repo_branches = get_github_branches()
+        # Define branch to fetch from (latest or current version)
+        version_branch = "grass{}".format(version[0])
+        if version_branch not in repo_branches:
+            version_branch = repo_branches[-1]
+
+        url = "https://grass.osgeo.org/addons/{}/".format(version_branch)
```
wenzeslaus, line 2193: Same as the other git_github_branches call.

scripts/g.extension/g.extension.py
```diff
@@ -234,6 +233,51 @@ def urlopen(url, *args, **kwargs):
...
+    req = urlrequest.urlopen(github_api_url)
+    content = json.loads(req.read())
+    branches = [repo_branch["name"] for repo_branch in content]
+    if version_only:
+        branches = [
+            version_branch
+            for version_branch in branches
+            if version_branch.startswith("grass")
+        ]
+    branches.sort()
```
wenzeslaus, line 251: This simple sort wouldn't not work beyond v9 for the purpose of picking the latest available branch. "natural sort" is needed. However, I'm suggesting to remove this anyway. (Useful in general, so perhaps keep the ocde it around.)

wenzeslaus (changes requested):

scripts/g.extension/g.extension.py
ninsbl: Yes, agreed. Usage of github API is abandoned for now. The rate limit for un-authenticated API calls in [github](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#increasing-the-unauthenticated-rate-limit-for-oauth-applications) seems to be 60/h (comaprd to [gitlab](https://docs.gitlab.com/ee/security/rate_limits.html) with a least 10 times as many).
As the fallback solution I went for fetching from version branch if no branch is given and from default branch, if the version branch does not exist. Maybe we should add in the documentation, that branches in forks from the official repo have to follow the structure of the default branch there (with `src` directory at the root)...

scripts/g.extension/g.extension.py
ninsbl: Has been removed...

scripts/g.extension/g.extension.py
```diff
@@ -2218,7 +2277,10 @@ def resolve_known_host_service(url, name, branch):
         else:
             actual_start = ""
         if "branch" in match["url_end"]:
-            suffix = match["url_end"].format(name=name, branch=branch)
+            suffix = match["url_end"].format(
+                name=name,
+                branch=branch if branch else "main",  # get_default_branch(url)
```
wenzeslaus, line 2282: Was your intention to call get_default_branch for GitLab? You could return "main" from get_default_branch for GitHub.

wenzeslaus (changes requested):

scripts/g.extension/g.extension.py
```diff
@@ -234,6 +233,67 @@ def urlopen(url, *args, **kwargs):
...
+            for version_branch in branches
+            if version_branch.startswith("grass")
+        ]
+    branches.sort()
+    return branches
+
+
+def get_default_branch(full_url):
+    """Get default branch for repo (currently only implemented for github API and gitlab API)"""
+    if "/github.com/" in full_url:
```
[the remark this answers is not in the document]
ninsbl: In 290cd96 the urlparse solution got implemented.

scripts/g.extension/g.extension.py
ninsbl: Actually, I was not sure myself. I solved it now by fetching the default branch for gitlab and using `main` for all other hosting services as default.

scripts/g.extension/g.extension.py
```diff
@@ -2299,47 +2361,38 @@ def resolve_source_code(url=None, name=None, branch=None, fork=False):
...
+        if not branch:
+            # Fetch from default branch
+            version_branch = get_version_branch(version[0])
+            try:
+                url = (
+                    url.rstrip("/") if url else "https://github.com/OSGeo/grass-addons"
+                )
+                urlrequest.urlopen(source_url=f"{url}/tree/{version_branch}/src")
+                svn_reference = "branches/{}".format(version_branch)
+            except Exception:
```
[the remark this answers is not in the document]
ninsbl: Yes. Now `URLError` is used.

scripts/g.extension/g.extension.py
```diff
@@ -234,6 +233,67 @@ def urlopen(url, *args, **kwargs):
...
+def get_version_branch(major_version):
+    """Check if version branch for the current GRASS version exists,
+    if not, take branch for the previous version
+    For the official repo we assume that at least one version branch is present"""
+    version_branch = f"grass{major_version}"
+    try:
+        urlrequest.urlopen(
+            source_url=f"https://github.com/OSGeo/grass-addons/tree/{version_branch}/src"
+        )
+    except Exception:
```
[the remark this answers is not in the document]
ninsbl: Yes. Now `URLError` is used.

ninsbl:
Should be ready to merge.
I fixed the issues from review (and some additional) and the download tests pass locally.
Given the blocker status, maybe best to merge now, and then backport all required changes of g.extension to the release branch. We should probably check if it would work with 7.6 as well...

scripts/g.extension/g.extension.py
```diff
@@ -270,25 +271,29 @@ def get_github_branches(
...
-            0:2
+    # Parse URL
+    url_parts = urlparse(full_url)
+    # Get organization and repository component
+    try:
+        organization, repository = url_parts.path.split("/")[
+            1:3
         ]
+    except URLError:
+        gscript.fatal(_("Cannot retrieve organization and repository from URL: <{}>.".format(full_url)))
```
wenzeslaus: Having this part here (and not e.g. in a separate, possibly nested function and called only when needed), prohibits general URLs to be used here (just to get `main`). I'm not sure how relevant that is, though.

scripts/g.extension/g.extension.py
wenzeslaus: Agreed that documentaion is needed for that, but maybe in another PR. The whole "fork" think here is suboptimal. Typical (soft) fork on GitHub will have the branches. Do we mean hard project fork? I think the aim is institutional repositories (private or public), right? That was the case in the past, I think. I'm not sure how to call "a repo with the same structure as the official repo", plus there is the issues with the generated metadata file which is separate if I recall correctly. ...next PR. Perhaps open an issue?

scripts/g.extension/g.extension.py
```diff
@@ -2218,7 +2285,18 @@ def resolve_known_host_service(url, name, branch):
...
             actual_start = ""
         if "branch" in match["url_end"]:
-            suffix = match["url_end"].format(name=name, branch=branch)
+            # Due to strict rate limits for the github API and lack of implementation
+            # for other hosting services, the default branch is currently only
+            # fetched for gitlab
+            default_branch = (
+                get_default_branch(actual_start)
+                if urlparse(actual_start).netloc == "gitlab.com"
+                else "main"
```
wenzeslaus, line 2294: Isn't this what get_default_branch is doing?

wenzeslaus (approved):
We can merge this as is. I have some additional comments which don't need to be addressed in this PR. I marked most of the conversations as resolved.

wenzeslaus:
This can be tested without compiling or download using Binder: [Binder](https://mybinder.org/v2/gh/ninsbl/grass/g_extension_version?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fexample_notebook.ipynb)

scripts/g.extension/g.extension.py
ninsbl: This function is only meant for known hosting services (currently: github, gitlab, bitbucket), so, this should not be an issue...

scripts/g.extension/g.extension.py
ninsbl: Yes. I moved it into the function in the latest commit.

ninsbl: As for the "fork" issue, I agree that it should be more clear.
The case I had in mind was an addon-dev contributing a new addon, for example. Until the fork-flag got implemented, it was not possible to test contributions to the official addon repo without cloning the repo of the addon-dev and installing from local URL. So, the main intention has been to allow to install directly from e.g. a feature branch of a (soft) fork of the official repo...
Should be probably added to the manual....

wenzeslaus: > As for the "fork" issue...The case I had in mind was an addon-dev contributing a new addon...

Just for the record, I think in the past (Subversion times) it was possible to install from repositories which behaved like the official repo (I think @neteler or @landam created some). Today, the "test from contributor's fork" use case you mentioned is definitively more important, esp. since individual modules can be installed from small repos (without a need for mirroring the official repo infrastructure).

scripts/g.extension/g.extension.py
wenzeslaus: I was thinking the `main` would be used as a the default for any other service, but in that case, use likely needs to (and can) specify branch explicitly anyway.

wenzeslaus (approved):

wenzeslaus:
@ninsbl Leaving the fork-related conversations open (I marked the rest as resolved). Please, resolve them as you see fit. Thank you.

ninsbl:
Thanks, @wenzeslaus for bearing with me here. I mentioned this PR in the ticket about general git related improvements of g.extension. Refactoring g.extension may be a good task for the next community sprint...

But an open question to me is how we backport best? Not all earlier changes in g.extension were backported, see:
https://github.com/OSGeo/grass/commits/master/scripts/g.extension/g.extension.py
vs.
https://github.com/OSGeo/grass/commits/releasebranch_7_8/scripts/g.extension/g.extension.py
Should the version from master be copied into the releasebranch? And should we try to support the new addon-repo also for GRASS 7.6 (and Python 2), as that is still the most recent version on some systems?

neteler:
Update: the `grass8` branch has recently been created in [grass-addons](https://github.com/OSGeo/grass-addons/tree/grass8) (new default branch).

Shall this PR be backported now? Would any changes be needed for that (i.e., in new PR), @ninsbl?

neteler:
Gentle ping, @ninsbl :)

neteler:
Ah, if I understand correctly, it has already been backported in #1762 (?)

wenzeslaus:
This is in 8.0 and 8.2 branches and seems to be in 7.8 as 42f015eb5e02c5e9f6d828dbe15a3120e8ad60d9 (#1700), so removing the backport label.

PR #1705, opened 2021-07-03
i18n: Get ready for GitHub to Transifex integration

HuidaeCho:
Based on my communication with Transifex support, to enable automatic integration (PRs) from Transifex to GitHub, we need to first sync from GitHub to Transifex. However, since we already have translated messages on their platform, we need to remove all PO files from the repo to avoid any overwriting. Then, the removed PO files will be generated by Transifex PRs. At least, that's my understanding. Also, POT files need to be added as source files.


HuidaeCho:
For our records,

> Hello,
> 
> Accidentally I clicked the send button to my previous message.
> Below you can find the whole response.
> 
> Regarding your first sentence, "As for the first question, my question was whether or not individual PO files can have non-empty msgstr's. I think it's obvious that they can have them unlike source files which should empty msgstr, but I wanted to confirm this with you.
> 
> The answer is what you described.
> 
> As far as the second question, it seems that you have already resources in Transifex and you want to Sync them with the existing files in your Github account.
> In that case, you have 2 options.
> 
> 1) From the following link https://docs.transifex.com/transifex-github-integrations/github-tx-ui#syncing-existing-resources-with-github you can find how to SYNC existing resources between Transifex and Github. I would recommend you in order to ensure that translations in Transifex will remain, to remove any translation file, from the path(repository path) you specify in the YAML configuration. So, during the first SYNC, translations will not exist in GITHUB so, will not be downloaded to Transifex. After the first SYNC, once you meet the SYNC content goal, translations will be upload to Github.
> 
> 2) The second option, is to create a new Project in Transifex.
> 
>     Enable TM in this project
>     Create a Translation Sharing Memory group, including the existing project with the translations and the New project.
>     Under the new project, configure the Github integration, but also, in that case, remove any translation file, from the path(repository path) you specify in the YAML configuration file.
>     Once the source files will be created automatically from Github to your new project, TM will start translate the strings that have a 100% similarity match
>     Then once you meet the SYNC content goal from Github Integration of the new project, translations will be upload automatically to Github.
> 
> 
> I hope that covers your questions.
> 
> Kind regards,
> Panagiotis

HuidaeCho:
This PR would be just too disruptive and I'm even sure if it'll work smoothly with Transifex. We're considering migrating to https://weblate.osgeo.org/projects/grass-gis/. Closing this PR for now. This PR is not merge-able anyway and I'll create a new one if needed.

PR #1708, opened 2021-07-04
r.mfilter: implement parallelization with OpenMP

aaronsms:
Checklists before merging:
- [x] code review
- [x] CI passes
- [x] rebase to main


aaronsms:
The previous workflow run has all passsing checks, except for OSGeo4W, where the download link is no longer available.

raster/r.mfilter/testsuite/test_r_mfilter.py
```diff
@@ -0,0 +1,434 @@
...
+        """Test output with sequential filter type with null mode enabled."""
+        test_case = "test_sequential_null"
+        output = "{}_raster".format(test_case)
+        output_z = "{}_z_raster".format(test_case)
+        self.to_remove.extend([output, output_z])
+
+        filter = self.create_filter(self.filter_options["sequential"])
+        self.assertModule(
+            "r.mfilter",
+            input="slope",
```
marisn, line 281: My version of nc_basic_spm_grass7 location does not have a slope map. Can it be replaced with something else to allow testing on a basic location?

marisn (approved):
Code looks OK. I only could not run some tests as I don't have "slope" map – would be nice to replace with one present in the basic version of NC location.

petrasovaa:
I tested it with bigger data and it crashes with `Error reading temporary file`. I run it on a slope map (derived from US 30m DEM) and it shows up with region with 
16.832.104.560 cells and nprocs=1 and with 900.000.000 cells and nprocs=2. I ran a benchmark with this data successfully for 1.000.000 cells.
I also ran it with the original version of r.mfilter and that works fine.

raster/r.mfilter/testsuite/test_r_mfilter.py
HuidaeCho: @aaronsms Slope is missing in [the basic location](https://grass.osgeo.org/sampledata/north_carolina/nc_basic_spm_grass7.zip). It can be a good idea to use the basic location for writing test scripts.

@marisn Any specific reasons why you don't use the full nc_spm_08_grass7?

HuidaeCho:
> 16.832.104.560 cells and nprocs=1

@aaronsms Please check nprocs=1 code again. In this case, I think we need to revert back to the original behavior for nprocs=1. I thought we're not creating more tempfiles for r.mfilter, but are they bigger now even for nrpcos=1?

petrasovaa:
Also, I ran it with valgrind and it gave me one error, and although it didn't look too related as far as I remember, it's worth checking that.

raster/r.mfilter/testsuite/test_r_mfilter.py
aaronsms: I will replace 'slope' with 'lakes'.

petrasovaa: Why not `elevation`?

aaronsms:
> I thought we're not creating more tempfiles for r.mfilter, but are they bigger now even for npcos=1?
For r.mfilter, we are not changing the default behavior. The original one uses a temporary file as buffer as well. 

> it shows up with region with
> 16.832.104.560 cells and nprocs=1 and with 900.000.000 cells and nprocs=2
@petrasovaa, can I clarify that whether you meant, it works with the mentioned two cases or, you achieve different results with (nprocs = 1 vs = 2)? Also, may I also ask at what cell count did it crash with the `Error reading temporary files`? Thanks

aaronsms:
Currently, I have found some issues when grass is compiled without OpenMP, will follow up soon. But I haven't found issues with the one compiled with OpenMP.

petrasovaa:
> @petrasovaa, can I clarify that whether you meant, it works with the mentioned two cases or, you achieve different results with (nprocs = 1 vs = 2)? Also, may I also ask at what cell count did it crash with the `Error reading temporary files`? Thanks


16.832.104.560 cells and nprocs=1
and 
900.000.000 cells and nprocs=2
failed with the `Error reading temporary files`

With just 1.000.000 cells everything worked, although I haven't looked at the resulting raster closely.

aaronsms:
@petrasovaa I have solved the overflow issues, so it should now work for larger raster map as per previous behavior.

petrasovaa:
Here is my benchmark on a slope map of 16.000.000.000 cells and 5x5 filter, only 3 repeats, on a machine with 28 cores:
results

petrasovaa:
Another benchmark, this time with larger filter (15x15), the rest is the same (slope map of 16.000.000.000 cells, 3 repeats, on a machine with 28 cores)
results15_2
:

aaronsms:
> Another benchmark, this time with larger filter (15x15), the rest is the same (slope map of 16.000.000.000 cells, 3 repeats, on a machine with 28 cores):
> results15_2

Hi @petrasovaa, I was wondering if I can make use of this benchmark result for the twitter post to show the result on large map? Thanks.

petrasovaa:
Question, I can see that in PR #1724 (r.neighbors) you use memory option but not here. Also I can see huge temporary file being created when I run this on large data. @aaronsms are there any improvements in the r.neighbors PR that would be relevant here?

aaronsms:
> Question, I can see that in PR #1724 (r.neighbors) you use memory option but not here. Also I can see huge temporary file being created when I run this on large data. @aaronsms are there any improvements in the r.neighbors PR that would be relevant here?

For r.mfilter, the original behavior is we use at most 2 temp files for processes requiring 2 or more sequential filters. This PR maintains that behavior of using temp files. 

I think there can be improvements from r.neighbor that would be relevant here (similar approach but will change the current behavior, e.g., if the users want to use memory instead of disk. For example, if a chunk of memory is used, then we can propagate from start all the way to finish, before moving on to the next chunk. Could be worth exploring, then it would remove the need for temp files altogether.

petrasovaa:
Thanks for replying, now I remember I ran into this before. Since this was the original behavior, I plan to merge this as it is now.

petrasovaa:
I updated manual page with performance section, not sure if there is something else that needs to go in there.
It would be useful to have a keyword for parallel modules, I think this was even discussed, not sure if there was any conclusion. I suggest `parallel`, but I am open to other options.

petrasovaa (approved):

neteler:
> It would be useful to have a keyword for parallel modules, I think this was even discussed, not sure if there was any conclusion. I suggest `parallel`, but I am open to other options.

Very good idea! Another keyword might be `multicore` but `parallel` sounds better to me. Others?

nilason:
> parallel

👍

PR #1710, opened 2021-07-05
Interactive vector maps for Jupyter Notebooks

chaedri:
* InteractiveMap class produces interactive folium maps centered on current g.region in Jupyter Notebooks
* Functions for adding vectors to folium map by
* Vectors are passed from GRASS to folium by reprojecting to WGS84 in a temporary Location, exporting as GeoJSON files and importing to folium.

This PR contains functions for displaying GRASS vector data in interactive folium maps for Jupyter Notebooks. It is part of an ongoing Google Summer of Code project, Improved Integration of GRASS and Jupyter Notebooks. You can find more information [here](https://trac.osgeo.org/grass/wiki/GSoC/2021/JupyterAndGRASS).


chaedri:
This is still just  a draft but I'm opening a PR to get the discussion rolling. Passing the GeoJson to folium isn't working yet and there is no clean-up for removal of the temporary location, mapset, and environment either.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,228 @@
...
+
+import os
+import sys
+import subprocess
+import folium
+import grass.script as gs
+import shutil
+import codecs
+
+def create_location(
```
wenzeslaus, line 21: What's wrong with `grass.script.create_location`?

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,228 @@
...
+    
+        
+    def add_vector(self, vector, mapset=None):
+        """Imports vector into temporary WGS84 location, 
+        re-formats to a GeoJSON and adds to folium map"""
+        if mapset is None:
+            mapset = self._rc_original["MAPSET"]
+        
+        # Import vector into new Location/Mapset
+        name = vector+"@"+mapset
```
wenzeslaus, line 205: Use f-string.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,228 @@
...
+        """This function returns the bounding box of the current region
+        in WGS84, the required projection for folium"""
+
+        # Get proj of current GRASS region
+        proj = gs.read_command("g.proj", flags="jf", env=env)
+
+        # Get extent
+        extent = gs.parse_command("g.region", flags="g", env=env)
+        extent_ne = "{}, {}".format(extent["e"], extent["n"])
+        extent_sw = "{}, {}".format(extent["w"], extent["s"])
```
wenzeslaus, line 179: Since you have the `_convert_coordinates` function, it is better to do (hide) this formatting stuff there.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,228 @@
...
+
+        # Convert extent to EPSG:3857, required projection for Folium
+        north, east = self._convert_coordinates(extent_ne, proj)
+        south, west = self._convert_coordinates(extent_sw, proj)
+
+        extent = {'North': north, 'South': south, 'East': east, 'West': west}
```
wenzeslaus, line 185: The upper case here is somewhat unexpected. Either all lower case or switch the dictionary to types.SimpleNamespace and use attributes where it is clear it needs to be lower case.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,228 @@
...
+                        [extent['South'], extent['East']]]
+        
+        return bounding_box
+    
+        
+    def add_vector(self, vector, mapset=None):
+        """Imports vector into temporary WGS84 location, 
+        re-formats to a GeoJSON and adds to folium map"""
+        if mapset is None:
+            mapset = self._rc_original["MAPSET"]
```
wenzeslaus, line 202: This will actually break for more advanced cases. Adding mapset is actually more sophisticated. If v.proj can handle that, that would be best.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,228 @@
...
+        name = vector+"@"+mapset
+        gs.run_command("v.proj",
+                       input=name,
+                       location=self._rc_original["LOCATION_NAME"],
+                       mapset=self._rc_original["MAPSET"],
+                       env=self._env
+                      )
+        
+        # Convert to GeoJSON
+        gs.run_command("v.out.ogr", input=name, output="tmp_{}.json".format(vector), format="GeoJSON", env=self._env)
```
wenzeslaus, line 214: The code will change, but `"tmp_{}.json".format(vector)` should be a variable.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,228 @@
...
+            encoding="utf-8",
+            mode="w",
+        )
+        
+        fd.write(os.linesep)
+        fd.close()
+    except OSError as e:
+        raise ScriptError(repr(e))
+
+class GrassRendererInteractive():
```
wenzeslaus, line 85: I still think there might be a better, completely different, name, but in any case, the adjectives should go first, noun last, e.g., GrassInteractiveRenderer.

wenzeslaus (changes requested):
For some reason, I agree with the general idea, so here are just some details, initial thoughts and questions.

veroandreo:
Trying the jupyter notebook, I get an error with `gj.display_settings()` since the function was removed. Also, I do not see any streams displayed, is that because of the above error?

image

And, the creation of the interactive display takes ~10 seconds. Is that normal or is it something most likely on my end?

python/grass/jupyter/interact_display.py
chaedri: `create_location` calls `create_environment`. The user cannot pass an environment to `create_location` so I couldn't get a new location to be created in `self._env`; it will always create a new environment with `os.environ.copy()` and doesn't return the new environment either. I think this means it wouldn't work? Or is there another workflow you had in mind?

chaedri:
@veroandreo The `display_settings` functions shouldn't have been there anymore - I just removed it. The contents of that function are now included in the GrassRenderer and GrassRendererInteractive class.

The streams don't show up for me either. I'm currently trying to figure out why but yes, that's what I meant by the GeoJson not getting properly passed to folium.

And, 10 seconds seems a little long. It was running faster on my computer (~2 sec) but when I ran it once in Binder and it was slower. I'm not sure why that is...

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,229 @@
...
+        # Create Folium Map 
+        self.map = folium.Map(
+            width=self.width,
+            height=self.height,
+            location=center,
+            tiles="cartodbpositron",
+        )
+
+        
+    def _get_rc(self, env=None):
```
wenzeslaus, line 128: Likely not needed.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,229 @@
...
+        north, east = self._convert_coordinates(extent_ne, proj)
+        south, west = self._convert_coordinates(extent_sw, proj)
+
+        extent = {'North': north, 'South': south, 'East': east, 'West': west}
+
+        return extent
+    
+    def _folium_bounding_box(self, extent):
+        """Reformats extent into bounding box to pass to folium"""
+
```
wenzeslaus, line 191: Extra line.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,229 @@
...
+        extent = {'North': north, 'South': south, 'East': east, 'West': west}
+
+        return extent
+    
+    def _folium_bounding_box(self, extent):
+        """Reformats extent into bounding box to pass to folium"""
+        
+        bounding_box = [[extent['North'], extent['West']],
+                        [extent['South'], extent['East']]]
+
```
wenzeslaus, line 194: I don't think this line improves readability.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,229 @@
...
+        raise ScriptError(repr(e))
+
+class GrassRendererInteractive():
+    """This class creates interative GRASS maps with folium"""
+
+    def __init__(self, width=400, height=400):
+        """This initiates an instance of GrassRenderer"""
+        self.width=width
+        self.height=height
+
```
wenzeslaus, line 92: I was saying this at another occasion somewhere, but I will use the same line: Despite the popular belief, Python does not require an empty line after each statement.

Well places empty lines here and there, may improve readability. More space around each line may be achieved in a many text editors by increasing line spacing.

wenzeslaus (changes requested):

python/grass/jupyter/interact_display.py
chaedri: Hmm, if I leave the mapset kwarg empty in v.proj, I get an error that the vector cannot be found in "user1" (because the vectors I've been using are all in the PERMANENT mapset). Is this what you had in mind? Not sure how else to do this...

python/grass/jupyter/interact_display.py
chaedri: What about InteractiveMap? The idea that is for GRASS is already implied since the library is called grass.jupyter.

python/grass/jupyter/interact_display.py
wenzeslaus: Right, v.proj can't do this automatically in this context. It just defaults to the current mapset (and it is suspicious that it is user1 in your case, but that's a separate issue).

You need to use `gs.find_file` which will do the correct lookup for you. This needs to run in the current mapset and than the mapset name you acquire will go to v.proj running in that WGS84 location.

python/grass/jupyter/interact_display.py
chaedri: I switched back to `gs.create_location` and it doesn't return an error anymore and it seems to work fine. I'm not sure why I was having so much trouble with that before... So, all set now - thanks!

python/grass/jupyter/interact_display.py
chaedri: Ah! Ok, that makes sense. I think I have it working with `gs.find_file` now.

It is "user1" because that's what I called in gj.init.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,229 @@
...
+    def __init__(self, width=400, height=400):
+        """This initiates an instance of GrassRenderer"""
+        self.width=width
+        self.height=height
+        
+        # Get the parent location/mapset
+        self._rc_original = self._get_rc(os.environ)
+        
+        # Create new environment for tmp WGS84 location
+        rcfile, self._env = gs.create_environment(self._rc_original["GISDBASE"], "temp_folium_WGS84", "PERMANENT")
```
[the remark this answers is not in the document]
chaedri: Maybe `self._folium_env`?

Noted. I think I'm creating a lot of things that probably have to be cleaned up later...

python/grass/jupyter/interact_display.py
chaedri: I'm not sure how else to get `GISDBASE` which is required for `create_location` and `create_environment`. It isn't stored in `os.environ`. I also think I need the original location and mapset for v.proj and I'm not sure how to get that besides in the rcfile.

wenzeslaus: > I'm not sure how else to get `GISDBASE` which is required for `create_location` and `create_environment`.

Actually, that's another issue. You should probably use some temporary directory for this. You can use standard Python library to get a temporary directory. (`GISDBASE` is just a directory - a directory where locations ("projects") are.) In standard GRASS session created by the _grass_ command, there is one already created and can be used for things like this, but it needs to be used with some care and _grass.script.setup.init_ is not setting it up unlike the _grass_ command.

> It isn't stored in `os.environ`.

The equivalent of the system environment in GRASS is so called _GIS environment_ (a legacy name) accessible through _g.gisenv_ (_gs.gisenv_ function in Python). It is interface to the rc file (session file).

> I also think I need the original location and mapset for v.proj and I'm not sure how to get that besides in the rcfile.

Right, but the mapset should come from the _find_file_ call.

python/grass/jupyter/interact_display.py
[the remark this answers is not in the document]
wenzeslaus: > Maybe `self._folium_env`?

You will need pseudo-mercator for the rasters/images, so perhaps `self._vector_env` or `self._wgs84_env`.

> I think I'm creating a lot of things that probably have to be cleaned up later...

Right, good to keep a track of those, but you can just create an issue for that when you are ready to merge this PR, i.e., it does not have to be part of this PR (this applies to all other PRs as well). Or you can list known issues in the PR description.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,158 @@
...
+        """Imports vector into temporary WGS84 location, 
+        re-formats to a GeoJSON and adds to folium map"""
+        
+        # Import vector into new Location/Mapset
+        name = gs.find_file(vector, element="vector")["fullname"]
+        # Reproject vector into WGS84 Location
+        gs.run_command("v.proj",
+                       input=name,
+                       location=self._rc_original["LOCATION_NAME"],
+                       mapset=self._rc_original["MAPSET"],
```
wenzeslaus, line 136: Maybe the mapset is not needed when v.proj gets a fully qualified name as input (since mapset is already specified in input). Read the documentation and/or test. Using the original/current mapset would be wrong (the vector is in the mapset returned by find_file, not necessarily in the current one).

I would also suggest to name the variable full_name just to be clear later what is happening or overwrite the vector variable by the result if you want shorter code (I think it would still be perfectly readable), but see my later comment about v.out.ogr. It is also somewhat confusing to name this name since one of the values find_file returns is under key `name`.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,158 @@
...
+                       env=self._env
+                      )
+        
+        # Convert to GeoJSON
+        json_file = f"tmp_{vector}.json"
+
+        gs.run_command("v.out.ogr", input=name, output=json_file, format="GeoJSON", env=self._env)
```
wenzeslaus, line 143: This works I think only because the vector happens to be in the PERMANENT mapset in the original/current location. This may or may not be the case generally. In the `self._env` location, it will always be in PERMANENT. Specifying only the name without mapset seems to me like the best way to go. You just reprojected that to whatever is the current mapset for `self._env`, so specifying the mapset is not needed.

To cover all the cases, you need to really have the original name here, but perhaps stripped from the mapset - right, you should support things like `add_vector("streams@user1")`. Fortunately, just passing that to _find_file_ should work (I just tested with _g.findfile_). It will resolve all the pieces for you. In other words, passing the fullname is wrong, passing the original parameter value won't work for `name@mapset` case, use full _find_file_ result to reliably get the pieces you need.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,158 @@
...
+
+        extent = {'north': north, 'south': south, 'east': east, 'west': west}
+
+        return extent
+    
+    def _folium_bounding_box(self, extent):
+        """Reformats extent into bounding box to pass to folium"""
+        return [[extent['north'], extent['west']], [extent['south'], extent['east']]]
+    
+    def add_vector(self, vector):
```
wenzeslaus, line 126: I'm generally see `add_vector(vector="abc")` as a missed opportunity to document what the parameter is. In a general Python context, I would say `add_vector(name="abc")`. You could also say it should always be called as `add_vector("abc")` and leave your hands free for later change of the parameter name.

wenzeslaus (changes requested):

python/grass/jupyter/interact_display.py
wenzeslaus: To clarify, the missed opportunity in `add_vector(vector="abc")` is that you are saying twice you are dealing with a vector. It is just repeating the information. Actually calling the function with the keyword argument then feels repetitive. Defining the function as `add_vector(self, name)` adds some information about what the argument actually is (it is a name, so string, and not some complex object - although in theory we could support things like that, in Python `arg` is often used as an argument name in that case from what I have seen). On the other hand, name does not add that much info in the actual call `add_vector(name="abc")`, so perhaps using it just as `add_vector("abc")` makes more sense. You could argue that name is name of the layer as displayed in the legend I guess. If you say in documentation, never use keyword argument for the first argument, you can change the name any time without breaking your API. That's the Python side of things.

In GRASS GIS, the parameters always have names like in Python and in some contexts, it is okay to leave out the name of the first one (in Python you can _generally_ (not always) leave out any name). The current practice is that if there is one raster or vector map as input and there is no output, it is called _map_, it is called _input_ if there is also output(s), and it is called _vector_ when there is also raster and vice versa. This was called into question because the rules are unclear to some users (defining "no output" is one of the tricky parts). The non-interactive d.vect-like functions will use this GRASS parameter name, however, we are not bound by these names here and can pick what works best.

chaedri: Ah, ok, I understand better now. So, let's go with `add_vector(self, name)`? That would match both Python and GRASS convention since it is the first parameter (see 1st sentence, 2nd paragraph. And, I agree that adding `name=` doesn't add much information.

chaedri: Also, reading your comment above again, I see your point. I need to use `["name"]` from `gs.find_file` instead of from whatever the user provides so that it will work if the user provides `"abc@PERMANENT"` - that way you don't end up with a file called "tmp_abc@PERMANENT.json".

wenzeslaus:
03f2dd6 seems to revert changes from the previous commit. `git restore -s e8e97b4 -- interact_display.py` would get you the old version if this was a mistake. If you have some changes you care about - it seems there are some - running `black interact_display.py` is a better option as long as you check that 03f2dd6 did not remove any of the changes from e8e97b4 you care about.

chaedri:
Ooops - thanks for catching that @wenzeslaus. I'm not sure how I did that.

chaedri:
@wenzeslaus Oops! Thanks for catching that. I've been working developing/testing in a Jupyter Notebook then copying and pasting into interact_display.py. So, I accidentally removed all the formatting I did.

python/grass/jupyter/Makefile
```diff
@@ -5,7 +5,7 @@ include $(MODULE_TOPDIR)/include/Make/Python.make
 
 DSTDIR = $(ETC)/python/grass/jupyter
 
-MODULES = setup display
+MODULES = setup display interact_display
```
wenzeslaus, line 8: To (partially) future proof this, you can list each module (name) on a separate line. I just did that for the Makefile in parent dir in this PR: https://github.com/OSGeo/grass/pull/1715/files#diff-ca3c8a522b85886bddf093dff36a30011e88cd6c5335d554d2f1ffb08582903d

python/grass/jupyter/setup.py
```diff
@@ -21,7 +21,6 @@ def _set_notebook_defaults():
     """
     This function sets several GRASS environment variables that are
     important for GRASS to run smoothly in Jupyter.
-
     It also allows GRASS to overwrite existing maps of the same name.
     """
```
wenzeslaus, line 25: This change is unrelated to this PR, but if you want to make your docstrings correct, you can follow [PEP 257 -- Docstring Conventions](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings):

```
def complex():
    """Form a complex number.

    More text. ...

    More text in more pragraphs. ...
    """
```

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,163 @@
...
+        """Reformats extent into bounding box to pass to folium"""
+        return [[extent["north"], extent["west"]], [extent["south"], extent["east"]]]
+
+    def add_vector(self, name):
+        """Imports vector into temporary WGS84 location,
+        re-formats to a GeoJSON and adds to folium map"""
+
+        # Find full name of vector
+        full_name = gs.find_file(name, element="vector")["fullname"]
+        name = gs.find_file(name, element="vector")["name"]
```
wenzeslaus, line 125: To much overhead, this is calling g.findfile twice and twice parsing its output in Python. Maybe small overhead compared to time spend with manipulating the vector, but still not a good practice.

Naming that variable is hard, but the code is short (scope where the variable lives), so it can be something generic, like result (as in result from querying the spatial database) or some combination of file, vector, info, names, dict... I'm afraid there is no term for this thing in GRASS.

wenzeslaus (changes requested):

python/grass/jupyter/interact_display.py
wenzeslaus: > ...let's go with `add_vector(self, name)`? ...adding `name=` doesn't add much information.

Sounds good, but say in the documentation that "*name* a positional-only parameter" (as, for example, *x* is described in [documentation of float buildin](https://docs.python.org/3/library/functions.html#float). There is a way in [new Python versions](https://docs.python.org/3/reference/compound_stmts.html#function-definitions) to do that on language-level rather than just in the documentation, but we need to support older version too (you can see full info in 
[PEP 570 -- Python Positional-Only Parameters](https://www.python.org/dev/peps/pep-0570/)).

This way you are not making any promises about the name of the parameter and are saying that `add_vector("abc")` is the right way of using this method.

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,163 @@
...
+        # Reproject vector into WGS84 Location
+        gs.run_command(
+            "v.proj",
+            input=full_name,
+            location=gs.gisenv(env=os.environ)["LOCATION_NAME"],
+            dbase=gs.gisenv(env=os.environ)["GISDBASE"],
```
wenzeslaus, line 131: Same as above for find_file. `gisenv = gs.gisenv...`

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,163 @@
...
+            env=self._vector_env,
+        )
+
+        # Convert to GeoJSON
+        json_file = f"{self.tmp_dir}/tmp_{name}.json"
```
wenzeslaus, line 136: Use Path from pathlib to manipulate the paths or os.path.join, i.e., do not construct paths using string formatting (filename and directory names yes, but not paths).

Assuming that self.tmp_dir is already a Path, the statement will look like:

```
        json_file = self.tmp_dir / f"tmp_{name}.json"
```

The disadvantage of Path, is that you need to convert it to string, but things like format strings work out of the box:

```
>>> from pathlib import Path
>>> a = Path("/tmp")
>>> f"{a}"
'/tmp'
```

wenzeslaus (changes requested):

veroandreo:
I get the following error in the notebook when reaching the interactive part:
```
ValueError: Cannot render objects with any missing geometries: PosixPath('tmp/tmp_boundary_region.json')
```

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -86,17 +86,38 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Demonstration of GrassRenderer for non-interactive map display\n",
     "\n",
-    "m.show()"
+    "r = gj.GrassRenderer(height=540, filename = \"streams_maps.png\")\n",
```
wenzeslaus, line 108: There is a `m = gj.GrassRenderer` line above the g.region line. So delete that one?

doc/notebooks/jupyter_integration.ipynb
```diff
@@ -120,18 +142,18 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Then, we return to the first instance and continue to modify/display\n",
+    "# Then, we return to the first instance and continue to modify and disply it\n",
```
wenzeslaus, line 145: display

python/grass/jupyter/interact_display.py
```diff
@@ -0,0 +1,155 @@
...
+        self.layer_control_object = folium.LayerControl(**kwargs)
+
+    def show(self):
+        """This function creates a folium map with a GRASS raster
+        overlayed on a basemap.
+
+        If map has layer control enabled, additional layers cannot be
+        added after calling show()."""
+
+        if self.layer_control == True:
```
wenzeslaus, line 148: You need to set this to False in init to make this line work or set self.layer_control_object to None and than test it here. In both cases, the test can be just `if self.layer_control:`.

wenzeslaus (changes requested):

python/grass/jupyter/interact_display.py
chaedri: I decreased the number of blank lines in later commits.

python/grass/jupyter/interact_display.py
chaedri: I think this is resolved?

python/grass/jupyter/setup.py
chaedri: To clarify here, I should open a new PR where I update all the docstrings to follow PEP 257? Or should I address this here?

python/grass/jupyter/interact_display.py
chaedri: I removed many empty lines in later commits.

python/grass/jupyter/Makefile
```diff
@@ -5,7 +5,10 @@ include $(MODULE_TOPDIR)/include/Make/Python.make
 
 DSTDIR = $(ETC)/python/grass/jupyter
 
-MODULES = setup display
+MODULES = \
+	 setup \
+	 display \
+	 interact_display
```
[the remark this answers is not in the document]
chaedri: Not sure which space you meant. I deleted a one space before each word on lines 9-11.

python/grass/jupyter/interact_display.py
chaedri: Perhaps we can talk about this at our next meeting. I have some lingering confusion and am not sure if I've done the correct way. I think the way I have it written is a "positional-only parameter"? From reading those links, it's my understanding that, with the way it's currently written, the user couldn't use `add_vector(self, name="abc")` which inherently means it's a "positional-only parameter."

chaedri:
> I get the following error in the notebook when reaching the interactive part:
> 
> ```
> ValueError: Cannot render objects with any missing geometries: PosixPath('tmp/tmp_boundary_region.json')
> ```

Oops - I have fixed this now and changed the way that I test/commit code. Now, it's easier for me to test before committing so hopefully, I will push fewer errors in the future.

python/grass/jupyter/setup.py
wenzeslaus: Fix the newly added docstrings. Leave the rest for later or for PRs where you are modifying the code anyway.

python/grass/jupyter/interact_display.py
wenzeslaus: Let's talk about it then. In the meantime, try running `add_vector(self, vector="abc")` yourself.

python/grass/jupyter/Makefile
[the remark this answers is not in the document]
wenzeslaus: That's exactly what I meant. Thanks. This makes it perfectly clear that the space is not needed there for the next person editing this file or copying code from here.

python/grass/jupyter/setup.py
chaedri: I removed the changes in setup.py and reformatted to docstrings in interact-display.py to reflect PEP 257 (since these are the newly added docstrings).

wenzeslaus (approved):
Ready to merge assuming we will get back to some of the issues such as temporary files and documentation.

@chaedri Please update the main PR description so that it reflects latest state and contains what was implemented, intention, decisions made (brief summary), and whatever might be relevant to future us, other maintainers, or feature announcements, see e.g., 86acc1cb080b312cfa89aed5d21897dbec79986c and 78feee7c26bf07796508fe96f2835cb882b4963c. I will use the text for the commit message when merging.

PR #1711, opened 2021-07-06
libpython: Save and load benchmark results

wenzeslaus:
Series of functions to save benchmark results to JSON, load them back, and combine them together.

This needs some command line interface to make it convenient, but I used it in a custom benchmark script.


wenzeslaus:
This has tests, it is self-contained, and passes even most of Pylint, so merging now.

PR #1712, opened 2021-07-06
pygrass: Add update parameters method to Module

wenzeslaus:
This adds a new update method which can update the parameters (and generally attributes) of the Module.
The current practice is to use the function call operator to do the update. This seems to go against
the function call operator semantics as the module is not executed unless explicit `run_=True` is passed.
It is also inconsistent with using the function call operator without parameters which runs the module unconditionally.

This PR is not changing the current behavior, but it adds a better way of expressing the same idea. The modification of parameters can now be done using a method called _update_. The name _update_ is based on update function in Python dict since this seems to be a similar case.

The change in doctest and in _grass.benchmark_ (original motivation for this change) shows how the new usage looks like. _grass.benchmark_ update of documentation shows also how clear is the interface with update compared to the very specific original interface.

I think the underlying issue - which created the situation when the function call operator can be used for update without actually running the module (depending on presence or absence of parameters) - is that a module by default runs unless explicit `run_=False` is passed in `__init__`. This is the behavior which allows for `from grass.pygrass.modules import Module as run_command` to work and the `grass.pygrass.modules.shortcuts` to work with Module class as is. Changing these and all places where Module parameters are updated and removing the update-only functionality from the function call operator is something to consider, but out of scope for this PR.

### DIFF
--- a/python/grass/benchmark/runners.py
+++ b/python/grass/benchmark/runners.py
@@ -23,7 +23,13 @@ import grass.script as gs
 def benchmark_nprocs(module, label, max_nprocs, repeat):
     """Benchmark module using values of nprocs up to *max_nprocs*.
 
-    *module* is an instance of PyGRASS Module class.
+    *module* is an instance of PyGRASS Module class or any object which
+    has a *update* method taking *nprocs* as a keyword argument,
+    a *run* which takes no arguments and executes the benchmarked code,
+    and attribute *time* which is set to execution time after the *run*
+    function returned. Additionally, the object should be convertible to *str*
+    for printing.
+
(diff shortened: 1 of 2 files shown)


wenzeslaus:
Description changed to provide full info about the change.

The doctest runs locally. I'll wait with merging for full tests running (locally or in CI).

In the mean time, any reviews or feedback are welcome, about this code or what can be done further in this direction.

wenzeslaus:
The CI errors for 9f1d1ea are unrelated to the changes in this PR (OSGeo4W: `(404) Not Found` and Ubuntu 20.04: `Unable to make mapset element .tmp/fv-az180-858`).

petrasovaa (approved):

PR #1715, opened 2021-07-08
Move g.download.location to core, create grass.utils lib

wenzeslaus:
* This moves g.download.location from addons to core to avoid need to install it in every CI.
* g.download.location partially matches functionality of download in Data tab, so the module is important for feature parity.
* Shared code is now in new subpackage of grass, grass.utils (original code is from g.extension which still has its own copy).
* location name is now optional in the is_location_valid function to support cases when the path is full path to the location.

### DIFF
--- a/scripts/g.download.location/g.download.location.py
+++ b/scripts/g.download.location/g.download.location.py
@@ -42,15 +42,15 @@
 # % multiple: no
 # %end
 
+import atexit
 import os
 import shutil
-import atexit
-
+from pathlib import Path
 
 import grass.script as gs
-from grass.script.utils import try_rmdir
-from grass.utils import download_and_extract, name_from_url, DownloadError
 from grass.grassdb.checks import is_location_valid
+from grass.script.utils import try_rmdir
+from grass.utils.download import DownloadError, download_and_extract, name_from_url
 
 
 def find_location_in_directory(path, recurse=0):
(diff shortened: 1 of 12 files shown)


wenzeslaus:
Notably, this also fixes tests in CI.

scripts/g.download.location/g.download.location.py
```diff
@@ -0,0 +1,148 @@
...
+
+    if not database:
+        # use current
+        database = gs.gisenv()["GISDBASE"]
+    if not name:
+        name = location_name_from_url(url)
+    destination = os.path.join(database, name)
+
+    if os.path.exists(destination):
+        gs.fatal(_("Location named <%s> already exists, download canceled") % name)
```
tmszi, line 103: Format string with `format()` function, please.

tmszi (approved):
Is there a reason why you don't use the new [`pathlib`](https://docs.python.org/3/library/pathlib.html) lib instead of `os.path`?

scripts/g.download.location/g.download.location.py
wenzeslaus: Sure. Done.

wenzeslaus:
Thanks for the feedback @tmszi. Merged. The tests on Ubuntu passed successfully and I tested the GUI locally.

> Is there a reason why you don't use the new [`pathlib`](https://docs.python.org/3/library/pathlib.html) lib instead of `os.path`?

No particular reason. This code is just old written to work with Python 2 or at least Python older than 3.6. Although, pathlib is 3.4, os.path supports Path objects only since 3.6 so using Path was too cumbersome. shutil seems to be the same case (given this [resolved issue](https://bugs.python.org/issue23241)), although the documentation reflects that only since 3.8.

Anyway, you are right, we should use it. I rewrote parts of the code to use pathlib.Path, esp. in the module and in the API, but left os.path in most cases internally. This os.path code works and is not obsolete, so I changed only the suffix/extension handling where Path seemed to simplify things.

Interestingly, os.path seem to be more advantageous for library functions as long as you don't need to manipulate the path much (which is where Path is better) or you do a lot of operations on it (which will also push you to use Path). Consider the following lines:

```python
os.path.isdir(x)
x.is_dir()
```

The first line works for both strings and Path (i.e., anything [path-like](https://docs.python.org/3/glossary.html#term-path-like-object)). If the second line is in a library function and _x_ is a parameter, you should make it work with strings too, so you need to ensure that _x_ is Path (by `x = Path(x)`).

Just to be clear, I don't think this is a reason not to use Path. It is just something to be aware of when picking the best solution. One notable case when to use Path is when library code does some string operations on the path. These should never be performed on the parameter value (which can be string or Path), but only one values derived from Path. String operations (like split or startswith) would fail with Path, so this will ensure that the code works with Path too. Likely, the code using Path will end up being simpler in cases when the path needs to be somehow parsed or checked for special suffixes/extensions which is exactly what happened to me when rewriting code in this PR based on feedback from @tmszi.

PR #1721, opened 2021-07-12
Temporary Directory support for Non-Interactive Jupyter displays

chaedri:
This PR contains modifications to display.py that move temporary files from the current working directory to a temporary directory that can be deleted later. The temporary directory is the same one used by the interactive mapping functions in PR #1710. You can test it out in Binder here:

https://mybinder.org/v2/gh/chaedri/grass/add-temp-directory?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb

### DIFF
--- a/python/grass/jupyter/display.py
+++ b/python/grass/jupyter/display.py
@@ -31,17 +31,23 @@ class GrassRenderer:
         else:
             self._env = os.environ.copy()
 
+	# Make temporary folder for all our files
+        self._tmp_dir = Path("./tmp/")
+        try:
+            os.mkdir(self._tmp_dir)
+        except FileExistsError:
+            pass
+        self._filepath = os.join(self._tmp_dir, filename)
+        
         self._env["GRASS_RENDER_WIDTH"] = str(width)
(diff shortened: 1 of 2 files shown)


python/grass/jupyter/display.py
```diff
@@ -31,18 +31,24 @@ def __init__(
         else:
             self._env = os.environ.copy()
 
+        # Make temporary folder for all our files
+        self._tmp_dir = Path("./tmp/")
+        try:
+            os.mkdir(self._tmp_dir)
+        except FileExistsError:
+            pass
```
wenzeslaus, line 39: If user passes _filename_, they likely expect that the file will be created somewhere they desire and that it is a file they will use later. Otherwise there would be no reason to provide filename in the first place.

wenzeslaus (changes requested):

chaedri:
After meeting with Vaclav Petras, I'm closing this PR because we are going to use Python temporary files and directories which I will put in a separate pull request.

PR #1723, opened 2021-07-13
__getattr__ shortcut for calling GRASS display modules

chaedri:
* Shortcut allowing users to call GRASS display modules ('d.*) with class method `.d_*`. 
* Example of usage in jupyter_integration.ipynb
* Typo fix and black formatting

These modifications are part of an on-going Google Summer of Code project, Improved Integration of GRASS GIS and Jupyter Notebooks. You can find more information [here](https://trac.osgeo.org/grass/wiki/GSoC/2021/JupyterAndGRASS). You can test out the functionality in Binder here:

https://mybinder.org/v2/gh/chaedri/grass/non-interactive-display-update?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb


chaedri:
There are a couple things that need improvement still:

1. Error Handling. If the user passes an incorrect or incorrectly formatted command, say `.d_raster` instead of `d.rast`, it returns two errors, one that results from the `try:` statement and one that results from the `except:` statement. I think this is unexpected behavior...
2. I haven't tested all the `d.* modules.

python/grass/jupyter/display.py
```diff
@@ -53,6 +53,28 @@ def run(self, module, **kwargs):
...
+            raise ValueError("Module must begin with 'd_'.")
+        # Now reformat string
+        grass_module = f"d.{attr[2:]}"
+
+        def wrapper(**kwargs):
+            # And try to run module
+            try:
+                self.run(grass_module, **kwargs)
+            except FileNotFoundError:
+                raise ModuleNotFoundError(
```
wenzeslaus, line 72: This is more difficult to decide. Not `ModuleNotFoundError` as that's meant for Python modules, not GRASS modules. `FileNotFoundError` might be okay, but confusing. It is a failed attribute access, so `AttributeError` seem right, but this is no longer attribute access, so maybe not. Perhaps raising a new `FileNotFoundError` with a custom message. This can include the original message or use `raise ... from ...`. You could also use `shutil.which` ahead of time and then `AttributeError` would be a clear choice. This goes a little against Python principles of trying first instead of testing, but it may result in much better behavior. There is more cases than just `FileNotFoundError`, for example `PermissionError`. `shutil.which` tries to check for all these things. It worked correctly for me when only the permissions were wrong. Other than that, it should be have like `run` (e.g. with wrong parameters).

python/grass/jupyter/display.py
```diff
@@ -53,6 +53,28 @@ def run(self, module, **kwargs):
         else:
             raise ValueError("Module must begin with letter 'd'.")
 
+    def __getattr__(self, attr, **kwargs):
+        """Parse attribute to GRASS display module. Attribute should be in
+        the form 'd_{module}'."""
+
+        # First check to make sure format is correct
+        if attr[:2] != "d_":
```
wenzeslaus, line 61: `attr.startswith(...)`

python/grass/jupyter/display.py
```diff
@@ -53,6 +53,28 @@ def run(self, module, **kwargs):
         else:
             raise ValueError("Module must begin with letter 'd'.")
 
+    def __getattr__(self, attr, **kwargs):
+        """Parse attribute to GRASS display module. Attribute should be in
+        the form 'd_{module}'."""
```
wenzeslaus, line 58: This text may use some improvement, but more importantly, add examples to the class docstring.

```
>>> m = ...
>>> m.d_vect(...)
````

**Review by wenzeslaus — CHANGES_REQUESTED** [review]
Sweet and simple, just the error handling needs some more work.

Please, include an example in the notebook, so that others can try easily.

File: python/grass/jupyter/display.py [hunk 4]
```diff
@@ -53,6 +53,28 @@ def run(self, module, **kwargs):
         else:
             raise ValueError("Module must begin with letter 'd'.")
 
+    def __getattr__(self, attr, **kwargs):
```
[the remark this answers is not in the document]
chaedri: Based off the docs, I think it should read:

`    def __getattr__(self, name, **kwargs):`

Is that right? Or should `**kwargs` be passed differently too?

python/grass/jupyter/display.py
chaedri: I went with 'FileNotFoundError ' using `raise ... from ...` in my newest commit. I wasn't quite sure how to implement the `shutil.which` method and what the advantage or expected behavior would be. But, if we feel that would be a better solution, let's talk about it at our next meeting and I'll switch.

python/grass/jupyter/display.py
wenzeslaus: I really meant docstring of the class, not the one for the method. `__getattr__` is too special for it to be an obvious place to look for documentation of `m.d_vect` behavior.

python/grass/jupyter/display.py
```diff
@@ -53,25 +53,40 @@ def run(self, module, **kwargs):
...
+        # Add legend with shortcut
+        >>> m.d_legend(raster="elevation")
+        # Show map
+        >>> m.show()
+        -------------------
+        """
+
+        # Check to make sure format is correct
+        if not name.startswith("d_"):
+            raise AttributeError("Module must begin with 'd_'.")
```
wenzeslaus, line 79: Python exception messages don't end with `.`, test e.g.:

```
>>> xxx
NameError: name 'xxx' is not defined
>>> def a():
...     pass
... 
>>> a(1)
TypeError: a() takes 0 positional arguments but 1 was given
```

I think exceptions raised in GRASS code are translatable (check the other code) which in GRASS means enclosing the string in `_(...)` (call to a function named `_` which is defined as a buildin and returns the translated message if locale language is not English and the message was translated).

python/grass/jupyter/display.py
```diff
@@ -53,25 +53,40 @@ def run(self, module, **kwargs):
...
-            print(attr[:2])
-            raise ValueError("Module must begin with 'd_'.")
-        # Now reformat string
-        grass_module = f"d.{attr[2:]}"
+        :param str name: display module in form 'd_{module}'
+        :param kwargs: arguments to be passed to display module
+
+        :return: output of display module
+
+        ----- Example -----
```
wenzeslaus, line 65: See how Sphinx recommends formatting examples. I think introducing the example with a sentence is probably enough. See what these docstrings might be turned into:
https://grass.osgeo.org/grass80/manuals/libpython/pygrass.modules.interface.html?highlight=module#module-pygrass.modules.interface.module (and the corresponding source code in grass/pygrass)

wenzeslaus (approved):

wenzeslaus:
@chaedri It seems to me that the message "Module must begin with 'd_'" needs little more clarification, but let's leave that for some documentation PR. (I'm thinking module vs GRASS module, "must begin with" vs display commands, and including the name of the command.

PR #1728, opened 2021-07-16
r.reclass.area: Add flag to optionally output NULL raster if no areas are left after reclass

griembauer:
Currently, `r.reclass.area` with `method=reclass` fails if no areas greater or less than the specified value are found. This PR adds a flag to create a NULL-raster as output instead for this case.


neteler:
(re-run black)

neteler:
@griembauer please see the black formatting error... thanks

scripts/r.reclass.area/r.reclass.area.py
```diff
@@ -163,18 +168,29 @@ def reclass(inf, outf, lim, clump, diag, les):
...
+    expression = "$outfile = $recfile"
+    mapcalc_kwargs = {"outfile": outfile, "recfile": recfile}
     if p2.returncode != 0:
         if lesser:
-            grass.fatal(
-                _("No areas of size less than or equal to %f " "hectares found.")
-                % limit
+            op_str = "less"
+        else:
+            op_str = "greater"
```
wenzeslaus, line 177: The original code is right. Sentences cannot be build from parts because that makes translations impossible. If really needed, you could create two sentences and then put them together using another, possibly translatable, string.

scripts/r.reclass.area/r.reclass.area.py
```diff
@@ -163,18 +168,29 @@ def reclass(inf, outf, lim, clump, diag, les):
...
+                % (op_str, limit)
             )
+            expression = "$outfile = null()"
+            del mapcalc_kwargs["recfile"]
         else:
             grass.fatal(
-                _("No areas of size greater than or equal to %f " "hectares found.")
-                % limit
+                _("No areas of size %s than or equal to %f " "hectares found.")
+                % (op_str, limit)
```
wenzeslaus, line 191: To improve the code overall, please use `.format()` when rewriting the messages.

scripts/r.reclass.area/r.reclass.area.py
```diff
@@ -163,18 +168,29 @@ def reclass(inf, outf, lim, clump, diag, les):
...
+        if nodata is True:
+            grass.warning(
+                _(
+                    "No areas of size %s than or equal to %f "
+                    "hectares found. Creating NULL raster..."
+                )
+                % (op_str, limit)
             )
+            expression = "$outfile = null()"
+            del mapcalc_kwargs["recfile"]
```
wenzeslaus, line 187: Maybe having two calls of mapcalc would result in a simpler code.

scripts/r.reclass.area/r.reclass.area.py
```diff
@@ -163,18 +168,29 @@ def reclass(inf, outf, lim, clump, diag, les):
...
     if p2.returncode != 0:
         if lesser:
-            grass.fatal(
-                _("No areas of size less than or equal to %f " "hectares found.")
-                % limit
+            op_str = "less"
+        else:
+            op_str = "greater"
+        if nodata is True:
+            grass.warning(
```
wenzeslaus, line 179: You are probably in the best position to decide that, but if user requests NULL raster in those cases, is warning needed? Isn't that a verbose message instead? In other words, the new flag seems to make the the "no areas" state to be a standard state with appropriate output, so why a warning?

scripts/r.reclass.area/r.reclass.area.py
```diff
@@ -237,13 +256,15 @@ def cleanup():
     TMPRAST.reverse()  # reclassed map first
     for mapp in TMPRAST:
         if method == "rmarea":
-            grass.run_command(
-                "g.remove", flags="f", type="vector", name=mapp, quiet=True
-            )
+            if grass.find_file(name=mapp, element="vector")["file"]:
```
wenzeslaus, line 259: Please comment on why you are making this change in the PR description.

wenzeslaus (changes requested):

wenzeslaus:
The code here needs some changes, but otherwise it is a very good improvement. I'm keeping it open, just moving the milestone.

scripts/r.reclass.area/r.reclass.area.html
echoix, line 1: Remember to apply the changes in the markdown docs when updating the PR.

echoix (changes requested):

scripts/r.reclass.area/r.reclass.area.py
echoix: Applied the suggestion, resolving the thread

scripts/r.reclass.area/r.reclass.area.py
echoix: Applied the suggestion, resolving the thread

PR #1730, opened 2021-07-16
v.rast.stats: note about vector overlap limitation

jfbourdon:
Short note warning the user that if the vector map contains overlapping vectors, only one category will be picked during the rasterization process.

I'm suggesting that note because it took me a while recently before understanding why I was not always getting the stats I was expecting...

### DIFF
--- a/scripts/v.rast.stats/v.rast.stats.html
+++ b/scripts/v.rast.stats/v.rast.stats.html
@@ -31,8 +31,10 @@ in the use of the column prefix when using DBF as any additional characters
 will be chopped off.
 <p>If a MASK is present, it will be restored after the script finished.
 The script changes temporarily to the resolution of the given raster map.
-<p>If some vectors overlap, only one category will be kept during the
-rasterization process. Statistics for these vectors will thus be partial.
+<p>If an area has several categories in the selected layer (equivalent
+to overlapping polygons in Simple Features), only one category will be
+kept during the rasterization process. Statistics for the skipped
+categories will thus be partial.
 <p><!-- r.univar limitation -->
 Large amounts of system memory can be used when extended statistics
 (<em>first_quartile,median,third_quartile,percentile </em>) are being requested
--- a/scripts/v.rast.stats/v.rast.stats.html
+++ b/scripts/v.rast.stats/v.rast.stats.html
@@ -31,6 +31,8 @@ in the use of the column prefix when using DBF as any additional characters
 will be chopped off.
 <p>If a MASK is present, it will be restored after the script finished.
 The script changes temporarily to the resolution of the given raster map.
+<p>If some vectors overlap, only one category will be kept during the
+rasterization process. Statistics for these vectors will thus be partial.
 <p><!-- r.univar limitation -->
 Large amounts of system memory can be used when extended statistics
 (<em>first_quartile,median,third_quartile,percentile </em>) are being requested
--- a/scripts/v.rast.stats/v.rast.stats.html
+++ b/scripts/v.rast.stats/v.rast.stats.html
@@ -35,6 +35,12 @@ The script changes temporarily to the resolution of the given raster map.
 to overlapping polygons in Simple Features), only one category will be
 kept during the rasterization process. Statistics for the skipped
 categories will thus be partial.
+<p>
+For example, if there are three areas: area 1 with cat 1, area 2 with
+cat 2, area 3 with cats 1, 2. Only one category value of area 3 will be
+used for rasterization, the other category value will be skipped. Thus
+statistics for the used category value will be complete, while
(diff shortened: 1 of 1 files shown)


neteler:
@metzm: may we merge this documentation change?

neteler:
Close/reopen to awake flake8 test

scripts/v.rast.stats/v.rast.stats.html
```diff
@@ -31,6 +31,8 @@ <h2>NOTES</h2>
 will be chopped off.
 <p>If a MASK is present, it will be restored after the script finished.
 The script changes temporarily to the resolution of the given raster map.
+<p>If some vectors overlap, only one category will be kept during the
+rasterization process. Statistics for these vectors will thus be partial.
```
metzm, line 35:
```suggestion
<p>If an area has several categories in the selected layer (equivalent
to overlapping polygons in Simple Features), only one category will be
kept during the rasterization process. Statistics for the skipped
categories will thus be partial.
```

scripts/v.rast.stats/v.rast.stats.html
```diff
@@ -31,8 +31,10 @@ <h2>NOTES</h2>
 will be chopped off.
 <p>If a MASK is present, it will be restored after the script finished.
 The script changes temporarily to the resolution of the given raster map.
-<p>If some vectors overlap, only one category will be kept during the
-rasterization process. Statistics for these vectors will thus be partial.
+<p>If an area has several categories in the selected layer (equivalent
+to overlapping polygons in Simple Features), only one category will be
+kept during the rasterization process. Statistics for the skipped
```
veroandreo, line 36: Shouldn't skipped there be replaced by kept? The skipped will be skipped, no? So the resulting stats for kept categories will be partial.

metzm: In the case of two polygons partially overlapping, you have three topological areas where one area has two categories, shared with the other two areas:

- area 1: category 1
- area 2: category 2
- area 3: categories 1, 2

Let's assume, for area 3 category 1 has been used: statistics for category 1 are complete and for category 2 incomplete because category 2 has been skipped for area 3. Thus resulting stats for skipped categories are partial.

veroandreo: Thanks for clarification, Markus. I re-created the plot in a paper here. So, you mean it's partial because category 2 will be considered only in area 2 in this case then?

scripts/v.rast.stats/v.rast.stats.html
```diff
@@ -31,6 +31,10 @@ <h2>NOTES</h2>
 will be chopped off.
 <p>If a MASK is present, it will be restored after the script finished.
 The script changes temporarily to the resolution of the given raster map.
+<p>If an area has several categories in the selected layer (equivalent
+to overlapping polygons in Simple Features), only one category will be
+kept during the rasterization process. Statistics for the skipped
+categories will thus be partial.
```
metzm, line 37:
```suggestion
categories will thus be partial.
<p>
For example, if there are three areas: area 1 with cat 1, area 2 with
cat 2, area 3 with cats 1, 2. Only one category value of area 3 will be
used for rasterization, the other category value will be skipped. Thus
statistics for the used category value will be complete, while
statistics for the skipped category value will be incomplete.
```

scripts/v.rast.stats/v.rast.stats.html
metzm: I have added a more detailed explanation. OK, @veroandreo?

veroandreo: Yes, thanks Markus :)

metzm:
An example to create a vector in nc_spm_08 where areas have multiple categories and to get the number of areas that have multiple categories (requires PR #2085):
```
# create a vector where some areas have multiple categories 
v.buffer input=hospitals output=hospitals_circled type=point distance=1000 -t
# add unique category values to a new layer
v.category in=hospitals_circled out=hospitals_circled_l2 type=centroid op=add layer=2
# add a table to layer 2 with a column to hold counts of categoty values in layer 1
v.db.addtable map=hospitals_circled_l2 layer=2 columns="ncats_l1 integer"
# check range of category values in both layers
# range is larger in layer 2 because of unique category values 
v.category in=hospitals_circled_l2 op=report
# load number of different category values in layer 1 to layer 2 
v.to.db map=hospitals_circled_l2 layer=2 op=query query_layer=1 query_column="count(cat)" column=ncats_l1
# print summary of multiple category values in layer 1
v.db.select map=hospitals_circled_l2 layer=2 col="ncats_l1,count(ncats_l1)" group=ncats_l1
```

PR #1733, opened 2021-07-19
libpython: Support non-parallel runs better

wenzeslaus:
* Function for possibly non-parallel repeated runs for writting benchmarking scripts.
* Better documentation of non-parallel runs in resolution-changing benchmark.
* CLI for joining JSON result files from multiple benchmarks and plotting from a file.
* CLI which is using argparse with subcommands (subparsers) is extensible and more can be added in the future.


wenzeslaus:
@aaronsms Just a ping for you to notice this. Please open an issue if something is unclear or PR if you have a fix right away.

PR #922, opened 2020-08-20
raster module memory: set new value globally

neteler:
This change allows to globally override (as being written to `$HOME/.grass7/rc`) the default 300 MB raster cache setting with a user defined value:

```
r.in.gdal --help
Imports raster data into a GRASS raster map using GDAL library.
Usage:
 r.in.gdal [-ojeflakcrp] input=name output=name
[...]
Parameters:
           input   Name of raster file to be imported
          output   Name for output raster map
            band   Band(s) to select (default is all bands)
          memory   Maximum memory to be used (in MB)
                   default: 300                            <=== original
          target   Name of GCPs target location
           title   Title for resultant raster map
          offset   Offset to be added to band numbers
                   default: 0
[...]

# change to new setting, valid for all modules (r.cost, r.walk, i.segment, ...):
g.gisenv set="MEMORYMB=6000"

# example
r.in.gdal --help
Imports raster data into a GRASS raster map using GDAL library.
Usage:
 r.in.gdal [-ojeflakcrp] input=name output=name
[...]
Parameters:
           input   Name of raster file to be imported
          output   Name for output raster map
            band   Band(s) to select (default is all bands)
          memory   Maximum memory to be used (in MB)
                   default: 6000                            <=== !
          target   Name of GCPs target location
           title   Title for resultant raster map
          offset   Offset to be added to band numbers
                   default: 0
[...]

# set back to default
g.gisenv set="MEMORYMB=300"
```

Based on demo patch my @metzm - thanks!

**Open issues:**

a) Compiler warning (casting needed?):

```
main.c:195:12: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  195 |     memstr = G_getenv_nofatal("MEMORYMB");
```

b) How to do the same for the Python scripts?

```
$ ag 'answer: 300'
scripts/r.fillnulls/r.fillnulls.py
107:#% answer: 300

scripts/r.import/r.import.py
44:#% answer: 300

temporal/t.rast.import/t.rast.import.py
84:#% answer: 300


$ ag 'memory=300'
lib/python/temporal/stds_import.py
55:                                  set_current_region=False, memory=300):
176:                memory=300):
```

PR #1734, opened 2021-07-19
feat(misc): set NPROCS environment variable

aaronsms:
This PR allows users to set nprocs parameter globally using g.gisenv. This is similar to the previous PR for raster module memory (MEMORYMB: #922).


wenzeslaus (changes requested):
This looks good, but it misses documentation in `lib/init/variables.html`.

wenzeslaus (approved):

PR #1739, opened 2021-07-19
Test Module for Non-interactive display in Jupyter Notebooks

chaedri:
* Added testsuite directory for `grass.jupyter`
* Tests for display.py

This PR contains a test module for non-interactive displays created in `grass.jupyter`. It is part of an ongoing Google Summer of Code project, Improved Integration of GRASS GIS and Jupyter Notebooks. You can find more information [here](https://trac.osgeo.org/grass/wiki/GSoC/2021/JupyterAndGRASS).


chaedri:
This is just a draft because after some of the other display-related PR's a merged, the test will need to modified slightly and expanded to test other functions (such as the shortcuts in #1723).

wenzeslaus (changes requested):
Looks good in general, but you need to lazy-import folium so that grass.jupyter can be imported without folium installed since folium is not a mandatory dependency of GRASS GIS. You will need to do something like:

```python
def __init__():
    import folium
    self._folium = folium
```

The tests will need to be skipped with `@skip_if` decorator (see, e.g., r.in.pdal test where it was recently used). You may need to write some helper function to see if you can import folium, something like this:

```python
def can_import_folium():
    try:
        import folium
        return True
    except ImportError:
        return False
```

See also comment where we first hit this issue: https://github.com/OSGeo/grass/pull/1668#discussion_r655064892

chaedri:
I added the @skipIf decorator but I didn't add the lazy folium import to `interactive_display.py`. I'm going to save that for another PR since there are so many changes to `interactive_display.py` in PR #1769.

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def tearDown(self):
+        """
+        Remove the PNG file created after testing with "filename =" option.
+        This is executed after each test run.
+        """
+        for f in self.files:
+            Path(f).unlink(missing_ok=True)
+
+    @unittest.skipIf(can_import_folium() is False, "Cannot import folium")
```
wenzeslaus, line 65: What about one of these instead of `is False`?

```python
@unittest.skipIf(not can_import_folium(), "Cannot import folium")
@unittest.skipIf(not folium_imports(), "Cannot import folium")
```

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,125 @@
...
+        """Remove temporary region"""
+        cls.del_temp_region()
+
+    def tearDown(self):
+        """
+        Remove the PNG file created after testing with "filename =" option.
+        This is executed after each test run.
+        """
+        for f in self.files:
+            Path(f).unlink(missing_ok=True)
```
wenzeslaus, line 63: Same issue as in the other PR. `missing_ok` parameter is not available in 3.6 and 3.7 if I recall correctly.

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,125 @@
...
+        grass_renderer.run("d.vect", map="roadsmajor")
+        grass_renderer.run("d.rast", map="elevation")
+
+    @unittest.skipIf(can_import_folium() is False, "Cannot import folium")
+    def test_shortcut(self):
+        """Test that we can use display shortcuts with __getattr__."""
+        # Create map
+        grass_renderer = gj.GrassRenderer()
+        # Add raster to map with shortcut
+        grass_renderer.d_rast(map="elevation")
```
wenzeslaus, line 121: Testing more modules than just d.rast would be nice, at least d.vect.

You can also test that it throws attr error when the module name is wrong. See [assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) and use it as a context manager (with `with`). You could also test with `assertRaisesRegex` and confirm that module name is in the message and things like that.

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,125 @@
...
+        grass_renderer.run("d.rast", map="elevation")
+        grass_renderer.run("d.vect", map="roadsmajor")
+        # Assert image exists
+        self.assertFileExists(grass_renderer._filename)
+
+    @unittest.skipIf(can_import_folium() is False, "Cannot import folium")
+    def test_filename(self):
+        """Test that GrassRenderer creates maps with unique filenames."""
+        # Create map with unique filename
+        grass_renderer = gj.GrassRenderer(filename="test_filename.png")
```
wenzeslaus, line 80: May be more clear if "test_filename.png" is in a variable.

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,125 @@
...
+# Dependencies
+import os
+import unittest
+from pathlib import Path
+import grass.jupyter as gj
+from grass.gunittest.case import TestCase
+from grass.gunittest.main import test
+
+
+# Helper function
```
wenzeslaus, line 28: Make this a docstring instead saying what the function does.

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,125 @@
...
+#
+# COPYRIGHT: (C) 2021 by Caitlin Haedrich and the GRASS Development Team
+#
+#            This program is free software under the GNU General Public
+#            License (>=v2). Read the file COPYING that comes with GRASS
+#            for details.
+#
+#############################################################################
+
+# Dependencies
```
wenzeslaus, line 19: No need for a comment (would be good for an example code for learning Python, but not code which will be maintained for people who went at least though the initial hurdles of getting to know Python).

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,125 @@
...
+def can_import_folium():
+    try:
+        import folium
+
+        return True
+    except ImportError:
+        return False
+
+
+# Tests
```
wenzeslaus, line 38: Remove.

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @unittest.skipIf(can_import_folium() is False, "Cannot import folium")
+    def test_defaults(self):
+        """Test that GrassRenderer can create a map with default settings."""
+        # Create a map with default inputs
+        grass_renderer = gj.GrassRenderer()
+        # Adding vectors and rasters to the map
+        grass_renderer.run("d.rast", map="elevation")
+        grass_renderer.run("d.vect", map="roadsmajor")
+        # Assert image exists
```
wenzeslaus, line 73: If the code below is not clear enough, I'm afraid that the word "assert" will not help much (not the most common word in English).

wenzeslaus (changes requested):

wenzeslaus:
> ...I'm going to save that for another PR since there are so many changes to `interactive_display.py` in PR #1769.

Since the tests are failing because of missing folium, we can't merge this before the lazy import change, but that's fine.

Unrelated to Jupyter, this seems to suggest we need two sets of test environments, a minimal one with only mandatory dependencies and and a one with all optional dependencies, but even if we would have that now, this PR would need to run without folium in that minimal environment.

wenzeslaus:
Here is the easiest pattern for the optional import of folium:

```python
# (module top level like any import)
try:
    import folium
except ImportError:
    folium = None
```

Will fail at any attempt to use folium with a somewhat strange error, but it is easy to implement and use.

python/grass/jupyter/testsuite/display_test.py
chaedri: I added d.vect and another test for the attr error.

chaedri:
I went for the lazy folium import, the first suggestion. I also added a test for InteractiveMap but it's pretty minimal - ideas for further tests welcome.

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,155 @@
...
+        self.assertFileExists(grass_renderer._filename)
+
+    @unittest.skipIf(not can_import_folium(), "Cannot import folium")
+    def test_filename(self):
+        """Test that GrassRenderer creates maps with unique filenames."""
+        # Create map with unique filename
+        unique_filename = "test_filename.png"
+        grass_renderer = gj.GrassRenderer(filename=unique_filename)
```
wenzeslaus, line 87: Do you mean custom? (not unique)

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,155 @@
...
+    def test_shortcut_error(self):
+        """Test that passing an incorrect attribute raises
+        appropriate error"""
+        # Create map
+        grass_renderer = gj.GrassRenderer()
+        # Pass bad shortcuts
+        with self.assertRaisesRegex(AttributeError, "Module must begin with 'd_'"):
+            grass_renderer.r_watersheds()
+        with self.assertRaisesRegex(AttributeError, "d.ModuleDoesNotExist"):
+            grass_renderer.d_ModuleDoesNotExist()
```
wenzeslaus, line 141: Possibly more readable when it looks like a module name, so underscores, not camel case.

wenzeslaus (changes requested):

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,155 @@
...
+        grass_renderer.run("d.vect", map="roadsmajor")
+        # Assert image exists
+        self.assertFileExists(grass_renderer._filename)
+
+    @unittest.skipIf(not can_import_folium(), "Cannot import folium")
+    def test_filename(self):
+        """Test that GrassRenderer creates maps with unique filenames."""
+        # Create map with unique filename
+        custom_filename = "test_filename.png"
+        grass_renderer = gj.GrassRenderer(filename=custon_filename)
```
wenzeslaus, line 87: typo

wenzeslaus (changes requested):
One more lazy import will be needed for IPython. Probably you just need to have it in one function.

```
Running ./python/grass/jupyter/testsuite/display_test.py...
  File "python/grass/jupyter/testsuite/display_test.py", line 23, in <module>
    import grass.jupyter as gj
  File "etc/python/grass/jupyter/__init__.py", line 2, in <module>
    from .interact_display import *
  File "etc/python/grass/jupyter/interact_display.py", line 19, in <module>
    from .display import GrassRenderer
  File "etc/python/grass/jupyter/display.py", line 16, in <module>
    from IPython.display import Image
ModuleNotFoundError: No module named 'IPython'
```

python/grass/jupyter/testsuite/display_test.py
wenzeslaus: Do you plan to update this? If you think it is good as is, just resolve this conversation.

wenzeslaus (changes requested):

python/grass/jupyter/testsuite/display_test.py
chaedri: There were two `#Assert image exists` and I changed the wrong/other one. Oops! Now they both say something better. Thanks for catching it.

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @unittest.skipIf(not can_import_folium(), "Cannot import folium")
+    def test_InteractiveMap(self):
+        # Create InteractiveMap
+        interactive_map = gj.InteractiveMap()
+        # Add raster and vector
+        interactive_map.add_raster("elevation")
+        interactive_map.add_vector("roadsmajor")
+        # Show
+        interactive_map.show()
```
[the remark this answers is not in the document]
chaedri: In the command line, `interactive_map.show()` returns a folium object (specifically, something like `branca.element.Figure object at <...>`. I added a comment to the `show()` docstring saying it returned  a folium figure object. I'm not sure if this is problematic or not - my guess is not because the object is just taking up Python memory and will be deleted after Python is exited?

python/grass/jupyter/testsuite/display_test.py
```diff
@@ -0,0 +1,155 @@
...
+        # Create map
+        grass_renderer = gj.GrassRenderer()
+        # Pass bad shortcuts
+        with self.assertRaisesRegex(AttributeError, "Module must begin with 'd_'"):
+            grass_renderer.r_watersheds()
+        with self.assertRaisesRegex(AttributeError, "d.module.does.not.exist"):
+            grass_renderer.d_module_does_not_exist()
+
+    @unittest.skipIf(not can_import_folium(), "Cannot import folium")
+    def test_InteractiveMap(self):
```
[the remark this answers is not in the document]
chaedri: I created a new file.

python/grass/jupyter/testsuite/display_test.py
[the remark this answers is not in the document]
wenzeslaus: That sounds good. `show()` here is little surprising as in e.g. Matplotlib it may mean showing a window and waiting for a user to close it, i.e., show() seems like something will be shown (not the case in the tests), rather than an independent (although perhaps useless) object being returned. It makes sense in the context of Jupyter notebook, so let's leave it as is.

wenzeslaus (approved):

PR #1200, opened 2020-12-21
r.in.pdal – a PDAL based replacement of r.in.lidar

marisn:
This is an enhanced version of r.in.pdal based on https://github.com/wenzeslaus/grass/tree/pdal-binning (PR #61). This PR makes PR #61 obsolete. This PR depends on #1148 Code has been rebased on top of current master for easier testing.

TODO:
- [x] Print file info (-p flag)
- [x] More tests for filters, selection, base_raster, ...
- [x] Review for missing functionality in comparison to r.in.lidar

Features to discuss:
- [ ] ~~Determine extent in scan mode (with all filters applied) (-s flag)~~ Postponed after 8.0
- [ ] ~~Output convex hull of imported data as a vector map (@neteler wish)~~ Postponed after 8.0
- [ ] ~~Add PDAL pipeline support (on-fly preprocessing)~~ Postponed after 8.0

PR #1749, opened 2021-07-20
ci: Remove duplication in Ubuntu workflow

wenzeslaus:
matrix.os has the name of the the system which is used in the YAML.
This basically returns the CI YAML files state to what it was before #1200
keeping the addition of a new build script and version-specific naming for the scripts.

Additionally, this now removes the build job and leaves only the test job in the workflow.


marisn:
I was looking at ci.yml file – did I read correctly that code for running tests is build for a second time? If so, we could get rid of build part as code building is tested also by the tests run.

wenzeslaus:
> ...code for running tests is build for a second time? If so, we could get rid of build part as code building is tested also by the tests run.

That is correct. I wrote it this way because at that time it was the only workflow there and I wanted to have a separate build result from a test result. Build runs fast and you see it succeeded even when tests failed. Now we are in a different situation since we have build-only jobs in the GCC workflow. I removed the build-only job from Ubuntu and renamed it from CI to Ubuntu.

wenzeslaus:
To perhaps clarify a bit what this is doing, here is the diff between the pre-PDAL-split state and pre-build-removal (`git df 94bcea421..862126ddc -- .`):

```diff
diff --git a/.github/workflows/build_ubuntu-18.04.sh b/.github/workflows/build_ubuntu-18.04.sh
new file mode 100755
index 000000000..68d10d6c8
--- /dev/null
+++ b/.github/workflows/build_ubuntu-18.04.sh
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+... [whole new file here]
diff --git a/.github/workflows/build.sh b/.github/workflows/build_ubuntu-20.04.sh
similarity index 100%
rename from .github/workflows/build.sh
rename to .github/workflows/build_ubuntu-20.04.sh
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 50fccaa5e..340a46775 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
         run: |
           echo "LD_LIBRARY_PATH=$HOME/install/lib" >> $GITHUB_ENV
       - name: Build
-        run: .github/workflows/build.sh $HOME/install
+        run: .github/workflows/build_${{ matrix.os }}.sh $HOME/install
 
   test:
     name: ${{ matrix.os }} tests
@@ -65,7 +65,7 @@ jobs:
         run: |
           echo "LD_LIBRARY_PATH=$HOME/install/lib" >> $GITHUB_ENV
       - name: Build
-        run: .github/workflows/build.sh $HOME/install
+        run: .github/workflows/build_${{ matrix.os }}.sh $HOME/install
       - name: Add the bin directory to PATH
         run: |
           echo "$HOME/install/bin" >> $GITHUB_PATH
diff --git a/.github/workflows/codeql-analysis.yml b/.github/workflows/codeql-analysis.yml
index f1a561102..f6e65a1cf 100644
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -50,7 +50,7 @@ jobs:
         env:
           CFLAGS: "-std=gnu11"
           CXXFLAGS: "-std=c++11"
-        run: .github/workflows/build.sh $HOME/install
+        run: .github/workflows/build_ubuntu-20.04.sh $HOME/install
 
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v1
diff --git a/.github/workflows/gcc.yml b/.github/workflows/gcc.yml
index c0dd6de94..0efc9e647 100644
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -45,4 +45,4 @@ jobs:
           CFLAGS: "-std=${{ matrix.c }}"
           # TODO: -pedantic-errors here won't compile
           CXXFLAGS: "-std=${{ matrix.cpp }}"
-        run: .github/workflows/build.sh $HOME/install
+        run: .github/workflows/build_ubuntu-20.04.sh $HOME/install
```

And then the other part is more or less simply removing the build job and leaving only test there (plus renames).

wenzeslaus:
@marisn Please see if it makes more sense now.

marisn (approved):
Yes, now this is much better than anything I've done.

PR #1751, opened 2021-07-20
libgis: add missing return keyword to G__make_mapset_element_misc()

nilason:


wenzeslaus (approved):

wenzeslaus:
Thanks @nilason. I don't know how I missed that locally, but this definitely looks like something what shouldn't pass the CI.

marisn (approved):

nilason:
I cannot take credit for spotting this, clang said:
```
mapset_msc.c:235:1: warning: non-void function does not return a value [-Wreturn-type]
}
^
1 warning generated.
```

PR #1752, opened 2021-07-20
r.in.pdal: use fabs for double values

nilason:


marisn (approved):
Swapping both summators plays a role only if both differ significantly. But lets go with this change as current code seems confusing for a casual code reader.

nilason:
Again, clang reported:
```
bin_update.c:105:9: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
    if (abs(old_sum) >= abs(value))
        ^
bin_update.c:105:9: note: use function 'fabs' instead
    if (abs(old_sum) >= abs(value))
        ^~~
        fabs
bin_update.c:105:9: note: include the header <math.h> or explicitly provide a declaration for 'fabs'
bin_update.c:105:25: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
    if (abs(old_sum) >= abs(value))
                        ^
bin_update.c:105:25: note: use function 'fabs' instead
    if (abs(old_sum) >= abs(value))
                        ^~~
                        fabs
bin_update.c:105:25: note: include the header <math.h> or explicitly provide a declaration for 'fabs'
```

PR #1754, opened 2021-07-21
Temporary Files for Interactive Folium Maps

chaedri:
* Creates temporary files for InteractiveMap
* Adds clean-up method (weakref.finalizer) for rcfile associated with temporary location

This short PR changes the files associated with each instance of InteractiveMap (GeoJSONs, Location, GRASS Geodatabase) to temporary files which will be automatically deleted. These modifications are part of an on-going Google Summer of Code project, Improved Integration of GRASS GIS and Jupyter Notebooks. You can find more information [here](https://trac.osgeo.org/grass/wiki/GSoC/2021/JupyterAndGRASS). 

You can also test the functionality at this binder link:
https://mybinder.org/v2/gh/chaedri/grass/InteractiveMaps-add-temp-files?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb


wenzeslaus (changes requested):
This doesn't make much sense to me in the current form. Why not go all the way and introduce the clean up here too? Also, using NamedTemporaryFile for a directory is suspicions.

chaedri:
According to the [docs](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile), cleanup is included, even in NamedTemporaryFiles. When I add clean-up using the finalizer() method we'd discussed (see #1727, commit 18d798b), I get an error message when when tempfile tries to clean-up the file but can't find it.

Also, I chose NameTemporaryFile here since there is no function NamedTemporaryDirectory. There is TemporaryDirectory though - should I use that one? I'm unclear the functional difference between the Named and un-named files but I remembered we'd discussed that Named should be used here.

python/grass/jupyter/interact_display.py
```diff
@@ -32,18 +32,16 @@ def __init__(self, width=400, height=400):
         self.width = width
         self.height = height
         # Make temporary folder for all our files
-        self.tmp_dir = Path("./tmp/")
-        try:
-            os.mkdir(self.tmp_dir)
-        except FileExistsError:
-            pass
+        self._tmp_dir = tempfile.TemporaryDirectory()
+        # self.tmp_dir = Path(self._tmp_dir_obj.name)
```
wenzeslaus, line 36: Please, remove this line.

wenzeslaus (changes requested):

wenzeslaus (changes requested):
When I do `del m` in the notebook, I get:

```
TypeError: unlink() got an unexpected keyword argument 'missing_ok'
```

According to [the doc](https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink), *missing_ok* is available only since Python 3.8, but Binder has 3.7 and minimal supported version by GRASS is currently 3.6.

One way how to do it is to wrap the unlink call in try-except with FileNotFoundError and pass, although at this point we expect the file to exist. You can add there an if statement to check Python version and do the other call instead which will be helpful when GRASS bump the minimal required Python to 3.8 because it will be clear the the try-except can be removed and the shorter *missing_ok* can be kept. See a similar example in [grass.benchmark](https://github.com/OSGeo/grass/blob/master/python/grass/benchmark/app.py#L124).

chaedri:
Closing because changes are included in PR #1769.

PR #1683, opened 2021-06-25
Address some MS Windows test failures

ninsbl:
Should be backported to the release branch.

PR #1757, opened 2021-07-22
Fix osgeo4w workflow

agiudiceandrea:
osgeo4w-setup-x86_64.exe executable was renamed osgeo4w-setup-x86_64-v1.exe

### DIFF
--- a/.github/workflows/osgeo4w.yml
+++ b/.github/workflows/osgeo4w.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install OSGeo4W
         run: |
-          $exe = 'osgeo4w-setup-x86_64.exe'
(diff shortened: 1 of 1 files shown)


ninsbl:
Thanks @agiudiceandrea !
While I curently try to get the next vesion of OSGeo4W to work for checks (see: #1683) this is a straight forward way to get the checks back with version 1. So, I would be very supportive to merge this at last as an interim solution, but maye even good to have tests running for both!?!

Interestingly, several test cases are failing for diferent reasons. Some e.g. possibly due to Python version. Woul be worth looking into why tests are failing on Windows...

wenzeslaus:
Well, this makes the previously broken CI pass, so I'm merging this now, although only the build step is in a good shape.

Thanks @agiudiceandrea for the update!

PR #1760, opened 2021-07-23
Improve G_open|find _misc function documentation

marisn:
lib/gis functions dealing with elements are hard to understand due to opposite layout in the mapset for "normal" and "misc" elements:
* in the "usual" case element = directory with named files or folders (e.g. vector/my_roads/; cell/my_raster)
* in the "misc" case element = file inside a named subfolder (e.g. cell_misc/my_raster/history; group/my_group/REF)

This PR just replaces factually wrong documentation of G_open_misc and adds notes to G_find_file and G_file_name.

### DIFF
--- a/lib/gis/file_name.c
+++ b/lib/gis/file_name.c
@@ -31,10 +31,25 @@ static void append_char(char*, char);
   Paths to files are in form:
   /home/user/GISDBASE/LOCATION/mapset/element/name
 
-  path input buffer memory must be preallocated by caller:
-  char path[GPATH_MAX];
+  path input buffer memory must be preallocated by caller.
 
-  \param[out] path buffer to hold resultant full path to file
+  C:
+  @code
+  char path[GPATH_MAX];
+  @endcode
+  Python:
+  @code
+  import ctypes
+  from grass.pygrass.utils import decode
+  from grass.lib.gis import G_file_name, GPATH_MAX
+
+  path = ctypes.create_string_buffer(GPATH_MAX)
+  path_str = decode(G_file_name(path, "elem", "name", "mapset"))
+  print(path_str)
+  /home/user/gisdb/loc/mapset/elem/name
+  @endcode
+
+  \param[out] path preallocated buffer to hold resultant full path to file
   \param element database element (eg, "cell", "cellhd", "vector", etc)
   \param name name of file to build path to (fully qualified names allowed)
   \param mapset mapset name
@@ -53,12 +68,27 @@ char *G_file_name(char *path,
   Paths to misc files are in form:
   /home/user/GISDBASE/LOCATION/mapset/dir/name/element
 
-  path input buffer memory must be preallocated by caller:
-  char path[GPATH_MAX];
+  path input buffer memory must be preallocated by caller.
 
-  \param[out] path buffer to hold resultant full path to file
-  \param dir misc directory
-  \param element database element (eg, "cell", "cellhd", "vector", etc)
+  C:
+  @code
+  char path[GPATH_MAX];
+  @endcode
+  Python:
+  @code
+  import ctypes
+  from grass.pygrass.utils import decode
+  from grass.lib.gis import G_file_name_misc, GPATH_MAX
+
+  path = ctypes.create_string_buffer(GPATH_MAX)
+  path_str = decode(G_file_name_misc(path, "dir", "elem", "name", "mapset"))
+  print(path_str)
+  /home/user/gisdb/loc/mapset/dir/name/elem
+  @endcode
+
+  \param[out] path preallocated buffer to hold resultant full path to file
+  \param dir misc directory (e.g., "cell_misc", "group")
+  \param element database element (e.g., "history", "REF")
   \param name name of file to build path to (fully qualified names allowed)
   \param mapset mapset name
 
--- a/lib/gis/file_name.c
+++ b/lib/gis/file_name.c
@@ -34,13 +34,13 @@ static void append_char(char*, char);
   path input buffer memory must be allocated by caller.
 
   C:
-  @code
+  @code{.c}
   char path[GPATH_MAX];
   G_file_name(path, "fcell", "my_raster", "my_mapset");
   // path now is "/full/path/to/my_mapset/fcell/my_raster"
   @endcode
   Python:
-  @code
+  @code{.py}
   import ctypes
   from grass.pygrass.utils import decode
   from grass.lib.gis import G_file_name, GPATH_MAX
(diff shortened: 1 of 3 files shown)


lib/gis/file_name.c
```diff
@@ -44,9 +65,30 @@ char *G_file_name(char *path,
...
-  \param element database element (eg, "cell", "cellhd", "vector", etc)
+  Paths to misc files are in form:
+  /home/user/GISDBASE/LOCATION/mapset/dir/name/element
+
+  path input buffer memory must be preallocated by caller.
+
+  C:
+  @code
+  char path[GPATH_MAX];
+  @endcode
```
wenzeslaus, line 76: Unfinished example? Maybe some spaces needed too, but the Python example might be enough.

lib/gis/file_name.c
```diff
@@ -27,8 +27,29 @@ static void append_char(char*, char);
   If <i>name</i> is of the form "nnn@ppp" then path is set as if name
   had been "nnn" and mapset had been "ppp" (mapset parameter itself is
   ignored in this case).
-  
-  \param[out] path buffer to hold resultant full path to file
+
+  Paths to files are in form:
+  /home/user/GISDBASE/LOCATION/mapset/element/name
```
wenzeslaus, line 32: This might be misleading. GISDBASE var contains /home/user or generally anything else. Some alternatives:

```
mapset/element/name
.../mapset/element/name
/path/to/mapset/element/name
/path/to/LOCATION/mapset/element/name
.../location/mapset/element/name
$GISDBASE/$LOCATION/mapset/element/name
$GISDBASE/$LOCATION_NAME/mapset/element/name
```

I think the intention here was to hide the concrete path, so what path it is is an implementation detail which the user should not depend on, so perhaps you can say something like _Paths to files are currently in form:_ One use case for the change of implementation would be changing mapset structure (to raster/elevation/fcell misc-like layout), but supporting old mapset structure too. "open old" would then check both possible paths while "open new" would use the new style and delete the files in the old structure if they exist.

lib/gis/file_name.c
```diff
@@ -27,8 +27,29 @@ static void append_char(char*, char);
...
+
+  path input buffer memory must be preallocated by caller.
+
+  C:
+  @code
+  char path[GPATH_MAX];
+  @endcode
+  Python:
+  @code
```
wenzeslaus, line 41: You can tell Doxgen this is Python: https://www.doxygen.nl/manual/commands.html#cmdcode

Or you can use just Markdown syntax with 3 backticks followed by simple `py` (although the doc for that is less specific).

lib/gis/file_name.c
```diff
@@ -27,8 +27,29 @@ static void append_char(char*, char);
...
+  import ctypes
+  from grass.pygrass.utils import decode
+  from grass.lib.gis import G_file_name, GPATH_MAX
+
+  path = ctypes.create_string_buffer(GPATH_MAX)
+  path_str = decode(G_file_name(path, "elem", "name", "mapset"))
+  print(path_str)
+  /home/user/gisdb/loc/mapset/elem/name
```
wenzeslaus, line 49: Python examples with outputs are typically written with the `>>>` prompt. Alternatively, a cross-language style would be `# prints /home/user/` or `# path_str now contains ...`

lib/gis/file_name.c
```diff
@@ -27,8 +27,29 @@ static void append_char(char*, char);
   If <i>name</i> is of the form "nnn@ppp" then path is set as if name
   had been "nnn" and mapset had been "ppp" (mapset parameter itself is
   ignored in this case).
-  
-  \param[out] path buffer to hold resultant full path to file
+
+  Paths to files are in form:
+  /home/user/GISDBASE/LOCATION/mapset/element/name
+
+  path input buffer memory must be preallocated by caller.
```
wenzeslaus, line 34: "must be allocated by caller" and "path is a preallocated buffer". I think _preallocating_ raises an unnecessary question: is is same as allocating?

lib/gis/file_name.c
```diff
@@ -44,9 +65,30 @@ char *G_file_name(char *path,
...
+  path = ctypes.create_string_buffer(GPATH_MAX)
+  path_str = decode(G_file_name_misc(path, "dir", "elem", "name", "mapset"))
+  print(path_str)
+  /home/user/gisdb/loc/mapset/dir/name/elem
+  @endcode
+
+  \param[out] path preallocated buffer to hold resultant full path to file
+  \param dir misc directory (e.g., "cell_misc", "group")
+  \param element database element (e.g., "history", "REF")
   \param name name of file to build path to (fully qualified names allowed)
```
wenzeslaus, line 92: Isn't _file_ here the confusing part? Name is name of raster map (or whatever we call it), not of a file. I think we lack a good term for that named thing, so that's why I went with _object_ in #1681, but I don't consider that naming issue resolved.

wenzeslaus (approved):
I think this could go in as is, but in the documentation, I would like to avoid creating an impression that the directory nesting is all that matters. That might be why Vlib is usage of these functions is rather strange and what required most changes in #1681. The functions are there to abstract the directory structure not to be used to create arbitrary directory structures. Clarifying the documentation by including examples, that's of course fine.

lib/gis/open_misc.c
```diff
@@ -97,9 +97,9 @@ static int G__open_misc(const char *dir,
 
 
 /*!
- * \brief open a new database file
+ * \brief open a new database misc file
  *
- * The database file <b>name</b> under the <b>element</b> in the
+ * The database file <b>element</b> under the <b>dir/name</b> in the
```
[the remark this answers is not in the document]
marisn: There is a reason why I asked you to do a review ;-)

lib/gis/file_name.c
marisn: I went with /full/path/to/mapset/element/name

lib/gis/file_name.c
marisn: Thanks for a hint!

lib/gis/file_name.c
marisn: Then plain allocating it is.

lib/gis/file_name.c
marisn: I added code hints to both C an Py parts. Doxygen was already doing a good job with highlighting though.

lib/gis/file_name.c
marisn: You are right. I should have read twice. I went with object until we come up with a new naming scheme.

PR #1762, opened 2021-07-24
g.extension: attempt to backport changes from master

ninsbl:
Backports all changes to g.extension (excluding `#%` -> `# %`)

### DIFF
--- a/scripts/.flake8
+++ b/scripts/.flake8
@@ -1,5 +1,6 @@
 [flake8]
 ignore =
+    E203, # white space before ':'
     E265, # block comment should start with '# '
     E266, # too many leading '#' for block comment
     E402, # module level import not at top of file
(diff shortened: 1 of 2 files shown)


neteler:
Thanks a lot for this backport, @ninsbl !

I have made shallow tests by installing some extensions from the main repo: all successfull.

Then I went to `grass_master/scripts/g.extension/testsuite/` and tried the three test scripts therein.

All working except for this one which also fails in master:

```
python test_addons_download.py
...
i.sentinel
...
ERROR: Option <branch> requires <url>
```

wenzeslaus:
> Should I replace f-strings with format() for the backport?

v7.8 is supposed to support Python 2.7, so yes, f-strings. Also black with `--target-version py27` just to be sure. Possibly also some URL related things, although there seems to be a lot "six" imports which should take care of that (@tmszi, do you know right away by any chance?).

> Also, flake8 checks are failing cause other configuration is used in the release branch, I guess...

You can add `E203` to ignored issues. Flake8 goes against PEP8 there if I understand correctly.

neteler:
No more CI test errors, congrats!

neteler:
Yes, please merge. The also the docker builds should be back

PR #1767, opened 2021-08-01
r.slope.aspect: implement parallelization with OpenMP

aaronsms:
Checklists before merging:
- [x] code review
- [x] CI passes
- [x] rebase to main


marisn (approved):
Didn't test the performance but code and output seems to be fine.

raster/r.slope.aspect/main.c
```diff
@@ -541,427 +568,515 @@ int main(int argc, char *argv[])
...
-        }
-        if (slope_fd >= 0) {
-            slp_ptr = slp_raster;
-        }
-
-        if (pcurv_fd >= 0) {
-            pcurv_ptr = pcurv_raster;
-        }
+    int computed = 0; /* for computing progress */
+    int written = 0;
```
HuidaeCho, line 575: `row = 0;` to avoid
```
main.c:1076:5: warning: ‘row’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     G_percent(row, nrows, 2);
```

petrasovaa:
Here is my large 16 billion cell banchmark result, no repetition.
results

raster/r.slope.aspect/main.c
aaronsms: Opps yeap it is supposed to be `G_percent(nrows, nrows, 2);`. Fixed it.

aaronsms:
Why are the Ubuntu 20.04 tests halted, maybe we can rerun that again?

wenzeslaus:
> Why are the Ubuntu 20.04 tests halted,

The cause is unknown, but it happens time to time. So far, I have observed different tests getting stuck, but I didn't check lately.

>  maybe we can rerun that again?

Done.

PR #1769, opened 2021-08-01
Raster Support for Interactive Jupyter maps with folium

chaedri:
* Add temporary WGS84-Pseudo-Mercator location for rasters
* Add `add_raster` method for `InteractiveMap` class
* Update demonstration notebook to include raster example
* Change `GrassRenderer` PNG background to transparent
* Re-arrange functions for clarity; added `utils.py'
* Added rcfile cleanup
* Added `save()` method for saving folium map as HTML file
* Added temporary directory with `tempfile`

These changes are part of an on-going Google Summer of Code project, Improved Integration of GRASS GIS and Jupyter Notebooks. You can find more information [here](https://trac.osgeo.org/grass/wiki/GSoC/2021/JupyterAndGRASS). 

You can also test the functionality at this binder link:
https://mybinder.org/v2/gh/chaedri/grass/interactive-rasters?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb


chaedri:
There are two issues that I'm aware of. First, it takes a long while for the raster to reproject. I'm not sure if that's typical or if there's ways to make it faster but, in its current state, it's pretty slow. Secondly, since I haven't implemented temporary files here (in a separate PR #1754 ), you can't have multiple instances of `InteractiveMap`.

wenzeslaus (changes requested):
The reason the reprojection takes a long time is that the target rows x cols is huge. Try to do the same from Data tab (data catalog) in GUI between pseudo-mercator and NC SPM (1 second for me, longer than I would like to, but acceptable). Data tab internally estimates an appropriate resolution and uses whole map extent. (Whole extent is not necessary the best thing to do in your case, but acceptable some intersection of computational region and map extent might be more appropriate.)

Perhaps someone can chime with what is the best way to ensure it does not over-sample the raster.

Calling `r.info map=elevation` and `g.region -p` from your Python code is more elegant way of seeing what you have there, but see what I got in the notebook by inspecting the files directly:

```
!cat tmp/temp_folium_WGS84_pmerc/PERMANENT/cellhd/elevation
...
north:      4274397.7286105
south:      4257765.15448084
east:       -8750642.28186897
west:       -8769144.64079992
cols:       18502
rows:       16633
...
!cat tmp/temp_folium_WGS84_pmerc/PERMANENT/WIND
...
north:      4274397.7286105
south:      4257765.15448084
east:       -8750636.28175257
west:       -8769150.64091632
cols:       18514
rows:       16633
e-w resol:  1.0000194
n-s resol:  0.9999744
```

python/grass/jupyter/utils.py
```diff
@@ -0,0 +1,100 @@
...
+    region["south"] = region["s"]
+    return region
+
+
+def get_location_proj_string(env=None):
+    out = gs.read_command("g.proj", flags="jf", env=env)
+    return out.strip()
+
+
+def reproject_region(region, from_proj, to_proj):
```
wenzeslaus, line 60: Ideally, you would have a third function which would have the m.proj call and surrounding string manipulation and that may avoid the need for two calls of `convert_coordinates_to_latlon` which are good API-wise, but not ideal performance-wise, but it might be too cumbersome in practice.

python/grass/jupyter/interact_display.py
```diff
@@ -142,11 +144,71 @@ def add_vector(self, name):
...
+        # Find full name of raster
+        file_info = gs.find_file(name, element="raster")
+        full_name = file_info["fullname"]
+        name = file_info["name"]
+
+        # Reproject raster into WGS84/epsg3857 location
+        env_info = gs.gisenv(env=self._src_env)
+        resolution = estimate_resolution(
+            full_name, env_info["GISDBASE"], env_info["LOCATION_NAME"], self._psmerc_env
+        )
```
wenzeslaus, line 170: That needs keyword arguments to clarify what is being passed to it.

python/grass/jupyter/utils.py
```diff
@@ -0,0 +1,100 @@
...
+    elon, nlat, unused = enws[0].split(" ")
+    wlon, slat, unused = enws[1].split(" ")
+    region["east"] = elon
+    region["north"] = nlat
+    region["west"] = wlon
+    region["south"] = slat
+    return region
+
+
+def estimate_resolution(raster, dbase, location, env):
```
wenzeslaus, line 90: Add docstrinfgs, run Pylint.

wenzeslaus (changes requested):

python/grass/jupyter/interact_display.py
```diff
@@ -70,70 +67,75 @@ def __init__(self, width=400, height=400):
...
-        # Reproject coordinates
-        coords_folium = gs.read_command(
-            "m.proj",
-            coordinates=coordinates,
-            proj_in=proj_in,
-            separator="comma",
-            flags="do",
+    def _setup_vector_location(self):
+        # Create new vector environment for tmp WGS84 location
+        rcfile, self._wgs84_env = gs.create_environment(
```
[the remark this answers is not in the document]
chaedri: I think you are right but I have all the cleanup in PR #1754. I think it makes most sense to move rcfile cleanup here and delete that PR since it (1) doesn't contain much new code and (2) this PR contains so many big changes, it's just going to cause a bunch of merge conflicts.

python/grass/jupyter/utils.py
```diff
@@ -88,6 +72,16 @@ def reproject_region(region, from_proj, to_proj):
 
 
 def estimate_resolution(raster, dbase, location, env):
+    """Estimates resolution of reprojected raster.
+
+    :param str raster: name of raster
+    :param str dbase: path to source database
+    :param str location: name of source location
+    :param str env: target environment
```
wenzeslaus, line 80: dict

python/grass/jupyter/interact_display.py
```diff
@@ -70,84 +76,126 @@ def __init__(self, width=400, height=400):
...
+        """Imports raster into temporary WGS84 location,
+        exports as png and overlays on folium map
+
+        :param str name: name of raster to add to display; positional-only parameter
+        :param float opacity: raster opacity, number between
+                              0 (transparent) and 1 (opaque)
+        """
+
+        # Find full name of raster
+        file_info = gs.find_file(name, element="raster")
```
wenzeslaus, line 140: raster makes sense, but it should be cell here. Find file is a low-level function looking for a file. Hopefully, this will change at some point, but now "raster" generates a verbose message, but "cell" does not. See implementation of find_file.

python/grass/jupyter/utils.py
chaedri: I managed to avoid calling `convert_coordinates_to_latlong` and deleted it. Do you think it's still necessary to have a third function here?

python/grass/jupyter/utils.py
```diff
@@ -0,0 +1,100 @@
...
+        coordinates=coordinates,
+        proj_in=proj_in,
+        separator="comma",
+        flags="do",
+    )
+    # Reformat from string to array
+    new_coords = new_coords.strip()  # Remove '\n' at end of string
+    new_coords = new_coords.split(",")  # Split on comma
+    new_coords = [float(value) for value in new_coords]  # Convert to floats
+    return new_coords[1], new_coords[0]  # Return Lat and Lon
```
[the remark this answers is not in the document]
chaedri: I deleted `convert_coordinates_to_latlon`.

python/grass/jupyter/interact_display.py
```diff
@@ -142,11 +144,71 @@ def add_vector(self, name):
...
+            output=name,
+            location=env_info["LOCATION_NAME"],
+            dbase=env_info["GISDBASE"],
+            resolution=resolution,
+            env=self._psmerc_env,
+        )
+
+        # Reprojects bounds of raster for overlaying png (THIS HAS TO BE IN WGS84)
+        bounds = gs.read_command("g.region", flags="g", env=self._src_env)
+        bounds = gs.parse_key_val(bounds, sep="=", vsep="\n")
```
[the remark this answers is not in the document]
chaedri: You're right - I switched to `get_region` here.

python/grass/jupyter/interact_display.py
```diff
@@ -70,70 +67,75 @@ def __init__(self, width=400, height=400):
...
+            n=new_region["north"],
+            s=new_region["south"],
+            e=new_region["east"],
+            w=new_region["west"],
+            env=self._wgs84_env,
+        )
+
+    def _setup_raster_location(self):
+        # Create new raster environment in PseudoMercator
+        rcfile, self._psmerc_env = gs.create_environment(
```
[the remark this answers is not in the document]
chaedri: I switch to `_setup_location()` and moved that functions to `utils.py`.

python/grass/jupyter/interact_display.py
```diff
@@ -70,70 +67,75 @@ def __init__(self, width=400, height=400):
...
+        to_proj = get_location_proj_string(env=self._psmerc_env)
+        new_region = reproject_region(region, from_proj, to_proj)
+        # Set raster region to match original region extent
+        gs.run_command(
+            "g.region",
+            n=new_region["north"],
+            s=new_region["south"],
+            e=new_region["east"],
+            w=new_region["west"],
+            env=self._psmerc_env,
```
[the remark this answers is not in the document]
chaedri: I'm not sure if I've addressed this comment - I got rid of the repetition with a new function `setup_location()`.

python/grass/jupyter/utils.py
```diff
@@ -0,0 +1,127 @@
...
+    params = gs.parse_key_val(output, vsep=" ")
+    output = gs.read_command("g.region", flags="ug", env=env, **params)
+    output = gs.parse_key_val(output, val_type=float)
+    cell_ns = (output["n"] - output["s"]) / output["rows"]
+    cell_ew = (output["e"] - output["w"]) / output["cols"]
+    estimate = (cell_ew + cell_ns) / 2.0
+    return estimate
+
+
+def setup_location(name, dbase_location, epsg, src_env):
```
wenzeslaus, line 97: dbase_location - In the context of the "classic" naming practice, dbase_location is confusing: Is it dbase or location? You can call it _path_ because that's what it is.

python/grass/jupyter/utils.py
```diff
@@ -0,0 +1,127 @@
...
+    """Setup temporary location with different projection but
+    same computational region as source location
+
+    :param str name: name of new location
+    :param path dbase_location: path to new location database
```
wenzeslaus, line 102: Maybe _of_ or _'s_ would clarify the description here.

python/grass/jupyter/utils.py
```diff
@@ -0,0 +1,127 @@
+#
+# AUTHOR(S): Caitlin Haedrich <caitlin DOT haedrich AT gmail>
+#
+# PURPOSE:   This module contains utility functions for InteractiveMap.
+#
+# COPYRIGHT: (C) 2021 Caitlin Haedrich, and by the GRASS Development Team
+#
+#            This program is free software under the GNU Gernal Public
+#            License (>=v2). Read teh file COPYING that comes with GRASS
```
wenzeslaus, line 9: typo: GNU **Gernal** Public...Read **teh** file COPYING that comes with GRASS

Same in python/grass/jupyter/interact_display.py.

You should not be writing these by hand.

python/grass/jupyter/interact_display.py
```diff
@@ -166,3 +216,10 @@ def show(self):
         fig.add_child(self.map)
 
         return fig
+
+    def save(self, filename="GRASS_map.html"):
```
wenzeslaus, line 220: What use case do you have in mind for the default? Is this trying to mimic the d-command behavior in creating `map.png` by default? How do you plan to document it?

wenzeslaus (changes requested):

python/grass/jupyter/utils.py
chaedri: Ah, good catch. I went hunting for the proper header to put here and found some conflicting methods. There's the [https://trac.osgeo.org/grass/wiki/Submitting/Python](guidelines on the wiki) which are similar to what I have but as a docstring instead of a comment. But there's also lots (most?) of python scripts on github that use comments (see [https://github.com/OSGeo/grass/blob/master/python/grass/utils/download.py](utils/download.py) or [https://github.com/OSGeo/grass/blob/master/lib/init/grass.py](lib/init/grass.py)). There's also examples I found in the python directory that have more a Sphinx docstring style (such as [https://github.com/OSGeo/grass/blob/master/python/grass/script/raster.py](raster.py)). What do you think is most appropriate here?

python/grass/jupyter/interact_display.py
chaedri: Yes, I think I was thinking of the d-command behavior here. `map.png` would probably be a better choice. Use-case-wise, I was imagining users might want to embed their map in a website or create a website from. They create their map then save as an html (i.e. `m.save(filename="map_for_website.html")`).

wenzeslaus: Right, so what is the use case for the default value? In other words, in what circumstances will the user know about the default and write save() while not saying where to save?

python/grass/jupyter/utils.py
wenzeslaus: This style of header is fine. Just don't write it by hand and if you do, don't include typos :-)

chaedri: Typos fixed. I'll copy and paste next time 👍

python/grass/jupyter/interact_display.py
chaedri: AH, ok, I see your point now. Having a default is not necessary. The line should be `def save(filename):`.

python/grass/jupyter/utils.py
wenzeslaus: No, with the current code, this work well.

python/grass/jupyter/interact_display.py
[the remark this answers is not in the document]
wenzeslaus: setup_location works for me.

python/grass/jupyter/interact_display.py
```diff
@@ -142,11 +144,71 @@ def add_vector(self, name):
...
+        # Reprojects bounds of raster for overlaying png (THIS HAS TO BE IN WGS84)
+        bounds = gs.read_command("g.region", flags="g", env=self._src_env)
+        bounds = gs.parse_key_val(bounds, sep="=", vsep="\n")
+        proj = gs.read_command("g.proj", flags="jf", env=self._src_env)
+        north, east = convert_coordinates_to_latlon(bounds["e"], bounds["n"], proj)
+        south, west = convert_coordinates_to_latlon(bounds["w"], bounds["s"], proj)
+        new_bounds = [[north, west], [south, east]]
+
+        # Write raster to png file with GrassRenderer
+        filename = os.path.join(self.tmp_dir, "map.png")
```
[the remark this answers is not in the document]
chaedri: I'm still not sure I fully understand why passing the original name could cause a problem here. Are you meaning that `elevation@PERMANENT` and `elevation@user1` are different rasters but we want to add both of them to our folium map. In that case, we'd end up with two `elevation`s in the temporary location which would cause a problem. Perhaps replacing the `@` in the `full_name` string to make something like `elevation_user1` as the new name would solve this?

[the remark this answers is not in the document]
wenzeslaus: I see two issues:

1. `self.tmp_dir` + `"map.png"` will be the same filename for each `add_raster` call. Is this a problem? For GrassRenderer? For folium.ImageOverlay`? I have no idea how ImageOverlay behaves or what is in its doc. When this image needs to be present? Using the new name (from 2.) might be the best option at this point.
2. Replacing `@` with `_` should work well enough. We will always do that so that should avoid any conflicts (mapname@mapsetname is a unique identifier for a map of a given type within a location). (If we would convert rasters to vectors or the other way around - like in the possible alternative workflows we talked about - that would create potential for conflicts, but we could avoid that by simply using different location or different mapset for that workflow.)

wenzeslaus (approved):

wenzeslaus:
Try online at https://mybinder.org/v2/gh/chaedri/grass/interactive-rasters?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_integration.ipynb

PR #1772, opened 2021-08-02
misc: fix indentation for r.series and r.patch

aaronsms:
- Reformat r.series
- Reformat r.patch


marisn (approved):

aaronsms:
I have added the two commits above to .git-blame-ignore-revs.

wenzeslaus (changes requested):
Please remove the commits from .git-blame-ignore-revs. Squashing the commits during merge will create a different commit (with a different hash). This is something which needs to be done separately after the merge, hence the need for the label.

wenzeslaus:
I have clarified the label description.

aaronsms:
I have reverted the last commit.

PR #1777, opened 2021-08-04
Fix valid location check, sync with other functions

wenzeslaus:
* Path to database was dropped when constructing the full path to location in 94bcea4216cdabcd18dbe5dc4686e38d4d3800de.
* This fixes the path join code in valid location check.
* Creating mapsets from command line with -c now works again.
* Other validitity and existence checks now use the same 1-or-3 parameter interface (path or db, loc, mapset).
* Tests added for modified functions and for couple of related ones.
* Import order fixed with isort not to have glob mixed GRASS imports as pointed out by Pylint.
* Two missing docstrings added to the is-current-x checks.

### DIFF
--- a/python/grass/grassdb/checks.py
+++ b/python/grass/grassdb/checks.py
@@ -9,50 +9,71 @@ for details.
 .. sectionauthor:: Vaclav Petras <wenzeslaus gmail com>
 """
 
+import datetime
+import glob
 import os
 import sys
-import datetime
 from pathlib import Path
-from grass.script import gisenv
-import grass.script as gs
-import glob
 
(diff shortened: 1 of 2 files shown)


marisn (approved):

wenzeslaus:
Thanks @marisn. There is a CodeQL failure, but that's unrelated (something newly detected by CodeQL in tests). Merging.

marisn:
> Thanks @marisn. There is a CodeQL failure, but that's unrelated (something newly detected by CodeQL in tests). Merging.

Yes, I noticed the same CodeQL issue all over the place.

PR #1782, opened 2021-08-10
r.patch: implement parallelization with OpenMP

aaronsms:
Checklists before merging:
- [x] code review
- [x] CI passes
- [x] rebase to main


raster/r.patch/main.c
```diff
@@ -130,46 +156,109 @@ int main(int argc, char *argv[])
...
-                continue;
-
-            Rast_get_row(infd[i], patch, row, out_type);
-            if (!do_patch(presult, patch, &statf[i], ncols, out_type,
-                          out_cell_size, use_zero))
-                break;
+
+    int computed = 0;
+    int written = 0;
+
```
HuidaeCho, line 188: `row = 0` to avoid
```
main.c:250:5: warning: ‘row’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     G_percent(row, nrows, 2);
```

HuidaeCho:
Please add a test script.

raster/r.patch/main.c
aaronsms: Yeap fixed it! Supposed to be `G_percent(nrows, nrows, 2)`.

wenzeslaus:
Compiling with GCC gives:

```
main.c: In function ‘main’:
main.c:250:5: warning: ‘row’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  250 |     G_percent(row, nrows, 2);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~
```

wenzeslaus:
This will also need a update/rebase/merge with the main branch because there is a new _flake8_ job expected which needs to be completed before the merge to main can happen.

aaronsms:
I have fixed the compilation issues as well as rebased to the main branch.

petrasovaa:
Results from the benchmark:
rpatch_benchmark

petrasovaa:
Here is a benchmark patching multiple tiles in 16 bil raster:
rpatch_benchmark_large

@aaronsms  Any idea why the memory has no effect?

aaronsms:
> Here is a benchmark patching multiple tiles in 16 bil raster: rpatch_benchmark_large
> 
> @aaronsms Any idea why the memory has no effect?

The result is interesting, I should expect the 4000MB and 300MB should be around the same. The memory option allows user to choose the amount of memory to allocate to enable parallelism, but it does not always improve the performance, as long as it is not below some level of memory, where the chunk size gets too small and the threads start waiting for each other.

This begs the question: is the memory option then needed, if it turns out that even at very large files (16B), this optimal threshold of memory to hit, does not grow very large, since the user should expect to use 200/300 MB anyway, perhaps that can be hidden from the user?

petrasovaa:
Here is a benchmark on a much smaller area, looks like with memory=0 there seems to be some slow down with a lot threads.

rpatch_benchmark_large

arsonic12:
For memory=0, it auto sets to the bare minimum required for it to work, which is the number of threads * 1 row.

petrasovaa (approved):

PR #1786, opened 2021-08-11
lib/gis: allow creation of temporary files outside the current mapset

metzm:
This PR introduces some new functions in lib/gis to create temporary files in a custom location. The new functions are `G_tempfile_basedir()`, `G_file_name_basedir()`, `G_make_mapset_object_group_basedir()`. These new functions are generic and not tied to a particular GRASS library or module.

If the new argument basedir is NULL, these new functions behave like the existing funtions without _basdir. Thus the new _basedir functions could replace the old functions if function calls are updated accordingly, e.g. `G_tempfile(NULL)` instead of `G_tempfile()`.

~~Example usage is implemented in the raster lib, where temporary data files for new raster maps can be stored in an existing folder specified with the env var GRASS_RASTER_TMPDIR. Setting GRASS_RASTER_TMPDIR to a folder on a fast block device while the current mapset is located on a slow block device might speed up raster processing. Upon closing a new raster map, these temporary data are moved to the current mapset, which could be slower than creating the temporary raster data files directly on the slower block device.~~

The most efficient way to speed up data processing with final data stored on a slow block device is to create a temporary mapset on a fast block device and copy the final results to the slow block device.

Replacing simply /path/to/mapset/.tmp/ with a link to some directory on another mount point can cause fatal errors when moving temporary files to their final destination in the current mapset, because moving or renaming files across mount points is not always supported.

The new `G_*_basedir()` functions should be only used where temporary files are created, moved, and deleted. That means, each GRASS library and module is responsible for handling these temporary files, not libgis because libgis can not know and does not need to know how these temporary files are used.

### DIFF
--- a/lib/init/variables.html
+++ b/lib/init/variables.html
@@ -328,6 +328,17 @@ PERMANENT
   exists, <em><a href="v.generalize.html">v.generalize</a></em> runs
   in extremely slow debug mode.</dd>
 
+  <dt>GRASS_RASTER_TMPDIR</dt>
+  <dd>[rasterlib]<br> By default the GRASS temporary directory is
+  located in <tt>$LOCATION/$MAPSET/.tmp/$HOSTNAME</tt>. If
+  GRASS_RASTER_TMPDIR is set to some existing directory, this directory 
+  will be used to store data files of newly created raster maps. This 
+  can be usefull if this temporary directory is located on a fast 
+  storage device. However, upon closing a new raster map, these files 
+  need to be transferred to the final destination (the current mapset) 
+  which might be slow or even slower than creating the temorary raster 
+  data files directly on the final storage device.</dd>
+
   <dt>GRASS_WXBUNDLED</dt>
   <dd>[wxGUI]<br>
     set to tell wxGUI that a bundled wxPython will be used.<br>
--- a/lib/init/variables.html
+++ b/lib/init/variables.html
@@ -328,17 +328,6 @@ PERMANENT
   exists, <em><a href="v.generalize.html">v.generalize</a></em> runs
   in extremely slow debug mode.</dd>
 
-  <dt>GRASS_RASTER_TMPDIR</dt>
-  <dd>[rasterlib]<br> By default the GRASS temporary directory is
-  located in <tt>$LOCATION/$MAPSET/.tmp/$HOSTNAME</tt>. If
-  GRASS_RASTER_TMPDIR is set to some existing directory, this directory
-  will be used to store data files of newly created raster maps. This
-  can be usefull if this temporary directory is located on a fast
-  storage device. However, upon closing a new raster map, these files
-  need to be transferred to the final destination (the current mapset)
-  which might be slow or even slower than creating the temorary raster
-  data files directly on the final storage device.</dd>
-
   <dt>GRASS_WXBUNDLED</dt>
   <dd>[wxGUI]<br>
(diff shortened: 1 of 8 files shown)


tmszi (approved):
Great work. Compiled and tested on OS GNU/Linux. It works as expected. Minor fixes are fixed in commits bellow.

lib/init/variables.html
```diff
@@ -331,12 +331,12 @@ <h2>List of selected (GRASS related) shell environment variables</h2>
...
-  storage device. However, upon closing a new raster map, these files 
-  need to be transferred to the final destination (the current mapset) 
-  which might be slow or even slower than creating the temorary raster 
+  GRASS_RASTER_TMPDIR is set to some existing directory, this directory
+  will be used to store data files of newly created raster maps. This
+  can be usefull if this temporary directory is located on a fast
+  storage device. However, upon closing a new raster map, these files
+  need to be transferred to the final destination (the current mapset)
+  which might be slow or even slower than creating the temorary raster
   data files directly on the final storage device.</dd>
```
ninsbl: This is indeed a good point and important documentation.

What I am wondering is, if this new functionality could be used specifically for really temporary data (like from `r.terraflow` or `r.watershed`, or temporary text files, ...)?

Will this f.ex. be available in g.tempfile?

metzm: > What I am wondering is, if this new functionality could be used specifically for really temporary data (like from `r.terraflow` or `r.watershed`, or temporary text files, ...)?

Yes, this PR was designed for modules creating potentially large temporary data. Modules such as `r.terraflow`, `r.watershed`, `i.segment`, `g.tempfile` would need a new option where to create files for temporary data.

However, the most efficient way to speed up data processing with final data stored on a slow block device is to create a temporary mapset on a fast block device, if available, and copy the final results to the slow block device.
> 
> Will this f.ex. be available in g.tempfile?

If `g.tempfile` becomes a new option for the base directory of temporary files, yes.

metzm:
> [...] Minor fixes are fixed in commits bellow.

Thanks!

lib/init/variables.html
ninsbl: Nice!

neteler:
@metzm @tmszi may this PR be merged?

tmszi:
> @metzm @tmszi may this PR be merged?

It seems that the original main intention in this PR has not been fulfilled, be able to define a raster temporary directory on another disk (fast NVMe disk) for raster modules, to speed up the whole process of creating the resulting raster maps as mentioned above. But it seems that it is slower than using a temporary directory on the same disk partition as is the structure of the GRASS GIS data directory. However, I do not have an exact comparison.

metzm:
The idea to define a raster temporary directory on another, faster disk for raster modules, to speed up the whole process of creating the resulting raster maps does not work. The raster maps still need to be transferred to the final destination. Writing out to a faster disk, copying to a slower disk, deleting on the faster disk will be slower than writing out directly and only once to the final destination. Standard raster modules, e.g. `r.mapcalc` or `r.neighbors` will not benefit from a raster temporary directory on another, faster disk. That means the paragraph in the original description referring to the raster lib is no longer valid.

The idea of a temporary directory on a faster disk makes only sense for temporary files that are deleted at the end, for example files created by the segment lib (e.g. `r.cost`, `r.walk`, `r.watershed`, `i.segment`) or the cache of `r.proj`.

PR #1792, opened 2021-08-13
r.clump: fix clumping of FCELL/DCELL maps with threshold=0

### LINKED ISSUE #1599: r.clump clumps cells that do not have the same cell values
**Describe the bug**

Running r.clump on a floating raster with gradually changing values, and with threshold value set to 0 (default). The result is a a map with clusters consisting of raster cells that do not have the same values. Likely cause, r.clump currently converts a floating point raster to integer first.

**To Reproduce**

```
r.random.surface output=test high=100
r.mapcalc "test2 = test/1000.0"
r.clump -d input=test2 output=test3
```

**Expected behavior**
Only contiguous cells with the same cell values receive the same category (clump) value

- Operating System: Windows 10
- GRASS GIS version [e.g. 7.8.5]

metzm:
With threshold=0, FCELL/DCELL maps are truncated to CELL, resulting in clumps of input cells with different values. Fixes #1599.


marisn (approved):

mlennert:
@metzm Can this be merged ?

neteler:
@ecodiv Do you see a chance to test this bugfix which addresses your issue?

wenzeslaus (approved):
I would appreciate an addition to documentation with what is basically the PR description, i.e., what to expect as a result, but the change seems correct in the context.

PR #1795, opened 2021-08-15
Define MIN() and MAX() correctly and exactly once.

infrastation:


marisn (approved):

nilason (approved):

include/grass/gis.h
```diff
@@ -129,6 +129,13 @@ static const char *GRASS_copyright __attribute__ ((unused))
 #define CONFIG_DIR ".grass8"
 #endif
 
+#ifndef MAX
```
metzm, line 132: Why use a previously defined MAX or MIN? You don't know what these previously defined MAX and MIN macros are doing, thus for protection add
```
#ifdef MAX
#undef MAX
#endif
``` 
for MIN accordingly.

metzm (changes requested):

include/grass/gis.h
infrastation: Yes, that's a valid point. libgeotiff before commit 9b9ea38 (the one in Ubuntu 20.04) defines these incorrectly, for instance.

metzm (approved):

nilason:
Any objections to merge this PR?

PR #1797, opened 2021-08-15
r.random.cells: treat distance as minimum distance

### LINKED ISSUE #1779: [Bug] Option distance in r.random.cells does not specify minimum distance
**Describe the bug**
According to the documentation, _r.random.cells_ should generate a random sets of raster cells that are at least **distance** apart. I.e., if run with **distance** set to region resolution and **ncells** set to number of region cells, one would expect the resulting map to contain no NULLs.
Instead, it seems to generate a set of raster cells that are _more than_ **distance** apart and with the abovementioned settings, the resulting map contains ~60% NULLs.

**To Reproduce**
(maps from NC dataset)
1. set computational region: `g.region -p geology_30m@PERMANENT`
2. distribute 225000 cells (ncells in geology_30m@PERMANENT) with 30m distance (resolution of geology_30m@PERMANENT): 
`r.random.cells output=rand_cells_30m distance=30 ncells=225000`
3. the output map contains only ~81000 non-null cells
4. distribute the same number of cells with 29.999m distance:
`r.random.cells output=rand_cells_29m distance=29.999 ncells=225000`
5. The output map contains exactly 225000 non-null cells

**Expected behavior**
Given the specified distance and number of cells, the output map should contain exactly 225000 non-null cells. Alternatively, the documentation could be changed to _r.random.cells_ generates a random sets of raster cells that are more than **distance** apart.

- Operating System: Ubuntu 18.04.5 LTS
- GRASS GIS version 7.8.5

metzm:
Currently, the documentation says that "`r.random.cells` generates a random sets of raster cells that are at least distance apart", but in fact `r.random.cells` generates a random sets of raster cells that are more than distance apart.
This PR syncs the behaviour to the documentation: random cells are at least distance apart. Fixes #1779.

### DIFF
--- a/raster/r.random.cells/init.c
+++ b/raster/r.random.cells/init.c
@@ -73,7 +73,7 @@ void Init()
     G_debug(1, "(CellCount):%d", CellCount);
 
     sscanf(Distance->answer, "%lf", &MaxDist);
-    if (!(MaxDist > 0.0))
+    if (MaxDist <= 0.0)
 	G_fatal_error(_("Distance must be > 0.0"));
     
     G_debug(3, "(MaxDist):%.12lf", MaxDist);
--- a/raster/r.random.cells/init.c
+++ b/raster/r.random.cells/init.c
@@ -73,8 +73,8 @@ void Init()
     G_debug(1, "(CellCount):%d", CellCount);
 
     sscanf(Distance->answer, "%lf", &MaxDist);
-    if (MaxDist < 0.0)
-	G_fatal_error(_("Distance must be >= 0.0"));
+    if (!(MaxDist > 0.0))
+	G_fatal_error(_("Distance must be > 0.0"));
     
     G_debug(3, "(MaxDist):%.12lf", MaxDist);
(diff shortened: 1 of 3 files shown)


marisn (approved):

mlennert:
@metzm, can this be merged ?

raster/r.random.cells/init.c
```diff
@@ -73,8 +73,8 @@ void Init()
     G_debug(1, "(CellCount):%d", CellCount);
 
     sscanf(Distance->answer, "%lf", &MaxDist);
-    if (MaxDist < 0.0)
-	G_fatal_error(_("Distance must be >= 0.0"));
+    if (!(MaxDist > 0.0))
```
susi, line 76: To me `MaxDist <= 0.0` is more readable, requires less thinking :sweat_smile:

neteler: @metzm do you have any opinion on this suggested change?

neteler, line 76:
```suggestion
    if (MaxDist <= 0.0)
```

PR #1800, opened 2021-08-17
Docstring improvements

chaedri:
* Minor improvements to the docstrings in `grass.jupyter`

Comments, improvements, suggestions welcome! Clear docstrings are important.

This PR part of an ongoing Google Summer of Code project, Improved Integration of GRASS GIS and Jupyter Notebooks. You can find more information [here](https://trac.osgeo.org/grass/wiki/GSoC/2021/JupyterAndGRASS).


python/grass/jupyter/display.py
```diff
@@ -92,7 +92,10 @@ def __init__(
         self._env["GRASS_LEGEND_FILE"] = str(self._legend_file)
 
     def run(self, module, **kwargs):
-        """Run modules from "d." GRASS library"""
+        """Run modules from "d." GRASS library using grass.script.run_command().
```
wenzeslaus, line 95: `using grass.script.run_command()` is relevant in the detailed description because it, for example, influences what syntax should be used. However, "using run_command" is not the purpose of the this function which is what this first sentence of documentaion should describe.

GRASS library is confusing: C library? Python library? Library of modules makes sense, but it is rarely used. No clear rule here. Things such as "display modules" and "modules from the display family" are used. "d-modules"/"d-commands" is not, but seemed okay to me lately, but not readable for novices similarly to the more traditional "`d.*` commands".

python/grass/jupyter/display.py
```diff
@@ -92,7 +92,10 @@ def __init__(
...
+
+        :param str module: name of GRASS module
+        :param **kwargs: named arguments passed to run_command()"""
```
wenzeslaus, line 98: Please check Sphinx for both the kwargs syntax and reffering to functions (I'm not sure if the `**` is needed and cursive may be recommend instead of `()` when referring to functions).

wenzeslaus (changes requested):
Please include somewhere (perhaps in `__init__.py`?) that this is experimental and that the API may change between 8.0 and 8.2. See grass.benchmark or grass.app which are different cases, but a good starting point.

python/grass/jupyter/interact_display.py
```diff
@@ -139,7 +139,14 @@ def add_vector(self, name):
 
     def add_raster(self, name, opacity=0.8):
         """Imports raster into temporary WGS84 location,
-        exports as png and overlays on folium map
+        exports as png and overlays on folium map.
+
+        Color table for the raster can be modified with `r.colors` before calling
+        this function. NOTE: this will only work if the raster is located in the
```
wenzeslaus, line 145: If you want something like NOTE, use Sphinx/Docutils way of doing a note.

wenzeslaus (changes requested):

wenzeslaus (approved):

wenzeslaus:
API documentation for grass.jupyter merged. Please, open new issues if there is anything unclear or needs other changes.

PR #1801, opened 2021-08-17
wxGUI: new design for GNotebooks

lindakladivova:
This PR creates a new design without using an old green color for GNotebooks.
It involves GNotebooks for lmgr, nviz, vnet and dmgr (all notebooks which use globalvar.FNPageColor variable).
I think it is a good idea to have the look of all these notebooks uniform.

Multi-Window GUI lmgr notebook before:
normalmulti


veroandreo:
I like the suggestion, the default not so much though. How does it look if the user chooses a dark theme system-wide?

petrasovaa:
I thought you would also try just regular notebook, so that would be FormNotebook, not GNotebook. The grey looks ok, but I would like to see the dark theme also as Vero suggests.

lindakladivova:
> 
> 
> I thought you would also try just regular notebook, so that would be FormNotebook, not GNotebook. The grey looks ok, but I would like to see the dark theme also as Vero suggests.

Ok, I will try but I cannot find where or how to set up the dark mode. Can you give me a hint?

petrasovaa:
> > I thought you would also try just regular notebook, so that would be FormNotebook, not GNotebook. The grey looks ok, but I would like to see the dark theme also as Vero suggests.
> 
> Ok, I will try but I cannot find where or how to set up the dark mode. Can you give me a hint?

Depends on your ubuntu version, I suggest just google it.

lindakladivova:
> 
> 
> > > I thought you would also try just regular notebook, so that would be FormNotebook, not GNotebook. The grey looks ok, but I would like to see the dark theme also as Vero suggests.
> > 
> > 
> > Ok, I will try but I cannot find where or how to set up the dark mode. Can you give me a hint?
> 
> Depends on your ubuntu version, I suggest just google it.

Great, I thought that it is something related directly to GRASS.
I think it looks good but probably a bigger problem is for MapDisplay AuiNotebook tabs... they really do not look good in the dark mode. What do you think?
darkmode180
darkmode

veroandreo:
Indeed, map display tabs in the single window output don't look any good in the dark mode. Is is possible to use plain colors instead of the gradient?

petrasovaa:
> Indeed, map display tabs in the single window output don't look any good in the dark mode. Is is possible to use plain colors instead of the gradient?

Not so easy, it's not easily customizable, one way would be to derive our own AuiTabArt class, that's for later.

petrasovaa:
> I thought you would also try just regular notebook, so that would be FormNotebook, not GNotebook.

I realized the wx.Notebook doesn't have the right click menu functionality (currently used for renaming). So we would have to deal with that somehow, not sure how, so we might need to stick with FlatNotebook.

lindakladivova:
Well. I have tried to use the bare style settings without color settings and it looks quite nice for Light mode, but not much for the Dark mode. :-(
Not sure how to go on with the dark mode... I have tried to use System settings, but I would need to use black text color for both cases - dark mode as well as light mode... but the dark one uses white instead of black.

Screenshots:
pr1801
pr1801b

Furthermore, I have noticed there are two interesting options - FNB_HIDE_ON_SINGLE_TAB and FNB_NAV_BUTTONS_WHEN_NEEDED. The first one hides the notebook if there is only one tab, the second one hides the navigation arrows when all tabs fit the space. @petrasovaa  , any suggestions?

lindakladivova:
I would add, that I personally think that in general tabs and background without gradient look much better so the way we go is likely right.. but for the dark mode it is problematic,

petrasovaa (approved):
I like it and I also like the option to hide the tab. One display covers majority of cases anyway. I tested that same option for the aui notebook in single window and it looked a little weird so am not sure there.

lindakladivova:
I removed fancy style for all GNotebooks and set it to normal. There are some screenshots:

**3D View tab:**
3dview
3dtabdark

**Web services:**
webservice2
web_service_dark

**Modeler:**
modeler

**Attribute manager:**
Figure 1:
attribute_manager
Figure 2:
webservice2

Everything looks the same basically - no fancy stuff, no gradients.

However, I am sure about _psmap/frame.py_, _nviz/tools.py_ and sqlNtb in _dbmg/base.py_ - here we have the base **FlatNotebooks**.
You can e.g. notice that the upper notebook in Attribute manager in Figure 1 has a different layout than in Figure 2 where I have set the same settings as for **GNotebooks**. 

Maybe, we could change these three FlatNotebooks to the GNotebooks. What do you think @petrasovaa ?

gui/wxpython/gui_core/forms.py
```diff
@@ -1053,11 +1053,10 @@ def __init__(self, parent, giface, task, id=wx.ID_ANY, frame=None, *args, **kwar
         elif style == 1:  # basic left
             self.notebook = FormNotebook(self, style=wx.BK_LEFT)
             self.notebook.Bind(wx.EVT_NOTEBOOK_PAGE_CHANGED, self.OnPageChange)
-        elif style == 2:  # fancy green
+        elif style == 2:  # normal style
```
petrasovaa, line 1056: Could you completely remove this style? It doesn't make sense to leave it there, it's very similar to the simple top style.

petrasovaa (changes requested):
I agree with the changes, and yes, making the FlatNotebooks into GNotebook would make sense.

lindakladivova:
@petrasovaa, I have a problem with the dbmgr InsertNBPage call. It throws this error when clicking on Attribute manager:

  File "/home/linwe/grass/dist.x86_64-pc-linux-
gnu/gui/wxpython/lmgr/frame.py", line 1649, in
OnShowAttributeTable

selection=selection,
  File "/home/linwe/grass/dist.x86_64-pc-linux-
gnu/gui/wxpython/dbmgr/manager.py", line 129, in __init__

self.CreateDbMgrPage(parent=self, pageName="browse")
  File "/home/linwe/grass/dist.x86_64-pc-linux-
gnu/gui/wxpython/dbmgr/base.py", line 831, in
CreateDbMgrPage

parent=parent, parentDbMgrBase=self, onlyLayer=onlyLayer
  File "/home/linwe/grass/dist.x86_64-pc-linux-
gnu/gui/wxpython/dbmgr/base.py", line 1112, in __init__

self.AddLayer(layer)
  File "/home/linwe/grass/dist.x86_64-pc-linux-
gnu/gui/wxpython/dbmgr/base.py", line 1177, in AddLayer

% (layer, label,
self.dbMgrData["mapDBInfo"].layers[layer]["table"]),
  File "/home/linwe/grass/dist.x86_64-pc-linux-
gnu/gui/wxpython/gui_core/widgets.py", line 311, in
InsertNBPage

self.controller.InsertPage(**kwargs)
  File "/home/linwe/grass/dist.x86_64-pc-linux-
gnu/gui/wxpython/gui_core/widgets.py", line 278, in
InsertPage

self.classObject.InsertPage(self.widget, **kwargs)
  File "/usr/lib/python3/dist-
packages/wx/lib/agw/flatnotebook.py", line 4318, in
InsertPage

self.AddPage(page, text, select, imageId)
TypeError
:
AddPage() takes 1 positional argument but 5 were given

I have looked at https://github.com/wxWidgets/Phoenix/blob/master/wx/lib/agw/flatnotebook.py on row 4318, where I obviously need to pass 5 parameters to AddPage method. I have also looked into lmgr/frame.py where this InsertNBPage method is also used. It seems to me, that I call it the same as in lmgr. I am probably missing something but not able to figure out what.

As for psmap and nviz, no special methods from GNotebook class are called. I tested them and it works well.

gui/wxpython/core/settings.py
```diff
@@ -808,8 +808,7 @@ def _internalSettings(self):
         self.internalSettings["appearance"]["gSelectPopupHeight"]["max"] = 1000
         self.internalSettings["appearance"]["commandNotebook"]["choices"] = (
             _("Basic top"),
-            _("Basic left"),
-            _("Fancy green"),
+            _("Normal style"),
             _("List left"),
```
petrasovaa, line 812: I wanted to remove the Fancy green, not Basic left, what is supposed to be Normal style?

gui/wxpython/gui_core/forms.py
```diff
@@ -1050,16 +1050,12 @@ def __init__(self, parent, giface, task, id=wx.ID_ANY, frame=None, *args, **kwar
         if style == 0:  # basic top
             self.notebook = FormNotebook(self, style=wx.BK_TOP)
             self.notebook.Bind(wx.EVT_NOTEBOOK_PAGE_CHANGED, self.OnPageChange)
-        elif style == 1:  # basic left
-            self.notebook = FormNotebook(self, style=wx.BK_LEFT)
-            self.notebook.Bind(wx.EVT_NOTEBOOK_PAGE_CHANGED, self.OnPageChange)
```
petrasovaa, line 1055: keep this

gui/wxpython/gui_core/forms.py
```diff
@@ -1050,16 +1050,12 @@ def __init__(self, parent, giface, task, id=wx.ID_ANY, frame=None, *args, **kwar
...
-        elif style == 2:  # fancy green
+        elif style == 1:  # normal style
             self.notebook = GNotebook(
                 self, style=globalvar.FNPageStyle | FN.FNB_NO_X_BUTTON
             )
```
petrasovaa, line 1056: remove this

gui/wxpython/gui_core/widgets.py
```diff
@@ -264,21 +264,22 @@ def GetPageIndexByName(self, page):
...
-        self.classObject.InsertPage(self.widget, **kwargs)
+        self.classObject.InsertPage(self.widget, *args, **kwargs)
 
 
 class GNotebook(FN.FlatNotebook):
     """Generic notebook widget.
 
     Enables advanced style settings.
+    Uses system colours for active tabs.
     Problems with hidden tabs and does not respect system colors (native look).
```
petrasovaa, line 283: Hmm, uses system colours and does not respect system colors at the same time?

petrasovaa:
> @petrasovaa, I have a problem with the dbmgr InsertNBPage call. It throws this error when clicking on Attribute manager:
> I have looked at https://github.com/wxWidgets/Phoenix/blob/master/wx/lib/agw/flatnotebook.py on row 4318, where I obviously need to pass 5 parameters to AddPage method. I have also looked into lmgr/frame.py where this InsertNBPage method is also used. It seems to me, that I call it the same as in lmgr. But probably missing something and not able to figure it out.

That was little tricky, we were missing *args, since they don't use keyword, but positional arguments in the notebook AddPage call. Please test, I tested it just briefly...

lindakladivova:
> 
> 
> > @petrasovaa, I have a problem with the dbmgr InsertNBPage call. It throws this error when clicking on Attribute manager:
> > I have looked at https://github.com/wxWidgets/Phoenix/blob/master/wx/lib/agw/flatnotebook.py on row 4318, where I obviously need to pass 5 parameters to AddPage method. I have also looked into lmgr/frame.py where this InsertNBPage method is also used. It seems to me, that I call it the same as in lmgr. But probably missing something and not able to figure it out.
> 
> That was little tricky, we were missing *args, since they don't use keyword, but positional arguments in the notebook AddPage call. Please test, I tested it just briefly...

Great, I will know for the next time when I encounter this. :-) I tested it and it works fine. I think it is ready for merge.

petrasovaa (approved):

PR #1271, opened 2021-01-25
Fix compiler warnings, part 5 (rst/interpfl lib & v.surf.rst)

nilason:
Output deviation argument to `IL_init_params_2d()` and in the `interp_params` struct (lib/rst/interp_float), was defined as FILE*. In practice a char* of the file path was given to the function in 'v.surf.rst/main.c' and this pointer's only use was to check whether it was NULL or not. It was used as a flag (and this is the reason it worked at all). This changes the argument to be ~~an integer, representing a boolean value~~ a `bool`.

Addresses one -Wincompatible-pointer-types compiler warning.

```
/usr/bin/clang  -g -pipe -arch x86_64 -DGL_SILENCE_DEPRECATION    -I/Volumes/grass/dist.x86_64-apple-darwin19.6.0/include -I/Volumes/grass/dist.x86_64-apple-darwin19.6.0/include   -I/opt/local/include -I/opt/local/include  -DPACKAGE=\""grassmods"\"  -I/opt/local/include/postgresql12  -I/Volumes/grass/dist.x86_64-apple-darwin19.6.0/include -I/Volumes/grass/dist.x86_64-apple-darwin19.6.0/include -DRELDIR=\"vector/v.surf.rst\" -o OBJ.x86_64-apple-darwin19.6.0/main.o -c main.c
main.c:646:31: warning: incompatible pointer types passing 'char *' to parameter of type 'FILE *' (aka 'struct __sFILE *') [-Wincompatible-pointer-types]
                      Tmp_fd_yy, Tmp_fd_xy, devi, NULL, cv,
                                            ^~~~
```

```
gcc  -std=gnu11   -I/home/runner/work/grass/grass/dist.x86_64-pc-linux-gnu/include -I/home/runner/work/grass/grass/dist.x86_64-pc-linux-gnu/include   -I/usr/include/gdal -I/usr/include -fopenmp -DPACKAGE=\""grassmods"\"    -I/home/runner/work/grass/grass/dist.x86_64-pc-linux-gnu/include -I/home/runner/work/grass/grass/dist.x86_64-pc-linux-gnu/include -DRELDIR=\"vector/v.surf.rst\" -o OBJ.x86_64-pc-linux-gnu/main.o -c main.c
main.c: In function ‘main’:
main.c:646:31: warning: passing argument 41 of ‘IL_init_params_2d’ from incompatible pointer type [-Wincompatible-pointer-types]
  646 |         Tmp_fd_yy, Tmp_fd_xy, devi, NULL, cv,
[...]

PR #1804, opened 2021-08-17
Add -Werror to "GCC C/C++ standards check" CI builds

nilason:
This adds compiler flag `-Werror` to "*GCC C/C++ standards check*" CI builds. Compiler warnings are treated as errors.

This should only be merged after #1271 and #1406 (which fixes all existing warnings) have been merged.


wenzeslaus (commented):
I think the best place to change this is the _GCC C/C++ standards check_ workflow (gcc.yml), not the build script which is used from multiple workflows.

https://github.com/OSGeo/grass/blob/8ce9a9e981281e00afdc2d3516d94ad28c3cb20a/.github/workflows/gcc.yml

.github/workflows/gcc.yml
```diff
@@ -45,4 +45,4 @@ jobs:
           CFLAGS: "-std=${{ matrix.c }}"
           # TODO: -pedantic-errors here won't compile
           CXXFLAGS: "-std=${{ matrix.cpp }}"
-        run: .github/workflows/build_ubuntu-20.04.sh $HOME/install
+        run: .github/workflows/build_ubuntu-20.04.sh $HOME/install -Werror
```
wenzeslaus, line 48: The CFLAGS right above is not working for you? That's what is determining the standard used, so it better works.

nilason: > The CFLAGS right above is not working for you? That's what is determining the standard used, so it better works.

No, it's not working locally, nor on the CIs.
My take is that the configure file is based on a too old version of autoconf (v2.13 from 1999). See `config.log` locally to see there are a number of warnings, which ends up as errors during configuration with `-Werror` causing configuration failure. 
For a permanent solution configure.in should be updated. While this only affects certain CI build for testing purposes I'd say this workaround is acceptable for the time being.

nilason:
With present workaround only the GCC CIs fail as expected.
For testing purposes I committed the fixes from #1271 and #1406, which should make all CIs pass.

nilason:
Apparently the `-fPIC` flag needs to be added for `-Werror` to work.
Also, #1795 is required before this is to be merged.

.github/workflows/gcc.yml
wenzeslaus: Then just add a note to the definition of `makecmd` that it can be removed once configure works with -Werror enabled.

PR #1813, opened 2021-08-22
ci: Fix ifs for jobs and rename master to main

wenzeslaus:
* Reference to branches (heads), not tags, to state the branch condition positively.
* Remove comparison to boolean in the condition for better readability.
* Fix comparison for master/main (use equal operator, startsWith may match too much).
* Update for rename from master to main for the default branch.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -38,7 +38,7 @@ jobs:
   # stable-alpine, stable-debian, stable-ubuntu
   docker-branch-os-matrix:
     name: build and push ${{ matrix.os }} for branch
-    if: startsWith(github.ref, 'refs/heads/') != true
+    if: startsWith(github.ref, 'refs/heads/')
     runs-on: ubuntu-latest
 
     strategy:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -38,7 +38,7 @@ jobs:
   # stable-alpine, stable-debian, stable-ubuntu
   docker-branch-os-matrix:
     name: build and push ${{ matrix.os }} for branch
-    if: ${{ startsWith(github.ref, 'refs/heads/') != true }}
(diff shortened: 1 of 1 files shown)


wenzeslaus:
Please review @mmacata.

wenzeslaus:
One change was not necessary. I missed it in the documentation. Reverted and the description updated accordingly.

.github/workflows/docker.yml
```diff
@@ -32,13 +32,13 @@ env:
...
   # Only run for push to configured branches, do not run for releases.
-  # Take care of different os. For master branch, created tags are:
+  # Take care of different os. For main branch, created tags are:
   # latest-alpine, latest-debian, latest-ubuntu
   # For releasebranch_7_8, created tags are:
   # stable-alpine, stable-debian, stable-ubuntu
   docker-branch-os-matrix:
     name: build and push ${{ matrix.os }} for branch
-    if: startsWith(github.ref, 'refs/tags/') != true
+    if: startsWith(github.ref, 'refs/heads/') != true
```
mmacata, line 41: This job is meant to run for everything except tags. Now it prevents the run for branches which is needed here. I would agree if the negotioation is removed (`!= true`) then maybe it is more clear. The check for `refs/heads/` should be sufficient, I don't think any other ref type is relevant here.

mmacata (changes requested):

.github/workflows/docker.yml
wenzeslaus: Right. I misunderstood the condition. As far as I can tell from Git and GitHub doc, `refs/heads/` is branches, so exactly what we want to query for. Keeping `startsWith...heads`, dropping the `!= true` part.

wenzeslaus:
Thanks @mmacata. Code and PR description updated. Please, review again if you can.

mmacata (approved):
Fine for me to merge. Thank you!

PR #1817, opened 2021-08-23
t.rast.export: use crossplatform tmp-directory

ninsbl:
The default for `directory` throws errors on MS Windows. This PR uses the directory returned by Python tempfile as cross-platform default...

### DIFF
--- a/temporal/t.rast.export/t.rast.export.py
+++ b/temporal/t.rast.export/t.rast.export.py
@@ -37,9 +37,8 @@
 
 # %option G_OPT_M_DIR
 # % key: directory
-# % description: Path to the work directory, default is /tmp
+# % description: Path to the work directory, standard temporary directory is used if not given
 # % required: no
-# % answer: /tmp
 # %end
 
 # %option
@@ -102,6 +101,7 @@
 # %option G_OPT_T_WHERE
 # %end
 
+import tempfile
 import grass.script as grass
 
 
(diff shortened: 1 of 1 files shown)


temporal/t.rast.export/t.rast.export.py
```diff
@@ -37,9 +37,8 @@
 
 # %option G_OPT_M_DIR
 # % key: directory
-# % description: Path to the work directory, default is /tmp
+# % description: Path to the work directory, standard temporary directory is used if not given
```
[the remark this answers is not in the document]
ninsbl: Makes sense! Split up and re-worded as label and description in 4b10e8a

temporal/t.rast.export/t.rast.export.py
```diff
@@ -37,9 +37,9 @@
 
 # %option G_OPT_M_DIR
 # % key: directory
-# % description: Path to the work directory, default is /tmp
+# % label: Path to the temporary working directory
```
veroandreo, line 40: No really strong opinion, but since this is the export directory why not saying just that, i.e., `Path to the export directory`? Temporary working seems a bit odd to me.

ninsbl:
Failing checks seem unrelated...

temporal/t.rast.export/t.rast.export.py
wenzeslaus: I think what is also expected is that the `./` stuff is not in the default answer but resolved in the program internally. It would have to be `.` and not `./` anyway.

PR #1831, opened 2021-08-29
jupyter: Render 3D images with m.nviz.image

wenzeslaus:
* 3D is rendered using m.nviz.image.
* The interface to m.nviz.image is direct/basic, i.e., the parameters are just passed as is.
* There is several special m.nviz.image parameters which are handled separately such as size and mode.
* The main function to pass the parameters is called render.
* The 2D renderer is used to create 'overlays' such as raster legend.
* In Binder and other headless environments, PyVirtualDisplay is used to call m.nviz.image.
* PyVirtualDisplay is not an optional dependency (in turn it depends on Xvfb).
* Given the API of PyVirtualDisplay, the 3D renderer cannot support custom environments (i.e., it always uses os.environ) although it makes effort not to modify it if possible.

### DIFF
--- a/.github/workflows/apt.txt
+++ b/.github/workflows/apt.txt
@@ -12,14 +12,15 @@ libopenblas-dev
 libpdal-dev
 libpng-dev
 libproj-dev
-pdal
-proj-bin
 libreadline-dev
 libzstd-dev
+pdal
+proj-bin
 python3-dateutil
 python3-matplotlib
 python3-numpy
 python3-ply
+python3-pyvirtualdisplay
 python3-six
 python3-termcolor
 sqlite3
--- a/.github/workflows/apt.txt
+++ b/.github/workflows/apt.txt
@@ -19,6 +19,7 @@ proj-bin
 python3-dateutil
 python3-matplotlib
 python3-numpy
+python3-pil
 python3-ply
 python3-pyvirtualdisplay
 python3-six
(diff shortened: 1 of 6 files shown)


wenzeslaus:
Available for testing in [Binder](https://mybinder.org/v2/gh/wenzeslaus/grass/main?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fgrass_jupyter.ipynb) (section _GRASS 3D Renderer_ in _grass_jupyter.ipynb_).

wenzeslaus:
@chaedri Please review if you can. The 3D renderer opens again some questions about naming (renderer, display, figure, ..., driver, backend) and API - I went here is `render(...)` instead of `run(...)` because it seemed different enough to require different name.

Adding some additional APIs like `add_raster(...)` (which would be really more add surface) or `add(Raster(...))` or even `add(Raster(..., color=my_grass_renderer_object))` will be possible, I think, while keeping the current basic API, but the complexity seems to be closer to folium than to the 2D renderer.

chaedri (commented):
This looks great. I see what you mean with the naming though - it's hard to know what is the clearest name for all these methods. I think the way you have it written is pretty intuitive (`render()` is different enough from `run`). Just to add to the list of possibilities, perhaps `add_surface` could be another alternative to `add_raster` or `add(Raster(...))`. I'm not sure if it's useful to have it named differently from the folium/`InteractiveMap` method or if it's confusing.

PR #1834, opened 2021-08-31
grass.jupyter: Return session from init

wenzeslaus:
* Calling grass.script.setup.init needs to paired with a call to finish.
* Explicit call to finish would have to be commented out in most notebooks with notes about nothing working once this is called, but an explanation of the need to call it at some point (confusing).
* atexit does not or may not work in Jupyter (IPython kernel). It does not for me locally and the Internet is full of various notes about not working in certain cases or versions.
* Returned session object uses weakref.finalize technique (used, e.g., in the Python standard library temporary directory object) to call the finish procedure.
* The session object takes care of switching the mapset although there is nothing in the object needed to do that since the state is global anyway.
* User needs to hold result of init, e.g., s = init(...), otherwise the session will end immediately.

### DIFF
--- a/doc/notebooks/basic_example_grass_jupyter.ipynb
+++ b/doc/notebooks/basic_example_grass_jupyter.ipynb
@@ -46,7 +46,7 @@
     "import grass.jupyter as gj\n",
     "\n",
     "# Start GRASS Session\n",
-    "gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")"
+    "session = gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")"
    ]
   },
   {
--- a/doc/notebooks/grass_jupyter.ipynb
+++ b/doc/notebooks/grass_jupyter.ipynb
@@ -357,45 +357,37 @@
   {
    "cell_type": "markdown",
    "source": [
-    "## Mapset and Session Management\n",
+    "## Switching Mapsets and Session Management\n",
     "\n",
-    "The `init` function returns a reference to a session object which can be used to manipulate the current session. The session is global (the global state of the environment is changed), but the session object is a handle for accessing this global session. Most importantly, keeping the reference means keeping the session active and that's all that is needed to properly use the session object.\n",
+    "The `init` function returns a reference to a session object which can be used to manipulate the current session. The session is global, i.e., the global state of the environment is changed. The session object is a handle for accessing this global session. When the kernel for the notebooks shuts down or is restarted, the session ends automatically. The session can be explicitly ended using `session.finish()`, but that's usually not needed in notebooks.\n",
     "\n",
     "Additionally, the session object can be used to change the current mapset. Here, we will switch to mapset called *PERMANENT*:"
-   ],
-   "metadata": {}
+   ]
   },
   {
    "cell_type": "code",
    "execution_count": null,
+   "metadata": {},
+   "outputs": [],
    "source": [
     "session.switch_mapset(\"PERMANENT\")"
-   ],
-   "outputs": [],
-   "metadata": {}
+   ]
   },
   {
    "cell_type": "markdown",
+   "metadata": {},
    "source": [
     "Now we could add more data to the PERMANENT mapset or modify the existing data there. We don't need to do anything there, so we switch back to the mapset we were in before:"
-   ],
-   "metadata": {}
+   ]
   },
   {
    "cell_type": "code",
    "execution_count": null,
+   "metadata": {},
+   "outputs": [],
    "source": [
     "session.switch_mapset(\"user1\")"
-   ],
-   "outputs": [],
-   "metadata": {}
-  },
-  {
-   "cell_type": "markdown",
-   "source": [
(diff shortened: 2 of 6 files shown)


wenzeslaus:
[binder link for testing](https://mybinder.org/v2/gh/wenzeslaus/grass/jupyter-session-object?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fbasic_example.ipynb)

wenzeslaus:
@chaedri Please see the finish/cleanup/close procedures here. The "at exit" approach does not work for Jupyter notebooks (as I found out, the "at exit" functions are not called when restarting a IPython kernel), so only "proper" solutions are possible. Try how it works and you can compare it to the clean up with the non-Jupyter code which I added to the basic example notebook in #1829.

chaedri (commented):
I think this way makes sense. When testing, I found you can run the basic_example_grass_jupyter notebook _without_ adding the `session = gj.init()`, therefore not cleaning up or unlocking the mapset. Is that intentional?

It looks great and makes sense otherwise. Perhaps in a separate PR, you or I should update the example notebook with the `session=gj.init()` line.

wenzeslaus:
You are right. I can't reproduce the behavior I observed earlier. However, now I see that it makes sense: According to the documentation the finalize function returns _a callable finalizer object which will be called when obj is garbage collected_. This means that the behavior I observed may or may not happen.

I can now reproduce only the behavior when the session object survives even when not assigned and it works as expected in the sense that it cleans up the temporary files when the kernel is restarted.

> I found you can run the basic_example_grass_jupyter notebook without adding the session = gj.init(), therefore not cleaning up or unlocking the mapset. Is that intentional?

It is actually cleaned up and happens when the kernel is restarted or sooner if the object is garbage collected. You can test this by running`!ls $GISRC`, recording the file somewhere, and then re-running the notebook and doing `!ls /the/old/gisrc/path` (or doing it in the side).

The troubling part now is that it may work without assigning the result to a variable, but it is not guaranteed to work and since it seems to mostly work, it will be harder to communicate that a user needs to do it to ensure it always works.

> Perhaps in a separate PR, you or I should update the example notebook with the session=gj.init() line.

I'll include it here since that's how it needs to be used for consistent results. I'm not sure why I didn't. Perhaps I because it will conflict with the grass.script.setup.init PR.

wenzeslaus:
It doesn't change anything, but I'm able to reproduce the failing behavior from a script quite reliably. In the following code, the g.region call fails with `ERROR: GISRC - variable not set`. Adding `s = gj.init...` makes it work and the file obtained from printing `GISRC` does not exist after the script finished.

```
import os
import subprocess
import sys
gisbase = subprocess.check_output([".../grass", "--config", "path"], text=True).strip()
os.environ["GISBASE"] = gisbase
sys.path.append(os.path.join(gisbase, "etc", "python"))
import grass.script as gs
import grass.jupyter as gj
gj.init(".../grassdata", "nc_basic_spm_grass7", "user1")
# print(os.environ["GISRC"])
gs.parse_command("g.region", raster="lakes", flags="pg")
gs.run_command("r.buffer", input="lakes", output="lakes_buff", distances=[60, 120, 240, 500])
```

wenzeslaus:
[a better binder link for testing going to `grass_jupyter.ipynb`](https://mybinder.org/v2/gh/wenzeslaus/grass/jupyter-session-object?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fgrass_jupyter.ipynb)

wenzeslaus:
The session is now stored in each notebook.

Latest changes also reflect that *g.mapset* can't be mixed with *grass.script.setup.init* (*grass.jupyter.init*). Multiple *g.gisenv* calls are used instead to switch mapsets.

wenzeslaus:
@chaedri please review again if you can.

chaedri:
Thank you for the new binder link. I think updating all the notebooks is a good call here. 

So, ok, I am now _unable_ to reproduce the behavior where I run `gj.init()` without the `session=` part and it works. Now, I consistently get the `ERROR: GISRC - variable not set` message. 

But, while experimenting with this, I also found some other behavior that doesn't seem right. If you run the first cell twice in the same kernel session, you will get an error and be unable to run other cells (i.e. launch binder link, restart kernel and run notebook, then run first cell again). This could be inconvenient because, every time you accidentally ran the first cell, you'd have to restart the kernel and lose all your variables.

wenzeslaus:
I reworked how the session is handled. The class represents a global GRASS session and now I keep a reference to one global session only, i.e., I treat it in singleton fashion with a private global module-level variable.

The _init_ function can be called more than once and the user doesn't need to keep the returned reference. I still kept the idea of _init_ returning the reference to the session object which can be used to handle the session. Although the same could be done with standalone functions, this may be good for consistency with some future version of the _grass.script.setup.init_ function.

Not assigning the reference to a variable results in `<grass.jupyter.setup._JupyterGlobalSession at 0x7f9a7c0e2ed0>` being printed. Not the nicest thing, but also common enough.

wenzeslaus:
@chaedri please review.

chaedri (approved):
This looks great to me - an elegant way of handling the session and switching mapsets. I tested it out in Binder and couldn't find any issues.

PR #1836, opened 2021-09-01
init: Default to Bash, fallback to sh

wenzeslaus:
* When SHELL env var was not set, sh would be used which likely unexpected if bash is on path.
* Now bash presence is tested and it is used if available.
* If there is no bash on path, sh is used as a fallback.
* As a bonus, this works also with an empty SHELL variable and avoids use of bare except.

### DIFF
--- a/lib/init/grass.py
+++ b/lib/init/grass.py
@@ -1664,12 +1664,20 @@ def get_shell():
         os.environ["SHELL"] = "/usr/bin/bash.exe"
         os.environ["OSTYPE"] = "cygwin"
     else:
-        # in docker the 'SHELL' variable may not be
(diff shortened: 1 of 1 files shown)


wenzeslaus:
@ocsmit Can you please have a look?

The motivation is to avoid lines [like this one (`ENV SHELL /bin/bash`)](https://gitlab.com/geoharmonizer_inea/eumap/-/blob/master/docker/00-gdal/v3.x.x-grass-saga.Dockerfile#L242) in Dockerfiles and make experimenting in Docker containers more convenient.

ocsmit (approved):
Tested by unsetting shell env variable (zsh) on my local machine. Works as intended with bash being chosen as shell in GRASS session.

PR #1838, opened 2021-09-01
init: Lazy-initialize translations, fix order dependency

wenzeslaus:
In the current code, if the translation domains (GUI, lib, modules) are
ordered differently, the GUI may not be translated. The Python gettext
documentation is silent on using install with multiple domains, but separate
install calls are used to switch between different languages, so even if
multiple install calls ever resulted in all domains being used, it is not
guarateed. The new code uses translation object directly and adds multiple
domains using fallbacks. Now GUI translations are loaded regardless of
the order of domains in the calls.

The current code requires the GISBASE variable to be set before first import
is made which makes it difficult to bootstrap. The new code delays the
initialization of translations until the first call to the translation
function removing the need to set GISBASE before the import.

The code now contains a detailed information about the current translation
approach to clarify some of the pitfalls of the current approach.

### DIFF
--- a/python/grass/__init__.py
+++ b/python/grass/__init__.py
@@ -1,32 +1,104 @@
-import gettext
+# MODULE:    grass
+#
+# AUTHOR(S): Vaclav Petras <wenzeslaus gmail com>
+#
+# PURPOSE:   Top level file of the grass package and its initialization
+#
+# COPYRIGHT: (C) 2021 Vaclav Petras, and by the GRASS Development Team
+#
+#           This program is free software under the GNU General Public
+#           License (>=v2). Read the file COPYING that comes with GRASS
+#           for details.
+
+"""Top-level GRASS GIS Python package
+
+Importing the package (or any subpackage) initializes translation functions
+so that the function ``_`` appears in the global namespace (as an additional build-in).
+"""
+
+import builtins as _builtins
 import os
 
-import six
 
-# Setup i18N
+# Setup translations
 #
-# Calling `gettext.install()` injects `_()` in the builtins namespace and
-# thus it becomes available globally (i.e. in the same process). Any python
-# code that needs to use _() should make sure that it firsts imports this
-# library.
+# The translations in the GRASS GIS Python package grass, GRASS Python modules
+# (scripts), and the wxPython GUI (wxGUI) are handled as application translations,
+# rather than Python module or package translations. This means that that translation
+# function called `_` (underscore) is added to buildins namespace. When the grass
+# package or any subpackage (or an object from there) is imported, the `_` function
+# becomes available globally (i.e. for all Python code in the same process). This is
+# the good part.
+
+# Unfortunately, creating `_` in the global namespace has unitended consequences such
+# as issues in interactive shells, with doctests, and it is conflicting with a common
+# practice of using `_` for unused variables. This is the same behavior as with the
+# common function `gettext.install()` which also adds `_` into the global namespace.
+# The solution to this is to use the module translations approach instead of
+# application translation approach. Without changing the overall approach, the current
+# code can be modified to allow for using imports of translation functions instead of
+# relying on the buildins when desired as a transitional state before removing the
+# modification of buildins.
 #
-# Note: we need to do this at the beginning of this module in order to
-# ensure that the injection happens before anything else gets imported.
+# The current code mitigates two other issues associated with `gettext.install()`
+# approach: First, it delays initialization of translations to the time when they are
+# needed instead of doing it during import time (and possibly failing when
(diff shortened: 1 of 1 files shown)


wenzeslaus:
@HuidaeCho Can you please test this? For me, installing the GUI domain first makes the GUI not translated. I was not able to confirm the behavior for modules or the lib domains and I'm not sure why. In any case, the new code works for me regardless of the order in which the domains are used.

wenzeslaus:
This fixes [Track Ticket 3772](https://trac.osgeo.org/grass/ticket/3772) by @pmav99 at least in terms of solving the `AttributeError: 'NoneType' object has no attribute 'endswith'` during import time:

```sh
$ PYTHONPATH=.../dist.x86_64-pc-linux-gnu/etc/python python -c 'import grass.script; print("OK")'
OK
```

HuidaeCho:
> 
> 
> @HuidaeCho Can you please test this? For me, installing the GUI domain first makes the GUI not translated. I was not able to confirm the behavior for modules or the lib domains and I'm not sure why. In any case, the new code works for me regardless of the order in which the domains are used.

I'll test it. Could you provide command lines to reproduce your issue with the current code?

wenzeslaus:
> Could you provide command lines to reproduce your issue with the current code?

It is two pieces, one, setting all locale related env vars (perhaps LC_ALL is enough?) and two, reordering the following lines in `python/grass/__init__.py` (and compiling):

```python
    gettext.install("grasslibs", _LOCALE_DIR)
    gettext.install("grassmods", _LOCALE_DIR)
    gettext.install("grasswxpy", _LOCALE_DIR)
```

Unfortunately, not a clear command line way of reproducing it. I would be happy if you can test that the translations work for you with the new code, although confirming the issue would be better.

I read the documentation, gettext package code, and the Internet and multiple `gettext.install` calls don't seem to be the way, but I don't understand why it is not showing as a serious problem.

wenzeslaus:
The failing CodeQL are new warnings detected by CodeQL unrelated to this PR.

marisn (approved):
Can not instantly gasp all Python finesses but running GRASS with this PR produces no bad side effects, messages are translated thus +1.

HuidaeCho:
> > Could you provide command lines to reproduce your issue with the current code?
> 
> It is two pieces, one, setting all locale related env vars (perhaps LC_ALL is enough?) and two, reordering the following lines in `python/grass/__init__.py` (and compiling):
> 
> ```python
>     gettext.install("grasslibs", _LOCALE_DIR)
>     gettext.install("grassmods", _LOCALE_DIR)
>     gettext.install("grasswxpy", _LOCALE_DIR)
> ```
> 
> Unfortunately, not a clear command line way of reproducing it. I would be happy if you can test that the translations work for you with the new code, although confirming the issue would be better.
> 
> I read the documentation, gettext package code, and the Internet and multiple `gettext.install` calls don't seem to be the way, but I don't understand why it is not showing as a serious problem.

OK, let me understand this issue first. I just tried without this PR:
```bash
LC_CTYPE=ko_KR.UTF-8 grass # to use the ko_KR.UTF-8 locale
g.list --ui # messages are translated
exit

vi dist.x86_64-pc-linux-gnu/etc/python/grass/__init__.py
# move gettext.install("grasswxpy", _LOCALE_DIR) above "grasslibs"
LC_CTYPE=ko_KR.UTF-8 grass # to use the ko_KR.UTF-8 locale
g.list --ui # messages are still translated
```
I think I'm missing something...

HuidaeCho:
> Try g.gui instead of g.list. I was running whole GRASS with GUI, so I didn't think about mentioning that explicitly. The texts in g.list are - at least in theory - mostly coming from from the grassmods domain, not grasswxpy,

Ah! Makes sense. I'll try g.gui.

HuidaeCho:
I confirmed the issue and fix. This PR actually addresses the welcome message as well. Awesome!

HuidaeCho (approved):

wenzeslaus:
This fixes some issues with translations, so we should consider backporting this to the 8.0 branch. On the other hand, it significantly changes the approach to translation setup.

neteler:
> This fixes some issues with translations, so we should consider backporting this to the 8.0 branch. On the other hand, it significantly changes the approach to translation setup.

@HuidaeCho Do you have any opinion about the backporting here?

marisn:
I would say we should backport this to the 8.0 branch. It removes Py2 code, is clear and simple, has good „smell“.

wenzeslaus:
I did the backport.

It is fixing some hard-to-define issues for the translations, so without this in 8.0, we would have to think about this PR with every bug report related to missing translations.

Issue #1630, opened 2021-06-09
[Bug] Bulk re-indentation leads to errors


neteler:
**Describe the bug**

It may be a good idea to run a bulk re-indentation of all C/C++/h files before releasing GRASS GIS 8.0.0 (similar to the Python reformatting already done).

The problem is that (in a local dry-run) errors occur:

```
utils/grass_indent_ALL.sh
Indenting *.c and *.h...
indent: ./vector/v.in.ogr/main.c:1457: Error:Stmt nesting error.
indent: ./vector/v.in.ogr/main.c:1779: Error:Stmt nesting error.
indent: ./vector/v.in.ogr/main.c:2042: Error:Stmt nesting error.
indent: ./imagery/i.eb.hsebal01/main.c:547: Warning:old style assignment ambiguity in "=-". Assuming "= -"

indent: ./imagery/i.eb.hsebal01/main.c:557: Warning:old style assignment ambiguity in "=-". Assuming "= -"

indent: ./imagery/i.evapo.pm/functions.c:161: Error:Unexpected end of file
indent: ./include/grass/iostream/empq_adaptive_impl.h:188: Warning:old style assignment ambiguity in "=-". Assuming "= -"

indent: ./include/grass/iostream/minmaxheap.h:198: Warning:old style assignment ambiguity in "=-". Assuming "= -"

indent: ./lib/vector/dglib/misc-template.c:574: Error:Stmt nesting error.
indent: ./lib/vector/dglib/misc-template.c:655: Error:Unmatched 'else'
indent: ./lib/vector/dglib/misc-template.c:684: Error:Stmt nesting error.
indent: ./raster/r.terraflow/3scan.h:123: Warning:old style assignment ambiguity in "=*". Assuming "= *"

indent: ./raster/r.terraflow/types.h:33: Warning:old style assignment ambiguity in "=-". Assuming "= -"

indent: ./raster/r.terraflow/types.h:75: Warning:old style assignment ambiguity in "=-". Assuming "= -"

cmp: EOF on ./display/d.vect/plot.h after byte 782
cmp: EOF on ./vector/v.mkgrid/local_proto.h after byte 279
cmp: EOF on ./ps/ps.map/decorate.h after byte 462
cmp: EOF on ./lib/display/clip.h after byte 451
cmp: EOF on ./lib/display/window.c after byte 594
cmp: EOF on ./lib/htmldriver/box.c after byte 211
[...]

PR #1850, opened 2021-09-09
Imagery: move signatures to subdirs

marisn:
Current structure of signature storage in `<mapset>/signatures/<type>/<file>` is OK for existing signature files but it lacks any flexibility. Other types of signatures might need to have more than one file per signature (e.g. HD5 for data and text files for metadata) and thus current structure is suboptimal.

This PR just moves signature files into a subdirectory in a similar fashion how the "cell_misc" directory is organized. Related internal functions are renamed to better reflect this semantic shift (element -> dir).

Full path to signature file now is:
`location/mapset/signatures/<type>/<name>/<element>`

Sorry for blocker, but as this PR changes the layout of mapset, it must be merged before 8.0 release. As there is no fundamental change in signature handling, reviewing should be fast. Sorry for unrelated changes introduced by grass_indent.sh script.

### DIFF
--- a/lib/imagery/sigfile.c
+++ b/lib/imagery/sigfile.c
@@ -17,7 +17,7 @@
 /*!
    \brief Create signature file
 
-   Returns a pointer to FILE for writting signature file.
+   Returns a pointer to FILE for writing signature file.
    Use fclose on the pointer to close after use.
 
    \param name signature filename
--- a/lib/imagery/sigfile.c
+++ b/lib/imagery/sigfile.c
@@ -17,6 +17,9 @@
 /*!
    \brief Create signature file
 
+   Returns a pointer to FILE for writting signature file.
+   Use fclose on the pointer to close after use.
+
    \param name signature filename
 
    \return pointer to FILE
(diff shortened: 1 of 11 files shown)


wenzeslaus (approved):
I agree with the intention and as far as I can tell it uses the right lower level functions to handle the files.

lib/imagery/sigfile.c
```diff
@@ -17,29 +17,34 @@
 /*!
    \brief Create signature file
 
+   Returns a pointer to FILE for writting signature file.
```
nilason, line 20: typo `writing`

nilason (commented):
`test_imagery_sigsetfile.py` fails on Mac (and probably most non-gnu platforms) with:
https://github.com/OSGeo/grass/blob/9b95bc121764bb838ded59d5d707f4964b46ca1d/lib/imagery/sigset.c#L186
as the 'm' assignment-allocation specifier is an ISO C extension and not available on all platforms.

```
...
WARNING: Error reading band references from sigset file
...

FAIL: test_roundtrip_sigset_v1_one_band (__main__.SigSetFileTestCase)
Test writing and reading back sigset file (v1)
  File "/Volumes/dev/grass/lib/imagery/testsuite/test_imagery_sigsetfile.py", line 96, in test_roundtrip_sigset_v1_one_band
    self.assertEqual(ret, 1)
AssertionError: -1 != 1
```

nilason:
The above mentioned issue with `test_imagery_sigsetfile.py` wasn't introduced with this PR, but it would be good to fix this here.

lib/imagery/sigfile.c
marisn: Thanks!

marisn:
> as the 'm' assignment-allocation specifier is an ISO C extension and not available on all platforms.

I was under impression that POSIX is fine. No problem, worked around. Please test.
Sorry for running grass_indent.sh on the code. We should run it on the main before release to tidy up our code base and not continue the pain also after 8.0 release :(

nilason:
> > as the 'm' assignment-allocation specifier is an ISO C extension and not available on all platforms.
> 
> I was under impression that POSIX is fine. No problem, worked around. Please test.

Would have been considerable less code if that were the case. Now compiled and run the tests successfully!

> Sorry for running grass_indent.sh on the code. We should run it on the main before release to tidy up our code base and not continue the pain also after 8.0 release :(

I agree, but that is not as smooth an operation as it first appears. See #1630.

PR #1856, opened 2021-09-10
r.external: add flag to read min/max from metadata

metzm:
This PR adds a new flag `-m` to read min/max from metadata. This is DANGEROUS because statistics in metadata are sometimes approximations: min and max can be wrong. Amongst other issues, display will not work, with cell values outside the wrong range being invisible. However, reading statistics from metadata happens instantly, while scanning the actual data to determine min/max can take some time for larger raster maps. Thus, if a user is confident that the statistics in the metadata of a GDAL raster band are accurate, this new flag can be used to speed up data registration.

### DIFF
--- a/raster/r.external/link.c
+++ b/raster/r.external/link.c
@@ -47,8 +47,21 @@ void query_band(GDALRasterBandH hBand, const char *output,
 	break;
     }
 
-    if (info->have_minmax)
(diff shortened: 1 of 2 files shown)


mmacata (approved):
As a warning is added to the documentation, I agree that it is dangerous but a cool feature to retrieve metadata quickly.

PR #1857, opened 2021-09-10
Added warning about region not being respected

CarlosGrohmann:
After a nice discussion here on GH, I prepared this paragraph about the module not respecting the computational region, and warning the user about it.

### DIFF
--- a/imagery/i.landsat.toar/i.landsat.toar.html
+++ b/imagery/i.landsat.toar/i.landsat.toar.html
@@ -24,6 +24,10 @@ not found in the metadata file then the command line values are used.
 raster is set to null in the output raster and it is not included in
 the equations.
 
+<p>
+<b>Attention</b>: This module does <b>not</b> respect the current region settings, in order to have the largest possible sample of pixels from where to get the darkest one of the scene and perform the DOS correction. To limit the results to a custom region, the user is advised to clip the results (with <a href="r.clip.html">r.clip</a>, for instance) or to define the region first, import the images with region cropping, and then running the module.
+
+
 <h2>Uncorrected at-sensor values (method=uncorrected, default)</h2>
 
 The standard geometric and radiometric corrections result in a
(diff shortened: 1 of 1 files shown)


imagery/i.landsat.toar/i.landsat.toar.html
```diff
@@ -24,6 +24,10 @@ <h2>DESCRIPTION</h2>
 raster is set to null in the output raster and it is not included in
 the equations.
 
+<p>
+<b>Attention</b>: This module does <b>not</b> respect the current region settings, in order to have the largest possible sample of pixels from where to get the darkest one of the scene and perform the DOS correction. To limit the results to a custom region, the user is advised to clip the results (with <a href="r.clip.html">r.clip</a>, for instance) or to define the region first, import the images with region cropping, and then running the module.
```
metzm, line 28: The line is a bit long, see [https://trac.osgeo.org/grass/wiki/Submitting/Docs](https://trac.osgeo.org/grass/wiki/Submitting/Docs): "please break a line at approximately 70-80 chars". Breaking up into several lines would look like:

```suggestion
<b>Attention</b>: This module does <b>not</b> respect the current 
region settings, in order to have the largest possible sample of pixels 
from where to get the darkest one of the scene and perform the DOS 
correction. To limit the results to a custom region, the user is 
advised to clip the results (with <a href="r.clip.html">r.clip</a>, for 
instance) or to define the region first, import the images with region 
cropping, and then running the module.
```

veroandreo: I'd remove the clipping suggestion. I don't think clipping is needed for the output... this is the beauty of GRASS, just change the region. All other modules will respect it, so your further outputs will be clipped...

CarlosGrohmann: I don't mind if you remove the clipping part.

Issue #1872, opened 2021-09-16
[Bug] GRASS misses GDAL dynamic library references on Windows


jef-n:
There are explicit references to GDAL DLL  that miss versions above GDAL 3.0 on Windows: https://github.com/OSGeo/grass/blob/75dd60a694198f7bca17b7e3d6387b6eefa9bb7a/lib/raster/gdal.c#L119

See also qgis/QGIS#45079

nilason:
Related issue #326 (in particular see rouault's https://github.com/OSGeo/grass/issues/326#issuecomment-619612799).
The big question, why is that part (`lib/raster/gdal.c`) needed at all?

ninsbl:
This obviously affects several modules (https://lists.osgeo.org/pipermail/grass-dev/2021-October/095447.html). Therefore, I take the liberty to set the priority to blocker...

hellik:
> 
> 
> This obviously affects several modules (https://lists.osgeo.org/pipermail/grass-dev/2021-October/095447.html). Therefore, I take the liberty to set the priority to blocker...

a quick rescue 7.8.7 release needed?

ninsbl:
Given that this is an MS Windows only issue and GRASS code is currently patched in OSGeo4W anyway, I hope we can avoid a 7.8.7 release...
Therefore I made a PR to OSGeo4W.

But it would be great if a core dev with the right skills could have a shot at a proper fix in the main code base...

wenzeslaus:
The proper fix according to the last comment in #326 means just removing all that code, no? Or is there something else needed to use the standard linking mechanism; it is probably used for all other libraries already.

nilason:
FWIW, I removed the gdal patch for macports for the 7.8.6 release, therefore I assume gdal is working fully dynamically. Yet so far there has been no complaints reported. In addition, the official mac binary is also built in similar way, without any issues I’m aware of.  Seems to me #326 is not addressing any actual problem (at least not anymore). No idea how/why/if at all this is a problem on windows or elsewhere.

nilason:
Quick browse through history seems to link this to `r.external` And `GDAL_LINK `.  Hope to get some time to dig into this. Anyone recollects anything on the history on this?

neteler:
The easiest way to look up the history of `lib/raster/gdal.c` might be here:

https://trac.osgeo.org/grass/log/grass/trunk/lib/raster/gdal.c

which points to (note the different path, initially it was part of `lib/gis/`)

[...]

PR #1874, opened 2021-09-16
lib/raster/gdal.c: add recent GDAL dlls

ninsbl:
This is a quick fix / work around for #1872
It kicks the can down the road (until after release of 7.8.6), if no one else comes up with a proper solution before that.
It is a rather simple change and should probably go into 7.8.6...

### DIFF
--- a/lib/raster/gdal.c
+++ b/lib/raster/gdal.c
@@ -116,6 +116,9 @@ static void load_library(void)
 	"libgdal1.7.0.so",
 # endif
 # ifdef _WIN32
+	"gdal303.dll",
+	"gdal302.dll",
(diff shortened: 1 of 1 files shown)


ninsbl:
This did not make it into RC3. It is a rather trivial change.

OK to merge anyway? Or should we update the milestone?

hellik:
OSGeo4W vers. 2 ships

```
C:\OSGeo4W>gdalinfo --version
GDAL 3.3.1, released 2021/06/28
```

I guess the PR is needed then.

ninsbl:
Backported in 5035673dc

PR #1882, opened 2021-09-20
wxGUI/vdigit: use 'FloatSpin' widget, which allows you to set the decimal snapping threshold value

tmszi:
**Describe the bug**
If LOCATION has WGS84 (EPSG:4326) coordinate system with degree map unit, it isn't possible set Vector Digitizer decimal snapping threshold value < 1 degree (map unit), 1 degree is approximatelly 111 km. Affected is Query tool Select lines widget too.

**To Reproduce**

1. Lauch wxGUI, create new  or switch to LOCATION with WGS84  (EPSG:4326) coordinate system
2. Switch to Vector Digitizer (edit some existed or create new vector map) and open Preference dialog
3. Try to set decimal snapping treshold value < 1 degree (map unit)  e.g.  0.0001 degree which is aproximatelly 11.11 m
4. Actually isn't possible set decimal snapping threshold value

**Expected behavior**
Allow setting decimal snapping threshold value.

### DIFF
--- a/gui/wxpython/vdigit/preferences.py
+++ b/gui/wxpython/vdigit/preferences.py
@@ -23,7 +23,7 @@ import wx.lib.colourselect as csel
 from gui_core.gselect import ColumnSelect
 from core.units import Units
 from core.settings import UserSettings
-from gui_core.wrap import Button, CheckBox, SpinCtrl, StaticBox, StaticText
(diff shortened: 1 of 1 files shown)


petrasovaa (approved):
Looks good!

PR #1883, opened 2021-09-21
configure: Fix ODBC configuration on Windows (branch 7.8)

nilason:
The update to autoconf 2.69 cause configure failure for including ODBC on Windows.

[Windows configure log](https://wingrass.fsv.cvut.cz/grass78/x86_64/logs/log-r341370e46-648/package.log):
```
...
checking whether to use ODBC... "yes"
checking for location of ODBC includes... 
checking sql.h usability... no
checking sql.h presence... yes
configure: WARNING: sql.h: present but cannot be compiled
configure: WARNING: sql.h:     check for missing prerequisite headers?
configure: WARNING: sql.h: see the Autoconf documentation
configure: WARNING: sql.h:     section "Present But Cannot Be Compiled"
configure: WARNING: sql.h: proceeding with the compiler's result
checking for sql.h... no
configure: error: *** Unable to locate ODBC includes.
```

This is caused by changes on how `AC_CHECK_HEADER` works, see [autoconf manual](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/autoconf.html#Present-But-Cannot-Be-Compiled).

Specifically in this case, on Windows the header `<windows.h>` must be included for the check to succeed.

This fix has been tested on non-Windows machine, with and without ODBC; as well as on Windows CI on my fork.

### DIFF
--- a/configure
+++ b/configure
@@ -12265,27 +12265,32 @@ if test -n "$with_odbc_includes" ; then
 fi
 
 
+ac_fn_c_check_header_mongrel "$LINENO" "windows.h" "ac_cv_header_windows_h" "$ac_includes_default"
+if test "x$ac_cv_header_windows_h" = xyes; then :
 
-ac_save_cppflags="$CPPFLAGS"
-CPPFLAGS="$ODBCINC $CPPFLAGS"
-for ac_header in sql.h
-do :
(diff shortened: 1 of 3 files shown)


nilason:
To be able to see if there are more issues with Windows down the line, I will now merge this fix.

PR #1884, opened 2021-09-22
g.region: Apply -a flag for n=, s=, w=, e=

HuidaeCho:
This PR adjusts the boundaries to the resolution when only `n=`, `s=`, `w=`, and/or `e=` are used.

### DIFF
--- a/general/g.region/main.c
+++ b/general/g.region/main.c
@@ -7,7 +7,7 @@
  * PURPOSE: 	Program to manage and print the boundary definitions for the
  *              geographic region.
  *
- * COPYRIGHT:  	(C) 2000 by the GRASS Development Team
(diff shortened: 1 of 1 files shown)


HuidaeCho:
> Could you share little bit more about the change, what is the use case and how the behavior changes with this PR?

Before this PR, `-a` had no effects when only n/s/w/e is used. n/s/w/e was strictly forced even with `-a`. With this PR, with`-a`, n/s/w/e is snapped to (aligned with?) multiples of resolution just like with `raster=` or `vector=`.

HuidaeCho:
Use case: I want to expand the region by *about* 9000 units while keeping the current resolution.
```bash
g.region -a n=n+9000 s=s-9000 e=e+9000 w=w-9000
```

Issue #1720, opened 2021-07-10
[Bug] python/grass/ctypes dir looks like ctypes package


wenzeslaus:
**Describe the bug**

The directory with ctypesgen is called ctypes and is under `python/grass/`. However, the code is not either the `ctypes` package nor it is meant to be imported as `grass.ctypes`. It contains `__init__.py` file, so it looks like a Python (sub-)package (package is Python is often defined as a directory with a file called `__init__.py`, but this file breaks it). It seems that some renaming is needed.

**Expected behavior**

- The directory name makes it clear that it is not the ctypes library. (Confusion issue, needed.)
- Possibly also the name could make more it clear that is set of file for generating code. Something like `build-helpers/ctypes-interface`. (Documentation issue, optional.)
- The file `__init__.py` has a different name and is renamed only later to `__init__.py`. (Avoids confusing both people and computers.)
- Maybe moving the code is outside of `python/grass` would help, too, although having there creates kind of a placeholder for the compiled lib. Anyway, `python` or `utils` dirs would be candidates.

**Screenshots**

```
python/grass/
├── bandref
│   ├── __init__.py
│   ├── Makefile
│   └── reader.py
├── ctypes
│   ├── ctypesgencore
│   ├── __init__.py
│   ├── Makefile
```

nilason:
Suggested fix with #1898 .

PR #1898, opened 2021-09-28
ctypes: move directory out of python/grass

nilason:
Move and rename `ctypes` directory, which has been placed in `python/grass`.

Addresses #1720.

Suggested new place and name:

```
python/
├── grass
│   ├── app
│   ├── bandref
│   └── ...
├── libgrass-interface-generator
│   ├── ctypesgen
│   ├── __init__.py
│   ├── ...
```

Putting this PR up as a draft, as this contains the changes of #1651, which will have to be dropped if/when #1651 is merged. Its destination and name may also be discussed.


marisn:
+1 for the idea.

wenzeslaus (approved):
I think renaming also `python/libgrass-interface-generator/__init__.py` would be even better, but renaming the directory is great and all what this PR needs to do. I like the new name.

neteler:
Just a naive question: dash versus underscore?

`libgrass-interface-generator` 
or `libgrass_interface_generator`?

My gut feeling prefers the underscore but maybe I overlook some convention here.

nilason:
> Just a naive question: dash versus underscore?
> 
> `libgrass-interface-generator` or `libgrass_interface_generator`?
> 
> My gut feeling prefers the underscore but maybe I overlook some convention here.

I was more focused on the naming as such, no particular reason for dashes, but I agree underscore is better.

nilason:
> I think renaming also `python/libgrass-interface-generator/__init__.py` would be even better, but renaming the directory is great and all what this PR needs to do. I like the new name.

Why not **delete** that `__init__.py` file; It's not being used anyway, is it? It never gets installed and GRASS compiles and runs just fine without it.

nilason:
> > I think renaming also `python/libgrass-interface-generator/__init__.py` would be even better, but renaming the directory is great and all what this PR needs to do. I like the new name.
> 
> Why not **delete** that `__init__.py` file; It's not being used anyway, is it? It never gets installed and GRASS compiles and runs just fine without it.

My bad! Installation of `__init__.py` got "lost in the translation" with https://github.com/OSGeo/grass/commit/ca2d28a592064b501c0eb6c4ec648984f386dd42, I have now re-introduced it (renamed to `libgrass__init__.py`) as well as re-added python compilation of *all* py files.

PR #1899, opened 2021-09-28
g.rename: avoid buffer overflow

metzm:
A string buffer was too small, set to size 50 instead of `GNAME_MAX`, causing buffer overflow and segmentation fault. This PR fixes it, using the same method like `g.copy`.

### DIFF
--- a/lib/manage/do_rename.c
+++ b/lib/manage/do_rename.c
@@ -11,6 +11,7 @@
   \author Original author CERL
 */
 
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -34,7 +35,6 @@
 int M_do_rename(int n, const char *old, const char *new)
 {
     int i, ret;
-    int len;
     const char *mapset;
     int result = 0;
     int renamed = 0;
@@ -44,8 +44,6 @@ int M_do_rename(int n, const char *old, const char *new)
     
     if (G_strcasecmp(old, new) == 0)
 	return 1;
-
-    len = M__get_description_len(n);
 
     M__hold_signals(1);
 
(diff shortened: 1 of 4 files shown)


lib/manage/do_rename.c
```diff
@@ -94,7 +94,7 @@ int M_do_rename(int n, const char *old, const char *new)
 	}
 
 	if (G_strcasecmp(list[n].element[0], "cell") == 0) {
-	    char colr2[50];
+	    char colr2[GNAME_MAX];
 
 	    sprintf(colr2, "colr2/%s", G_mapset());
```
nilason, line 99: Quick glance at this: `G_mapset()` may return a string of `GMAPSET_MAX` (= GNAME_MAX). This may potentially case segfault, no?
Perhaps a solution like:
https://github.com/OSGeo/grass/blob/38c084ad0b45bc2a46e333eaf4ec5ecabde5d315/lib/vector/Vlib/map.c#L494
could be considered?

metzm: Right, but then other instances in `lib/manage` need to be updated as well to use `G_snprintf()`. I will change it accordingly. Maybe also use
```
	    char colr2[GNAME_MAX + 6];
```
?

nilason: `colr2` is supposed to be of max `GNAME_MAX` size, isn't it. If so, then `GNAME_MAX + 6` wouldn't be the best choice.

Re.: G_snprintf(), I believe the C standard `snprintf()` is nowadays well supported and preferable. Remember our discussion on https://github.com/OSGeo/grass/pull/1406#discussion_r594719601? I have even a local branch for replace usage and remove/deprecate G_snprintf().

nilason:
I see that similar problem has been addressed in:
https://github.com/OSGeo/grass/blob/38c084ad0b45bc2a46e333eaf4ec5ecabde5d315/lib/manage/do_remove.c#L102

however, the return value was ignored, with potential string truncation.

lib/manage/do_rename.c
nilason: Or isn’t `colr2 ` size of `GNAME_MAX + GNAME_MAX `?

nilason: It sure was a quick glance. Sorry for the noice, but now I looked a bit closer at what's what:

`char colr2[GNAME_MAX + 6];` should be enough. No real need for snprintf() in this case.

metzm:
The string buffer sizes should now be correct. The return value of `snprintf()` is also checked.

nilason (approved):

PR #1907, opened 2021-09-30
v.in.pdal: remove obsolete PCL functionality

### LINKED ISSUE #851: [Bug] v.in.pdal and r.path not in menus
**Describe the bug**
You can run v.in.pdal and r.path from the terminal or GUI console but they are not anywhere in the menu system. Need to be added. Maybe there are other recent modules that need adding too? 

- Operating System: all]
- GRASS GIS version 7.8 and up at least

### LINKED ISSUE #1049: v.in.pdal: remove ground filtering capability
PCL support was removed from PDAL 2.0.0. Thus also "filters.ground" is gone. Now there is a set of new filters for ground classification ("filters.csf", "filters.pmf", "filters.skewnessbalancing" and "filters.smrf"). Unfortunately all of them come with different set of parameters.

As trying to include all filter parameters in GRASS UI will be always just a game of chasing a runaway train, I would like to see ground filtering capabilities to be completely removed.

An alternative could be ability to provide pipeline support, still I am not certain if v/r.in.pdal should try to clone pdal utility.

marisn:
Fixes  #1049 and partially fixes  #851

### DIFF
--- a/vector/v.in.pdal/v.in.pdal.html
+++ b/vector/v.in.pdal/v.in.pdal.html
@@ -1,15 +1,8 @@
 <h2>DESCRIPTION</h2>
 
-<em>v.in.pdal</em> supports the following PDAL filters:
+<p>Imports point cloud (e.g. from a LAS file) as vector points with PDAL library.
 
-<ul>
-<li>groud detection with Point Cloud Library (PCL) Progressive Morphological Filter (Zhang 2003)
-    (<a href="http://www.pdal.io/stages/filters.ground.html">filters.ground</a>)
-<li><a href="http://www.pdal.io/stages/filters.height.html">filters.height</a>
-    (requires PCL)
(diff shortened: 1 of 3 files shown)


vector/v.in.pdal/v.in.pdal.html
```diff
@@ -1,15 +1,8 @@
 <h2>DESCRIPTION</h2>
 
-<em>v.in.pdal</em> supports the following PDAL filters:
+<p>Imports point cloud (e.g. from a LAS file) as vector points with PDAL library.
```
susi, line 3: Imports **a** point cloud (e.g. from a LAS file) as vector points with **the** PDAL library.

susi (changes requested):

wenzeslaus (approved):

PR #1909, opened 2021-10-01
macOS/install: do not install system level docs

nilason:
It is no longer (never became) custom to install documentation in `/Library/Documentation/Help/` in general for Mac.
This behaviour is being patched out of all packages to my knowledge (official Mac GRASS release, MacPorts, Homebrew, QGIS).
This removes this installation (of a symlink) and make further patching unnecessary.


PeterPetrik:
👏 👏

PR #1915, opened 2021-10-02
CI: Do not run Docker job on forks

wenzeslaus:
When a main branch is updated in fork, the workflow runs, but jobs fail because of missing credentials.
The workflow should not run in the first place. Currently, jobs need to be disabled individually.


wenzeslaus:
@mmacata Can you please see if this makes sense to you?

mmacata (approved):
Makes totally sense, yes! When I created it, I was glad it worked to be able to test it but it doesn't make much sense now, so thanks for this enhancement!

Issue #962, opened 2020-09-11
[Bug] GUI should use its own history and not Bash's


wenzeslaus:
**Describe the bug**
Every shell is using its own history file. GUI should do that too, but instead it is using .bash_history file. This does not even work across different platforms or shells, it syncs with Bash, but does not with Z shell or others which now more relevant when macOS switched to Z shell from Bash.

**Expected behavior**

wxGUI (or the Console specifically) is another "shell", so it should keep history in `.wxgui_history`.

I'm suggesting wxgui_history over gui_history because of the idea of specific shell (with potentially different features and history records) and over wxgui_console_history because the idea in wxGUI is anyway to run all the user-issued commands through the console.

**System description (please complete the following information):**

- Operating System:
  * Windows (no conflict with Bash at this point I guess)
  * Linux with Bash (shared with bash now, will change behavior)
  * macOS with Z shell (no change of behavior, but clarification of files in mapset)

**Additional context**

See discussion in #930 which triggered this: https://github.com/OSGeo/grass/pull/930#issuecomment-678896317.

susi:
I'm going to start to look at this  :)

susi:
Should this be targeted for 8.0.0 instead, since we have a PR that fixes it?

PR #1917, opened 2021-10-02
Use .wxgui_history as history file for the gui (#962)

susi:
Changed .bash_history to .wxgui_history as per issue #962


veroandreo:
Maybe we could ship this with 8.0.0 already, no?

susi:
> Maybe we could ship this with 8.0.0 already, no?

Sure! I don't know why the bug was moved to 8.2.0? :thinking:

wenzeslaus (approved):
`.wxgui_history` is what I had in mind and the two places is what seems to be sufficient to make it work.

PR #1919, opened 2021-10-02
g.gui.gmodeler: add PyWPS export option

pesekon2:


pesekon2:
Figures containing screenshots of the Python editor should be updated - is there anyone who has the same visual as those that are already there? The GUI looks a bit different in my case and it would be nice to have the look more-or-less standardized (and I would like to avoid taking all the pictures again). @landam?

gui/wxpython/gmodeler/g.gui.gmodeler.html
```diff
@@ -251,16 +253,16 @@ <h3>Managing model parameters</h3>
...
+<i>Figure: Average elevation for ZIP codes using North Carolina sample dataset as
 an automatic calculation performed by Graphical Modeler.</i>
 </center>
 
-<h3>Managing model properties</h3>  
-When one wants to run model again with the same data or the same names, it is 
-necessary to use <tt>--overwrite</tt> option. It will cause maps with identical 
-names to be overwritten. Instead of setting it for every 
+<h3>Managing model properties</h3>
+When one wants to run model again with the same data or the same names, it is
```
veroandreo, line 261:
```suggestion
When the user wants to run model again with the same data or the same names, it is
```

gui/wxpython/gmodeler/g.gui.gmodeler.html
```diff
@@ -272,16 +274,16 @@ <h3>Managing model properties</h3>
...
+Another useful trick is the possibility to set variables. Their content can be used
+as a substitute for other items. Value of variables can be values such as
+raster or vector data, integer, float, string value or they may constitute some
+region, mapset, file or direction data type.
 
 Then it is not
-necessary to set any parameters for input data. The dialog with variable settings 
-is automatically displayed after model is run. So, instead of model parameters 
+necessary to set any parameters for input data. The dialog with variable settings
+is automatically displayed after model is run. So, instead of model parameters
```
veroandreo, line 284:
```suggestion
is automatically displayed after the model is run. So, instead of model parameters
```

gui/wxpython/gmodeler/g.gui.gmodeler.html
```diff
@@ -292,9 +294,9 @@ <h3>Defining variables</h3>
 </center>
 
 <p>
-The key point is the usage of <tt>%</tt> before the substituting variable and 
-settings in <tt>Variables</tt> dialog. For example, in case of a model variable 
-<tt>raster</tt> that points to an input file path and which value is required to be 
+The key point is the usage of <tt>%</tt> before the substituting variable and
+settings in <tt>Variables</tt> dialog. For example, in case of a model variable
```
veroandreo, line 298:
```suggestion
settings in the <tt>Variables</tt> dialog. For example, in the case of a model variable
```

gui/wxpython/gmodeler/g.gui.gmodeler.html
```diff
@@ -342,9 +344,9 @@ <h3>Handling intermediate data</h3>
 <br>
 <i>Figure: Usage and definition of intermediate data in model.</i>
 </center>
- 
+
 <h3>Using the Python editor</h3>
-By using the Python editor in the Graphical Modeler one can add Python code and then 
+By using the Python editor in the Graphical Modeler one can add Python code and then
```
veroandreo, line 349:
```suggestion
By using the Python editor in the Graphical Modeler the user can add Python code and then
```

gui/wxpython/gmodeler/g.gui.gmodeler.html
```diff
@@ -357,17 +359,23 @@ <h3>Using the Python editor</h3>
 <i>Figure: Python editor in the wxGUI Graphical Modeler.</i>
 </center>
 
+In the <i>Script type</i> combobox, the user can also choose a PyWPS export
+instead of a basic Python script. A PyWPS process based on the model will be
+generated then; for the PyWPS script, the <i>Run</i> button is disabled as
+the user is expected to include this script in his web processing service and
```
veroandreo, line 365:
```suggestion
users are expected to include this script in their web processing service and
```

veroandreo (changes requested):
minor manual fixes - the rest, I didn't check

landam:
Looks nice

modeler-wps

```py
    process = Model()

    processes = [Model()]
    application = Service(processes)
```

@pesekon2 Question: What is purpose of line

```
 process = Model()
```

?

landam:
Suggestion: line

```py
from pywps.app.Service import Service
```

could be moved into `if __name__` block.

landam:
Also consider improving comment

```
# here you could also specify the GRASS location, for example:
```

1. You **must** specify GRASS location
2. Either by providing path to the existing GRASS location 
3. Or by EPSG code in order to create a new one

gui/wxpython/gmodeler/g.gui.gmodeler.html
```diff
@@ -82,10 +83,11 @@ <h3>Main dialog</h3>
...
 define default values and descriptions. The Python editor dialog window
-allows seeing workflows written in Python code. The rightmost tab of
-the bottom menu is automatically triggered when the model is activated and
-shows all the steps of running GRASS modeler modules. In case of errors
-in the calculation process, it is written at that place.
+allows seeing workflows written in Python code, either as a basic Python
+script, or as a PyWPS script. The rightmost tab of the bottom menu is
+automatically triggered when the model is activated and shows all the steps
+of running GRASS modeler modules. In case of errors in the calculation
+process, it is written at that place.
```
veroandreo, line 90: they are wrtten? As errors is plural.. or what does "written in that place" refer to?

gui/wxpython/gmodeler/g.gui.gmodeler.html
```diff
@@ -143,7 +145,7 @@ <h3>Components of models</h3>
 <a href="g_gui_gmodeler_model_landslides.png">
 <img src="g_gui_gmodeler_model_landslides.png" width="600" height="290"></a>
 <br>
-<i>Figure: A model to perform creation of parametric maps used by geologists 
+<i>Figure: A model to perform creation of parametric maps used by geologists
```
veroandreo, line 148:
```suggestion
<i>Figure: A model to create parametric maps used by geologists
```

gui/wxpython/gmodeler/g.gui.gmodeler.html
```diff
@@ -251,16 +253,16 @@ <h3>Managing model parameters</h3>
...
+<i>Figure: Average elevation for ZIP codes using North Carolina sample dataset as
 an automatic calculation performed by Graphical Modeler.</i>
 </center>
 
-<h3>Managing model properties</h3>  
-When one wants to run model again with the same data or the same names, it is 
-necessary to use <tt>--overwrite</tt> option. It will cause maps with identical 
-names to be overwritten. Instead of setting it for every 
+<h3>Managing model properties</h3>
+When the user wants to run model again with the same data or the same names, it is
```
veroandreo, line 261:
```suggestion
When the user wants to run the model again with the same data or the same names, it is
```

veroandreo (commented):
tiny grammar fixes :)

pesekon2:
> 
> ```python
>     process = Model()
> 
>     processes = [Model()]
>     application = Service(processes)
> ```
> 
> @pesekon2 Question: What is purpose of line
> 
> ```
>  process = Model()
> ```
> 
> ?

@landam: Probably just some residuum from one of the previous versions. Fixed in 024d3cde1f1c8d1aab48400fd6501e5da9c96661, thanks for pointing it out.

pesekon2:
> Suggestion: line
> 
> ```python
> from pywps.app.Service import Service
> ```
> 
> could be moved into `if __name__` block.

@landam: Moved in bdc5f35a9438e1c1607f53d8d325d0382c418211

pesekon2:
> Also consider improving comment
> 
> ```
> # here you could also specify the GRASS location, for example:
> ```
> 
>     1. You **must** specify GRASS location
> 
>     2. Either by providing path to the existing GRASS location
> 
>     3. Or by EPSG code in order to create a new one

@landam: Here I am not sure. Although in most cases you need to specify the location, in the special case you're running the process inside GRASS (being in a mapset and then running the PyWPS process), you don't have to specify it. It is probably not such a very common case, but could happen - and then putting there "must" would be a wrong precept.

gui/wxpython/gmodeler/g.gui.gmodeler.html
pesekon2: Hard to say as I have just copied that part from the previous version of the docs, but I guess that you're right and it refers to the errors. Reformulated in cb431fcd74cca7750fa3721e283d7b3e2846c061 (and one more little update in 1634cec288369f94099567eb81267532c6c623c6 to make it clearer) - does it make sense now?

landam:
I tested PyWPS export functionality on several models. @pesekon2 There is a problem with output options which are not defined. See attached model [pywps_grass_test.zip](https://github.com/OSGeo/grass/files/8695956/pywps_grass_test.zip) (Gismentors location).

pywps_grass

Modeler exports all output options even those which are not set up. Note: output option of v.buffer is skipped (marked in the modeler as intermediate) - fine. But this approach cannot be applied on outputs which are not defined. 

```py
       outputs.append(ComplexOutput(
            identifier="gregion2_save",
            title="Save current region settings in named region file",
            supported_formats=sup_formats))



        outputs.append(ComplexOutput(
            identifier="rslopeaspect4_slope",
            title="Name for output slope raster map",
            supported_formats=sup_formats,
            default="slope"))

        outputs.append(ComplexOutput(
            identifier="rslopeaspect4_aspect",
            title="Name for output aspect raster map",
            supported_formats=sup_formats))

        outputs.append(ComplexOutput(
            identifier="rslopeaspect4_pcurvature",
            title="Name for output profile curvature raster map",
            supported_formats=sup_formats))

        outputs.append(ComplexOutput(
            identifier="rslopeaspect4_tcurvature",
            title="Name for output tangential curvature raster map",
            supported_formats=sup_formats))

        outputs.append(ComplexOutput(
            identifier="rslopeaspect4_dx",
            title="Name for output first order partial derivative dx (E-W slope) raster map",
            supported_formats=sup_formats))

        outputs.append(ComplexOutput(
            identifier="rslopeaspect4_dy",
            title="Name for output first order partial derivative dy (N-S slope) raster map",
            supported_formats=sup_formats))

        outputs.append(ComplexOutput(
            identifier="rslopeaspect4_dxx",
            title="Name for output second order partial derivative dxx raster map",
            supported_formats=sup_formats))

        outputs.append(ComplexOutput(
            identifier="rslopeaspect4_dyy",
            title="Name for output second order partial derivative dyy raster map",
            supported_formats=sup_formats))

        outputs.append(ComplexOutput(
            identifier="rslopeaspect4_dxy",
            title="Name for output second order partial derivative dxy raster map",
            supported_formats=sup_formats))
```

I would expect here only r.slope.aspect's slope option.

landam (changes requested):
Output options, which are not set up, should be skipped.

pesekon2:
@landam The code has been changed (see 7028d12) in the following way: Output options are skipped if they are intermediate or (blank and unparameterized).

landam:
Great, now only relevant output parameter is included:

```py
        inputs.append(LiteralInput(
            identifier="vbuffer1_distance",
            title="Buffer distance along major axis in map units",
            data_type="float",
            default=1000))




        outputs.append(ComplexOutput(
            identifier="rslopeaspect4_slope",
            title="Name for output slope raster map",
            supported_formats=sup_formats,
            default="slope"))
```

I just wonder why there are so many empty lines. @pesekon2 I would suggest to keep it consistent (1/2 lines). Similarly here:

```py
        return response



if __name__ == "__main__":
```

landam:
@pesekon2 Unfortunately the process fails with:

```py
  File "/home/martin/Downloads/pywps-flask-master/demo.py", line 53, in <module>
    Model()
  File "/home/martin/Downloads/pywps-flask-master/processes/pywps_grass_test.py", line 29, in __init__
    outputs.append(ComplexOutput(
TypeError: __init__() got an unexpected keyword argument 'default'
```

Tested with pywps version 4.2.11 and 4.5.2.

pesekon2:
@landam According to the PyWPS documenatation, PyWPS 4.x does not support parameter `default` for `ComplexOutput` objects. Updated in 2fc8692 to skip this parameter for `ComplexOutput` objects - fortunately the code is written in a way to handle it later in `_handler`, so the default value is still used, just later.

landam:
Currently it's failing with

```py
 File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/gui/wxpython/gmodeler/model.py", line 2837, in _write_input_outputs
    self._write_input_output_object(
TypeError: WritePyWPSFile._write_input_output_object() missing 1 required positional argument: 'value'
```

@pesekon2 Method `WritePyWPSFile._write_input_output_object()` has 7 arguments (https://github.com/pesekon2/grass/blob/gmodeler_export_pywps/gui/wxpython/gmodeler/model.py#L2843) defined, but only called with 6 arguments (https://github.com/pesekon2/grass/blob/gmodeler_export_pywps/gui/wxpython/gmodeler/model.py#L2837).

pesekon2:
@landam: Thanks for noticing that. The origin of this bug is actually sort of funny. The latest state of this PR wasn't cool enough according to the Black code style suppression, so I pushed a minimal reformatting commit 8868761. Apparently, it made the code stylish enough, but a small typo broke it to be unusable. 

(what to learn: test the code even after those stupid reformatting commits)

Fixed in 0282c38.

landam:
New tasks based on our testing

- [ ] add `overwrite` flag to `r.out.gdal` command (only if overwrite is enabled in preferences)

pesekon2:
> landam:
> add `overwrite` flag to `r.out.gdal` command (only if overwrite is enabled in preferences)

Solved in [a249d85](https://github.com/OSGeo/grass/pull/1919/commits/a249d8575c7b75a2c1e78d98763dcc2b99ec74ae)

pesekon2:
>  landam:
>  limit supported formats (raster -> `image/tif`, vector -> `application/gml+xml`)

Solved in b176b82

gui/wxpython/gmodeler/model.py
```diff
@@ -2883,6 +2883,7 @@ def _writePythonAction(self, item, variables={}, intermediates=None):
         for param in item.GetParams()["params"]:
             value = param["value"]
             age = param["age"]
+            overwrite = self.model.GetProperties().get("overwrite", "False")
```
landam, line 2883: @pesekon2 To be consistent with other commands, `overwrite` should be defined if overwrite option is enabled in the model.

pesekon2: Fixed in [3aabc6f](https://github.com/OSGeo/grass/pull/1919/commits/3aabc6f9c5abb97822c34bbc79122be2ef9eb898)

landam:
I tested sample model with pywps 4.5.2 and it works.

landam (approved):

landam:
@veroandreo Can we merge this PR?

veroandreo (approved):

Issue #1911, opened 2021-10-01
[Bug] GRASS GIS 8.0 doesn't launch


tmszi:
**Describe the bug**
GRASS GIS 8.0 doesn't launch.

**To Reproduce**

1. Download [Standalone Win GRASS GIS 8.0 installer](https://wingrass.fsv.cvut.cz/grass80/x86_64/WinGRASS-8.0.dev-257-Setup-x86_64.exe)
2. Install GRASS GIS
3. Try launch GRASS GIS 8.0
4.  Quickly appear and disappear GRASS GIS session Command Prompt without error message

**Expected behavior**
Launch GRASS GIS.

**Screenshots**

![mswin_grass80_py_startup_file](https://user-images.githubusercontent.com/50632337/135665004-91551f9f-f44f-442a-8a19-e6833074e8c3.png)

**System description (please complete the following information):**
- MS Windows
- Standalone [WinGRASS-8.0.dev-257-Setup-x86_64.exe installer](https://wingrass.fsv.cvut.cz/grass80/x86_64/WinGRASS-8.0.dev-257-Setup-x86_64.exe)

**Additional context**
GISBASE directory `C:\Program Files\GRASS GIS 8.0\etc\` contains 2 startup GRASS GIS Python files `grass80.py` (which is empty) and `grass.py`. Startup command inside `grass80.bat` file point to` "%GRASS_PYTHON%" "%GISBASE%\etc\grass80.py" %*`.

neteler:
Does this persist with the latest version?
https://wingrass.fsv.cvut.cz/grass80/x86_64/

dnewcomb:
Yes, still a problem with the binary installer from 10/13

tmszi:
Yes, it still doesn't work (Standalone Win GRASS GIS installer 2021-10-13).

dnewcomb:
Apologies for my post on 12th, it was meant for 7.8.6-1.  However, the nightly standalone installer for 8.0 for 11/14 https://wingrass.fsv.cvut.cz/grass80/x86_64/WinGRASS-8.0.dev-7be7cf9b0-2-Setup.exe has the same issue as 7.8.6-1.
There are no read permissions for the c:\program files\grass 8.0\etc\grass8.0.py for standard users
This may be a different bug than the original.
When trying to launch grass as a standard user ( dlocal in this case)  

[...]

PR #1880, opened 2021-09-19
Apply patches from OSGeo4W upstreams

ninsbl:
In this PR I would like to suggests to apply patches from OSGeo4W (https://github.com/jef-n/OSGeo4W/blob/master/src/grass/osgeo4w/patch) in the code base.

This would reduce maintenance in OSGeo4W and in turn also here if OSGeo4W would become the basis for CI on Windows.

Co-authored-by: jef-n <jef@norbit.de>

PR #1921, opened 2021-10-02
Apply more osgeo4w patches

ninsbl:
This PR builds upon #1880 and adds some more functionality.

It adresses also #1911

As well as: https://trac.osgeo.org/grass/ticket/3759

It builds in OSGeo4W and GRASS starts (tests fail because of shell scripts in the testsuite that are not supported on CMD): https://github.com/ninsbl/OSGeo4W/runs/3778418720?check_suite_focus=true

### DIFF
--- a/mswindows/osgeo4w/mklibs.sh
+++ b/mswindows/osgeo4w/mklibs.sh
@@ -1,30 +1,13 @@
 #!/bin/sh
 
 set -e
-
-if [ -z "$OSGEO4W_POSTFIX" ]; then
-    OSGEO4W_POSTFIX=""
-fi
-
-if [ "$OSGEO4W_POSTFIX" != "64" ]; then
-    MACHINE=x86
-else
-    MACHINE=x64
-fi
-
-PROGRAMFILES="/c/Program Files (x86)"
-VSDIR="$PROGRAMFILES/Microsoft Visual Studio 14.0"
-# add example for MS Visual Studio 2017 Community 64 bit
-# VSDIR="$PROGRAMFILES/Microsoft Visual Studio/2017/Community"
-PATH="$VSDIR/Common7/IDE:$PATH"
-PATH="$VSDIR/VC/bin:$PATH"
-# add example for MS Visual Studio 2017 Community 64 bit
-# PATH="$VSDIR/VC/Tools/MSVC/14.12.25827/bin/Hostx64/x64:$PATH"
-PATH="$VSDIR/Common7/Tools:$PATH"
-PATH="$PATH:/c/OSGeo4W${OSGEO4W_POSTFIX}/bin"
-export PATH
+set -x
 
 [ -d mswindows/osgeo4w/vc ] || mkdir mswindows/osgeo4w/vc
+
+if [ -n "$VCPATH" ]; then
+	PATH=$PATH:$VCPATH
+fi
 
 for dllfile in "$@"; do
 	dlldir=${dllfile%/*}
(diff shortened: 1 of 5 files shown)


mswindows/osgeo4w/mklibs.sh
```diff
@@ -2,6 +2,15 @@
 
 set -e
 
+if [ "$CI" ] ; then
+	DUMPBIN_PATH=$(cygpath -ua "C:/Program Files (x86)\Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.29.30133/bin/HostX64/x64/")
```
landam, line 6: What is diference between `DUMPBIN_PATH` and `VCPATH`? I thought that `VCPATH` is basically the path to `dumpbin` binary.

mswindows/osgeo4w/package.sh
```diff
@@ -234,6 +241,19 @@ if [ -n "$PACKAGE_PATCH" ]; then
     unix2dos etc/preremove/grass${PACKAGE_POSTFIX}.bat
 
     # copy dependencies (TODO: to be reduced)
+    blas_dll=../msys64/mingw64/bin/libblas.dll
```
landam, line 244: Is it needed to treat new dependecies separately? Why not include them to `DLL` variable (https://github.com/OSGeo/grass/blob/main/mswindows/osgeo4w/package.sh#L109)?

mswindows/osgeo4w/package.sh
```diff
@@ -203,6 +208,8 @@ mv $OSGEO4W_ROOT_MSYS/apps/grass/grass$POSTFIX/include/grass/config.h \
    $OSGEO4W_ROOT_MSYS/apps/grass/grass$POSTFIX/include/grass/config.h.mingw
 cp mswindows/osgeo4w/config.h.switch $OSGEO4W_ROOT_MSYS/apps/grass/grass$POSTFIX/include/grass/config.h
 cp mswindows/osgeo4w/config.h.vc $OSGEO4W_ROOT_MSYS/apps/grass/grass$POSTFIX/include/grass
+mv $OSGEO4W_ROOT_MSYS/apps/grass/grass$POSTFIX/etc/grass.py $OSGEO4W_ROOT_MSYS/apps/grass/grass$POSTFIX/etc/grass${POSTFIX}.py
```
landam, line 211: What about fixing https://github.com/OSGeo/grass/blob/main/mswindows/osgeo4w/grass.bat.tmpl#L18 instead of introducing `mv` statement?

landam (changes requested):

landam:
Setting blocker due to #1911

mswindows/osgeo4w/package.sh
landam: I tried to modify `grass.bat` locally but it end up with error
VirtualBox_Windows 10 (ArcGIS Pro)_05_10_2021_14_09_54

So the move statement approach seems to be correct.

mswindows/osgeo4w/package.sh
```diff
@@ -128,7 +128,11 @@ DLLS="
 	/mingw64/bin/libpcre-1.dll
 	/mingw64/bin/libstdc++-6.dll
 	/mingw64/bin/libgcc_s_seh-1.dll
+    /mingw64/bin/libgcc_s_dw2-1.dll
```
hellik, line 131: indendation correct?

ninsbl: good catch. 
Actually, `libgcc_dw2` seems to be 32bit and the 64bit equivalent is `libgcc_s_seh`, so I removed that line. But indentation was inconsistent also other places...

A test build based on OSGeo4W V2 is running here:
https://github.com/ninsbl/OSGeo4W/runs/3806467054?check_suite_focus=true

mswindows/osgeo4w/package.sh
ninsbl: Included dependencies in DLL variable in c29f606

landam:
@ninsbl I extracted critical fix (https://github.com/OSGeo/grass/pull/1933) from this PR in order to apply fix in short time.

ninsbl:
The DUMPBINPATH / VCPATH issue is addressed in 4e163bb
It seems I also missed out a patch for config.h.vc which now is included in 460f414
@landam please let me know if anything else should be adjusted here.

ninsbl:
https://github.com/jef-n/OSGeo4W/pull/3 depends on this PR...

Issue #1868, opened 2021-09-15
[Bug] Band references need a better name and definition


wenzeslaus:
## The issue

The term _band reference_ was introduced with #63. There, perhaps, the word _reference_ was referring to the fact that the text is referring to the band description stored in the file g.bands gives access to. However, now an arbitrary _band reference_ can be stored for any raster map, i.e., it is more general than just bands from a predefined (or possibly in the future, registered) list of bands associated with particular sensors.

The band references can now be used (optionally assuming #1866) to label any rasters (raster maps or bands) in an imagery group so that signatures are associated with given set of bands rather than just a fixed set of raster data.

Going even further, the usage does not have to limited to image processing. For example, in modeling, standardized names such as `topographic__elevation` and `sea_surface_air__temperature` are used to describe inputs and outputs of models (see [Landlab Standard Name Definitions](https://landlab.readthedocs.io/en/latest/user_guide/field_definitions.html) and [CSDMS Standard Names](https://csdms.colorado.edu/wiki/CSN_Examples)).

Overall, band references quickly outgrew their original definition and there is a potential to have good name suggesting the potential use. At the same time, there is a potential for ending up with a cryptic term which is difficult to explain and difficult to link to its application.

The issues with the current name are:

1. The word _band_, although often used quite generally, might be tight too much to remote sensing and may overlap with other usages of the term.
2. The word _reference_ is the cryptic part. Name, label, band and id are words people may have different ideas about, but they generally be somewhat right. Reference, on the other hand, is a very special term, loaded for some, meaningless for others.
3. The term _band reference_ seems to be too long, resulting in modules and functions using different combinations of _band_ and _bandref_ in the interface instead of full _band reference_. Notably, _bandref_ is not a word. The word _band_ by itself may be too ambiguous. Interestingly, I don't think there is _reference_ used as the only word, although that's really the noun in the term.

Since the current system is not part of any release yet, now is the time to get it right.

## Naming options

- band label
- band name
- band tag
- band reference
- band id
- band identifier
- semantic label
- semantic name
- semantic tag
- standard variable name
- standard field name
- ...

## Expected resolution

1. We decide on a name for what these things are and even if it stays as _band references_ we are sure that that's how we want to call them.
2. The usage is consistent in documentation and in the interface (option names, function names).

## Additional context

We discussed this during the last call, but decided to open this as an issue to discuss this in written form and asynchronously.

mlennert:
Thanks, Maris, for launching this.

[...]

PR #1928, opened 2021-10-05
Rename band references to semantic labels

marisn:
See discussion in #1868


wenzeslaus (approved):
Great.

I'm just not sure if you meant to change the temporal modules or not. For example, t.rast.list.html still talks about band references, or is that on purpose?

marisn:
> I'm just not sure if you meant to change the temporal modules or not. For example, t.rast.list.html still talks about band references, or is that on purpose?

Hard to tell if I messed up during rebase (again) or just skipped. Went over and replaced even more locations of band references with semantic label.
Now should only i.bands and g.bands remain. What should be the new name for each of them?

wenzeslaus:
> Now should only i.bands and g.bands remain. What should be the new name for each of them?

During the PSC meeting today we talked about options such as my suggestion:

* i.bands -> r.semantic.label (Assigns semantic label to a raster)
* g.bands -> i.band.library (Register band metadata for a semantic label)

...but there was no final decision made.

temporal/t.register/t.register.html
```diff
@@ -137,16 +137,16 @@ <h3>Support for band references</h3>
 T33UYP_20190331T094039_B09|2019-03-31 09:40:39|S2_9
 </pre></div>
 
-In this case <em>t.register</em> assignes to given raster maps band
-references similarly as
+In this case <em>t.register</em> assigns to given raster maps
+semantic label similarly as
```
wenzeslaus, line 141: ...assigns ... _a_ semantic label...

wenzeslaus (approved):

temporal/t.register/t.register.html
marisn: Thanks.

marisn:
> During the PSC meeting today we talked about options such as my suggestion:
> ...but there was no final decision made.

Then lets do it in a separate step. I'll merge this one and then we can handle renaming (or deactivating compilation) of modules as a separate PR.

landam:
Sorry for being late with review. Let's backport to `releasebranch_8_0`.

neteler:
BTW: Shall we have separate G7 and G8 backport labels?

landam:
I think that it's not needed. Milestone is set to `8.0.0`, so only backport to `releasebranch_8_0` is relevant.

neteler:
> I think that it's not needed. Milestone is set to `8.0.0`, so only backport to `releasebranch_8_0` is relevant.

Backport to  `releasebranch_8_0` done by cherry-picking.

PR #1929, opened 2021-10-06
register.py: allow registering from io objects

ninsbl:
Currently, `register_maps_in_space_time_dataset` takes a file as input. This PR allowes to use the function in Python also with `io` objects (like `StinrgIO`)

It is a small change that would be nice to have in 8.0, but feel free to bump up the milestone...


metzm:
This is a nice enhancement! I don't see a need to bump up the milestone because it is not a breaking change.

I noticed that `fd.close()` is missing from `register_maps_in_space_time_dataset()`. `fd.close()` is not strictly needed but clean code would close a fd if no longer needed, particularly in library function to avoid having dangling open file descriptors.

PR #1934, opened 2021-10-08
python/grass/__init__.py: fix load user defined language (internationalization)

### LINKED ISSUE #1489: [Bug] Unable to change language settings
**Describe the bug**
When I set "non-system" language, then quit GRASS and start again, it is still in the "system" language.

**To Reproduce**

1. Go to Settings
2. Choose the "non-system" language 
3. Click save
4. Quit GRASS
5. Start GRASS again, the GRASS is still in the "system" language

**Expected behavior**
It should switch to "non-system" language.

**Screenshots**
VirtualBox_Windows_10__ArcGIS_Pro__31_03_2021_16_19_36

**System description:**

- Operating System: Windows
- GRASS GIS version master

tmszi:
Closes #1489.

### DIFF
--- a/python/grass/__init__.py
+++ b/python/grass/__init__.py
@@ -110,7 +110,8 @@ def _translate(text):
                     key="locale",
                     subkey="lc_all",
                 )
-            except NameError:  # if _UserSettings isn't imported during tests (without wxGUI)
+            except NameError:
+                # If _UserSettings isn't imported during tests (without wxGUI)
                 lang = None
 
             if lang and lang not in ("en", "system"):
--- a/python/grass/__init__.py
+++ b/python/grass/__init__.py
@@ -54,6 +54,24 @@ import os
 # - https://www.wefearchange.org/2012/06/the-right-way-to-internationalize-your.html
 
 
+def _import_user_settings():
+    """Import UserSettings
+
+    Required for getting user defined language.
+    """
+    try:
+        import gettext
+        from grass.script.setup import set_gui_path
+
+        set_gui_path()
+        # Initialize translatation func before import UserSettings
+        _builtins.__dict__["_"] = gettext.gettext
+        global _UserSettings
+        from core.settings import UserSettings as _UserSettings
+    except ModuleNotFoundError:  # during compilation, and tests (without wxGUI)
+        pass
+
+
 def _translate(text):
     """Get translated version of text
 
@@ -75,16 +93,47 @@ def _translate(text):
             # does not raise an exception even if the locale settings is broken
             # or the translation files were not installed.
             fallback = True
-            translation = gettext.translation(
-                "grasslibs", locale_dir, fallback=fallback
-            )
-            # Add other domains as fallback.
-            translation.add_fallback(
-                gettext.translation("grassmods", locale_dir, fallback=fallback)
-            )
-            translation.add_fallback(
-                gettext.translation("grasswxpy", locale_dir, fallback=fallback)
-            )
+            translation = None
+            domains = ("grasslibs", "grassmods", "grasswxpy")
+
+            def apply_trans(kwargs):
+                nonlocal translation
+                if kwargs["domain"] == "grasslibs":
+                    translation = gettext.translation(**kwargs)
+                else:
+                    translation.add_fallback(gettext.translation(**kwargs))
+
+            try:
+                # Get user defined lang
+                lang = _UserSettings.Get(
+                    group="language",
+                    key="locale",
+                    subkey="lc_all",
+                )
+            except NameError:  # if _UserSettings isn't imported during tests (without wxGUI)
+                lang = None
+
+            if lang and lang not in ("en", "system"):
+                for domain in domains:
+                    apply_trans(
+                        {
+                            "domain": domain,
+                            "localedir": locale_dir,
+                            "fallback": fallback,
+                            "languages": (lang,),
+                        }
+                    )
+            elif lang == "en":
+                translation = gettext.NullTranslations()
+            else:
+                for domain in domains:
+                    apply_trans(
+                        {
+                            "domain": domain,
+                            "localedir": locale_dir,
+                            "fallback": fallback,
+                        }
+                    )
             # Store the resulting translation object.
             _translate.translation = translation
         except (KeyError, ImportError):
@@ -97,7 +146,7 @@ def _translate(text):
 # Initialize the translation attribute of the translate function to indicate
 # that the translations are not initialized.
 _translate.translation = None
-
+_import_user_settings()
 _builtins.__dict__["_"] = _translate
 
 
--- a/python/grass/__init__.py
+++ b/python/grass/__init__.py
@@ -54,11 +54,8 @@ import os
 # - https://www.wefearchange.org/2012/06/the-right-way-to-internationalize-your.html
 
 
-def _import_user_settings():
-    """Import UserSettings
-
-    Required for getting user defined language.
-    """
+def _get_lang_from_user_settings():
+    """Get language from user settings"""
     try:
         import gettext
         from grass.script.setup import set_gui_path
@@ -69,7 +66,7 @@ def _import_user_settings():
         global _UserSettings
         from core.settings import UserSettings as _UserSettings
     except ModuleNotFoundError:  # during compilation, and tests (without wxGUI)
-        pass
+        _UserSettings = None
 
 
 def _translate(text):
@@ -93,48 +90,43 @@ def _translate(text):
             # does not raise an exception even if the locale settings is broken
             # or the translation files were not installed.
             fallback = True
-            translation = None
-            domains = ("grasslibs", "grassmods", "grasswxpy")
+            lang = None
 
-            def apply_trans(kwargs):
-                nonlocal translation
-                if kwargs["domain"] == "grasslibs":
-                    translation = gettext.translation(**kwargs)
-                else:
-                    translation.add_fallback(gettext.translation(**kwargs))
-
-            try:
+            if _UserSettings:
                 # Get user defined lang
-                lang = _UserSettings.Get(
-                    group="language",
-                    key="locale",
-                    subkey="lc_all",
-                )
-            except NameError:
-                # If _UserSettings isn't imported during tests (without wxGUI)
-                lang = None
+                lang = [
+                    _UserSettings.Get(
+                        group="language",
+                        key="locale",
+                        subkey="lc_all",
+                    )
+                ]
+                if lang.count(None):
+                    lang = None
 
-            if lang and lang not in ("en", "system"):
-                for domain in domains:
-                    apply_trans(
-                        {
-                            "domain": domain,
-                            "localedir": locale_dir,
-                            "fallback": fallback,
-                            "languages": (lang,),
-                        }
-                    )
-            elif lang == "en":
-                translation = gettext.NullTranslations()
-            else:
-                for domain in domains:
-                    apply_trans(
-                        {
-                            "domain": domain,
-                            "localedir": locale_dir,
-                            "fallback": fallback,
-                        }
-                    )
+            translation = gettext.translation(
+                "grasslibs",
+                locale_dir,
+                fallback=fallback,
+                languages=lang,
+            )
+            # Add other domains as fallback.
+            translation.add_fallback(
+                gettext.translation(
+                    "grassmods",
+                    locale_dir,
+                    fallback=fallback,
+                    languages=lang,
+                )
+            )
+            translation.add_fallback(
+                gettext.translation(
+                    "grasswxpy",
+                    locale_dir,
+                    fallback=fallback,
+                    languages=lang,
+                )
+            )
             # Store the resulting translation object.
             _translate.translation = translation
         except (KeyError, ImportError):
@@ -147,7 +139,7 @@ def _translate(text):
 # Initialize the translation attribute of the translate function to indicate
 # that the translations are not initialized.
 _translate.translation = None
-_import_user_settings()
+_get_lang_from_user_settings()
 _builtins.__dict__["_"] = _translate
 
 


tmszi:
There are the following two issues with this PR:

1. Due to the fact that the class `UserSettings` that solves parsing the user settings is imported into the file `
python/grass/__init__.py` (place from which is installs the gettext function `_()` in Python’s builtins namespace, based on domain, localedir, and codeset) , and is dependent on the `wx` module, it is necessary that GRASS GIS is always compiled with wxPython. See the results of the Git workflow test [Ubuntu 18.04](https://github.com/OSGeo/grass/pull/1934/checks?check_run_id=3840857814#step:9:1), [Ubuntu 20.04](https://github.com/OSGeo/grass/pull/1934/checks?check_run_id=3840857864).

2. Another problem is that when compiled on the MS Windows platform, the `locale.getlocale()` function in the `python/grass/__init__.py` file, it seems that returns a different value of locales than C code than on GNU/Linux platform, and therefore the if branch evaluates to `True` will try to import the `UserSettings` class, which depends on the `wx` module. See the results of the Git workflow [Osgeo4W build](https://github.com/OSGeo/grass/pull/1934/checks?check_run_id=3840857819).

wenzeslaus:
This conflicts with #1838 in terms of code changes. I'm not sure about the conceptual level. #1838 is approved and fixes some of the issues. Since this is draft, I suggest to merge #1838 first. Are you okay with that @tmszi or do you have another suggestion on how to proceed?

This one is marked for backport, but #1838 is not a good backport candidate due to the nature of the changes.

tmszi:
> This conflicts with #1838 in terms of code changes. I'm not sure about the conceptual level. #1838 is approved and fixes some of the issues. Since this is draft, I suggest to merge #1838 first. Are you okay with that @tmszi or do you have another suggestion on how to proceed?

I appologize, for my not quickly response. I adapted code of this PR for your merged PR #1838. Import `UserSettings` class instance is solved  outside of `_translate` function via global variable because import inside this function it isn't possible due to  to a circular import error.

```
Starting GRASS GIS...
  File "/usr/bin/grass80", line 2729, in <module>
    main()
  File "/usr/bin/grass80", line 2643, in main
    clean_temp()
  File "/usr/bin/grass80", line 2104, in clean_temp
    message(_("Cleaning up temporary files..."))
  File "/usr/lib64/grass80/etc/python/grass/__init__.py", line 79, in _translate
    from core.settings import UserSettings as UserSettings
  File "/usr/lib64/grass80/gui/wxpython/core/settings.py", line 1191, in <module>
    UserSettings = Settings()
  File "/usr/lib64/grass80/gui/wxpython/core/settings.py", line 95, in __init__
    self._defaultSettings()  # -> self.defaultSettings
  File "/usr/lib64/grass80/gui/wxpython/core/settings.py", line 435, in _defaultSettings
    "legend": _("Segment break"),
  File "/usr/lib64/grass80/etc/python/grass/__init__.py", line 79, in _translate
    from core.settings import UserSettings as UserSettings
ImportError: cannot import name 'UserSettings' from partially initialized module 'core.settings' (most likely due to a circular import) (/usr/lib64/grass80/gui/wxpython/core/settings.py)
```

> This one is marked for backport, but #1838 is not a good backport candidate due to the nature of the changes.

Yes, allright.

python/grass/__init__.py
```diff
@@ -75,16 +93,48 @@ def _translate(text):
...
-            )
-            translation.add_fallback(
-                gettext.translation("grasswxpy", locale_dir, fallback=fallback)
-            )
+            translation = None
+            domains = ("grasslibs", "grassmods", "grasswxpy")
+
+            def apply_trans(kwargs):
+                nonlocal translation
+                if kwargs["domain"] == "grasslibs":
```
wenzeslaus, line 101: This might be better expressed as `translation is None`.

python/grass/__init__.py
```diff
@@ -75,16 +93,48 @@ def _translate(text):
...
-            translation.add_fallback(
-                gettext.translation("grassmods", locale_dir, fallback=fallback)
-            )
-            translation.add_fallback(
-                gettext.translation("grasswxpy", locale_dir, fallback=fallback)
-            )
+            translation = None
+            domains = ("grasslibs", "grassmods", "grasswxpy")
+
+            def apply_trans(kwargs):
```
wenzeslaus, line 99: You can pass the kwargs parameter really as `**kwargs` which will simplify the _apply_trans_ function calls.

python/grass/__init__.py
```diff
@@ -75,16 +93,48 @@ def _translate(text):
...
+                lang = None
+
+            if lang and lang not in ("en", "system"):
+                for domain in domains:
+                    apply_trans(
+                        {
+                            "domain": domain,
+                            "localedir": locale_dir,
+                            "fallback": fallback,
+                            "languages": (lang,),
```
wenzeslaus, line 124: This can probably become something like `languages=[lang] if lang else None`:

* [gettext](https://docs.python.org/3.10/library/gettext.html#gettext.find) doc says it is a list: _Optional languages is a list of strings, where each string is a language code._ (Probably doesn't matter much, but increasingly popular practice is to have things which are lists as lists even when they are fixed size.)
* You can resolve any language assignment special needs/reassignments before the apply_translation/gettext.translation call.
* The dict syntax should no be needed if kwargs is handled fully or if gettext.translation is called directly.

python/grass/__init__.py
```diff
@@ -75,16 +93,48 @@ def _translate(text):
...
+                lang = _UserSettings.Get(
+                    group="language",
+                    key="locale",
+                    subkey="lc_all",
+                )
+            except NameError:
+                # If _UserSettings isn't imported during tests (without wxGUI)
+                lang = None
+
+            if lang and lang not in ("en", "system"):
```
wenzeslaus, line 117: Please add a comment on why these two need a special treatment.

python/grass/__init__.py
```diff
@@ -75,16 +93,48 @@ def _translate(text):
...
+                    apply_trans(
+                        {
+                            "domain": domain,
+                            "localedir": locale_dir,
+                            "fallback": fallback,
+                            "languages": (lang,),
+                        }
+                    )
+            elif lang == "en":
+                translation = gettext.NullTranslations()
```
wenzeslaus, line 128: Is this really needed? How that worked before? It is not the case of GRASS GIS, but in a very theoretical level, there could be a translation from developer English to a proper one or to a its dialect.

python/grass/__init__.py
```diff
@@ -75,16 +93,48 @@ def _translate(text):
...
+            else:
```
wenzeslaus, line 129: This is meant for system, right? Likely the same call with language/lang being set to None will remove the need for another loop in the code (see the comment for the first loop).

python/grass/__init__.py
```diff
@@ -75,16 +93,48 @@ def _translate(text):
...
+                    apply_trans(
+                        {
+                            "domain": domain,
+                            "localedir": locale_dir,
+                            "fallback": fallback,
+                        }
+                    )
             # Store the resulting translation object.
             _translate.translation = translation
         except (KeyError, ImportError):
```
wenzeslaus, line 140: This except followed by NullTranslations may just handle your import and English trouble as is, but I would prefer not to even get into those in the first place.

python/grass/__init__.py
```diff
@@ -97,7 +147,7 @@ def _translate(text):
 # Initialize the translation attribute of the translate function to indicate
 # that the translations are not initialized.
 _translate.translation = None
-
+_import_user_settings()
 _builtins.__dict__["_"] = _translate
```
wenzeslaus, line 151: If all you need is to call import in a function, _translate is already a function and it get called only when a translation is needed, not before that, so likely long after imports. Whatever is needed can be done there instead of the top-level. I understand you wrote the code before you saw #1838, but I think it is relevant.

wenzeslaus (changes requested):
Thanks for merging it with the new code.

I think the dependence of library code on GUI code is a big issue, so this may require completely different approach; perhaps refactoring of the GUI settings code and moving it to the library.

I also feel I need more explanation on the motivation for the change. I assume it fixes the referenced issue, but is this a workaround or the best solution which should have been in the code already? Does it have some limitations?

python/grass/__init__.py
```diff
@@ -54,6 +54,24 @@
...
+    """
+    try:
+        import gettext
+        from grass.script.setup import set_gui_path
+
+        set_gui_path()
+        # Initialize translatation func before import UserSettings
+        _builtins.__dict__["_"] = gettext.gettext
+        global _UserSettings
+        from core.settings import UserSettings as _UserSettings
```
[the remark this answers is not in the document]
marisn: As settings set in GUI are managed by GUI, there is no other way at the moment. To have a core settings management functionality that is also exposed in GUI is the right way to go but it is out of scope of this PR.

python/grass/__init__.py
marisn: +1 for a comment. The rationale – we do not ship "en" or "system" translations and thus they need a different logic in setting up.

python/grass/__init__.py
marisn: At the moment translations in GUI are identified by string that should match suffix of translation file. Thus if somebody decides to translate GRASS to some flavour of English, he'd use that flavour in a file name e.g. en_GB and thus it would show up in GUI as en_GB (see pt_BR for an example). IIRC "en" option was introduced as Markus had problems using GRASS in Japan :D

marisn:
The main problem is that this PR will not fix the #1489 as the root of problem is not here but in init.py file. At some point GUI has migrated to a different settings file but init.py still is searching for the old settings file and thus it never gets setting set in GUI. One more +1 for taking a closer look at handling of settings.

marisn:
@tmszi  please examine this PR in the context of #1990.

python/grass/__init__.py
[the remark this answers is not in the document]
wenzeslaus: One of the advantages of using JSON is that it is easy enough to parse (with a library), so there is no need to duplicate parsing as with the previous format. However, it still makes sense to have an API for this as demonstrated by #1990. `open(..."wx...` outside of wxGUI is a code smell I was not aware of. Thanks for figuring this out.

python/grass/__init__.py
tmszi: Yes you're right in context of existed translation PO/MO files. But user has option via GUI settings change language to "en" (which is the language if no translation is used) or "system" (OS locale language) language and is justified.

A. Use case:

1. User has system with "cz_CZ.utf8" locale settings (wxGUI start with Czech language)
2. User want wxGUI with English language (choose "en" option from wxGUI Settings dialog Choose language combobox  widget)

B. Use case

1. User has system with "cz_CZ.utf8" locale settings (wxGUI start with Czech language)
2. User want wxGUI with English language (choose "en" option from wxGUI Settings dialog Choose language combobox  widget)
3. Later user want use wxGUI with default system language (default system locale which is "cz_CZ.utf8", choose "system" option from wxGUI Settings dialog Choose language combobox widget)

python/grass/__init__.py
```diff
@@ -75,16 +93,48 @@ def _translate(text):
...
+
+            try:
+                # Get user defined lang
+                lang = _UserSettings.Get(
+                    group="language",
+                    key="locale",
+                    subkey="lc_all",
+                )
+            except NameError:
+                # If _UserSettings isn't imported during tests (without wxGUI)
```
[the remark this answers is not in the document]
tmszi: Yes I agree.

python/grass/__init__.py
```diff
@@ -54,6 +54,24 @@
 # - https://www.wefearchange.org/2012/06/the-right-way-to-internationalize-your.html
 
 
+def _import_user_settings():
```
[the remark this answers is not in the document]
tmszi: Yes I agree.

python/grass/__init__.py
```diff
@@ -75,16 +93,48 @@ def _translate(text):
...
-                gettext.translation("grassmods", locale_dir, fallback=fallback)
-            )
-            translation.add_fallback(
-                gettext.translation("grasswxpy", locale_dir, fallback=fallback)
-            )
+            translation = None
+            domains = ("grasslibs", "grassmods", "grasswxpy")
+
+            def apply_trans(kwargs):
+                nonlocal translation
```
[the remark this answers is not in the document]
tmszi: I have completely simplified the code.

python/grass/__init__.py
tmszi: Yes, you're right.

python/grass/__init__.py
tmszi: Yes I understand, but "en" language code represent language if no translation is used. Check my previous comment with use case examples.

python/grass/__init__.py
tmszi: Yes, I understand how lazy translate work and I completely checked merged code from #1838 PR, before I update my PR. I can't import `UserSettings` class instance  inside `_translate()` function due to a circular import, check my first comment.

python/grass/__init__.py
```diff
@@ -54,6 +54,24 @@
...
+
+    Required for getting user defined language.
+    """
+    try:
+        import gettext
+        from grass.script.setup import set_gui_path
+
+        set_gui_path()
+        # Initialize translatation func before import UserSettings
+        _builtins.__dict__["_"] = gettext.gettext
```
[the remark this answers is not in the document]
tmszi: I need initialize basic translation function here, because `UserSettings` class contains translation strings, before import it.

tmszi:
> The main problem is that this PR will not fix the #1489 as the root of problem is not here but in init.py file. At some point GUI has migrated to a different settings file but init.py still is searching for the old settings file and thus it never gets setting set in GUI. One more +1 for taking a closer look at handling of settings.

I tested it and it works as expected and it fixes this issue #1489.

PR #1962, opened 2021-10-30
remove .setpdfwrite option on MS Windows

ninsbl:
.setpdfwrite is no longer supported with ghostscript 9.5 shipped with OSGeo4W
See: https://trac.osgeo.org/osgeo4w/ticket/697

### DIFF
--- a/gui/wxpython/psmap/frame.py
+++ b/gui/wxpython/psmap/frame.py
@@ -404,6 +404,7 @@ class PsMapFrame(wx.Frame):
                            '-sOutputFile=%s' % event.userData['pdfname'],
                            '-P-', '-dSAFER',
                            '-dCompatibilityLevel=1.4',
+                           '-c', '30000000', 'setvmthreshold',
                            '-f', event.userData['filename']]
             else:
                 command = [
--- a/gui/wxpython/psmap/frame.py
+++ b/gui/wxpython/psmap/frame.py
@@ -404,8 +404,7 @@ class PsMapFrame(wx.Frame):
                            '-sOutputFile=%s' % event.userData['pdfname'],
                            '-P-', '-dSAFER',
                            '-dCompatibilityLevel=1.4',
-                           '-c', '.setpdfwrite', '-f',
-                           event.userData['filename']]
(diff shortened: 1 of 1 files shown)


hellik:
taken from [gs 9.50 manual](https://www.ghostscript.com/doc/9.50/Language.htm)

```
.setpdfwrite
    This operator is now deprecated, and its use is discouraged

    This operator conditions the environment for the pdfwrite output device. It is a shorthand for setting parameters that have been deemed benificial. While not strictly necessary, it is usually helpful to set call this when using the pdfwrite device. For example, this is how the ps2pdf script calls Ghostscript:

        gs -q -dSAFER -dNOPAUSE -dBATCH -sOutputFile=file.pdf [more options] \
          -sDEVICE=pdfwrite -c .setpdfwrite -f source1.ps [more files] 

    Currently, the operator just sets a minimum 3 MB vmthreshold to allow for accumulating shared object data and to reduce the incidence of garbage collection as a performance improvement. Additional settings may be added in the future.
```

taken from [gs 9.26 manual](https://www.ghostscript.com/doc/9.26/Use.htm)

```
With some PDF files, or if you are using Chinese, Japanese, or other fonts with very large character sets, adding the following sequence of switches before the first file name may dramatically improve performance at the cost of an additional memory. For example, to allow use of 30Mb of extra RAM use: -c 30000000 setvmthreshold -f.

This can also be useful in processing large documents when using a high-level (vector) output device (like pdfwrite) that maintains significant internal state. In fact, the .setpdfwrite operator used by the ps2pdf script and others sets a vmthreshold value of 3 MB to account for this.
```
do we need to set **setvmthreshold** instead?

neteler:
Since #1961 has been merged and backported, can this be, too?

ninsbl:
> @ninsbl Since #1961 has been merged and backported, can this be, too?

This is actually the backport PR to the 7.8 releasebranch (as direct backport was not possible due to code formating (black)...

Shold be safe to merge...

neteler:
Thanks, so I merge it.

PR #1978, opened 2021-11-14
display/d.linegraph: fix generate graph image if 'x_title', 'y_title' param arg isn't defined

tmszi:
**Describe the bug**
`d.linegraph` module doesn't generate graph image if `x_title`, `y_title` param arg isn't defined

**To Reproduce**
Steps to reproduce the behavior (according `d.linegraph` manual page):

1. Launch GRASS GIS `grass80 --text`
2. Create some example text data files.
```
cat > /tmp/x.txt <<EOF
1
3
4
6
9
EOF
cat > /tmp/y1.txt <<EOF
50
58
65
34
27
EOF
cat > /tmp/y2.txt <<EOF
10
20
35
50
45
EOF
``` 
3. `d.mon start=cairo output=/tmp/plot.png`
4. `d.linegraph x_file=/tmp/x.txt y_file=/tmp/y1.txt,/tmp/y2.txt`
5. `d.mon stop=cairo`
6. Result 'plot.png' image file doesn't exists `file /tmp/plot.png`

**Expected behavior**
Result 'plot.png' image file exists with rendered graph.

**Screenshots**

plot

- GRASS GIS version 8.0 dev is affected with this bug only

### DIFF
--- a/display/d.linegraph/main.c
+++ b/display/d.linegraph/main.c
@@ -193,10 +193,9 @@ int main(int argc, char **argv)
     float xscale;
     float yscale;
 
-    char xlabel;
     char txt[1024];
     char tic_name[1024];
-    char *name, *pxlabel = &xlabel;
+    char *name, *xlabel;
     char color_name[20];
 
     FILE *fopen();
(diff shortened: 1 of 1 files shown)


display/d.linegraph/main.c
```diff
@@ -889,16 +890,16 @@ int main(int argc, char **argv)
 
     /* draw the x-axis label */
     if ((strcmp(title[0]->answer, "") == 0) && (strcmp(tic_name, "") == 0))
-        *xlabel = '\0';
+        *pxlabel = '\0';
```
nilason, line 893:
```suggestion
        xlabel = G_store("");
```

display/d.linegraph/main.c
```diff
@@ -1001,30 +1002,30 @@ int main(int argc, char **argv)
 
     /* draw the y-axis label */
     if ((strcmp(title[1]->answer, "") == 0) && (strcmp(tic_name, "") == 0))
-        *xlabel = '\0';
+        *pxlabel = '\0';
```
nilason, line 1005:
```suggestion
        xlabel = G_store("");
```

nilason (changes requested):
The problem was that the empty string of `xlabel` never got allocated. If you try my suggested changes, along with reverting *all* other changes (pxlabel etc.), I believe it should work again.

display/d.linegraph/main.c
tmszi: Yes, of course you're right, I completely overlooked the need for a dynamic memory allocation.

nilason (approved):
Thanks for catching this! Looks good to me now.

PR #1990, opened 2021-11-25
Fix loading of language override setting. Fixes #1489

### LINKED ISSUE #1489: [Bug] Unable to change language settings

marisn:
Setting of language override was broken as GUI now stores settings in a different file. This PR restores back old functionality by reading the setting from the new settings file.

### DIFF
--- a/lib/init/grass.py
+++ b/lib/init/grass.py
@@ -55,6 +55,7 @@ import locale
 import uuid
 import unicodedata
 import argparse
+import json
 
 
 # mechanism meant for debugging this script (only)
@@ -1373,19 +1374,11 @@ def set_language(grass_config_dir):
     encoding = None
 
     # Override value is stored in wxGUI preferences file.
-    # As it's the only thing required, we'll just grep it out.
-    try:
-        fd = open(os.path.join(grass_config_dir, "wx"), "r")
-    except:
-        # Even if there is no override, we still need to set locale.
-        pass
-    else:
-        for line in fd:
-            if re.search("^language", line):
-                line = line.rstrip(" %s" % os.linesep)
-                language = "".join(line.split(";")[-1:])
-                break
-        fd.close()
+    with open(os.path.join(grass_config_dir, "wx.json"), "r") as json_file:
+        try:
+            language = json.load(json_file)["language"]["locale"]["lc_all"]
+        except KeyError:
+            language = None
 
     # Backwards compatibility with old wx preferences files
     if language == "C":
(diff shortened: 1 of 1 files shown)


lib/init/grass.py
```diff
@@ -1373,19 +1374,11 @@ def set_language(grass_config_dir):
...
-        # Even if there is no override, we still need to set locale.
-        pass
-    else:
-        for line in fd:
-            if re.search("^language", line):
-                line = line.rstrip(" %s" % os.linesep)
-                language = "".join(line.split(";")[-1:])
-                break
-        fd.close()
+    with open(os.path.join(grass_config_dir, "wx.json"), "r") as json_file:
```
wenzeslaus, line 1377: Will fail if if wx.json does not exist like at a first start of GRASS by a user. Perhaps an additional/larger try-except with whatever open raises is the way to go.

lib/init/grass.py
```diff
@@ -1373,19 +1374,11 @@ def set_language(grass_config_dir):
...
+        try:
+            language = json.load(json_file)["language"]["locale"]["lc_all"]
+        except KeyError:
+            language = None
```
wenzeslaus, line 1381: +1, more clear than the previous code. I'm not sure if `language == "None"` is still relevant with the new format. I would hope it is not.

wenzeslaus (changes requested):
Looks good as far as replacing the outdated code except for that error handling of the _open_  function. Just for the record, I'm not looking much further (the function is almost 200 lines long).

marisn:
> Just for the record, I'm not looking much further (the function is almost 200 lines long).

Better don't ;-) There is a reason why the code is so long – for every line there is a system where another path to failure was found.

lib/init/grass.py
marisn: Thanks for catching this!

lib/init/grass.py
marisn: Lets hope so. I removed this old workaround.

wenzeslaus (approved):

landam:
@marisn Feel free to backport. Thanks.

marisn:
Thanks, @neteler

PR #2000, opened 2021-12-02
pythonlib: update script.setup docs

nilason:
The documentation example:

~- now works for Python 3~
- has correct and tested setup for mac
- is Black formatted
~- is flake8 validated~

### DIFF
--- a/python/grass/script/setup.py
+++ b/python/grass/script/setup.py
@@ -24,17 +24,16 @@ Usage::
     # query GRASS itself for its GISBASE
     # (with fixes for specific platforms)
     # needs to be edited by the user
-    grass8bin = 'grass'
-    if sys.platform.startswith('win'):
+    grass8bin = "grass"
+    if sys.platform.startswith("win"):
         # MS Windows
-        grass8bin = r'C:\OSGeo4W\bin\grass.bat'
+        grass8bin = r"C:\OSGeo4W\bin\grass.bat"
         # uncomment when using standalone WinGRASS installer
         # grass8bin = r'C:\Program Files (x86)\GRASS GIS 8.0.0\grass.bat'
         # this can be avoided if GRASS executable is added to PATH
-    elif sys.platform == 'darwin':
+    elif sys.platform == "darwin":
         # Mac OS X
-        # TODO: this have to be checked, maybe unix way is good enough
-        grass8bin = '/Applications/GRASS/GRASS-8.0.app/'
+        grass8bin = "/Applications/GRASS-8.0.app/Contents/Resources/bin/grass"
 
     # query GRASS GIS itself for its Python package path
     grass_cmd = [grass8bin, "--config", "python_path"]
@@ -51,15 +50,15 @@ Usage::
     rcfile = gsetup.init(gisdb, location, mapset)
 
     # example calls
-    gs.message('Current GRASS GIS 8 environment:')
+    gs.message("Current GRASS GIS 8 environment:")
     print(gs.gisenv())
 
-    gs.message('Available raster maps:')
-    for rast in gs.list_strings(type='raster'):
+    gs.message("Available raster maps:")
+    for rast in gs.list_strings(type="raster"):
         print(rast)
 
-    gs.message('Available vector maps:')
-    for vect in gs.list_strings(type='vector'):
+    gs.message("Available vector maps:")
+    for vect in gs.list_strings(type="vector"):
         print(vect)
 
     # clean up at the end


wenzeslaus:
Are you sure this takes into account c810f67ab21e94f17c0ad9cfa40312f40d5f4ac1 (#1835) and other recent PRs? For example, `~` should work out of the box (if not, it is a bug).

nilason:
> Are you sure this takes into account [c810f67](https://github.com/OSGeo/grass/commit/c810f67ab21e94f17c0ad9cfa40312f40d5f4ac1) (#1835) and other recent PRs? For example, `~` should work out of the box (if not, it is a bug).

Good you noticed, thanks! I accidentally picked up the code from an earlier checkout. I have now corrected.

python/grass/script/setup.py
```diff
@@ -24,17 +24,24 @@
...
         # Mac OS X
-        # TODO: this have to be checked, maybe unix way is good enough
-        grass8bin = '/Applications/GRASS/GRASS-8.0.app/'
+        mac_gisbase = "/Applications/GRASS-8.0.app/Contents/Resources"
+        usr_module_dir = "~/Library/GRASS/8.0/Modules"
+        grass8bin = f"{mac_gisbase}/bin/grass"
+        os.environ["PATH"] = (
+            f"{mac_gisbase}/bin:{mac_gisbase}/scripts:"
+            f"{usr_module_dir}/bin:{usr_module_dir}/scripts:"
+            "/usr/bin:/bin:/usr/sbin:/etc:/usr/lib"
```
wenzeslaus, line 42: There are [functions to do this](https://stackoverflow.com/a/1681244) in a better way, or is resetting the path your intention?

python/grass/script/setup.py
```diff
@@ -24,17 +24,24 @@
...
+        grass8bin = r"C:\OSGeo4W\bin\grass.bat"
         # uncomment when using standalone WinGRASS installer
         # grass8bin = r'C:\Program Files (x86)\GRASS GIS 8.0.0\grass.bat'
         # this can be avoided if GRASS executable is added to PATH
-    elif sys.platform == 'darwin':
+    elif sys.platform == "darwin":
         # Mac OS X
-        # TODO: this have to be checked, maybe unix way is good enough
-        grass8bin = '/Applications/GRASS/GRASS-8.0.app/'
+        mac_gisbase = "/Applications/GRASS-8.0.app/Contents/Resources"
```
wenzeslaus, line 36: The `mac_` part is not needed here since this is already under macOS, but I hope the whole variable is not needed.

python/grass/script/setup.py
```diff
@@ -24,17 +24,24 @@
...
+        usr_module_dir = "~/Library/GRASS/8.0/Modules"
+        grass8bin = f"{mac_gisbase}/bin/grass"
+        os.environ["PATH"] = (
+            f"{mac_gisbase}/bin:{mac_gisbase}/scripts:"
+            f"{usr_module_dir}/bin:{usr_module_dir}/scripts:"
```
wenzeslaus, line 41: Ideally, this should be resolved by the init function. The idea is that all what you need to call the init function is one call to the binary. (Ideally, you would not needed even that with grass package on (Python) path.)

wenzeslaus (changes requested):

python/grass/script/setup.py
nilason: The official binary `GRASS-8.0.app` type of program, is a contained conda environment, with the above set PATH on execution only (but not otherwise) . So yes, it was an intentional complete reset.

However, the example test works also with **only** setting `grass8bin`:

```python
elif sys.platform == "darwin":
    # Mac OS X
    grass8bin = "/Applications/GRASS-8.0.app/Contents/Resources/bin/grass"
```

Is setting PATH needed at all for this?

wenzeslaus: The new logic is the following:

1. You call the binary to obtain the path needed to import the _grass_ package.
2. Leave the rest of the setup on the _grass.script.setup.init_ function.

With the these two notes:

1. On Windows and macOS, number 1 might be a challenge (unlike Linux where executable called _grass_ is assumed to be on path).
2. Ideally, the _grass_ package would be on Python path, so you would just need to get the right Python set up (NOP on standard Linux installation) instead of dealing with the executable in the script, i.e., ideally you would just `import grass` (#658).

So, if `/Applications/GRASS-8.0.app/Contents/Resources/bin/grass` runs GRASS GIS from command line on your machine, that's all what you (should) need to do before calling the _init_ function.

nilason: Thanks for the explanation!

nilason:
Works fine with just setting `grass8bin` to the correct path. Updated accordingly. Think this may be merged now.

I realised there is an issue with addons dir for Mac in general. I will return with that in a new PR.
Meanwhile, I noted `GRASS_ADDON_BASE` isn't taken into account (if set) with `init()`, is that intentional? And why not (in `setup_runtime_env()`) prepend to PATH instead of append as present code does?

nilason:
Not much remained of this... :-)
Merged and backported.

wenzeslaus (commented):
+1. I just wanted to clarify that the issues were resolved and that #2004 is the followup.

PR #2009, opened 2021-12-08
CI: Remove Ubuntu 16.04 from GH Actions for 7.8 branch

wenzeslaus:
Ubuntu 16.04 environment was removed from GitHub in September. The job for the 7.8 branch now always fails with time out.


landam (approved):

landam:
Let's support only last two Ubuntu LTS.

PR #2010, opened 2021-12-09
Jupyter: timeseries visualizations

chaedri:
`TimeSeries` class for `grass.jupyter`:
* Interactive timeline slider
* GIF animation
* Tests

Binder Test Link:

https://mybinder.org/v2/gh/chaedri/grass/Jupyter-Timeseries?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Ftemporal.ipynb


python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,87 @@
+# MODULE:    grass.jupyter.timeseries
+#
+# AUTHOR(S): Caitlin Haedrich <caitlin DOT haedrich AT gmail>
+#
+# PURPOSE:   This module contains functions visualizing time-space raster
```
veroandreo, line 5:
```suggestion
# PURPOSE:   This module contains functions for visualizing raster and vector
```

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,87 @@
...
+#            and vector datasets in Jupyter Notebooks
```
veroandreo, line 6:
```suggestion
#            space-time datasets in Jupyter Notebooks
```

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,87 @@
...
+
+class timeseries:
+    """timeseries creates visualization of time-space raster and
+    vector dataset in Jupyter Notebooks"""
+
+    def __init__(self, map, basemap, overlay):
+        self.map = map
+        self.basemap = basemap
+        self.overlay = overlay
+        # Check that map ends with tsrds or tsvds
```
veroandreo, line 29: The abbreviations we use are: strds, stvds and str3ds. I'd suggest to keep it consistent if possible.

chaedri: Oops! Thank you for catching this. I'll make sure they are consistent.

wenzeslaus: I'm afraid that in the Python API for Jupyter, the abbreviations won't look really good and won't fit with the rest in terms of readability. Suggestions welcome.

veroandreo: My point was to use the same terms we use in the temporal framework as letters were exchanged and that might create confusion. 

Sorry for my ignorance here but, what do you mean by "in the Python API for Jupyter, the abbreviations won't look really good and won't fit with the rest"? Not using any abbreviations at all? Or finding new abbreviations to refer to strds and such? We can say raster time series and vector time series for example. Still, in the notebook or documentation I'd clarify that they refer to the temporal framework objects known as space-time raster dataset and such... If that's indeed what will be used in this context of course. Otherwise, ignore the noise :)

wenzeslaus: > ...letters were exchanged...

Yes, right, that's correct.

> ...Not using any abbreviations at all?...

Maybe. The tsrds and tsvds in the code just looks as arbitrary as the strds-stvds-str3ds mumble jumble. In grass.jupyter, the idea is that the names fit with Python naming convention unless there is a reason not to, so _d.rast_ could be hidden under _display_raster_, but "d-rast" is pronounceable and you can match it with "display raster" or it is at least easy enough to document and definitively it is easy to document that the _d_rast_ function is the _d.rast_ tool/module, so at the end, _d_rast_ looks like a descriptive, but pragmatically shortened name in Python (like e.g. _os.environ_ or _sys.path_). It now seems to me that with temporal datasets, we need to have similar discussion we had before for d-commands in grass.jupyter.

veroandreo:
Looking cool so far :)

image

ninsbl:
Really nice!

Tested in Binder. 

It currently prints:
`<function grass.jupyter.timeseries.TimeSeries.TimeSlider.<locals>.view_image(date)>`
at the bottom of the last cell though...

chaedri:
> Really nice!
> 
> Tested in Binder.
> 
> It currently prints: `<function grass.jupyter.timeseries.TimeSeries.TimeSlider.<locals>.view_image(date)>` at the bottom of the last cell though...

Thanks for checking it out! The printout at the bottom should be gone with the latest commit.

doc/notebooks/temporal.ipynb
```diff
@@ -0,0 +1,276 @@
...
+    "import shutil\n",
+    "\n",
+    "# Download zip\n",
+    "!curl http://fatra.cnr.ncsu.edu/temporal-grass-workshop/NC_spm_temporal_workshop.zip -o ../../data/NC_spm_temporal_workshop.zip\n",
+    "\n",
+    "# Unpack zip to grassdata\n",
+    "shutil.unpack_archive(\"../../data/NC_spm_temporal_workshop.zip\", \"../../data/grassdata\", \"zip\")\n",
+    "\n",
+    "# Delete Zip\n",
+    "os.remove(\"../../data/NC_spm_temporal_workshop.zip\")"
```
wenzeslaus, line 39: No need for this. Adds 3 lines.

doc/notebooks/temporal.ipynb
```diff
@@ -0,0 +1,276 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "gs.run_command(\"t.register\",\n",
+    "               input=\"tempmean\",\n",
+    "               type=\"raster\",\n",
+    "               start=\"2000-01-01\",\n",
+    "               increment=\"1 months\", \n",
+    "               maps=tempmean_list,\n",
+    "               flags=\"i\")\n",
```
wenzeslaus, line 142: Since you are formatting these anyway like this, try the latest Black with extension for Jupyter and format the notebook too.

wenzeslaus (changes requested):
Skipping the `render_layers()` step ends with traceback in both cases. Why no check for that and do it automatically if needed. The other objects work this way. It will also create shorter user code. Perhaps the difficult question is how to determine when to call render_layers() again.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+        param str text_color: color to use for the text.
+        param str filename: name of output GIF file
+        """
+        # Create a GIF from the PNG images
+        from PIL import Image  # pylint: disable=import-outside-toplevel
+        from PIL import ImageFont  # pylint: disable=import-outside-toplevel
+        from PIL import ImageDraw  # pylint: disable=import-outside-toplevel
+        from IPython.display import (  # pylint: disable=import-outside-toplevel
+            Image as ipyImage,
+        )
```
wenzeslaus, line 243: You can `import IPython.display` and use IPython.display.Image to avoid issues with creating a good enough name.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+        param bool label: include date/time stamp on each frame
+        param str font: font file
+        param int text_size: size of date/time text
+        param str text_color: color to use for the text.
+        param str filename: name of output GIF file
+        """
+        # Create a GIF from the PNG images
+        from PIL import Image  # pylint: disable=import-outside-toplevel
+        from PIL import ImageFont  # pylint: disable=import-outside-toplevel
+        from PIL import ImageDraw  # pylint: disable=import-outside-toplevel
```
wenzeslaus, line 240: Since this function uses both images, you can `import PIL` and use `PIL.Image...`.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+                    date,
+                    fill=text_color,
+                    font=ImageFont.truetype(font, text_size),
+                )
+            imgs.append(img)
+
+        img.save(
+            fp=filename,
+            format="GIF",
+            append_images=imgs[:-1],
```
wenzeslaus, line 267: Please add comment here. The order is not clear to me.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        # Lazy Imports
+        import ipywidgets as widgets  # pylint: disable=import-outside-toplevel
+        from IPython.display import Image  # pylint: disable=import-outside-toplevel
+
+        # Datetime selection slider
+        slider = widgets.SelectionSlider(
+            options=self._dates,
+            value=self._dates[0],
+            description="Date/Time",
```
wenzeslaus, line 196: Should be translatable.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Must be run before creating a visualization (i.e. time_slider or animate).
+
+        Can be time-consuming to run with large space-time datasets.
+        """
+
+        for name in self._renderlist:
+            # Create image file
+            filename = os.path.join(self._tmpdir.name, f"{name}.png")
+            # self._filenames.append(filename)
```
wenzeslaus, line 160: Remove if not needed.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+            "t.vect.list", method="gran", input=timeseries
+        ).splitlines()
+    else:
+        raise NameError(
+            _("Dataset {} must be element type 'strds' or 'stvds'").format(timeseries)
+        )
+
+    # Parse string
+    new_rows = [row.split("|") for row in rows]
+    new_array = [list(row) for row in zip(*new_rows)]
```
wenzeslaus, line 49: This transposes? Please add a comment.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+    # use the next non-None value
+    for i, name in enumerate(names):
+        if name == "None":
+            if i > 0:
+                names[i] = names[i - 1]
+            else:
+                search_count = 1
+                while name[i + search_count] == "None":
+                    search_count += 1
+                names[i] = name[i + 1]
```
wenzeslaus, line 69: I don't get this. Perhaps name versus names?

The while should be for or otherwise better checked. You don't 100% know what the input is, so it can run until IndexError is raised.

Finally, this for loop should be a standalone function so you can write a test for it.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+        if not test:
+            raise NameError(
+                _(
+                    "Could not find space time raster or vector " "dataset named {}"
+                ).format(timeseries)
+            )
+
+        # Create a temporary directory for our PNG images
+        self._tmpdir = (
+            # pylint: disable=consider-using-with
```
wenzeslaus, line 126: Copy the comment from the renderers about using finalize for cleanup to provide reasoning for the pylint: disable.

wenzeslaus (changes requested):
Please add tests. (The notebooks are not running automatically (possible bot not implemented yet), so we need at least some tests.)

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+
+    # Collect layer name and start time
+    for column in new_array:
+        if column[0] == "name":
+            names = column[1:]
+        if column[0] == "start_time":
+            dates = column[1:]
+
+    # For variable timestep datasets, fill in None values with
+    # previous time step value. If first time step is missing data,
```
petrasovaa, line 59: Could you give an example when the first will be None?

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+    # use the next non-None value
+    for i, name in enumerate(names):
+        if name == "None":
+            if i > 0:
+                names[i] = names[i - 1]
```
petrasovaa, line 64: I think the filling should be optional. In your example when you unregister one map, it feels like the user should know the gap is there, the behavior seems misleading. But I would be interested in other opinions about that.

veroandreo: I wanted to say this too, but was unsure of what happens in g.gui.animate for example (as to compare), when a date is missing in a time series with regular intervals. If I understand the current behavior in the notebook correctly, filling a missing date with the map from the previous date would be wrong IMHO... would it be possible to leave it blank? Maybe not nice but it would be correct as there's no map for that date, otherwise we would be making assumptions about the data (indeed, another possibility would be to interpolate, but again we would be making assumptions). I darkly remember to have created vector data animations where there were missing vector maps for certain dates and the user just see the time passing (in the time stamps) but no vector map was displayed until there were dates with data again.

doc/notebooks/temporal.ipynb
chaedri: Cool extension!

python/grass/jupyter/timeseries.py
chaedri: I wasn't sure if this was possible so I wrote it that way out of an abundance of caution. I was thinking that since t.register lets you set a start datetime and increment, you might be able to have a space-time dataset with an empty first map.  But now, I'm doubting that it's possible to do that. @petrasovaa , are you aware of any way to create a dataset with None in the first step?

python/grass/jupyter/timeseries.py
chaedri: I also wasn't sure that this behavior was ideal but wasn't sure of better alternatives. I think having it be an option makes a lot of sense, maybe something like:

`gj.TimeSeries( ... , gapfill=True,  method="previous" or "next")`. If `gapfill=False`, it returns a blank image?

Interpolation seems complex enough that it should be left to the user to interpolate the gaps and pass the filled dataset.  I do think it makes sense to offer the option to "fill gap with previous or next map"  since otherwise, some sparse and irregularly-spaced datasets will create visuals that are mostly blank (i.e. see [Nags Head example](http://ncsu-geoforall-lab.github.io/grass-temporal-workshop/))

python/grass/jupyter/timeseries.py
chaedri: `description=_("Date/Time")` should solve this, right?

chaedri:
> Please add tests. (The notebooks are not running automatically (possible bot not implemented yet), so we need at least some tests.)

@wenzeslaus Should this go in a separate PR? None of the other `grass.jupyter` functions/classes have tests yet.

wenzeslaus:
> Should this go in a separate PR?

It is generally better when a change is associated with its tests. Since you are actively working on this in multiple PRs, it might be advantageous to make a separate PR later with tests, but I wouldn't do it unless there is a reason for it. 

> None of the other `grass.jupyter` functions/classes have tests yet.

At last some already do:

https://github.com/OSGeo/grass/tree/main/python/grass/jupyter/testsuite

python/grass/jupyter/timeseries.py
wenzeslaus: If gapfill and method are needed, I would collapse that into `gap_fill={method}` with method being `"previous"`, `"next"`, `None`.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,383 @@
...
+                self._render_blank_layer(filename)
+            else:
+                # Create image file
+                filename = os.path.join(self._tmpdir.name, f"{layer}.png")
+                self._date_filename_dict[date] = filename
+                # Render image
+                img = GrassRenderer(filename=filename, use_region=True, env=self._env)
+                # Fill image background
+                img.d_erase(bgcolor=self._bgcolor)
+                for grass_module, kwargs in self._baselayers.calls:
```
wenzeslaus, line 272: An idea for rendering the background just once: Specify the image to render for a separate instance of GrassRenderer (base layers + background color). Then use copy that image and use copies for the time series renderers, but tell them not to draw the background color (no d.erase), i.e., use the image as is.

The blank layer also needs to be rendered just once. It seems you are reusing the name already, so there is in fact just one file. It's just rendered more that once now.

wenzeslaus (changes requested):

python/grass/jupyter/tests/timeseries_test.py
```diff
@@ -0,0 +1,87 @@
...
+@pytest.mark.skipif(IPython is None, reason="IPython package not available")
+@pytest.mark.skipif(ipywidgets is None, reason="ipywidgets package not available")
+def test_animate_time_slider(space_time_raster_dataset):
+    """Test returns from animate and time_slider are correct object types"""
+    img = gj.TimeSeries(space_time_raster_dataset.name)
+    assert isinstance(img.animate(), IPython.core.display.Image)
+    # This doesn't work
+    #  assert isinstance(img.time_slider(),
+    #                   (IPython.core.display.Image,
+    #                    ipywidgets.widgets.interaction.interactive))
```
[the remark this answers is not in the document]
chaedri: This fails with AssertionError (assert None)

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,435 @@
...
+    def render(self):
+        """Renders image for each time-step in space-time dataset.
+
+        Save PNGs to temporary directory. Must be run before creating a visualization
+        (i.e. time_slider or animate). Can be time-consuming to run with large
+        space-time datasets.
+        """
+
+        # Make base image (background and baselayers)
+        random_name_base = gs.append_random("base", 8) + ".png"
```
wenzeslaus, line 290: I forgot why this has a randomized name. Perhaps a comment explaining it if it is needed.

Maybe not a big deal, but given that the old files are still around perhaps delete them or overwrite them.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,435 @@
...
+                )
+                if self._element_type == "strds":
+                    img.d_rast(map=layer)
+                elif self._element_type == "stvds":
+                    img.d_vect(map=layer)
+                for grass_module, kwargs in self._overlays.calls:
+                    img.run(grass_module, **kwargs)
+                # Add legend if needed
+                if self._legend:
+                    self._render_legend(img)
```
wenzeslaus, line 331: Move this code to separate function, e.g., __render_one_layer_, to have a comparable function to __render_blank_layer_. Maybe same for overlays.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,435 @@
...
+        self._layers_rendered = True
+
+    def time_slider(self, slider_width=None):
+        """Create interactive timeline slider.
+
+        param str slider_width: width of datetime selection slider as a
+                                percentage (%) or pixels (px)
```
wenzeslaus, line 338: Maybe just short description here and then describe the format in a separate paragraph. It needs some more doc because it is a string with a suffix and that's not clear.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,435 @@
...
+
+    For datasets with variable time steps, makes step regular with
+    "gran" method for t.rast.list or t.vect.list then fills in
+    missing layers with previous time step layer.
+
+    :param str timeseries: name of space-time dataset
+    :param str element_type: element type, "stvds" or "strds"
+    :param bool fill_gaps: fill empty time steps with data from previous step
+    """
+    # NEW WAY: Comment in after PR 2258 is merged
```
wenzeslaus, line 47: Note that #2258 is only for _t.rast.list_, the module level code from _t.rast.list_ needs to be ported to _t.vect.list_ (the library code is the shared).

wenzeslaus (changes requested):

python/grass/jupyter/tests/timeseries_test.py
```diff
@@ -0,0 +1,82 @@
...
+    assert kwargs["map"] == "elevation"
+
+
+def test_default_init(space_time_raster_dataset):
+    """Check that TimeSeries init runs with default parameters"""
+    img = gj.TimeSeries(space_time_raster_dataset.name)
+    assert img.timeseries == space_time_raster_dataset.name
+
+
+@pytest.mark.parametrize("fill_gaps", ["False", "True"])
```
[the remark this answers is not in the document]
chaedri: I'm not sure what you mean - `test_renderer_layers` runs twice, once with `fill_gaps=True` and again with `fill_gaps=False`. The `test_fill_none_values` tests replacing `None` values with the previous values and `test_collect_layers` uses `fill_gaps=False` too.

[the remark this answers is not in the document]
wenzeslaus: > I'm not sure what you mean - `test_renderer_layers` runs twice, once with `fill_gaps=True` and again with `fill_gaps=False`.

The issue is that it runs with `fill_gaps="True"` and again with `fill_gaps="False"` so `bool(fill_gaps)` is always True. You need to remove the quotes.

```python
>>> bool("False")  # This is what you have there now.
True
>>> bool(False)
False
>>> False  # This is all you need here.
False
>>> import ast
>>> ast.literal_eval("False")
False
```

> The `test_fill_none_values` tests replacing `None` values with the previous values and `test_collect_layers` uses `fill_gaps=False` too.

Thanks. I didn't notice test_collect_layers=False, so that part is resolved for me now. We have a more focused test with `fill_gaps=False`.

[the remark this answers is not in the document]
chaedri: Ah! Ok, I see now. I removed the quotes.

python/grass/jupyter/tests/timeseries_test.py
```diff
@@ -0,0 +1,82 @@
...
+    # time-intensive)
+    img.baselayer.d_rast(map=space_time_raster_dataset.raster_names[0])
+    img.overlay.d_barscale()
+    img.d_legend()
+    # Render layers
+    img.render()
+    # check files exist
+    # We need to check values which are only in protected attributes
+    # pylint: disable=protected-access
+    for (_date, filename) in img._date_filename_dict.items():
```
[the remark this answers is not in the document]
wenzeslaus: Underscore alone or underscore as a prefix are used to mark unused variables. Underscore alone has two different uses already, so it is best avoided. It also has worse readability since there is no name.

The underscore prefix seems strange at least to me because that's used for the private (protected) attributes.

Hence, my suggestion is `unused_` as a prefix, e.g., `unused_date`.

python/grass/jupyter/timeseries.py
chaedri: Before we change to random, the file was called "base.png". This would create an issue if there was a layer called "base" (which would also create a file called "base.png").

wenzeslaus: I see. Can you please add this to a comment, so it is clear why you are creating a random, sort of unique name?

A comment on how it overwrites existing files when it runs render again would be helpful too. (To answer, e.g., how does it ensure that old file content is not read when new rendering starts?)

python/grass/jupyter/timeseries.py
chaedri: Right! Should I wait for that to be ported? It seems confusing and long to keep one method for t.rast.list and another for t.vect.list.

wenzeslaus: Just keep it as is, only change the comment. It works as is, so I probably won't change t.vect.list in time for 8.2.

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,465 @@
...
+        self._layers_rendered = True
+
+    def time_slider(self, slider_width=None):
+        """Create interactive timeline slider.
+
+        param str slider_width: width of datetime selection slider
+
+        The slider_width parameter sets the width of the slider in the output cell.
+        It should be formatted as a percentage (%) between 0 and 100 of the cell width
+        or in pixels (px). Values should be formatted as strings and include the "%"%
```
wenzeslaus, line 366:
```suggestion
        or in pixels (px). Values should be formatted as strings and include the "%"
```

python/grass/jupyter/timeseries.py
```diff
@@ -0,0 +1,465 @@
...
+        or "px" suffix. For example, slider_width="80%" or slider_width="500px")
```
wenzeslaus, line 367:
```suggestion
        or "px" suffix. For example, slider_width="80%" or slider_width="500px".
```

python/grass/jupyter/tests/conftest.py
```diff
@@ -8,6 +14,59 @@
...
+        times.pop(1)
+        full_names = [f"{name}@PERMANENT" for name in names]
+        full_names.pop(1)
+        names.pop(1)
+        yield SimpleNamespace(
+            name=dataset_name,
+            raster_names=names,
+            full_raster_names=full_names,
+            start_times=times,
+
```
wenzeslaus, line 68:
```suggestion
        )
```

python/grass/jupyter/tests/conftest.py
```diff
@@ -8,6 +14,59 @@
...
+        full_names = [f"{name}@PERMANENT" for name in names]
+        full_names.pop(1)
+        names.pop(1)
+        yield SimpleNamespace(
+            name=dataset_name,
+            raster_names=names,
+            full_raster_names=full_names,
+            start_times=times,
+        )
+
```
wenzeslaus, line 69:
```suggestion


```

python/grass/jupyter/Makefile
```diff
@@ -11,10 +11,11 @@ MODULES = \
 	interact_display \
 	region \
 	render3d \
-    reprojection_renderer \
+  reprojection_renderer \
+	utils \
+	timeseries \
```
wenzeslaus, line 16:
```suggestion
	reprojection_renderer \
	utils \
	timeseries \
```

wenzeslaus (approved):

PR #2012, opened 2021-12-10
Jupyter: Improved error message formatting

chaedri:
* Formatted grass.jupyter errors to format:  f"string{var}"

### DIFF
--- a/python/grass/jupyter/display.py
+++ b/python/grass/jupyter/display.py
@@ -150,7 +150,7 @@ class GrassRenderer:
         grass_module = name.replace("_", ".")
         # Assert module exists
         if not shutil.which(grass_module):
-            raise AttributeError(_("Cannot find GRASS module {}").format(grass_module))
+            raise AttributeError(_(f"Cannot find GRASS module {grass_module}"))
 
         def wrapper(**kwargs):
             # Run module
--- a/python/grass/jupyter/display.py
+++ b/python/grass/jupyter/display.py
@@ -150,7 +150,7 @@ class GrassRenderer:
         grass_module = name.replace("_", ".")
         # Assert module exists
         if not shutil.which(grass_module):
-            raise AttributeError(_(f"Cannot find GRASS module {grass_module}"))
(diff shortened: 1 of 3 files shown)


neteler:
Should this be assigned to Milestone 8.0.0 or later?

wenzeslaus:
Official release of grass.jupyter will with 8.2.

python/grass/jupyter/utils.py
```diff
@@ -60,7 +60,7 @@ def reproject_region(region, from_proj, to_proj):
     proc.stdin = None
     proj_output, stderr = proc.communicate()
     if proc.returncode:
-        raise RuntimeError("reprojecting region: m.proj error: " + stderr)
+        raise RuntimeError(f"Encountered error while running m.proj: {stderr}")
```
wenzeslaus, line 63: This is definitively needed to avoid `TypeError: can't concat str to bytes`.

It still leaves the issue of bytes unresolved. It will show something like: `"Encountered error while running m.proj: b'ERROR: Bad things...'"`, i.e., the error message will be in `b'...'` and it will work well for English/ASCII characters only, but put here a note about a need to decode `stderr` for now.

python/grass/jupyter/display.py
```diff
@@ -150,7 +150,7 @@ def __getattr__(self, name):
         grass_module = name.replace("_", ".")
         # Assert module exists
         if not shutil.which(grass_module):
-            raise AttributeError(_("Cannot find GRASS module {}").format(grass_module))
+            raise AttributeError(_(f"Cannot find GRASS module {grass_module}"))
```
wenzeslaus, line 153: This message and the ones in `render3d.py` are actually translatable (surrounded with `_(...)`), so f-string won't work well here because it will fail to translate. You could   change these to not translated to make it consistent with the rest (where we intentionally ignored the issue of translating to avoid the need to translate messages in experimental code) However, since these are already translatable, you can keep them and that will be consistent with the rest of GRASS code.

wenzeslaus (changes requested):

Issue #2017, opened 2021-12-14
[Feat] wxGUI: Redesigning Map Display status bar combo box into a new Map Display Settings dialog


lindakladivova:
**Is your feature request related to a problem? Please describe.**

Today's state of Map Display statusbar is unsatisfactory especially in terms of the planned Single-Window mode. However, the truth is that it hasn't changed for a very long time, so it is kind of obsolete also in terms of Multi-Window mode.
The main problem is that the status bar is overloaded because in addition to the status bar function, it also performs the function of a local Map Display settings. The current state of Map Display frame is captured in the following picture:

![Screenshot from 2021-12-13 01-29-45](https://user-images.githubusercontent.com/49241681/145948341-c5865b2d-2bb3-40a7-898f-b626d824134e.png)

The status bar at the bottom contains an extensive combo box that allows you to display static text items - such as Coordinates, Map window extent, Computational region extent. However, it also allows you to check several check boxes that relate to the local settings of the given map display. A list of all 10 items sorted by widget types can be seen here:

Text widgets:
![old coords](https://user-images.githubusercontent.com/49241681/145948547-c5afbeee-a975-4e1a-a60c-8dcd60c33cf9.PNG)
![old extent](https://user-images.githubusercontent.com/49241681/145948596-db9399a6-88d0-4324-b8e2-75ef6eec7e03.PNG)
![old computational](https://user-images.githubusercontent.com/49241681/145948699-8f1a9399-4c0b-4d77-848f-d8616e3fce24.PNG)
![old geometry](https://user-images.githubusercontent.com/49241681/145948782-2114035f-4d0e-4d32-919a-618a311b5d9d.PNG)

Check box widgets:
![old show_extent](https://user-images.githubusercontent.com/49241681/145948837-867777b3-e1ae-4764-9c89-20895fabbebd.PNG)
![old align](https://user-images.githubusercontent.com/49241681/145948679-831723d7-11e9-4652-98a7-e6bb06f2469b.PNG)
![old constrain](https://user-images.githubusercontent.com/49241681/145948654-0697ab82-a6d5-4963-bc33-50177328fd39.PNG)
![old projection](https://user-images.githubusercontent.com/49241681/145948758-6a7aa4bc-ac23-4e16-b492-707a7a26cd19.PNG)
+ the check box item is also the Render check box, which is located on the right side of the status bar

Other widgets:
[...]

PR #2031, opened 2021-12-23
wxGUI: Move statusbar checkboxes related to Map Display to a new Map Display Settings dialog

lindakladivova:
This is the starting point of changes that are gonna lead to complete statusbar combobox removal.

I am creating the new Map Display Settings here, accessible from the map display toolbar and move there four statusbar checkbox items - autoRendering, AlignExtent, Resolution and ShowRegion. I have not moved the last checkbox related to the Projection shown in the status bar since it is more complex and needs the special PR.

Screenshot from 2022-01-02 08-57-03

For above-described purpose, a new file mapdisp/settings.py has been proposed. Here we can prepare checkbox items (and other needed widgets) and then fill the new Map Display settings dialog.

**Please feel free to discuss!**

- We need to think about the Map Display base class. I used PreferencesBaseDialog but I am not sure how the "Set to default" option should work. When I tried it in General Settings, I discovered that this option does not work even there, so I have set up the Issue: https://github.com/OSGeo/grass/issues/2032.

- Second thing is the Map Display Settings icon. I have used the base Settings icon from General settings, but it is not final solution. Probably only changing the color is not the solution either as commented here https://github.com/OSGeo/grass/issues/2017. ..

Might be other things, also the form of Map Display settings is an important question, for more info please see https://github.com/OSGeo/grass/issues/2017.

### DIFF
--- a/gui/wxpython/mapdisp/preferences.py
+++ b/gui/wxpython/mapdisp/preferences.py
@@ -255,6 +255,7 @@ class MapDisplayPreferencesDialog(wx.Dialog):
         self._createDisplayPage(parent=self.notebook)
 
         self.btnCancel = Button(self, wx.ID_CANCEL, label="Close")
+        self.btnCancel.SetDefault()
         self.btnCancel.SetToolTip(_("Close dialog"))
 
         self._layout()
@@ -281,34 +282,29 @@ class MapDisplayPreferencesDialog(wx.Dialog):
         panel.SetupScrolling(scroll_x=False, scroll_y=True)
         parent.AddPage(page=panel, text=_("General"))
 
-        border = wx.BoxSizer(wx.VERTICAL)
-
         # General settings
-        gridSizer = wx.GridBagSizer(hgap=4, vgap=4)
+        sizer = wx.BoxSizer(wx.VERTICAL)
 
         # Auto-rendering
         self.autoRendering = ChBRender(panel, self.mapWindowProperties)
-        gridSizer.Add(self.autoRendering.GetWidget(), pos=(0, 0), span=(1, 2))
+        sizer.Add(self.autoRendering.GetWidget(), proportion=0, flag=wx.EXPAND | wx.ALL, border=5)
 
         # Align extent to display size
         self.alignExtent = ChBAlignExtent(panel, self.mapWindowProperties)
-        gridSizer.Add(self.alignExtent.GetWidget(), pos=(1, 0), span=(1, 2))
+        sizer.Add(self.alignExtent.GetWidget(), proportion=0, flag=wx.EXPAND | wx.ALL, border=5)
 
         # Use computation resolution
         self.compResolution = ChBResolution(
             panel, self.giface, self.mapWindowProperties
         )
-        gridSizer.Add(self.compResolution.GetWidget(), pos=(2, 0), span=(1, 2))
+        sizer.Add(self.compResolution.GetWidget(), proportion=0, flag=wx.EXPAND | wx.ALL, border=5)
 
         # Show computation extent
         self.showCompExtent = ChBShowRegion(
             panel, self.giface, self.mapWindowProperties
         )
-        gridSizer.Add(self.showCompExtent.GetWidget(), pos=(3, 0), span=(1, 2))
+        sizer.Add(self.showCompExtent.GetWidget(), proportion=0, flag=wx.EXPAND | wx.ALL, border=5)
 
-        gridSizer.AddGrowableCol(0)
-        border.Add(gridSizer, proportion=1, flag=wx.ALL | wx.EXPAND, border=3)
-
-        panel.SetSizer(border)
+        panel.SetSizer(sizer)
 
         return panel
--- a/gui/wxpython/mapdisp/preferences.py
+++ b/gui/wxpython/mapdisp/preferences.py
@@ -65,6 +65,9 @@ class ChBItem:
         self.mapWindowProperty = self.widget.GetValue()
         self._connect()
 
+    def _onMouseOver(self, event):
+        self.widget.SetToolTip(wx.ToolTip(self.tooltip))
+
 
 class ChBRender(ChBItem):
     """Checkbox to enable and disable auto-rendering."""
@@ -76,9 +79,10 @@ class ChBRender(ChBItem):
             parent=parent, id=wx.ID_ANY, label=_("Enable auto-rendering")
         )
         self.widget.SetValue(self.mapWindowProperty)
-        self.widget.SetToolTip(wx.ToolTip(_("Enable/disable auto-rendering")))
+        self.tooltip = _("Enable/disable auto-rendering")
 
         self.widget.Bind(wx.EVT_CHECKBOX, self._onToggleCheckBox)
+        self.widget.Bind(wx.EVT_MOTION, self._onMouseOver)
         self._connect()
 
     @property
@@ -109,17 +113,15 @@ class ChBAlignExtent(ChBItem):
             label=_("Align region extent based on display size"),
         )
         self.widget.SetValue(self.mapWindowProperty)
-        self.widget.SetToolTip(
-            wx.ToolTip(
-                _(
-                    "Align region extent based on display "
-                    "size from center point. "
-                    "Default value for new map displays can "
-                    "be set up in 'User GUI settings' dialog."
-                )
-            )
+        self.tooltip = _(
+            "Align region extent based on display "
+            "size from center point. "
+            "Default value for new map displays can "
+            "be set up in 'User GUI settings' dialog."
         )
+
         self.widget.Bind(wx.EVT_CHECKBOX, self._onToggleCheckBox)
+        self.widget.Bind(wx.EVT_MOTION, self._onMouseOver)
         self._connect()
 
     @property
@@ -147,17 +149,15 @@ class ChBResolution(ChBItem):
             label=_("Constrain display resolution to computational settings"),
         )
         self.widget.SetValue(self.mapWindowProperty)
-        self.widget.SetToolTip(
-            wx.ToolTip(
-                _(
-                    "Constrain display resolution "
-                    "to computational region settings. "
-                    "Default value for new map displays can "
-                    "be set up in 'User GUI settings' dialog."
-                )
-            )
+        self.tooltip = _(
+            "Constrain display resolution "
+            "to computational region settings. "
+            "Default value for new map displays can "
+            "be set up in 'User GUI settings' dialog."
         )
+
         self.widget.Bind(wx.EVT_CHECKBOX, self._onToggleCheckBox)
+        self.widget.Bind(wx.EVT_MOTION, self._onMouseOver)
         self._connect()
 
     @property
@@ -191,19 +191,16 @@ class ChBShowRegion(ChBItem):
             parent=parent, id=wx.ID_ANY, label=_("Show computational extent")
         )
         self.widget.SetValue(self.mapWindowProperty)
-        self.widget.SetToolTip(
-            wx.ToolTip(
-                _(
-                    "Show/hide computational "
-                    "region extent (set with g.region). "
-                    "Display region drawn as a blue box inside the "
-                    "computational region, "
-                    "computational region inside a display region "
-                    "as a red box)."
-                )
-            )
+        self.tooltip = _(
+            "Show/hide computational "
+            "region extent (set with g.region). "
+            "Display region drawn as a blue box inside the "
+            "computational region, "
+            "computational region inside a display region "
+            "as a red box)."
         )
         self.widget.Bind(wx.EVT_CHECKBOX, self._onToggleCheckBox)
+        self.widget.Bind(wx.EVT_MOTION, self._onMouseOver)
         self._connect()
 
     @property
--- a/gui/wxpython/mapdisp/preferences.py
+++ b/gui/wxpython/mapdisp/preferences.py
@@ -23,7 +23,7 @@ This program is free software under the GNU General Public License
 
 import wx
 import wx.lib.scrolledpanel as SP
-from gui_core.wrap import StaticBox, Button
+from gui_core.wrap import Button
 
 
 class ChBItem:
--- a/gui/wxpython/mapdisp/preferences.py
+++ b/gui/wxpython/mapdisp/preferences.py
@@ -65,9 +65,6 @@ class ChBItem:
         self.mapWindowProperty = self.widget.GetValue()
         self._connect()
 
-    def _onMouseOver(self, event):
-        self.widget.SetToolTip(wx.ToolTip(self.tooltip))
-
 
 class ChBRender(ChBItem):
     """Checkbox to enable and disable auto-rendering."""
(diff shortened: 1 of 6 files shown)


gui/wxpython/mapdisp/settings.py
```diff
@@ -0,0 +1,305 @@
...
+    @property
+    def _property(self):
+        pass
+
+    @_property.setter
+    def _property(self, value):
+        pass
+
+    @property
+    def _propertyChanged(self):
```
petrasovaa, line 65: Shouldn't this be a signal?

gui/wxpython/mapdisp/settings.py
```diff
@@ -0,0 +1,305 @@
...
+from gui_core.wrap import StaticBox
+
+from core.settings import UserSettings
+from gui_core.preferences import PreferencesBaseDialog
+
+
+class ChBItem:
+    """Base class for checkbox settings items"""
+
+    def __init__(self, parent, mapframe):
```
petrasovaa, line 34: Ideally, they shouldn't know about mapFrame and use signals or so.

gui/wxpython/mapdisp/settings.py
```diff
@@ -0,0 +1,305 @@
...
+
+    @property
+    def tooltip(self):
+        pass
+
+    @property
+    def _property(self):
+        pass
+
+    @_property.setter
```
petrasovaa, line 60: This probably shouldn't be private when you provide a setter. Or remove the setter.

gui/wxpython/mapdisp/settings.py
lindakladivova: Ok, I've incorporated giface..

gui/wxpython/mapdisp/settings.py
lindakladivova: Actually, it is a signal, and is returned in the child property. But you are right that we do not need any mapFrame dependency. I have got rid of it.

lindakladivova:
@petrasovaa  Save button and Set to default do not work so far, I am working on it. :-)

lindakladivova:
We need to think carefully about the Map Display Settings icon. We could use some from QGIS:
e.g. 
https://github.com/qgis/QGIS/blob/master/images/themes/default/mActionMapSettings.svg or
https://github.com/qgis/QGIS/blob/master/images/themes/default/mActionOptions.svg
or I would prefer to create the new one and apply the https://github.com/OSGeo/grass/blob/main/gui/icons/grass/map-settings.png settings theme on the monitor icon (https://github.com/OSGeo/grass/blob/main/gui/icons/grass/monitor-create.png). Any opinions on that?

petrasovaa:
Overall, I expected fewer changes. It looks like the SbItem-derived classes could be minimally adjusted and just work. The Python properties are not useful here, maybe except `mapWindowProperty`, they just blow out the code. Also, using underscore means the attribute is private, so you are not supposed to access it from outside (so property doesn't make sense).

I would first keep the statusbar.py untouched, they are used in other standalone map displays.

petrasovaa:
> We need to think carefully about the Map Display Settings icon. We could use some from QGIS: e.g. https://github.com/qgis/QGIS/blob/master/images/themes/default/mActionMapSettings.svg or https://github.com/qgis/QGIS/blob/master/images/themes/default/mActionOptions.svg or I would prefer to create the new one and apply the https://github.com/OSGeo/grass/blob/main/gui/icons/grass/map-settings.png settings theme on the monitor icon (https://github.com/OSGeo/grass/blob/main/gui/icons/grass/monitor-create.png). Any opinions on that?

No strong opinion here, I am fine with the same icon, or perhaps the last solution you suggest.

lindakladivova:
> Overall, I expected fewer changes. It looks like the SbItem-derived classes could be minimally adjusted and just work. The Python properties are not useful here, maybe except `mapWindowProperty`, they just blow out the code. Also, using underscore means the attribute is private, so you are not supposed to access it from outside (so property doesn't make sense).
> 
Ok, I will preserve just mapWindowProperty. 
> I would first keep the statusbar.py untouched, they are used in other standalone map displays.
Ou, you are right. I have not realized that. But I would keep the ChBRender in settings.py because it has the same form as other checkboxes and I think it is more clear to have it together with settings. We might also want to have the auto-render checkbox in Map Display settings.

lindakladivova:
> > Overall, I expected fewer changes. It looks like the SbItem-derived classes could be minimally adjusted and just work. The Python properties are not useful here, maybe except `mapWindowProperty`, they just blow out the code. Also, using underscore means the attribute is private, so you are not supposed to access it from outside (so property doesn't make sense).
> 
> Ok, I will preserve just mapWindowProperty.
> 
> > I would first keep the statusbar.py untouched, they are used in other standalone map displays.
> 
> Ou, you are right. I have not realized that. But I would keep the ChBRender in settings.py because it has the same form as other checkboxes and I think it is more clear to have it together with settings. We might also want to have the auto-render checkbox in Map Display settings.
 
Mmmm as I think about it, I let the statusbar.py entirely as it was - it is probably the best solution now even though we duplicate the code.

gui/wxpython/mapdisp/settings.py
```diff
@@ -0,0 +1,276 @@
...
+        pass
+
+    def _setValue(self, value):
+        self.SetValue(value)
+
+    def SetValue(self, value):
+        self._disconnect()
+        self.mapWindowProperty = value
+        wx.CheckBox.SetValue(self, value)
+        self._connect()
```
petrasovaa, line 65: I don't see the need for both `_setValue` and `SetValue`.

Shouldn't SetValue just set the checkbox, not the property? Setting the property would be done in `_onToggleCheckBox`.

petrasovaa (changes requested):

petrasovaa:
> @petrasovaa Save button and Set to default do not work so far, I am working on it. :-)

I don't think that's what we want. It should behave the same way as the widgets in statusbar, so when you check it, it immediately changes the property. No save buttons there. Set to default may make sense but is perhaps not needed.

gui/wxpython/mapdisp/settings.py
```diff
@@ -0,0 +1,276 @@
...
+        Shows or hides according to checkbox value.
+        """
+        super()._onToggleCheckBox(event)
+
+        # redraw map if auto-rendering is enabled
+        if self._properties.autoRender:
+            self.giface.updateMap.emit(render=False)
+
+
+class MapDisplayPreferencesDialog(PreferencesBaseDialog):
```
petrasovaa, line 210: I am not sure we should be using PreferencesBaseDialog at all, we are not setting GUI settings here.

lindakladivova: I was thinking that these settings could replace the same settings in Map Display. If you test this PR it looks very intuitive, you just set up something and if you want, you can save it and use it later on.

gui/wxpython/mapdisp/settings.py
```diff
@@ -0,0 +1,276 @@
...
+        mapWindowProperties,
+        label=_("Align region extent based on display size"),
+    ):
+        tooltip = _(
+            "Align region extent based on display "
+            "size from center point. "
+            "Default value for new map displays can "
+            "be set up in 'User GUI settings' dialog."
+        )
+        ChBItem.__init__(self, parent, mapWindowProperties, label, tooltip)
```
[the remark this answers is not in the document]
lindakladivova: The reason was the SbRender class that could inherit from ChBRender but has a different label. But you are right if label and tooltip would be set in the derived classes that would be better.

gui/wxpython/mapdisp/settings.py
lindakladivova: I am still thinking about it. You are probably right that It should influence only the given mapdisplay or statusbar. But then I am not sure about the Projection page. We agreed to move it to Map Display settings, but it is also in the form of global settings. So generally speaking, do we want to get rid of the global settings a little bit? Or do we just want to add some functions from GUI settings in the form of Local settings to Map Display settings?...

petrasovaa: You could potentially set the defaults from there, but I think it would be confusing and more work. Regarding the projection, I would remove it from general settings altogether. I would keep the rest there. Ultimately we are trying to reduce the amount of work.

lindakladivova:
We have only one button Close now, so the the way how it works is the same as in the status bar - just local settings.
Výstřižek

petrasovaa:
The tooltips don't work because the checkbox is not a child of the staticbox, so either remove the staticbox or change the parent of the checkboxes. Perhaps the staticbox might not be needed here.

gui/wxpython/mapdisp/statusbar.py
```diff
@@ -380,7 +380,7 @@ def GetPosition(self):
         return self.position
 
     def GetWidget(self):
-        """Returns underlaying winget.
+        """Returns underlaying widget.
```
petrasovaa, line 383: If you are fixing the typo, fix also underlying

lindakladivova: :-) right. I did not notice.

lindakladivova:
> The tooltips don't work because the checkbox is not a child of the staticbox, so either remove the staticbox or change the parent of the checkboxes. Perhaps the staticbox might not be needed here.

I would keep the Customize Map Display static box because there is the second static box below for Customize status bar with radio buttons planned. So, it would be better to change parent of the checkboxes.

petrasovaa:
> > The tooltips don't work because the checkbox is not a child of the staticbox, so either remove the staticbox or change the parent of the checkboxes. Perhaps the staticbox might not be needed here.
> 
> I would keep the Customize Map Display static box because there is the second static box below for Customize status bar with radio buttons planned. So, it would be better to change parent of the checkboxes.

So what are the tabs for then?

lindakladivova:
@petrasovaa 

> > > The tooltips don't work because the checkbox is not a child of the staticbox, so either remove the staticbox or change the parent of the checkboxes. Perhaps the staticbox might not be needed here.
> > 
> > 
> > I would keep the Customize Map Display static box because there is the second static box below for Customize status bar with radio buttons planned. So, it would be better to change parent of the checkboxes.
> 
> So what are the tabs for then?

It is prepared for the Projection settings. You wrote: Regarding the projection, I would remove it from general settings altogether. So you had the notion that the Projection page will be entirely removed from the whole GRASS or just from GUI settings?

I had the feeling from our video call with you,  Vashek and Martin that we plan to have a Projection page in Map Display settings (which sets a status bar CRS) and also mentions a global location coordinate system because it is something not visible anywhere and this place would be suitable. But maybe I understand it wrong.

Or maybe you thought that the Projection page in Map Display settings would be actually the Status bar page and we would have the radio buttons for status bar settings here and also the status bar projection. And maybe the global location projection could then be a part of the Map Display page. Probably this paragraph is a better option, or what did (do) you think?

petrasovaa:
> Or maybe you thought that the Projection page in Map Display settings would be actually the Status bar page and we would have the radio buttons for status bar settings here and also the status bar projection

I din't think about it too much yet, but perhaps this would make most sense.



> And maybe the global location projection could then be a part of the Map Display page.

Not quite sure what you mean here.

petrasovaa:
> Or maybe you thought that the Projection page in Map Display settings would be actually the Status bar page and we would have the radio buttons for status bar settings here and also the status bar projection

I din't think about it too much yet, but perhaps this would make most sense.



> And maybe the global location projection could then be a part of the Map Display page.

Not quite sure what you mean here.

lindakladivova:
> > Or maybe you thought that the Projection page in Map Display settings would be actually the Status bar page and we would have the radio buttons for status bar settings here and also the status bar projection
> 
> I din't think about it too much yet, but perhaps this would make most sense.

Good, then I will remove the static box widget and let the notebook prepared for the Status bar page :-).
> 
> > And maybe the global location projection could then be a part of the Map Display page.
> 
> Not quite sure what you mean here.

I mean that somebody set the CRS e.g. 5514 and then forgot what CRS he set. So, the Map Display settings dialog is, in my opinion, a good place to show it. Just as a non-editable text widget. There were also suggestions in the surveys that users would assume that they read the set CRS somewhere. They mentioned two ways - adding it in the context menu in Location tree mode (probably something like the Info context menu option) or directly as a part of the Location node text. But I think that for the beginning it would be also helpful to add it to Map Display settings.

lindakladivova:
Now, it looks as follows:
Výstřižek

lindakladivova:
No it looks as follows:
Výstřižek

petrasovaa (changes requested):
I would at this point remove the checkboxes from the map Display statusbar.

lindakladivova:
> Now it looks as follows: Výstřižek

After changes, it looks exactly the same.

gui/wxpython/mapdisp/preferences.py
```diff
@@ -0,0 +1,329 @@
+"""
+@package mapdisp.settings
```
petrasovaa, line 2: I just noticed the `settings` here but the file is named `preferences.py`. I would perhaps go with `settings` here (not strong opinion), so please rename the class as well.

petrasovaa (changes requested):
Almost there! Maybe also make the dialog smaller, we can enlarge it once we need it.

gui/wxpython/mapdisp/preferences.py
lindakladivova: Well, I changed it to preferences.py because in gui_core it is also called preferences.py. But anyway, settings.py could be anything (very general), I think preferences.py more points on the purpose. I would let it as it is, and change just the @package name if it okey. :-)

lindakladivova:
> Almost there! Maybe also make the dialog smaller, we can enlarge it once we need it.
Výstřižek

gui/wxpython/mapdisp/preferences.py
petrasovaa: I guess my point was to avoid preferences, because that's used for something else, this is not actually setting any preferences, you changing the actual behavior for specific mp display. Settings is used elsewhere, so that may not be the right choice either. How about rename this to properties.py and use *property* for naming of the classes, because that's the terminology we use for this already?

lindakladivova: Right, I think properties are better than preferences. Good idea! :-)

lindakladivova:
I have noticed one thing and have not succeeded to find out the solution. When you open the dialog for the second time (or any other time) it is smaller (a bit shrank). And the self.GetBestSize() is different. Probably not too important but I am just trying to figure out why it is like that...

lindakladivova:
> > I have noticed one thing and have not succeeded to find the solution. When you open the dialog for the second time (or any other time) it is smaller (a bit shrank). And the self.GetBestSize() is different. Probably not too important but I am just trying to figure out why it is like that...
> 
> Hard to say, sizing is always a pain... You can try not using scrolled panel, that could simplify the problem little bit.

It did the same but maybe worse because theres no scrolling so it is a bit cut off. Generally I think it is rather a detail and when I make a dialog a bit bigger in terms of the height I think nobody will notice. What do you think? I could set up the height of 250.

petrasovaa (approved):

PR #2035, opened 2021-12-25
i18N: fix long dash confusing xgettext

neteler:
Fixes

```
Generating grassmods...
xgettext --keyword=_ --keyword=n_:1,2 -cGTC -o ./templates/grassmods.pot `find ../ -name '*.c' | grep -v '../lib' | xargs grep -l "_(\"\|n_(\""`
xgettext: Non-ASCII string at ../imagery/i.smap/openfiles.c:43.
          Please specify the source encoding through --from-code.
make: *** [Makefile:52: pot] Error 1
```

This (probably) requires the update of the messages on Transifex.

### DIFF
--- a/imagery/i.cca/main.c
+++ b/imagery/i.cca/main.c
@@ -137,7 +137,7 @@ int main(int argc, char *argv[])
 
     err = I_sort_signatures_by_semantic_label(&sigs, &refs);
     if (err)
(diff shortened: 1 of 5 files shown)


marisn (approved):
My bad. I always forget to use hyphen where en or em dash should be used.

PR #2073, opened 2022-01-07
Update alpine version in Dockerfile

mmacata:
This PR suggests to update the alpine version from 3.12 to 3.15 in the alpine Dockerfile. It was released in November.
As pdal is available via alpine package manager now, the `mundialis/docker-pdal` is not needed anymore.

Two questions / problems appear:

1. the package `proj-datumgrid` does not exist anymore. When comparing the content and version of the differen alpine versions here: https://pkgs.alpinelinux.org/contents?branch=v3.12&name=proj-datumgrid&arch=x86_64&repo=community, it shows:
=> 3.12. contains proj 7.0.1-r0
=> 3.15. contains proj 8.2.0-r0 - is proj-datumgrid not needed here any more?

2. The test stage fails where `r.learn.ml` is installed:
```
Step 34/40 : RUN grass --tmp-location EPSG:4326 --exec g.extension extension=r.learn.ml
 ---> Running in 639e3fcf8360
Starting GRASS GIS...
Creating new GRASS GIS location <tmploc>...
Cleaning up temporary files...
Executing <g.extension extension=r.learn.ml> ...
Fetching <r.learn.ml> from GRASS GIS Addons repository (be patient)...
Compiling...
Traceback (most recent call last):
  File "/tmp/grass8-root-1/tmpac7nraye/r.learn.ml/scripts/r.learn.ml", line 414, in <module>
    from grass.pygrass.gis.region import Region
  File "/usr/local/grass80/etc/python/grass/pygrass/gis/__init__.py", line 19, in <module>
    import grass.lib.gis as libgis
  File "/usr/local/grass80/etc/python/grass/lib/gis.py", line 581, in <module>
    struct_timespec._fields_ = [
ValueError: number of bits invalid for bit field
make: *** [/usr/local/grass80/include/Make/Html.make:14: r.learn.ml.tmp.html] Error 1
ERROR: Compilation failed, sorry. Please check above error messages.
Execution of <g.extension extension=r.learn.ml> finished.
Cleaning up temporary files...
The command '/bin/sh -c grass --tmp-location EPSG:4326 --exec g.extension extension=r.learn.ml' returned a non-zero code: 1
```

This happens for building with `main` branch as well as `releasebranch_8_0`. As the test stage is ommited when building with github actions (as this stage is not needed for the final image), it can be tested only locally, e.g. with
```
docker build --file docker/alpine/Dockerfile --tag grass-py3-pdal:latest-alpine .
```

[...]

PR #2079, opened 2022-01-11
ctypesgen: Remove unnamed zero bit sized structure members

nilason:
Remove unnamed zero bit sized structure members  from ctypesgen generated python files, as they are not supported by Ctypes.

Addresses issue reported in #2073

What I initially thought to be a trivial fix, turned out not being so.
E.g. to be able to evaluate the expression
```
((8 * (sizeof(time_t) - sizeof(c_long))) * (1234 == 4321))
```
, `sizeof`, `time_t` and `c_long` need to be known (imported/defined). This was not possible within ctyepesgen, if we want code for a more general use. All these functions/definitions are known in the generated python file.
My solution was therefore to inject a "self healing" code in the ctypesgen generated python file, removing unneeded and unsupported zero bitsized structure members (and in the process this self healing code is itself removed, think Mission Impossible :).

### DIFF
--- a/python/libgrass_interface_generator/README.md
+++ b/python/libgrass_interface_generator/README.md
@@ -62,6 +62,164 @@ file generation.
 
 ```
 
+#### Ctypes "unnamed structure member with 0 bit size"-patch
+
+Using unnamed zero bit sized structure members, e.g.:
+
+```c
+struct timespec {
+    time_t tv_sec;
+    int :8*(sizeof(time_t)-sizeof(long))*(__BYTE_ORDER==4321);
+    long tv_nsec;
+    int :8*(sizeof(time_t)-sizeof(long))*(__BYTE_ORDER!=4321);
+};
+```
+
+is not supported by Ctypes. Ctypesgen generates this code to:
+
+```py
+struct_timespec._fields_ = [                         
+    ('tv_sec', time_t),                           
+    ('unnamed_1', c_int, ((8 * (sizeof(time_t) - sizeof(c_long))) * (1234 == 4321))),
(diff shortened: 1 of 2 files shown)


neteler:
@nilason may this PR be merged?

nilason:
> @nilason may this PR be merged?

I think so, didn't plan any changes. 
Should wait with backport till after 8.0.0 release though.

nilason:
I see there is problem with Windows on CI. Let's see if rebasing make any difference.

neteler:
@nilason may this PR be merged into main branch?

nilason:
Merged. Although this fix is designed to not kick in if not needed, we should wait a bit on back porting.

neteler:
@nilason Do you want to see this backported 8.0.1 or 8.0.2?

nilason:
Depends a bit on what will happen with #2073, if that one goes for 8.0.1 then this one might too.

mmacata:
In #2073 there is still the issue with the large `proj-datumgrid` package (559.58 MB) which I don't want to include in the Dockerimage. So it is not urgent from my side.

nilason:
Ok, I'd suggest let's bump the milestone for both to 8.0.2.

neteler:
> Merged. Although this fix is designed to not kick in if not needed, we should wait a bit on back porting.

Backport to 8.0.2 now done.

nilason:
> > Merged. Although this fix is designed to not kick in if not needed, we should wait a bit on back porting.
> 
> Backport to 8.0.2 now done.

@neteler Thanks!

PR #2082, opened 2022-01-12
scripts/g.extension.all: fix reinstall multi-addons

tmszi:
**Describe the bug**
Reinstall multi-addon e.g. wx.metadata (which contains multiple individual modules), doesn't work.

**To Reproduce**

1. Install wx.metadata addon `g.extension wx.metadata`
2. Reinstall all installed addons `g.extension.all -f`
3. See error

```
------------------------------------------------------------
Reinstalling extension <v.info.iso>...
------------------------------------------------------------
WARNING: Extension <v.info.iso> already installed. Re-installing...
Fetching <v.info.iso> from GRASS GIS Addons repository (be patient)...
svn: E170000: URL 'https://github.com/OSGeo/grass-addons/branches/grass8/src/vector/v.info.iso' doesn't exist
ERROR: GRASS Addons <v.info.iso> not found
ERROR: Unable to process extension:v.info.iso
------------------------------------------------------------
Reinstalling extension <t.info.iso>...
------------------------------------------------------------
WARNING: Extension <t.info.iso> already installed. Re-installing...
Fetching <t.info.iso> from GRASS GIS Addons repository (be patient)...
svn: E170000: URL 'https://github.com/OSGeo/grass-addons/branches/grass8/src/temporal/t.info.iso' doesn't exist
ERROR: GRASS Addons <t.info.iso> not found
ERROR: Unable to process extension:t.info.iso
------------------------------------------------------------
Reinstalling extension <r.info.iso>...
------------------------------------------------------------
WARNING: Extension <r.info.iso> already installed. Re-installing...
Fetching <r.info.iso> from GRASS GIS Addons repository (be patient)...
svn: E170000: URL 'https://github.com/OSGeo/grass-addons/branches/grass8/src/raster/r.info.iso' doesn't exist
ERROR: GRASS Addons <r.info.iso> not found
ERROR: Unable to process extension:r.info.iso
------------------------------------------------------------
Reinstalling extension <m.csw.update>...
------------------------------------------------------------
WARNING: Extension <m.csw.update> already installed. Re-installing...
Fetching <m.csw.update> from GRASS GIS Addons repository (be patient)...
svn: E170000: URL 'https://github.com/OSGeo/grass-addons/branches/grass8/src/misc/m.csw.update' doesn't exist
ERROR: GRASS Addons <m.csw.update> not found
ERROR: Unable to process extension:m.csw.update
------------------------------------------------------------
Reinstalling extension <g.gui.metadata>...
------------------------------------------------------------
[...]

PR #2080, opened 2022-01-11
bump version to 8.1

landam:

PR #2083, opened 2022-01-12
.github/workflows/osgeo4w.yml: bump GRASS GIS executable version to 8.1

tmszi:
Fixes OSGeo4W / windows-2019 build and tests CI error mentioned here https://github.com/OSGeo/grass/pull/2082#issuecomment-1011250868. Related with this commit https://github.com/OSGeo/grass/commit/029cd8ea8e9c98b575f57f148b1552c8821a0494.

### DIFF
--- a/.github/workflows/osgeo4w.yml
+++ b/.github/workflows/osgeo4w.yml
@@ -36,10 +36,10 @@ jobs:
         run: C:\msys64\usr\bin\bash.exe -l (''+(Get-Location)+'\.github\workflows\build_osgeo4w.sh') (Get-Location)
 
       - name: Test executing of the grass command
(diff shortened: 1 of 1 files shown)


wenzeslaus:
Maybe we should open an issue for each of those changes which are needed besides #2080 and treat them as bugs so we can aim at having only one place which needs changing.

nilason:
https://github.com/OSGeo/grass/search?q=grass80

neteler:
> Maybe we should open an issue for each of those changes which are needed besides #2080 and treat them as bugs so we can aim at having only one place which needs changing.

Addressed in #2084 (sorry for having merged #2080 without thinking...).

neteler (approved):

nilason (approved):

PR #2084, opened 2022-01-12
GRASS GIS 8 dev: version bump to grass81

neteler:
Fixes #2080 and completes 029cd8ea8e9c98b575f57f148b1552c8821a0494. In addition to #2083.

- grass80 -> grass81 updates in HTML pages, Dockerfiles and other files across the code tree
- minor cosmetics

(note: the milestone isn't exactly 8.2.0 but it needs to be redone in the upcoming `release_branch_8_2` branch)


docker/debian/Dockerfile
```diff
@@ -209,7 +209,7 @@ ENV LC_ALL "en_US.UTF-8"
 ENV GRASS_SKIP_MAPSET_OWNER_CHECK 1
 
 # Create generic GRASS GIS lib name regardless of version number
-RUN ln -sf /usr/local/grass80 /usr/local/grass
+RUN ln -sf /usr/local/grass81 /usr/local/grass
```
wenzeslaus, line 212: Candidate for a bigger change? Something like `ln -sf /usr/local/grass*` would work too, no? There is only one GRASS in that Docker image. (Same for all Dockerfiles.)

Or ever bigger change, why it is installed to grassXY in the first place? I have `/usr/local/share/zsh/` and `/usr/local/lib/R` there (although I have `/usr/local/lib/python2.7/` there too). Obviously, at least this piece is outside of scope of this PR and thus just a side note.

general/g.message/g.message.html
```diff
@@ -47,7 +47,7 @@ <h2>NOTES</h2>
 
 <h3>Usage in Python scripts</h3>
 
-<a href="https://grass.osgeo.org/grass80/manuals/libpython/">GRASS
+<a href="https://grass.osgeo.org/grass81/manuals/libpython/">GRASS
```
wenzeslaus, line 50: We have grass-stable, so why not grass-development, grass-preview (following naming on the website), or grass-main (following branch name)? (It makes sense to me to merge this PR without/before implementing this suggestion.)

wenzeslaus (approved):
All look good to me including the changes to something else than 8.1. I especially appreciate the fix of `grass.itc.it` :-) and other outdated links.

docker/debian/Dockerfile
neteler: > Candidate for a bigger change? Something like `ln -sf /usr/local/grass*` would work too, no? There is only one GRASS in that Docker image. (Same for all Dockerfiles.)

We had that already in the past and it failed (don't recall well why, perhaps it let to a recursive effect somehow). See f1a46d63fb3628d7ea506a146390f21d681c6d0d for the revert in last summer.

> Or ever bigger change, why it is installed to grassXY in the first place? I have `/usr/local/share/zsh/` and `/usr/local/lib/R` there (although I have `/usr/local/lib/python2.7/` there too). Obviously, at least this piece is outside of scope of this PR and thus just a side note.

The point is: how to then have distinct GRASS versions installed? But perhaps for another PR...

general/g.message/g.message.html
neteler: Indeed we also have 
https://grass.osgeo.org/grass-devel/manuals/

At time it doesn't point to anything useful but once this PR is merged we will have grass81 on the server shortly.

So I'll update this PR accordingly.

neteler: Done in 5af69faf5890248607d33620a230ffc864e72434

nilason (approved):
Looks good to me!

docker/debian/Dockerfile
nilason: > > Candidate for a bigger change? Something like `ln -sf /usr/local/grass*` would work too, no? There is only one GRASS in that Docker image. (Same for all Dockerfiles.)
> 
> We had that already in the past and it failed (don't recall well why, perhaps it let to a recursive effect somehow). See [f1a46d6](https://github.com/OSGeo/grass/commit/f1a46d63fb3628d7ea506a146390f21d681c6d0d) for the revert in last summer.
> 
> > Or ever bigger change, why it is installed to grassXY in the first place? I have `/usr/local/share/zsh/` and `/usr/local/lib/R` there (although I have `/usr/local/lib/python2.7/` there too). Obviously, at least this piece is outside of scope of this PR and thus just a side note.
> 
> The point is: how to then have distinct GRASS versions installed? But perhaps for another PR...

Agree this is a question for another PR, but we have already stripped the version from grass executable. For multiple version-installations alternative `--prefix` and `--exec-prefix` may be set.

wenzeslaus: > > ...Something like `ln -sf /usr/local/grass*` would work too, no?...
> 
> ...it failed (don't recall well why, perhaps it let to a recursive effect somehow). See [f1a46d6](https://github.com/OSGeo/grass/commit/f1a46d63fb3628d7ea506a146390f21d681c6d0d)...

Hm, I don't see a PR or an error related to that, so hard to say. It worked in other places at least for the executable. I have seen (less general) `grass??` elsewhere, so perhaps you or someone else successfully addressed the issue, whatever it was, this way.

> > ...why it is installed to grassXY in the first place? I have `/usr/local/share/zsh/` and `/usr/local/lib/R` there...
> 
> The point is: how to then have distinct GRASS versions installed?

Definitively another PR. But what is the use case when you need to have two GRASS versions installed side by side like this? Certainly, it is not these Dockerfiles. I would think that developers don't install the software with sudo to a system location like this either, or? PATH/environment modification and aliases can be used to have custom setups of multiple versions of the same software. Environment modification is exactly what tools like Python virtual environments are using.

general/g.message/g.message.html
wenzeslaus: Great! Thanks.

PR #2085, opened 2022-01-12
v.to.db: fix option=query

metzm:
with `option=query`, query columns can not only be columns, but also SQL expressions like count(x) or avg(x). In these cases the check for column type fails, and it is up to the user to first create a new column of the appropriate type for storing results, before running `v.to.db`.


veroandreo (approved):
Tested with the example added in v.rast.stats :+1:

metzm:
To be backported: https://github.com/OSGeo/grass/commit/3f7bd7eb5ec4c69a4c15d72bfb205693f3e747bc

neteler:
Backport done

PR #2086, opened 2022-01-12
odbc driver: fix pointer-sign compiler warrnings

nilason:
Just happened to see a bunch of `-Wpointer-sign` compiler warnings on the CI Travis Clang build log.
I guess nothing critical, but good to fix.


wenzeslaus:
> Just happened to see a bunch of...

BTW, do you have access to _Security > Code scanning alerts_?

> ...on the CI Travis Clang build log.

Also, feel free to enable more checks in the gcc.yml workflow or add a Clang one (marking enabled warnings as error to prevent introduction of new ones).

nilason:
> > Just happened to see a bunch of...
> 
> BTW, do you have access to _Security > Code scanning alerts_?
> 

Yes, there’s some work waiting there… :)

> > ...on the CI Travis Clang build log.
> 
> Also, feel free to enable more checks in the gcc.yml workflow or add a Clang one (marking enabled warnings as error to prevent introduction of new ones).

It’s on my todo list. 👍

nilason:
So, with latest push: no more compiler warnings on Travis!
Please review.

wenzeslaus (approved):
+1 Thank you.

Many these are clear, just the right type is needed. The SQLCHAR signed/unsigned char issues seems to be solved by casts to whatever is needed in other projects, too (PHP, unixODBC), although some discussion in Postgres seems to argue for more caution, but for ODBC it says:

> We're bound to need some casts, because the ODBC API tends to use
unsigned char pointers while all the libc string functions use signed. 

https://github.com/php/php-src/commit/aa58db723221ec891d4432621003bfa55dc15edf
https://github.com/lurcher/unixODBC/commit/78fb6bc3b4194d477a938318b3303232d204e095
https://postgrespro.com/list/thread-id/1985760

PR #2100, opened 2022-01-16
mkhtml.py: get date from gitlog

ninsbl:
Trying to get the latest change and commit from gitlog.
Requires git and is not yet tested more broadly.
Please test. Feedback on HTML formating is very welcome!

PR #2113, opened 2022-01-20
utils/mkhtml.py: fix get core modules commit from local git repo and remote GitHub API server

tmszi:
Fix get core modules commit from local git repository and remote GitHub API server. Bug related with PR #2100.

### DIFF
--- a/utils/mkhtml.py
+++ b/utils/mkhtml.py
@@ -68,10 +68,16 @@ if not PY2:
 grass_version = os.getenv("VERSION_NUMBER", "unknown")
 trunk_url = ""
 addons_url = ""
+grass_git_branch = "main"
 if grass_version != "unknown":
     major, minor, patch = grass_version.split(".")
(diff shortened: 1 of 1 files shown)


ninsbl:
Checking the manuals here:
https://grass.osgeo.org/grass78/manuals/r.info.html
https://grass.osgeo.org/grass80/manuals/r.info.html
They seem to pick up the latest commit to the branch, not the latest commit to a module...
Might be OK, just wnded to give a heads up...

tmszi:
> Checking the manuals here: https://grass.osgeo.org/grass78/manuals/r.info.html https://grass.osgeo.org/grass80/manuals/r.info.html They seem to pick up the latest commit to the branch, not the latest commit to a module... Might be OK, just wnded to give a heads up...

Yes you are right. I'm fixed this bug with this PR, but it is not yet updated on the server side as mentioned above.

Issue #2126, opened 2022-01-26
v.distance: required paramters do not allow creation of vector map with lines


ninsbl:
**Describe the bug**
Combinaion of required paramters in v.distance makes it impossible to produce a output vector lines map.
Likely introduced with #601 (which I reviewd).

Now, `upload` is a required parameter. When I add it, I get the following error:
```
ERROR: Option <upload> requires at least one of <-p>, <table> or <columns>
```
When I leave it out, I get the error message that the required parameter `upload` is missing.
 
**To Reproduce**
Try a version of the "Create a new vector map" example from the manual.
```
v.distance from=pnt to=ln out=connections upload=dist column=dist
```
Also the other example
```
v.distance from=pnt to=ln out=connections upload=cat,dist column=to_cat,dist table=connections
```
Does not work because `output` and `table` are mutually exclusive.

**Expected behavior**
The module allows creation of vector maps and examples from the manual work.

**System description (please complete the following information):**

- Operating System: Ubuntu 18.04
- GRASS GIS version: 8.1

**Additional context**
Tests should be added that cover these issues.

Personally, I think we should not ship a release with this issue, so I take the liberty to flag it as a blocker. Please feel free to change the label if you disagree.
Will try if I can fix it fast.

wenzeslaus:
On the other hand, we do plan to release a bugfix release 8.0.1...

PR #1265, opened 2021-01-21
Fix compiler warnings, part 4

nilason:
Fixes compiler warnings:

- -Wpointer-compare
- -Wpointer-sign
- -Wtautological-pointer-compare

Fourth part addressing #1247.

Modules / code parts directly affected:

- lib/cairodriver
- lib/db/dbmi_client
- lib/dspf
- lib/gis
- lib/pngdriver
- lib/raster3d

PR #601, opened 2020-05-08
v.distance: Linear matrix becomes default, new -s flag for square matrix [API change]

mlennert:
This PR is commits a patch that has been in trac for years ([#3638](https://trac.osgeo.org/grass/ticket/3638)). **It changes v.distances default API**.
Up to now, when using the -pa flag for printing the output of a complete matrix, this matrix would default to a square matrix. This is not always the easiest form for reuse, and it also means that you cannot use -pa with more than one upload option.
This PR makes the linear matrix default for all output forms, and allows multiple upload options even with -a. A new -s allows the user to request a square matrix when only one upload option is requested.

PR #2127, opened 2022-01-26
v.distance: relax parser rules

ninsbl:
This is a basic quick-fix for #2126

Tests should be added...

### DIFF
--- a/vector/v.distance/main.c
+++ b/vector/v.distance/main.c
@@ -266,7 +266,8 @@ int main(int argc, char *argv[])
     opt.to_field->guidependency = G_store(opt.to_column->key);
 
     G_option_required(opt.out, opt.table, flag.print, opt.column, NULL);
-    G_option_exclusive(opt.out, opt.table, flag.print, NULL);
+    G_option_exclusive(opt.out, flag.print, NULL);
+    G_option_exclusive(opt.table, flag.print, NULL);
     G_option_requires(opt.upload, flag.print, opt.table, opt.column, NULL);
     G_option_requires(opt.column, opt.upload, NULL);
     G_option_requires(flag.print, opt.upload, NULL);
--- a/vector/v.distance/main.c
+++ b/vector/v.distance/main.c
@@ -266,8 +266,7 @@ int main(int argc, char *argv[])
     opt.to_field->guidependency = G_store(opt.to_column->key);
 
     G_option_required(opt.out, opt.table, flag.print, opt.column, NULL);
-    G_option_exclusive(opt.out, flag.print, NULL);
-    G_option_exclusive(opt.table, flag.print, NULL);
+    G_option_excludes(flag.print, opt.out, opt.table, NULL);
     G_option_requires(opt.upload, flag.print, opt.table, opt.column, NULL);
     G_option_requires(opt.column, opt.upload, NULL);
     G_option_requires(flag.print, opt.upload, NULL);


vector/v.distance/main.c
```diff
@@ -266,7 +266,8 @@ int main(int argc, char *argv[])
     opt.to_field->guidependency = G_store(opt.to_column->key);
 
     G_option_required(opt.out, opt.table, flag.print, opt.column, NULL);
-    G_option_exclusive(opt.out, opt.table, flag.print, NULL);
+    G_option_exclusive(opt.out, flag.print, NULL);
+    G_option_exclusive(opt.table, flag.print, NULL);
```
wenzeslaus, line 270: Wasn't the idea in the original #601 code _G_option_excludes_ (when first is present, none of the other options should be present)?

```c
G_option_excludes(flag.print, opt.out, opt.table, NULL);
```

wenzeslaus (approved):
Thanks for quickly addressing this.

The original PR #601 changes the rules, but does not comment on in the the description nor the commit message and with all these options it is hard for me to get a full picture fast enough.

However, the current code here follows the manual and fits with the state before #601 where `opt.table` actually requires `opt.out`:

```c
G_option_requires_all(opt.table, opt.out, opt.upload, NULL);
```

I think it can be rewritten using one function call, but that's optional. Perhaps for a later PR with test.

vector/v.distance/main.c
ninsbl: Yes, yu are right. Implemented in 3f7f04f

ninsbl:
Unfortunately, now I get a `segmentation fault` error...

ninsbl:
OK. Now I ran it through `gdb`:

```
Starting program: /opt/src/grass_ninsbl/dist.x86_64-pc-linux-gnu/bin/v.distance --overwrite --verbose from=railroads@PERMANENT to=railroads@PERMANENT output=t dmin=0.1 upload=dist table=t
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
WARNING: Vector map <t> already exists and will be overwritten
Using native format
Finding nearest features...
 100%

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7f2518c in db_start_driver (name=0x0) at start.c:97
97	    if (name[0] == '\0') {
```


So, the culprit seems:
https://github.com/OSGeo/grass/pull/1265
in particular:
https://github.com/OSGeo/grass/blob/882597123599767abd02f1486724ea02914372f1/lib/db/dbmi_client/start.c#L97

Replacing `name[0]` with `name` makes `v.distance` produce output again. However, it was changed on purpose, so, I wonder if a different solution would be needed. Any idea @nilason ?
Changing the line to:
```
    if (name == NULL) {
```
both removes the compiler warning and makes the module run again...
Should I add the latter approach to the PR?

wenzeslaus:
Maybe `if(name == NULL || name[0] == '\0') {` is needed? Check whether or not the pointer is null point and then check that the provided string is actually filled. That would be equivalent with Python `if not name` which works like `if name is None or len(name) == 0`. The intention in the original code seems to be to check if the string is empty. Well, that's at least the understanding of the compiler producing the warning which says "did you mean to dereference the pointer?" However, according to the documentation, NULL is a valid value of _name_, so that needs to be checked and that's probably the actual intention there (and `\0` converts to NULL (`(char *)'\0'` so the original code worked).

So `name == NULL` would probably do, but covering also the "is empty" case (`name[0] == '\0'`) seems appropriate.

Since this is in the library (unrelated to v.distance parser rules), I'm afraid another PR is needed.

For the record, I was quite confused before I understood the relation of #1265 and f0261d556c577048f3e30546c0a324e6f44ad3d3. Let's not do that again :-)

ninsbl:
OK. The library fix is now implemented with `if(name == NULL || name[0] == '\0')` in #2131 
Is it OK to backport this to 8.0 (or should we wait for 8.0.1)?

PR #2131, opened 2022-01-27
dbmi client: check if driver name is null or empty

ninsbl:
This PR fixes an issue discovered with `v.distance`, where the start of the driver caused a `segmentation fault` when output should be produced, rendering `v.distance` unusable for creating connection lines.

Co-authored-by: Vaclav Petras wenzeslaus@gmail.com

### DIFF
--- a/lib/db/dbmi_client/start.c
+++ b/lib/db/dbmi_client/start.c
@@ -94,7 +94,7 @@ dbDriver *db_start_driver(const char *name)
 	return (dbDriver *) NULL;
 
     /* if name is empty use connection.driverName, added by RB 4/2000 */
-    if (name[0] == '\0') {
+    if (name == NULL || name[0] == '\0') {
 	db_get_connection(&connection);
 	if (NULL == (name = connection.driverName))
 	    return (dbDriver *) NULL;


ninsbl:
OK to backport to 8.0?

nilason (approved):
The `db_start_driver()` documentation states:
> If <i>name</i> is NULL, the db name will be assigned connection.driverName.

so `name == NULL` should be enough, which is what I assume was originally intended.

However the inline comment refer to "if name is empty ..." which is what `name[0] == '\0'` checks.

In any case, I think the double check proposed by this PR is acceptable. Incoming arg `name` may be either NULL or an empty string, in both cases the name need to be taken  from `connection.driverName`.

I see no reason not to backport to 8.0.

ninsbl:
Thanks @nilason for the swift reply.

Now, the unittest for `r.object.geometry` on Ubuntu 18.04 fails with:

```
  File "etc/python/grass/gunittest/case.py", line 1403, in assertModule
    module.run()
  File "etc/python/grass/pygrass/modules/interface/module.py", line 823, in run
    self.wait()
grass.exceptions.CalledModuleError: Module run `r.object.geometry input=test_objects1 separator=pipe output=output_file_meter.csv -m` ended with an error.
The subprocess ended with a non-zero return code: 1. See the following errors:
b'ERROR: Input raster must be of type CELL\n'
```
But I cannot see how that could be related... The same test does not fail on Ubuntu 20.04...

nilason:
> Thanks @nilason for the swift reply.
> 
> Now, the unittest for `r.object.geometry` on Ubuntu 18.04 fails with:
> 
> ```
>   File "etc/python/grass/gunittest/case.py", line 1403, in assertModule
>     module.run()
>   File "etc/python/grass/pygrass/modules/interface/module.py", line 823, in run
>     self.wait()
> grass.exceptions.CalledModuleError: Module run `r.object.geometry input=test_objects1 separator=pipe output=output_file_meter.csv -m` ended with an error.
> The subprocess ended with a non-zero return code: 1. See the following errors:
> b'ERROR: Input raster must be of type CELL\n'
> ```
> 
> But I cannot see how that could be related... The same test does not fail on Ubuntu 20.04...

Strange... just restarted the Ubuntu CI runs, let's see ...

ninsbl:
In my fork, the tests pass:
https://github.com/ninsbl/grass/tree/dbmi_client_start_driver

nilason:
> In my fork, the tests pass:
> https://github.com/ninsbl/grass/tree/dbmi_client_start_driver

Yea, cannot see how that would be related, probably because it isn't :-)
My bet would be a CI hiccup, did happen before.

nilason:
Latest commit to main also caused Ubuntu 18 failure. But the same commit passed as PR (#2125).

wenzeslaus:
+1 for the backport to 8.0.0.

And yes, r.object.geometry sometimes produces `ERROR: Input raster must be of type CELL` but there is no issue open for it.

Issue #2128, opened 2022-01-26
[Bug] Warnings for -Wunused-result in GCC build


wenzeslaus:
**Describe the bug**

A build with GCC can produce warnings _ignoring return value_ (`-Wunused-result`) coming from library functions declared with attribute `warn_unused_result` (such as _fgets_ and _fread_).

```
warning: ignoring return value of ‘fgets’, declared with attribute warn_unused_result [-Wunused-result]
```

This happens with the default setting on Ubuntu (so likely other systems too) and with the current development practices, today's warnings are tomorrow's errors.

**To Reproduce**

Compile with GGC with `-Wunused-result`, `-Werror=unused-result`, or default settings on Ubuntu 20.04. See also build in the _ubuntu-20.04 tests_ job in the CI (e.g., [run 4947257936](https://github.com/OSGeo/grass/runs/4947257936?check_suite_focus=true)) or a binder build ([link for main](https://mybinder.org/v2/gh/OSGeo/grass/main?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fbasic_example.ipynb), may not rebuild).

**Expected behavior**

No warnings. (Return values checked I guess.)

**Screenshots**

```
main.c: In function ‘update_reclass_maps’:
main.c:138:2: warning: ignoring return value of ‘fgets’, declared with attribute warn_unused_result [-Wunused-result]
  fgets(buf2, 255, fp);
  ^~~~~~~~~~~~~~~~~~~~
...
main.c:148:2: warning: ignoring return value of ‘fread’, declared with attribute warn_unused_result [-Wunused-result]
  fread(str, l, 1, fp);
  ^~~~~~~~~~~~~~~~~~~~
```

**System description**

- Operating System: Ubuntu 20.04
- GRASS GIS version: releasebranch_8_0, main

**Additional context**

Besides the potential of this being configures as an error in the future, the warning is currently visible in several places, namely in some of the CI builds and during a build on mybinder.org where end users may see it too.

PR #2135, opened 2022-01-28
libdisplay: remove redundant GRASS_NOTIFY

nilason:
This addresses one of the '-Wunused-result' compiler warnings reported in #2128.

The warning is caused by a call of `system(cmd)` upon display driver closure, where `cmd` is set by the environment variable `GRASS_NOTIFY` (if set). The warning in itself is not critical.

This PR is a suggestion to address this warning by removing this GRASS_NOTIFY "hack" or whatever it is.
This is the only place in GRASS sources where `GRASS_NOTIFY` is used at all.

The big question is, is this GRASS_NOTIFY mechanism used by anyone, or any platform?
My research into the motivation of its introduction to GRASS 7.0 [1], or of any of its use, turned up empty. I assume there was a good reason for this in the transition to G7, but now? The GRASS_NOTIFY environment variable is also [not documented](https://grass.osgeo.org/grass80/manuals/variables.html).

[1] initially committed to trunk: https://trac.osgeo.org/grass/changeset/32602/grass/trunk/lib/raster/raster.c

### DIFF
--- a/lib/display/r_raster.c
+++ b/lib/display/r_raster.c
@@ -158,17 +158,10 @@ int D_open_driver(void)
 
 /*!
   \brief Close display driver
-
-  If GRASS_NOTIFY is defined, run notifier.
 */
 void D_close_driver(void)
 {
-    const char *cmd = getenv("GRASS_NOTIFY");
-
     COM_Graph_close();
-
-    if (cmd)
-	system(cmd);
 }
 
 /*!


wenzeslaus (approved):

wenzeslaus:
It is there as one of the many attempts to improve the rendering workflow. It is not documented and not used in the code and there is no motivation in the commit message.

However, I double checked and some or all (?) usage of ximgview and wxpyimgview can be broken if removed. See:

> A suitable (although rather primitive) viewer is available as visualisation/ximgview. This will refresh its display from the file either at regular intervals, or when it receives SIGUSR1. If the environment variable GRASS_NOTIFY exists, its value will be executed via system() whenever a d.* command completes, so you can use e.g.:
>
> ```
> export "GRASS_NOTIFY=kill -USR1 `pidof ximgview`"
> ```

[[GRASS-dev] Shell scripts, Glynn Clements, Dec 2009](https://lists.osgeo.org/pipermail/grass-dev/2009-December/047194.html) (https://narkive.com/9nUlmRam.12)

SIGUSR1 is used in ximgview and wxpyimgview. wxpyimgview could probably use watchdog as wxGUI is using but it may not be the same.

nilason:
I did see that single mention in ML, but that’s all. I’m still convinced this is a hack/debug/test thing. Besides, it is not particularly safe to pass an env var unchecked to a system call.

wenzeslaus:
I'm for removing it, but doesn't that mean we need to remove SIGUSR1 from in ximgview and wxpyimgview, too? Then the question is how they will update? Both ximgview and wxpyimgview stand somewhat outside of the standard system as alternatives when the main GUI or `d.mon wx*` does not suite your needs. It is little hard to make decisions about these. I think it is fair to say that they are ad hoc additions. This might be something to discuss on grass-dev.

nilason:
Closes this in favour of #2705.

PR #2140, opened 2022-01-29
utils/mkhtml.py: generate local JSON file with core modules and their last commit

tmszi:
In the context of creating and using a core_modules_with_last_commit.json file after creating and submitting a new release tag to remote repository e.g. `git tag -a 8.2.0 -m "Release 8.2.0"` and `git push origin 8.2.0` a new draft is automatically created with GitHub action workflow "Create new release draft" #2224 with attached two files core_modules_with_last_commit.json and core_modules_with_last_commit.patch.
    
JSON file data structure:
    
```json
"r.pack": {
    "commit": "547ff44e6aecfb4c9cbf6a4717fc14e521bec0be",
    "date": "2022-02-03T11:10:06+01:00"
},
```
    
During compilation source code without .git dir, this local file "core_modules_with_last_commit.json" is used instead of getting Git commits from remote GitHub API server.


wenzeslaus (commented):
I need to look at this more, but it seems that something like this is the only way to get all the info from Git. I assume others already looked into bending Git to be more like Subversion and that didn't work. I would still ask though, why is carrying this generated file in the code necessary?

As for the implementation, the automatic commit won't work well here as far as I can tell. Doing on the PR branch will record wrong commit hashes. Doing it on the main branch will result in doubling the amount of commits as every commit (merge) will require another commit as an update for this file. (And also disabling some of the mandatory checks may be needed to make it work.)

I came up with two scenarios which could work:

1. Make the workflow a dispatch workflow (or whatever these are called) so that it is called manually. Part of the release procedure would be to trigger this workflow, it would create a PR against the release branch (or commit, but PR sounds safer!). This is closer to your current proposal, than the other option.
2. Make the generated file a GitHub Actions artifact (possible for all commits on all branches) and let people download it from GitHub when needed or, for convenience, upload it somewhere from GitHub actions (using scp or FTP). This could be part of creating the release tarball. Requires more manual work, but doesn't require to have the generated file in the source code.

tmszi:
> As for the implementation, the automatic commit won't work well here as far as I can tell. Doing on the PR branch will record wrong commit hashes. Doing it on the main branch will result in doubling the amount of commits as every commit (merge) will require another commit as an update for this file. (And also disabling some of the mandatory checks may be needed to make it work.)

Auto commit has limitation but in our case Git repository allow Squash and Merge and if this GitHub action is allowed on the only main branch, it is work as expected (update this JSON file is triggered on  push to main branch and one auto commit is pushed). I tested it on my test repository and it worked as expected.

lucadelu:
Tested with 7.8.7 release branch and it work properly

```
Started compilation: Fri Feb 11 09:20:13 UTC 2022
--
Errors in:
No errors detected.
--
Finished compilation: Fri Feb 11 09:29:52 UTC 2022
```

tmszi:
> I need to look at this more, but it seems that something like this is the only way to get all the info from Git. I assume others already looked into bending Git to be more like Subversion and that didn't work. I would still ask though, why is carrying this generated file in the code necessary?

Yes you are right it is not necessary.

> 2. Make the generated file a GitHub Actions artifact (possible for all commits on all branches) and let people download it from GitHub when needed or, for convenience, upload it somewhere from GitHub actions (using scp or FTP). This could be part of creating the release tarball. Requires more manual work, but doesn't require to have the generated file in the source code.

I decided to implement this procedure. The resulting GitHub workflow pgms-with-last-commit-patch-file artifact will create a pgms_with_last_commit.patch (new pgms_with_last_commit.json file) that will be downloadable as zip file for the GRASS GIS GNU/Linux distribution maintainer and applicable as a patch in its package build process.

tmszi:
@neteler, I think we should include this PR in the next release 8.0.1 and 7.8.7 and solve GitHub REST API rate limit completely (which means every core module manual page will have commit and commit date, when GRASS GIS is compile from release archive which doesn't have Git).

The generated `pgms_with_last_commit.json` file can then be downloaded from the GitHub action workflow  pgms-with-last-commit-patch-file artifact (for branch) and pasted into the release archive during creation.

neteler:
> I think we should include this PR in the next release 8.0.1 and 7.8.7 and solve GitHub REST API rate limit completely

Thanks for your work on this PR! I am a bit undecided if 8.0.1 and 7.8.7 or better 8.0.2 and 7.8.8? If the former, then we will need a longer test phase (I just published 8.0.1RC1 yesterday, see [grass-dev mailing list](https://lists.osgeo.org/pipermail/grass-dev/2022-February/095556.html)). I suggest to discuss it on grass-dev to reach more people.

lucadelu:
I would prefer 7.8.7 and 8.0.1

wenzeslaus:
I prefer after 7.8.7 and 8.0.1 or later. This is not even reviewed, while 8.0.1 is ready to go.

Let's not backport large or invasive changes and let's not backport quickly or right before a release. As far as I'm concerned, this can sit on the main branch all the way to 8.2.0. Date in a man page is not important enough to risk breaking compilation on a release branch.

tmszi:
> Date in a man page is not important enough to risk breaking compilation on a release branch.

I will try to explain in more detail what **other consequences** the exceeded GitHub REST API rate limit has during compilation, which is reported in the issue #2186, and that is why I suggest integrating it into 8.0.1 and 7.8.7 as soon as possible **after thorough testing**:

1. Compilation GRASS GIS from source code without Git was successfull on computer which never has installed GRASS GIS before (.grass8 dir doesn't exists)
2. GitHub REST API rate limit was exceeded during compilation
3. After GRASS GIS installation you want install some new addon: `g.extension i.sentinel`
4. You get error message GitHub REST API rate limit was exceeded, and **user must wait (1 hour)**. **Information when rate limit is freed is not included**, and this is reason why I created PR #2193

```
GRASS nc_basic_spm_grass7/PERMANENT:~ > g.extension i.sentinel
ERROR: The download of the JSON file with add-ons paths from the GitHub
       REST API server wasn't successful, rate limit exceeded, 60 requests
       per hour per your computer IP address. The previous downloaded JSON
       file will not be used, because does not exists. You can try it again
       on Saturday Feb 12 14:57:22 2022.
```

5.  Non technical user don't understand what is GitHub REST API rate limit without better explanation.

As [already mentioned](https://github.com/OSGeo/grass/issues/2186#issuecomment-1037008205), the main problem is that we are **depend on proprietary GitHub REST API** (g.extension module) which has [limitation](https://github.com/OSGeo/grass/issues/2186#issuecomment-1036336726) and **who has control over us, not we over him**.


> Let's not backport large or invasive changes and let's not backport quickly or right before a release. 

**It is just a sugestion** that explains in much more detail why it will be good to implement these changes as soon as possible.

Code changes with this PR:

1. New GitHub action workflow artifact, which I tested separately on my test repository
2. Simple new script which generate new JSON file
3. Open and read JSON file 

Code changes with  PR #2193:

1. Changed error message string

.github/workflows/gen_pgms_last_commit_json_file.yml
```diff
@@ -0,0 +1,30 @@
+name: Generate PGMS last commit JSON file
+
+on:
+  push:
+    branches:
+      - 'main'
+      - 'releasebranch_7_8'
+      - 'releasebranch_8_0'
```
ninsbl, line 8: Would`nt this be needed also for "nightly" builds?

ninsbl (approved):
OK. I invstigated the possibility to deploy to a gist to circumvent the API (content could be downloaded as raw file directly), but only users seem to be able to own gists (not organizations (here osgeo) and repos (grass)...
So, I guess artifacts are the best option if we do not publish the JSON file other places (which would have other drawbacks).
Another question is if we could find a mechanism where the JSON file is attempted to be downloaded only once (i.e. at start of compilation) or if it has been updated since the last download or lder than e.g one hour.

That said, this PR definately improves the current situation.

.github/workflows/gen_pgms_last_commit_json_file.yml
ninsbl: Sorry, I see main is covered...

utils/gen_pgms_last_commit.py
```diff
@@ -0,0 +1,60 @@
...
+    :param str src_dir: root src dir
+
+    :return JSON obj result: JSON object
+    """
+    result = {}
+    if not shutil.which("git"):
+        sys.exit("Git command was not found. Please install it.")
+    for root, dirs, files in os.walk(src_dir):
+        for f in files:
+            if f.endswith(".html"):
```
wenzeslaus, line 35: Why to limit that to HTML files? We are interested in history of the whole directory, no?

utils/gen_pgms_last_commit.py
```diff
@@ -0,0 +1,60 @@
...
+                rel_path = os.path.relpath(root)
+                stdout, stderr = subprocess.Popen(
+                    ["git", "log", "-1", rel_path],
```
wenzeslaus, line 38: Have you considered using `--format=`? It may help you parse the output in a robust way.

.github/workflows/gen_pgms_last_commit_json_file.yml
wenzeslaus, line 8: Perhaps `releasebranch*` is enough or `releasebranch_7_8` and `releasebranch_8_*`? Less updates needed.

wenzeslaus (changes requested):
This seems like a step in the right direction if we want to harvest all the information from Git.

If I remember correctly, GitHub artifacts don't have predictable URLs. Any plans for that? We could upload somewhere from GitHub Actions using something like FTP/SCP and secrets.

I would like to see this running on PRs as well for testing purposes, esp., for cases when you make change here. The additional_checks.yml workflow/job is meant for random things like this, so you could add the generator script there with some check of the JSON content.

wenzeslaus:
> ...As already mentioned, the main problem is that we are depend on proprietary GitHub REST API (g.extension module) which has limitation and who has control over us, not we over him.

Right, I think everyone agrees with that. My point is that there is no reason to have anything like that in the source code in the first place when it brings no benefit and since the code as is doesn't work, we can just drop it and, in some cases, live with some missing metadata in manual pages for 7.8 and 8.0. For 8.2, we can see how this PR works.

tmszi:
> If I remember correctly, GitHub artifacts don't have predictable URLs. Any plans for that? We could upload somewhere from GitHub Actions using something like FTP/SCP and secrets.

I assumed that this one manual step would be enoug. When we want make new release (source code without Git), person which make this release via [GitHub page](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository), first download `pgms_with_last_commit.json` file from last GitHub workflow pgms-with-last-commit-patch-file artifact to him computer and than upload as attach into process of making GitHub GRASS GIS release. File will be part of archive e.g. https://github.com/OSGeo/grass/archive/refs/tags/8.0.1RC1.tar.gz.

wenzeslaus:
> ....we want make new release (source code without Git)...first download `pgms_with_last_commit.json` file from last GitHub workflow...than upload as attach into process...

The [release procedure](https://github.com/OSGeo/grass/blob/main/doc/howto_release.md) is already a lot of work. We should be working towards minimizing the steps and automating ideally everything.

> ...file will be part of archive...

The GitHub page says you can _include binary files...in the binaries box_. I don' see anything about combining that with the tarball.

However, the download links in a release are stable and it seems you can upload to release from GitHub actions ([softprops/action-gh-release](https://github.com/softprops/action-gh-release), [svenstaro/upload-release-action](https://github.com/svenstaro/upload-release-action), [JasonEtco/upload-to-release](https://github.com/JasonEtco/upload-to-release), ...).

utils/gen_pgms_last_commit.py
tmszi: It is a directory filter that filters directories that are not core modules based on the assumption that each core module has a manual page in the root directory.

wenzeslaus: Oh, I see, f is not used to get the path. Can you rewrite the code so that the operation on the whole directory is not a inside of a loop over files and add comments? Perhaps something like:

```python
if not contains_html_file(files):
    # No HTML file in the directory, no need to store metadata.
    continue
# ...git here

**Comment by tmszi** [comment]
I have implemented all your suggestions.

**Comment by tmszi** [comment]
For testing purpose (preparing source code for compilation GRASS GIS from source code without Git):

```bash
tomas@gentoo-gnu-linux:/tmp$ git clone -b generate-local-pgms-json-file-with-last-commit --single-branch https://github.com/tmszi/grass

tomas@gentoo-gnu-linux:/tmp$ cd grass && python utils/generate_core_modules_with_last_commit_json_file.py .

tomas@gentoo-gnu-linux:/tmp/grass$ file core_modules_with_last_commit.json 
core_modules_with_last_commit.json: JSON data

tomas@gentoo-gnu-linux:/tmp/grass$ rm -rf .git

```

wenzeslaus (changes requested):
The release draft action is a great improvement. I'm not sure about the actions/create-release part since it is kind of buried in the context of this PR. The actions/upload-release-asset is more clear, but the this specific action is not maintained (of the ones I linked is a suggested replacement). Would you consider splitting the PR? 

I still think something the like `python utils/generate_core_modules_with_last_commit_json_file.py .` should run also in additional_checks.yaml with some subsequent check (like load to Python and check basic structure or one of the values).

Thanks for the rename. Perhaps the new one is little too long? `generate_last_commit_file.py` gets most of the important things in.

tmszi:
> The release draft action is a great improvement. I'm not sure about the actions/create-release part since it is kind of buried in the context of this PR. The actions/upload-release-asset is more clear, but the this specific action is not maintained (of the ones I linked is a suggested replacement). Would you consider splitting the PR?

Ok. I opened new PR #2224.


> I still think something the like `python utils/generate_core_modules_with_last_commit_json_file.py .` should run also in additional_checks.yaml with some subsequent check (like load to Python and check basic structure or one of the values).

Ok. I created pytest  file `pytest utils/test_generate_last_commit_file.py` and incrorporate into additional_checks.yaml file GitHub workflow action.

> Thanks for the rename. Perhaps the new one is little too long? `generate_last_commit_file.py` gets most of the important things in.

Yes, I agree with you.

 I have included all your previous suggestions.

utils/generate_core_modules_with_last_commit_json_file.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def contains_html_file(files):
+    """Contains HTML file
+
+    :param list files: list of files
+
+    :return bool: True if *.html file found
+    """
+    return ".html" in ",".join(files)
```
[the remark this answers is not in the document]
tmszi: This [adjustment](https://github.com/OSGeo/grass/pull/2140/commits/0ffd81dbac6d462561432f43230d772e1b013bc4) should be enough.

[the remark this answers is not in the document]
wenzeslaus: Probably best if you either hardcode the separator everywhere or add that to the ".html," string.

utils/test_generate_last_commit_file.py
```diff
@@ -0,0 +1,87 @@
...
+    # Get Git commit and commit date from local Git
+    process_result = subprocess.run(
+        ["git", "log", "-1", "--format=%H,%at", core_module_path],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )  # --format=%H,%at commit hash,author date (UNIX timestamp)
+    if process_result.returncode == 0:
+        commit, date = process_result.stdout.decode().strip().split(",")
+    else:
+        sys.exit(process_result.stderr.decode())
```
wenzeslaus, line 80: I think `check=True` is generally preferable and here is no reason to do otherwise.

utils/test_generate_last_commit_file.py
```diff
@@ -0,0 +1,87 @@
...
+    core_module = os.path.basename(core_module_path)
+    if core_module in read_json_file:
+        # Compare commit and commit date
+        assert (
+            read_json_file[core_module]["commit"] == commit
+            and read_json_file[core_module]["date"] == date
+        )
```
wenzeslaus, line 87: Any reason to use one assert? Two asserts seem more readable, are shorter, and likely easier to understand in case of errors.

utils/mkhtml.py
```diff
@@ -188,71 +179,169 @@ def download_git_commit(url, response_format, *args, **kwargs):
...
+    :return dict git_log: dict with key commit and date, if not
+                          possible download commit from GitHub REST API
+                          server values of keys have "unknown" string
+    """
+    process_result = has_src_code_git(src_dir=src_dir)
+    if process_result:
+        stdout = process_result.stdout.decode()
+        if stdout:
+            return parse_git_commit(commit=stdout)
+    else:
```
wenzeslaus, line 335: I'm still not sure if this else should be here. What happens when stdout above is empty? If you just not test it, that may be fine given the context, but this seems like missing return (or raise).

utils/mkhtml.py
```diff
@@ -188,71 +179,169 @@ def download_git_commit(url, response_format, *args, **kwargs):
...
+    git_log["commit"], commit_date = commit.strip().split(",")
+    git_log["date"] = format_git_commit_date_from_local_git(
+        commit_datetime=commit_date,
+    )
+    return git_log
+
+
+def get_git_commit_from_file(
+    src_dir,
+    git_log={"commit": "unknown", "date": "unknown"},
```
wenzeslaus, line 203: Here perhaps okay, but it is generally bad practice to use mutable values as defaults.

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

wenzeslaus (changes requested):
Getting there. Awesome example of full pytest usage.

utils/mkhtml.py
tmszi: Sure, you're right.

tmszi:
I also integrated the remaining suggestions.

wenzeslaus (approved):
I'm not sure about the use of UNIX timestamp instead of a readable date and use of sophisticated pytest with depends on an external state, but I think this can be merged.

I don't know where documentation should go, esp. in context of #2224 (automated upload to release draft).

tmszi:
> I'm not sure about the use of UNIX timestamp instead of a readable date and use of sophisticated pytest with depends on an external state, but I think this can be merged.

All right, I'll change to the readable date format [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601).

wenzeslaus (approved):

wenzeslaus:
This is in 8.2 branch and it is not a general backport material (so no 8.0 which has another fix), so removing the label.

PR #2142, opened 2022-01-30
Fix typo in v.import documentation

agiudiceandrea:
Substitute 'GDAL needs to be installed to use **v.in.ogr**' with 'GDAL needs to be installed to use **v.import**' in the [v.import](https://grass.osgeo.org/grass78/manuals/v.import.html) documentation.


neteler:
Thanks!

PR #2149, opened 2022-02-01
i.atcorr: Replace CRLF by LF in sensor CSVs

wenzeslaus:
This repleaces CRLF (dos) by LF (unix) line ending in geoeye1 and sentinel_2A_msi CSV files. See also #2144 and #2147.


wenzeslaus:
Can someone please review these changes in i.atcorr files? The CSV files seem unused, but I don't see documentation for it, so I would like to confirm that there is not anything special about them.

BadAssassin:
The python file 'create_iwave.py' in i.atcorr may need some massaging to handle the changes. Otherwise, I see no issue.

Best Regards,
--Brad

------- Original Message -------

neteler:
> The CSV files seem unused, but I don't see documentation for it, 

The sensor-specific CSV files are the input to reconstruct the CPP source code via `create_iwave.py` and should be kept:
https://github.com/OSGeo/grass/blob/eba61a1ca122758c9aae4ff0c28bab895609d9f6/imagery/i.atcorr/create_iwave.py#L16

wenzeslaus:
I have tested the script with the `geoeye1.csv` file and it gives same result as what is currently in the `.cpp` file. The documentation for this is in the parent directory in a README file which I overlooked before that (there is a separate README file just for the CSV files in the CSV directory).

PR #2153, opened 2022-02-03
wxGUI: Statusbar settings as a part of Map Display Settings

lindakladivova:
This PR creates several radiobuttons in Map Display settings that control which widget will be shown in statusbar.
Screenshot from 2022-02-03 03-58-09
This PR needs to wait for some refactoring that has to be done for mapdisp/statusbar.py file.


petrasovaa:
Show everywhere seems redundant. Main slot settings is unclear, maybe Displayed content/information?

lindakladivova:
> Show everywhere seems redundant. Main slot settings is unclear, maybe Displayed content/information?

Yes, that would be better.

lindakladivova:
> > Show everywhere seems redundant. Main slot settings is unclear, maybe Displayed content/information?
> 
> Yes, that would be better. Screenshot from 2022-02-03 11-07-52

Maybe, we should also adapt terms Display extent, Computational region and Computational extent. I think they are quite confusing. Display extent is called Display Geometry, but if I understand it well it should display Display resolution. And Region extent vs. Computational region extent... I am not sure at all what is the difference between them apart from the fact that computational region displays two extra information. What do your think? Do you have some ideas for better naming?
We could probably use **Display resolution,
Region extent and
Computational region extent**. But I think it is still confusing.

petrasovaa:
> We could probably use **Display resolution, Region extent and Computational region extent**. But I think it is still confusing.

We could use: Display resolution, Display extent, Computational region

gui/wxpython/mapdisp/properties.py
```diff
@@ -277,6 +283,42 @@ def _onToggleCheckBox(self, event):
...
+    displayGeometry = 6
+
+
+class RBShowInStatusbar(PropertyItem):
+    """General class for radiobuttons managing widgets in statusbar."""
+
+    def __init__(self, parent, mapWindowProperties, value, label, style=0):
+        PropertyItem.__init__(self, mapWindowProperties)
+        self.value = value
+        self.widget = wx.RadioButton(
```
petrasovaa, line 303: This should be RadioBox, that should simplify a lot of the code below.

gui/wxpython/mapdisp/properties.py
```diff
@@ -277,6 +283,42 @@ def _onToggleCheckBox(self, event):
...
 
+class RBEnum(Enum):
+    """Radiobutton types and their numbers"""
+
+    coordinates = (1,)
+    mapscale = (2,)
+    compRegion = (3,)
+    goTo = (4,)
+    compExtent = (5,)
+    displayGeometry = 6
```
petrasovaa, line 294: Why tuples here? And the last one is int?

gui/wxpython/mapwin/base.py
```diff
@@ -44,6 +44,10 @@ def __init__(self):
         self.useDefinedProjectionChanged = Signal(
             "MapWindowProperties.useDefinedProjectionChanged"
         )
+        self.shownWidgetInStatusbar = None
```
petrasovaa, line 47:
```suggestion
        self._shownWidgetInStatusbar = None
```

gui/wxpython/mapdisp/properties.py
lindakladivova: Interesting. I did not make it, that had to be caused by Black.

lindakladivova:
> > We could probably use **Display resolution, Region extent and Computational region extent**. But I think it is still confusing.
> 
> We could use: Display resolution, Display extent, Computational region

I was thinking about it more and I like a "canvas" word. It seems to me a bit clearer than display but the criticism is welcome. :-)

Now it looks as follows:
Screenshot from 2022-02-07 09-44-14

veroandreo:
That would be a new word for GRASS, no? What would it refer to: the map display region, right? Why call it Geographical? 

Also, what would canvas resolution refer to? Currently only the computational region option in the status bar shows resolution in between parenthesis. Is there a way to have a map display resolution different than that of the computational region?  

> > > We could probably use **Display resolution, Region extent and Computational region extent**. But I think it is still confusing.
> > 
> > 
> > We could use: Display resolution, Display extent, Computational region
> 
> I was thinking about it more and I like a "canvas" word. It seems to me a bit clearer than display but the criticism is welcome. :-)
> 
> Now it looks as follows: Screenshot from 2022-02-07 09-44-14

That would be a new word for GRASS, no? It refers to the "map display region/extent", right? I think the term is used in QGIS (at least I've seen it when exporting maps from there) so users might be familiar with it, but to my knowledge, it is new to GRASS. I wonder if this will have then further implications in terms of GRASS terminology? If we use canvas here, should we also change "Map display" to canvas everywhere? Also, why call it Geographical? Will there be other kind of canvas? 

Currently only the computational region option in the status bar shows resolution in between parenthesis. 
What would canvas resolution refer to? Is there a way to have/set a map display resolution different than that of the computational region? 

I'm thinking out loud here, apologies if these are silly questions

petrasovaa:
> > > We could probably use **Display resolution, Region extent and Computational region extent**. But I think it is still confusing.
> > 
> > 
> > We could use: Display resolution, Display extent, Computational region
> 
> I was thinking about it more and I like a "canvas" word. It seems to me a bit clearer than display but the criticism is welcome. :-)
> 
> Now it looks as follows: Screenshot from 2022-02-07 09-44-14

I would still go with my suggestion, as Vero points out, canvas is not a term we use and I don't think it particularly helps here.

I thought about merging the Display extent and Display resolution into Display region, which would be more consistent with Computational region, but if we would do that, that would be a separate PR.

petrasovaa:
> What would canvas resolution refer to? Is there a way to have/set a map display resolution different than that of the computational region?

By default the rendering (display) resolution is different than computational, it's based on the extent and display size, e.g. (n - s) / height in px.

lindakladivova:
> That would be a new word for GRASS, no? What would it refer to: the map display region, right? Why call it Geographical?
> 
> Also, what would canvas resolution refer to? Currently only the computational region option in the status bar shows resolution in between parenthesis. Is there a way to have a map display resolution different than that of the computational region?

I think you hit a nail on its head, @veroandreo  :-) That "Display resolution" is exactly the point when it becomes confusing. If I get it right it is given in pixels whereas Display Extent is given in coordinates. So, that's the reason why I suggested adding 'Geographical' before a word 'canvas extent' .... I agree that the term 'canvas' is too big intervention but let's try to think about this issue a bit. For example, I haven't found anywhere in the docs the explanation of what Display resolution actually means. Maybe it would help to add 'Map' and have 'Map Display resolution'. Or we could have 'Map Display info', 'Map Display technical info'..... anyway, I think we probably need to explain it more.

> 
> > > > We could probably use **Display resolution, Region extent and Computational region extent**. But I think it is still confusing.
> > > 
> > > 
> > > We could use: Display resolution, Display extent, Computational region
> > 
> > 
> > I was thinking about it more and I like a "canvas" word. It seems to me a bit clearer than display but the criticism is welcome. :-)
> > Now it looks as follows: Screenshot from 2022-02-07 09-44-14
> 
> That would be a new word for GRASS, no? It refers to the "map display region/extent", right? I think the term is used in QGIS (at least I've seen it when exporting maps from there) so users might be familiar with it, but to my knowledge, it is new to GRASS. I wonder if this will have then further implications in terms of GRASS terminology? If we use canvas here, should we also change "Map display" to canvas everywhere? Also, why call it Geographical? Will there be other kind of canvas?
> 
> Currently only the computational region option in the status bar shows resolution in between parenthesis. What would canvas resolution refer to? Is there a way to have/set a map display resolution different than that of the computational region?
> 
> I'm thinking out loud here, apologies if these are silly questions

lindakladivova:
> > > > We could probably use **Display resolution, Region extent and Computational region extent**. But I think it is still confusing.
> > > 
> > > 
> > > We could use: Display resolution, Display extent, Computational region
> > 
> > 
> > I was thinking about it more and I like a "canvas" word. It seems to me a bit clearer than display but the criticism is welcome. :-)
> > Now it looks as follows: Screenshot from 2022-02-07 09-44-14
> 
> I would still go with my suggestion, as Vero points out, canvas is not a term we use and I don't think it particularly helps here.
> 
> I thought about merging the Display extent and Display resolution into Display region, which would be more consistent with Computational region, but if we would do that, that would be a separate PR.

I am not sure if it helps. We can try it but I think we might need more to improve the documentation..

wenzeslaus (commented):
I think the two columns should not be there. It should be just one column.

I went through Preferences/Settings in several applications (Gimp, Inkscape, Spyder, QtCreator, VSCodium) and found two columns only in QtCreator in settings for Debugger and for Vim, so clearly something extreme. And even there it was for more than 10 items and they were checkboxes and exclusive choice radio buttons, so it is questionable if that even applies here.

petrasovaa:
I don't like the `mode` thing. Not sure if i don't overlook something, but perhaps creating a new class inheriting from `RBShowInStatusbar` and override e.g.,  a `_getOptions` method (containing the labels) would be better, or?

gui/wxpython/mapdisp/properties.py
```diff
@@ -277,6 +279,59 @@ def _onToggleCheckBox(self, event):
...
+    def GetWidget(self):
+        """Returns underlying widget.
+
+        :return: widget or None if doesn't exist
+        """
+        return self.widget
+
+    def _getChoices(self):
+        attributes = []
+        for value in self.statusbarManager.statusbarItems.values():
```
petrasovaa, line 318: I suggest adding a new method to sbManager to simplify this.

gui/wxpython/gcp/statusbar.py
```diff
@@ -34,7 +34,7 @@ class SbGoToGCP(SbItem):
     def __init__(self, mapframe, statusbar, position=0):
         SbItem.__init__(self, mapframe, statusbar, position)
         self.name = "gotoGCP"
-        self.label = _("Go to GCP No.")
+        self.label = _("Go to GCP number")
```
petrasovaa, line 37: Let's not put this change in this PR, it's unrelated, see also the ii2t case.

gui/wxpython/mapdisp/statusbar.py
```diff
@@ -182,35 +182,43 @@ def ShowStatusbarChoiceItemsByClass(self, itemClasses):
             if item.__class__ in itemClasses:
                 self.choice.Insert(item.label, pos, item)
 
+    def GetItemLabels(self):
+        """Get list of item labels"""
+        attributes = []
+        for value in self.statusbarItems.values():
+            if value.GetPosition() == 0:
+                attributes.append(value.label)
+        return attributes
```
petrasovaa, line 191: This is shorter, list comprehensions are typically preferred and once you get used to them, they are more readable:
```suggestion
        return [value.label for value in self.statusbarItems.values() if value.GetPosition() == 0]
```

gui/wxpython/mapdisp/statusbar.py
```diff
@@ -285,27 +289,22 @@ def GetProgressBar(self):
...
 
     def SetMode(self, modeIndex):
         """Sets current mode
 
         Mode is usually driven by user through choice.
         """
+        self._mode = modeIndex
         self._modeIndexSet = True
-        self.choice.SetSelection(modeIndex)
+        self.shownWidgetInStatusbarChanged.emit(value=modeIndex)
```
petrasovaa, line 303: Perhaps calling Update here would be more straightforward?

petrasovaa:
Is this still draft? It seems it's close to finishing, no?

lindakladivova:
The 3D/2D stuff is missing. We need to adapt ShowStatusbarChoiceItemsByClass and HideStatusbarChoiceItemsByClass. I think It should emit the signal which shops/hides buttons in the Map Display settings radiobox. The Combobox had reacted as well on 2D/3D switching so we should probably preserve that thought.

lindakladivova:
> The 3D/2D stuff is missing. We need to adapt ShowStatusbarChoiceItemsByClass and HideStatusbarChoiceItemsByClass. I think It should emit the signal which shows/hides buttons in the Map Display settings radiobox. The Combobox had reacted as well on 2D/3D switching so we should probably preserve that thought.

I was thinking about it a bit more and we do not probably need any signals, the dialog does not have to response dynamically... What do you think @petrasovaa ?
Btw, it is still not finished I need to test other Map Display apps, hopefully, tomorrow, it will be ready for review. :-)

petrasovaa:
> I was thinking about it a bit more and we do not probably need any signals, the dialog does not have to response dynamically... What do you think @petrasovaa ? Btw, it is still not finished I need to test other Map Display apps, hopefully, tomorrow, it will be ready for review. :-)
I agree, let's keep it simple.

lindakladivova:
> > I was thinking about it a bit more and we do not probably need any signals, the dialog does not have to response dynamically... What do you think @petrasovaa ? Btw, it is still not finished I need to test other Map Display apps, hopefully, tomorrow, it will be ready for review. :-)
> 
> I agree, let's keep it simple.

@petrasovaa I think it is ready to merge. I think disabling items is probably more readable than hiding them so I changed that. I think the code is also a bit more readable and we do not need to change the layout of the radiobox either.
So it looks as follows now:
**2D (basic):**
2d
**3D** (two items in settings disabled):
3D
**GCP apps** (as Image2Target as others):
gcp

veroandreo:
> @petrasovaa I think it is ready to merge. I think disabling items is probably more readable than hiding them so I changed that. I think the code is also simpler as we do not need to change the layout of the radiobox either. So it looks as follows now: 
> **2D (basic):** 
> 2d 

Given that it says `Display content` above, is it needed to write display again in the `Display extent` and `Display geometry` items? (same comment applies to those below)

> **3D** (two items in settings disabled): 
> 3D 

> **GCP apps** (as Image2Target and others): 
> gcp

Would it be possible to write `Number` instead of `No.`?

wenzeslaus:
> Given that it says `Display content` above, is it needed to write display again in the `Display extent` and `Display geometry` items? (same comment applies to those below)

I think we have to keep _display_ there. It is the display extent we are showing, not the computational region extent.

> Would it be possible to write `Number` instead of `No.`?

Yes, please! But perhaps the label is somewhat incomplete anyway. `Go to GCP number...` or `Pan to GCP by number`?

veroandreo:
> > Given that it says `Display content` above, is it needed to write display again in the `Display extent` and `Display geometry` items? (same comment applies to those below)
> 
> I think we have to keep _display_ there. It is the display extent we are showing, not the computational region extent.

Now I get it, it refers to the map display (I read it as the verb display and not the noun referring to map display)... I'd suggest then to use `Map display extent` and `Map display geometry` to be crystal clear
 
> > Would it be possible to write `Number` instead of `No.`?
> 
> Yes, please! But perhaps the label is somewhat incomplete anyway. `Go to GCP number...` or `Pan to GCP by number`?

lindakladivova:
> > > Given that it says `Display content` above, is it needed to write display again in the `Display extent` and `Display geometry` items? (same comment applies to those below)
> > 
> > 
> > I think we have to keep _display_ there. It is the display extent we are showing, not the computational region extent.
> 
> Now I get it, it refers to the map display (I read it as the verb display and not the noun referring to map display)... I'd suggest then to use `Map display extent` and `Map display geometry` to be crystal clear
> 
> > > Would it be possible to write `Number` instead of `No.`?
> > 
> > 
> > Yes, please! But perhaps the label is somewhat incomplete anyway. `Go to GCP number...` or `Pan to GCP by number`?

I agree with you :-) Adding "Map" could help to understand those concepts. I previously suggested renaming the No. to number but we needed to focus on the implementation more so I removed that then.. Now it is a good time to discuss it again and change labels if needed. I think "Pan to GCP by number" is also good. I was thinking about "Focus GCP by number". but @wenzeslaus 's proposal is probably better.

petrasovaa (changes requested):

petrasovaa:
I attached a diff with my changes, I didn't want to apply it directly since it's not completely finished and you may have your changes locally.
[sbdiff.txt](https://github.com/OSGeo/grass/files/8150711/sbdiff.txt)

The changes are adding the context menu in statusbar. Currently the item in the context menu is not always selected correctly and there may be other issues, but this is what I have now.

petrasovaa (approved):

Issue #2156, opened 2022-02-04
[Feat] Fix more compiler warnings


nilason:
This is a followup feature request to #1247, which addressed a great number of compiler warnings on both gcc and clang.
Those warnings were the compiler's default warnings.
Unleashing the compiler with `-Wall` will result in a new wave of warnings, which in many cases are perhaps harmless. But who knows what hidden or potential bugs are waiting there.

(Incomplete) list of warnings:

- [x] -Wtautological-overlap-compare #2157
- [x] -Wunused-variable #2158
- [x] -Wunused-const-variable #2159
- [x] -Wunused-lambda-capture #2160
- [x] -Wunused-private-field #2161
- [x] -Wself-assign #2162
- [x] -Wunused-function #2163
- [x] -Wuninitialized #2164
- [x] -Wsometimes-uninitialized #2165
- [x] -Wunused-result #2705  #2135 #2166 (see also #2128)
- [x] -Wstringop-overflow= #2694 #2733
- [x] -Wformat-truncation= #2694
- [x] -Wformat-overflow= #2700
- [x] -Wunused-but-set-variable #2648 #2650
- [x] -Wunused-variable #2651 
- [x] -Wparentheses #2652
- [x] -Wmaybe-uninitialized #2653
- [x] -Wunneeded-internal-declaration #2654
- [x] -Wclass-memaccess #2660
- [x] -Wdeprecated-declaration #2660
- [x] -Warray-parameter= #2733

nilason:
The CI runners *GCC C/C++ standards check* (GCC 9.4) and "*macOS / macOS build* (Clang 14) both build successfully with `-Wall` and `-Werror`, as well as the *Ubuntu test* runners also builds with `-Werror`.

PR #2157, opened 2022-02-04
Fix -Wtautological-overlap-compare compiler warning

nilason:
As reported in #2156.

**Affects:**

- r.slope.aspect

### DIFF
--- a/raster/r.slope.aspect/main.c
+++ b/raster/r.slope.aspect/main.c
@@ -1030,7 +1030,7 @@ int main(int argc, char *argv[])
         }
         else {
             for (i = ceil(max_asp); i >= 1; i--) {
-                if (i == 0 && i == 360)
+                if (i == 0 || i == 360)
                     sprintf(buf, "north");
                 else if (i == 45)
                     sprintf(buf, "north-east");


wenzeslaus:
Isn't the original code a bug? It never prints north, right?

```c
                if (i == 0 && i == 360)
                    sprintf(buf, "north");
```

I suggest re-framing it this PR as a bug fix for r.slope.aspect instead of a warning fix to draw attention to this issue.

nilason:
> Isn't the original code a bug? It never prints north, right?
> 
> ```c
>                 if (i == 0 && i == 360)
>                     sprintf(buf, "north");
> ```
> 
> I suggest re-framing it this PR as a bug fix for r.slope.aspect instead of a warning fix to draw attention to this issue.

You're right, this is one of those warnings that are more than noise.
It's soon a year since I prepared the bulk of these PRs, so good you mention it, I now remember this standing out.

This was introduced with: https://github.com/OSGeo/grass/commit/772baae662190808c480b39a2beb9721a35fdffc by @metzm .

nilason:
@metzm Could you please check if my fix is correct.

Whatever will be the solution, this have to be backported to 7.8.7 (as well as to 8.0.1).

PR #2160, opened 2022-02-04
Fix -Wunused-lambda-capture compiler warning

nilason:
As reported in #2156.

**Affects:**

- v.in.pdal

### DIFF
--- a/vector/v.in.pdal/main.cpp
+++ b/vector/v.in.pdal/main.cpp
@@ -456,7 +456,6 @@ int main(int argc, char *argv[])
     gpoint_count n_class_filtered = 0;
 
     int cat = 1;
-    bool cat_max_reached = false;
 
     // define callback
     // Capture all values for reading by value, except for the ones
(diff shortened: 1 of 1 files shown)


wenzeslaus (changes requested):
I'm the author of the original code and I think that my idea was that category number needs to be checked. (There can be much more points than we can have categories in a vector map, so we should tell the user when this issue arises and categories are being assigned. I think there is similar code in v.in.lidar.)

Adding a todo here or opening an issue is probably appropriate.

nilason:
> I'm the author of the original code and I think that my idea was that category number needs to be checked. (There can be much more points than we can have categories in a vector map, so we should tell the user when this issue arises and categories are being assigned. I think there is similar code in v.in.lidar.)
> 
> Adding a todo here or opening an issue is probably appropriate.

Ok, I’ll comment it out and add a TODO, leave it to you to open an issue.

wenzeslaus (approved):

PR #2163, opened 2022-02-04
Fix -Wunused-function compiler warnings

nilason:
As reported in #2156.

Initially this is a draft, I commented in code to highlight the unused functions. We need to decide which ones to keep (in one way or other) and which ones that can be removed.

**Affects:**

- include/grass/iostream

Modules

- d.vect.thematic
- i.segment
- r.mapcalc

GRASS Library parts

- lib/cairodriver
- lib/gis
- lib/raster3d
- lib/vector/rtree
- lib/vector/Vlib

### DIFF
--- a/display/d.vect.thematic/local_proto.h
+++ b/display/d.vect.thematic/local_proto.h
@@ -12,8 +12,6 @@ int dareatheme(struct Map_info *, struct cat_list *, dbCatValArray *,
                const struct color_rgb *, int, struct Cell_head *, int);
 
 int dcmp(const void *, const void *);
-static int cmp(const void *, const void *);
-static char *icon_files(void);
 
 /* display.c */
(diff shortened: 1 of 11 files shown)


nilason:
Rebased.

nilason:
Opening this for review. A number of cases I choose to `#if 0 …#endif` out, instead of deleting. Please check if any could safely be removed.

neteler:
Just checked after locally applying this PR:

- compiles fine
- only very few warnings are remaining in the entire code base when checking with  `cppcheck`, great work!

nilason:
> Just checked after locally applying this PR:
> 
> * compiles fine
> * only very few warnings are remaining in the entire code base when checking with  `cppcheck`, great work!

Great, thanks for checking!

PR #2165, opened 2022-02-04
Fix -Wsometimes-uninitialized compiler warnings

nilason:
As reported in #2156.

Initially this is a draft, I commented in code to highlight the *sometimes* uninitialised variables.
We need to decide how to address them.

**Affects:**

Modules

- db/drivers/ogr
- d.vect
- d.vect.thematic
- i.eb.soilheatflux
- r.buffer
- r.fill.stats
- r.resamp.rst
- r3.in.bin
- r3.in.v5d
- r3.out.bin
- r3.out.v5d
- v.normal
- v.proj
- v.sample
- v.to.rast3

GRASS Library parts

- lib/imagery
- lib/ogsf
- lib/raster3d
- lib/vector/Vlib

### DIFF
--- a/vector/v.sample/main.c
+++ b/vector/v.sample/main.c
@@ -205,7 +205,7 @@ int main(int argc, char **argv)
     nlines = Vect_get_num_lines(&In);
 
     for (line = 1; line <= nlines; line++) {
-        int type, cat = 0, ret, cval;
+        int type, cat = -1, ret, cval;
         double dval;
 
         G_debug(3, "line = %d", line);
@@ -215,7 +215,7 @@ int main(int argc, char **argv)
         type = Vect_read_line(&In, Points, Cats, line);
         if (!(type & GV_POINT))
             continue;
-        if (field != -1 && !Vect_cat_get(Cats, field, &cat)) // 'cat' is used uninitialized whenever '&&' condition is false
+        if (field != -1 && !Vect_cat_get(Cats, field, &cat))
             continue;
 
         G_debug(4, "cat = %d", cat);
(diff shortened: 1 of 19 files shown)


wenzeslaus:
I merged this with main which resolved conflicts, but it needs some additional clean up because now it would revert some changes which happened on main in the same files. Given that this is a draft without any actual code changes, I'm leaving it as is for now.

nilason:
Finalised the rebase, to be able to continue with clean sheets.

nilason:
Opening this for review. I left the comments temporarily for context, I will remove them before merge.

vector/v.sample/main.c
```diff
@@ -215,7 +215,7 @@ int main(int argc, char **argv)
         type = Vect_read_line(&In, Points, Cats, line);
         if (!(type & GV_POINT))
             continue;
-        if (field != -1 && !Vect_cat_get(Cats, field, &cat))
+        if (field != -1 && !Vect_cat_get(Cats, field, &cat)) // 'cat' is used uninitialized whenever '&&' condition is false
```
wenzeslaus, line 218: Here, it's actually never used in that case, right? Just making sure cat = 0 (or the previous undefined) does not leak anywhere. (Maybe similar question will apply to other parts, too.)

nilason: [Vect_cat_get()](https://grass.osgeo.org/programming8/vector_2Vlib_2cats_8c.html#a3d9d5db7d953e0a1b36b63e19a91b903j) will set `cat`:

> to first category found or -1 if field was not found

so it doesn’t matter what value we initialize it to (but I’ll change this one to `-1`). The compiler doesn’t know this.

Re the other cases: there are quite a few with if..else or similar that covers all options. Again, the compiler doesn’t know this.

nilason:
This is now ready, I'll proceed to merge soon if there are no objections.

marisn (approved):

PR #2168, opened 2022-02-04
Fix some code scanning CodeQL overflow warnings

nilason:
Touching files with PR #2165 triggered CodeQL overflow warnings.

This PR addresses these issues:

- Multiplication result may overflow 'int' before it is converted to 'long'.
- Multiplication result may overflow 'int' before it is converted to 'unsigned long'.

### DIFF
--- a/lib/ogsf/gs3.c
+++ b/lib/ogsf/gs3.c
@@ -1156,7 +1156,7 @@ int Gs_update_attrange(geosurf * gs, int desc)
 		return (-1);
 	    }
 
-	    size = gs->rows * gs->cols;
+	    size = (long)gs->rows * gs->cols;
 	    p = (char *)tb->cb;
 	    SET_MINMAX(p, nm, size, min, max);
 	}
--- a/lib/ogsf/gs3.c
+++ b/lib/ogsf/gs3.c
@@ -1086,7 +1086,7 @@ int Gs_load_3dview(const char *vname, geoview * gv, geodisplay * gd,
  */
 int Gs_update_attrange(geosurf * gs, int desc)
 {
-    long size;
+    size_t size;
     float min, max;
     typbuff *tb;
     struct BM *nm;
@@ -1115,7 +1115,7 @@ int Gs_update_attrange(geosurf * gs, int desc)
 	if (tb->ib) {
 	    int *p;
 
-	    size = gs->rows * gs->cols;
+	    size = (size_t)gs->rows * gs->cols;
 	    p = tb->ib;
 	    INIT_MINMAX(p, nm, size, min, max, found);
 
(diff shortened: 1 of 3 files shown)


lib/ogsf/gs3.c
```diff
@@ -1156,7 +1156,7 @@ int Gs_update_attrange(geosurf * gs, int desc)
 		return (-1);
 	    }
 
-	    size = gs->rows * gs->cols;
+	    size = (long)gs->rows * gs->cols;
```
metzm, line 1159:
```suggestion
	    size = (size_t)gs->rows * gs->cols;
```
Please use `size_t` for `size` and type casting because long is on MS Windows always a 32 bit integer type: no difference to int.

There are many more instances of `size = gs->rows * gs->cols;` in this file which would need to be fixed.

Be aware of the macros `INIT_MINMAX()` and `SET_MINMAX()` in this file.

nilason: I chose to cast to  `long` because `size` is defined as such:
https://github.com/OSGeo/grass/blob/b8ae3bb419368dc331133ccca2f110f43c71922a/lib/ogsf/gs3.c#L1089
and this was the least invasive approach.

Alternative fix might be changing `size` to `int` or `size_t`. (Note `gs->rows` and `gs->cols` are both ints), but as I'm not at all familiar with this code I can't tell with what consequences.

By the look of it `size` is really used as "size" (ie. size_t) in the macros.

> There are many more instances of size = gs->rows * gs->cols; in this file which would need to be fixed.

Yes I realised that (this is true also for the other files of this PR) and I plan to do them too. For some reason only these cases I initially addressed were errored.

wenzeslaus: I think the answer to "long may mean different bits on different platforms" are [fixed width integer types (since C99)](https://en.cppreference.com/w/c/types/integer).

nilason: What is needed for this particular piece of code, as I see it, is that the variable `size` is big enough for the result of multiplication of two ints, which is exactly what `size_t` is.

wenzeslaus: The rows, cols variables are signed while size_t is unsigned, so a signed type would still make sense to me here. I'm not sure about C world, but in C++ world there seems to be some shift to using signed types for everything including indices, although the C++ standard library uses unsigned types extensively. This comes from C++ Core Guidelines (which are work in progress).

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es106-dont-try-to-avoid-negative-values-by-using-unsigned
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-unsigned
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es102-use-signed-types-for-arithmetic

metzm: > I think the answer to "long may mean different bits on different platforms" are [fixed width integer types (since C99)](https://en.cppreference.com/w/c/types/integer).

Here, `size_t` is the right choice because it is used to calculate a pointer address.

Unrelated to this code: some time ago I have introduced `grass_int64` as an integer type that is guaranteed to be a 64 bit integer on all platforms. It seems that MS Windows does not know `int64_t` but `__int64`? I.e. C99 is not fully implemented in MS Windows regarding integer types.

nilason: I believe we all agree on `size_t` is the one for this case.

Side note re windows and fixed sized integers, seems MS has finally caught up on this. I’ll return with a PR for addressing that.

nilason:
If there are no objections, I’ll merge this soon.

Issue #2003, opened 2021-12-06
[Bug] r.watershed gives nodata value for 0 value cells with -a flag


ar-siddiqui:
**Describe the bug**
r.watershed returns nodata value for 0 value cells with -a flag. Without -a flag, it will return 0 value for the same cells.

**To Reproduce**

1. Run r.watershed with elevation and flow from attached zip using -a flag
2. There will be no data values at various places in the resulting flow accumulation layer

**Expected behavior**
0 values being recorded as 0 values like it is without -a flag

**Screenshots**
![image](https://user-images.githubusercontent.com/53625184/144925461-7ff8732a-26f2-48d3-ba7e-76fabe48e9f2.png)

**System description (please complete the following information):**

- Operating System: Windouws
- GRASS GIS version [7.8.6]

[grass_issue.zip](https://github.com/OSGeo/grass/files/7663626/grass_issue.zip)

PR #2169, opened 2022-02-04
r.watershed: Keep 0 accumulation with -a as 0 as opposed to null

HuidaeCho:
This PR addresses #2003 by saving 0 accumulated cells as 0, not as null, for the -a flag. It does not change the default behavior without -a.

Its side effect is that cells outside non-null input are also saved as 0 (was null before this PR), which is in fact more consistent with the default output without the -a flag. I think with/without -a, null input cells should produce null output cells (TODO).

### DIFF
--- a/raster/r.watershed/ram/close_maps.c
+++ b/raster/r.watershed/ram/close_maps.c
@@ -34,34 +34,25 @@ int close_maps(void)
 	    G_message("Writing out only positive flow accumulation values.");
 	    G_message
 		("Cells with a likely underestimate for flow accumulation can no longer be identified.");
-	    for (r = 0; r < nrows; r++) {
-		Rast_set_d_null_value(dbuf, ncols);	/* reset row to all NULL */
-		for (c = 0; c < ncols; c++) {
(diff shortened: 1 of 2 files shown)


metzm (approved):
Looks good to me

ar-siddiqui:
@HuidaeCho thanks for taking care of the original issue. Yes, I also noticed this as well that without -a flag original null cells are being converted to 0 value. This is lesser of a problem because the user can use a raster calculator this way `(input raster * 0) + grass output` to get the null values back.

Having said that can r.watershed do that internally as a last step to fix nulls being converted to 0? Perhaps this should be discussed in a separate issue.

HuidaeCho:
> @HuidaeCho thanks for taking care of the original issue. Yes, I also noticed this as well that without -a flag original null cells are being converted to 0 value. This is lesser of a problem because the user can use a raster calculator this way `(input raster * 0) + grass output` to get the null values back.
> 
> Having said that can r.watershed do that internally as a last step to fix nulls being converted to 0? Perhaps this should be discussed in a separate issue.

Check https://github.com/OSGeo/grass/pull/2221.

Issue #1794, opened 2021-08-13
[Feat] Automatically cancel outdated CI jobs


ninsbl:
**Is your feature request related to a problem? Please describe.**
There are 18 CI jobs that run for each PR. Some of them take quite some time to finish. When a PR is updated, queued and in progress CI jobs for that PR continue running, even when the code is outdated. First when those are finished, new actions are run for the new committ(s). So the queue of pending CI jobs (and thus waiting time for results) can become quite long.

**Describe the solution you'd like**
Github actions now feature `concurrency` (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency) meaning that pending CI jobs for a PR that gets updated can be automatically canceled. That should be implemented esp. for long running actions (like tests on Ubuntu and OSGeo4W).

**Describe alternatives you've considered**
There are a couple of possible alternative solutions:
- https://github.com/marketplace/actions/skip-duplicate-actions#cancel-outdated-workflow-runs
- https://github.com/marketplace/actions/cancel-workflow-action

wenzeslaus:
This seems to be well suited for the case when you are developing against the CI and you are doing a lot of changes in a short period of time. However, when you are developing locally, you may appreciate knowing that until a certain commit it was passing, but then it is not. On the other hand, in this case, you could wait with pushing for the jobs to finish in order to see the result.

Now, the jobs can be canceled manually, but the push request jobs run in the OSGeo org, not your fork, so I assume you need to have high access rights in order to do that.

ninsbl:
Fair enough. Tests are also running in my fork when I push to my feature branch. There I can cancel runs and the queue is not a problem... Thus closing this...

wenzeslaus:
With more PRs and more actions/workflows/jobs this is more desired than having precise history of runs. It may also more reflect the PR branch is used most of the times. Reopening.

PR #2170, opened 2022-02-05
CI: Cancel in-progress runs on update for PR

wenzeslaus:
Using GitHub Actions concurrency, runs from the previous commits are canceled within the same group which is defined by workflow and head_ref/branch name. It explicitly checks for event being pull request and if it is not, it uses commit hash which makes it unique, so commits to the main and release branches will not be canceled.

See also #1794.


wenzeslaus:
Replaced by #2868.

nilason:
Sorry, I overlooked this one totally!

wenzeslaus:
That's fine. There was something wrong with this one and I never went back.

PR #2171, opened 2022-02-05
d.legend: Fixed possible buffer overflow in sprintf() formatting.

BadAssassin:


display/d.legend.vect/main.c
```diff
@@ -256,7 +256,7 @@ int main(int argc, char **argv)
         sep = G_option_to_separator(opt_sep);
         file_name = opt_input->answer;
         if (!file_name)
-            G_fatal_error(_("Unable to open input file <%s>"), file_name);
+            G_fatal_error(_("Unable to open input legend file."));
```
wenzeslaus, line 259: I agree that the formatting is wrong, but isn't file_name set because opt_input->answer is set? I think the library ensures that the string, when set, is also non-empty, but I would have to check and perhaps that's something automated checkers can't see anyway which would be another reason for checking emptiness.

wenzeslaus:
FYI, the failed test on Ubuntu 18.04 is unrelated. Sometimes they get stuck and are canceled (after 6 hours = 360 mins).

nilason:
FYI there is an issue #2156 dealing with compiler warnings and a PR #2163 addressing the unused results specifically.

PR #2172, opened 2022-02-06
grass.gunittest: Set max time (timeout) per test

wenzeslaus:
This sets max execution time (timeout) per test file. After that, the test is ended. The subprocess timeout mechanism is used for that. Sub-processes of the test process may (likely will) keep running, but the testing continues.

Test which times out is counted as a failure. Additional value is passed around in the code to indicate timeout, but it is not treated as a separate type of result. However, it is clearly visible in the outputs.

The terminology (such as timeout versus max seconds per test) and the interface (config and messages) is similar to pytest.

### DIFF
--- a/python/grass/gunittest/main.py
+++ b/python/grass/gunittest/main.py
@@ -224,6 +224,7 @@ def main():
     invoker = GrassTestFilesInvoker(
         start_dir=start_dir,
         file_anonymizer=FileAnonymizer(paths_to_remove=[abs_start_dir]),
+        timeout=config.getfloat("timeout", None),
     )
     # TODO: remove also results dir from files
     # as an enhancemnt
--- a/.gunittest.cfg
+++ b/.gunittest.cfg
@@ -37,3 +37,10 @@ exclude =
     vector/v.out.lidar/testsuite/test_v_out_lidar.py
     vector/v.what/testsuite/test_vwhat_layers.py
     vector/v.what/testsuite/test_vwhat_ncspm.py
+
+# Maximum time for execution of one test file (not a test function)
+# after which test is terminated (which may not terminate child processes
+# from that test).
+# Using 5 minutes as maximum (average test time should be much shorter,
+# couple seconds per file, median should be around one second).
+timeout = 300
--- a/python/grass/gunittest/invoker.py
+++ b/python/grass/gunittest/invoker.py
@@ -64,7 +64,11 @@ def update_keyval_file(filename, module, returncode):
     keyval["name"] = module.name
     keyval["tested_dir"] = module.tested_dir
     if "status" not in keyval.keys():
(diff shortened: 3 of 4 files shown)


wenzeslaus:
For the record, I used capture_output which is only since Python 3.7, so in the Ubuntu 18.04  test which has Python 3.6, I got:

```
   File "/usr/lib/python3.6/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
TypeError: __init__() got an unexpected keyword argument 'capture_output'
```

wenzeslaus:
All checks passed. Let's see how this change to tests behaves in CI on main.

If there is a test getting stuck, we should see that test (or tests) failing time to time. This will be like before, but instead of run being canceled after 6 hours, the tests will fail. If the issue is elsewhere, we will still see a run being canceled after 6 hours time to time.

python/grass/gunittest/main.py
```diff
@@ -224,6 +224,7 @@ def main():
     invoker = GrassTestFilesInvoker(
         start_dir=start_dir,
         file_anonymizer=FileAnonymizer(paths_to_remove=[abs_start_dir]),
+        timeout=config.getfloat("timeout", None),
```
neteler, line 227: This seems to cause troubles, see

https://github.com/OSGeo/grass-addons/runs/5097252329?check_suite_focus=true

```
 Executing <python3 -m grass.gunittest.main --grassdata /home/runner --location nc_spm_full_v2alpha2 --location-type nc --min-success 60> ...
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/runner/install/grass81/etc/python/grass/gunittest/main.py", line 248, in <module>
    sys.exit(main())
  File "/home/runner/install/grass81/etc/python/grass/gunittest/main.py", line 227, in main
    timeout=config.getfloat("timeout", None),
AttributeError: 'dict' object has no attribute 'getfloat'
Execution of <python3 -m grass.gunittest.main --grassdata /home/runner --location nc_spm_full_v2alpha2 --location-type nc --min-success 60> finished.
Cleaning up temporary files...
```

wenzeslaus: Thanks for reporting this. I'll look into that and submit a fix. Python 3.8 is what is used here, but `'dict' object has no attribute 'getfloat'` is pretty clear. Probably happens when the config file is not there.

wenzeslaus: Fix should be in #2180.

PR #2175, opened 2022-02-06
CI: enable skip hint [ci skip] in GitHub actions

neteler:
Using `github.event.head_commit.message` the GitHub actions may be skipped with `[ci skip]` or `[skip ci]` in the commit message.

Inspired by GDAL, https://github.com/OSGeo/gdal/commit/635faad3988df177896b16cfa6da6a346ec7253e (author: @rouault).

Fixes #2174

PR #1778, opened 2021-08-05
Ignore paths in CodeQL

wenzeslaus:
* Do not run CodeQL on PR when only HTML, Markdown, or txt is changed.
* Ignore bin and dist dir in the CodeQL analysis.
* Add custom CodeQL config to have paths-ignore settings.
* Ignore node_modules dir also in Git.

PR #2176, opened 2022-02-06
CI: Ignore in GHA CI yml files HTML, Markdown, or Text changes

neteler:
Do not run CI runners on PR when only HTML, Markdown, or txt is changed (inspired by #1778).

Closes #2175 (replacement)

Fixes issue #2174

### DIFF
--- a/.github/workflows/additional_checks.yml
+++ b/.github/workflows/additional_checks.yml
@@ -8,6 +8,10 @@ name: Additional Checks
 on:
   - push
   - pull_request
+      paths-ignore:
+        - "**/*.html"
+        - "**/*.md"
+        - "**/*.txt"
 
 jobs:
   additional-checks:
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -3,6 +3,10 @@ name: Python Black Formatting
 on:
   - push
   - pull_request
+      paths-ignore:
+        - "**/*.html"
+        - "**/*.md"
+        - "**/*.txt"
 
 jobs:
   run-black:
--- a/.github/workflows/centos.yml
+++ b/.github/workflows/centos.yml
@@ -3,6 +3,10 @@ name: CentOS
 on:
   - push
   - pull_request
+      paths-ignore:
+        - "**/*.html"
+        - "**/*.md"
+        - "**/*.txt"
 
 jobs:
   build:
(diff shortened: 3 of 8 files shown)


neteler:
For some reason the `flake8` and `gnu99` pipelines are not triggered...

neteler:
@wenzeslaus is there a way to trigger the CI? I'd like to merge this PR but that's blocked...

wenzeslaus:
The tests never executed. Similar situation is in #2170. Both change workflows, but otherwise unrelated, so perhaps there was or is an issue with GitHub Actions. What seemed to help in #2170 at one point was to make additional commits to the branch. You can also rebase or merge main to add a real commit without making real changes.

wenzeslaus (changes requested):
What is the reasoning for ignores in each of these workflows? For example, compilation was many times broken in the past due to bad HTML. Additionally, some checks may check some of these file types in the future; thinking Super-Linter and Markdown here.

I don't know if a required check is ignored, whether a PR merge is allowed or impossible?

The syntax is wrong (which might be part of the reason the workflows don't show up at all). It is not valid YAML (try e.g., http://www.yamllint.com/ which gives: _mapping values are not allowed in this context_). `paths-ignore` is part of a mapping (dictionary), but there is no mapping here, only a list of strings.

[GitHub documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths) has the following example which suggests that you need to switch the whole content of `on` into a mapping:

```yaml
on:
  push:
    paths-ignore:
      - 'docs/**'
```

WordPress has a good example of a workflow in their [trac](https://core.trac.wordpress.org/browser/trunk/.github/workflows/coding-standards.yml).

neteler:
Thanks for your explanations.
However, as being not experiences enough with the workflow code I'll drop this PR (I was just annoyed to wait several hours for being able to merge a small update in `doc/howto_release.md`).

wenzeslaus:
To address your issue, I changed the settings to not require status checks, but only require pull requires for the main branch. This was not possible in GitHub in the past.

Now you don't have to wait for any checks to run because no checks are required. The idea is, of course, that you wait for the relevant checks.

Previously, with Flake8 and one GCC checks required, we ensured that the code at least compiles and it is a valid Python. Black or other checks would practically finish in the meantime too raising red flags if needed. Now you have the responsibility to determine what is needed.

Generally, PRs feature asynchronous workflows, so the expectation is that the code will sit in the PR for some time.

neteler:
Thanks, @wenzeslaus - so: "handle with care!".

PR #2206, opened 2022-02-13
 grass.jupyter: Use weakref.finalize to manage TemporaryDirectory

wenzeslaus:
Although tempfile.TemporaryDirectory uses weakref.finalize to delete the directory and its documentation does mention deletion during garbage-collection and during interpreter shutdown, the intention in its source code is that a with statement or explicit cleanup call is made because otherwise it generates a warning about implicit deletion. Pylint correctly generates warning suggesting with statement (consider-using-with).

In the future, we may consider implementing context-manager and cleanup method to the render classes, but even with that we need to stretch the existence of the temporary directory beyond one method, so the with statement is not applicable. For now, the classes are using only weakref.finalize which is sufficient for a notebook where the object life time is linked with the notebook's kernel life time which is relatively short. The Pylint warning is disabled and an explanatory comment is above.

### DIFF
--- a/python/grass/benchmark/results.py
+++ b/python/grass/benchmark/results.py
@@ -48,7 +48,7 @@ def save_results_to_file(results, filename):
     See :func:`save_results` for details.
     """
     text = save_results(results)
-    with open(filename, "w") as file:
+    with open(filename, "w", encoding="utf-8") as file:
         file.write(text)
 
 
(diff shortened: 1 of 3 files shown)


chaedri (commented):
@wenzeslaus Looks good to me! I'll implement something similar in the space time dataset visualizations and interactive displays.

PR #2208, opened 2022-02-14
Jupyter: Improved GRASS-folium Integration

chaedri:
Restructures and improves interactive mapping in Jupyter Notebooks with grass.jupyter:

* New `Raster` and `Vector` classes allow direct interaction with folium:
```
import folium
m = folium.Map()
gj.Raster("elevation").add_to(m)
gj.Vector("roadsmajor").add_to(m)
m
```
* Restructures InteractiveMap to use new Raster and Vector classes
* Create new ReprojectionRenderer that reprojects and exports raster/vectors for folium

Binder Test Link:
https://mybinder.org/v2/gh/chaedri/grass/GRASS-folium?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fgrass_jupyter.ipynb


veroandreo:
After running the linked notebook, I have a some comments/questions potentially not relevant to the PR itself, but well, I post them anyway :)
- is it possible to add also the openstreetmap basemap in the InteractiveMap display to then have the option to switch among basemaps?
- can the folium map be shrunk to a size similar to the others? (I guess these are folium.Map() properties, no?)
- I'd move the "Save InteractiveMap as HTML" section above the folium.Map() example. I was confused that the map saved was `fig` and not `m`. Also because `m` cannot be saved in the same way as the InteractiveMap object.

chaedri:
@veroandreo  Those are quite relevant here! I just pushed some changes to address these.

> After running the linked notebook, I have a some comments/questions potentially not relevant to the PR itself, but well, I post them anyway :)
> 
> * is it possible to add also the openstreetmap basemap in the InteractiveMap display to then have the option to switch among basemaps?

This is a simple addition - the most recent commit includes a `tiles` parameter for `InteractiveMap`. You should be able to use
```
fig=gj.InteractiveMap(tiles="OpenStreetMap") 
```
Here's a related excerpt from the docstring:
```
tiles parameter is passed directly to folium.Map() which supports several
built-in tilesets (including "OpenStreetMap", "Stamen Toner", "Stamen Terrain",
"Stamen Watercolor", "Mapbox Bright", "Mapbox Control Room", "CartoDB positron",
"CartoDB dark_matter") as well as custom tileset URL (i.e.
"http://{s}.yourtiles.com/{z}/{x}/{y}.png"). For more information, visit
folium documentation:
https://python-visualization.github.io/folium/modules.html
```


> * can the folium map be shrunk to a size similar to the others? (I guess these are folium.Map() properties, no?)

The size of the in-line figure is controlled by folium.Figure (which would be a parent of the Map). I updated the example to include the Figure creation and, therefore, a smaller map. I also changed the example to include a few other useful folium functions.

> * I'd move the "Save InteractiveMap as HTML" section above the folium.Map() example. I was confused that the map saved was `fig` and not `m`. Also because `m` cannot be saved in the same way as the InteractiveMap object.

Great catch - I just swapped these around.

python/grass/jupyter/interact_display.py
```diff
@@ -40,113 +145,68 @@ class InteractiveMap:
...
+    ):
         """Creates a blank folium map centered on g.region.
 
+        tiles parameter is passed directly to folium.Map() which supports several
+        built-in tilesets (including "OpenStreetMap", "Stamen Toner", "Stamen Terrain",
+        "Stamen Watercolor", "Mapbox Bright", "Mapbox Control Room", "CartoDB positron",
+        "CartoDB dark_matter") as well as custom tileset URL (i.e.
+        "http://{s}.yourtiles.com/{z}/{x}/{y}.png"). For more information, visit
+        folium documentation:
+        https://python-visualization.github.io/folium/modules.html
```
wenzeslaus, line 164: I think you need to support API_key at least to support _Cloudmade and Mapbox custom tilesets_ (https://pypi.org/project/folium/0.1.4/). Perhaps this is for a different PR/issue.

doc/notebooks/grass_jupyter.ipynb
```diff
@@ -261,6 +261,48 @@
...
+    "\n",
+    "# Do some cool folium stuff!\n",
+    "# Like make a tooltip\n",
+    "tooltip = \"Click me!\"\n",
+    "# and add a marker\n",
+    "folium.Marker(\n",
+    "    [35.781608,-78.675800], popup=\"<i>Point of Interest</i>\", tooltip=tooltip\n",
+    ").add_to(m)\n",
+    "\n",
+    "\n",
```
wenzeslaus, line 298: Remove one line?

python/grass/jupyter/interact_display.py
```diff
@@ -40,113 +145,68 @@ class InteractiveMap:
...
-
-        import folium
+        import folium  # pylint: disable=import-outside-toplevel
 
         self._folium = folium
 
+        # Store region settings
+        self._use_region = use_region
+        self._saved_region = saved_region
+
```
wenzeslaus, line 179: Seems like you don't need them as attributes.

wenzeslaus (approved):

wenzeslaus (changes requested):
Sorry, I meant to select request changes.

Also, a direct test of reprojection renderer would be nice to have.

python/grass/jupyter/tests/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    tmp_path = tmp_path_factory.mktemp("simple_dataset")
+    location = "test"
+    gs.core._create_location_xy(tmp_path, location)  # pylint: disable=protected-access
+    with grass_setup.init(tmp_path / location):
+        gs.run_command("g.proj", flags="c", epsg=26917)
+        gs.run_command("g.region", s=0, n=80, w=0, e=120, b=0, t=50, res=10, res3=10)
+        # Vector
+        vector_name = "point"
+        full_vector_name = f"{vector_name}@PERMANENT"
```
wenzeslaus, line 24: When the value is used only as a parameter in a call where the items are already keyword arguments line by line, it seems to me that removing the variable and doing the work in the keyword argument assignment is more readable because it saves lines and variables one needs to track.

python/grass/jupyter/tests/reprojection_renderer_test.py
```diff
@@ -0,0 +1,32 @@
...
+    bbox = renderer.get_bbox()
+    assert bbox == [[90, 180], [-90, -180]]
+
+
+# render_raster produces filename and new_bounds
+def test_render_raster(simple_dataset):
+    """Check render_raster returns image and bbox"""
+    renderer = gj.ReprojectionRenderer()
+    filename, bbox = renderer.render_raster(simple_dataset.raster_name)
+    print(bbox)
```
wenzeslaus, line 20: Extra print.

wenzeslaus (changes requested):

wenzeslaus (approved):

Issue #2210, opened 2022-02-15
[Feat] Allow to shrink Data Catalog pane in the Single-window GUI


veroandreo:
Currently, there is a limit to shrink the Data Catalog pane in the Single-window GUI (the limit is given by the Data Catalog width, see https://github.com/OSGeo/grass/pull/2207#issuecomment-1039227848). Would it be possible to allow to shrink it further? I try to exemplify the desired solution in the screenshot below. Being able to shrink data catalog further than the current limit would allow to maximize space for map display.

![image](https://user-images.githubusercontent.com/20075188/154165197-ac817bb2-af36-41c4-b64e-0e7ac1db7165.png)

PR #2216, opened 2022-02-16
wxGUI Single-Window: Make Tools pane wider after startup

lindakladivova:
Before:
Screenshot from 2022-02-17 01-21-45

After:
Screenshot from 2022-02-17 01-18-37
The width is the same for both side panels.

### DIFF
--- a/gui/wxpython/main_window/frame.py
+++ b/gui/wxpython/main_window/frame.py
@@ -157,7 +157,7 @@ class GMFrame(wx.Frame):
 
         # set pane sizes according to the full screen size of the primary monitor
         self.PANE_BEST_SIZE = tuple(t // 5 for t in self.size)
-        self.PANE_MIN_SIZE = tuple(t // 6 for t in self.size)
+        self.PANE_MIN_SIZE = tuple(t // 8 for t in self.size)
 
         # create widgets and build panes
         self.CreateMenuBar()
--- a/gui/wxpython/main_window/frame.py
+++ b/gui/wxpython/main_window/frame.py
@@ -679,7 +679,8 @@ class GMFrame(wx.Frame):
 
         # Set the size for automatic notebook
         pane = self._auimgr.GetPane(notebook)
-        pane.BestSize(self.PANE_MIN_SIZE)
+        pane.BestSize(self.PANE_BEST_SIZE)
+        pane.MinSize(self.PANE_MIN_SIZE)
 
         wx.CallAfter(self.datacatalog.LoadItems)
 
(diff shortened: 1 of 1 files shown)


veroandreo:
I changed milestone to 8.2 as this lives only in main so far, right?

petrasovaa:
@veroandreo does this address #2210? I am thinking they could be shrinked even more, but no strong opinion.

lindakladivova:
> I changed milestone to 8.2 as this lives only in main so far, right?

Oh sure, the previous plan :-) I switched 0 and 2 by mistake.

lindakladivova:
> @veroandreo does this address #2210? I am thinking they could be shrinked even more, but no strong opinion.

I have tried one eighth of display size as a minimum pane size. It looks as follows:
Screenshot from 2022-02-20 07-22-29

probably better, I would go with this. What do you think?

veroandreo:
> > @veroandreo does this address #2210? I am thinking they could be shrinked even more, but no strong opinion.
> 
> I have tried one eighth of display size as a minimum pane size. It looks as follows: Screenshot from 2022-02-20 07-22-29
> 
> probably better, I would go with this. What do you think?

Yes! I like this one better! For more space, as you said earlier, the user can directly minimize.

lindakladivova:
> Before: Screenshot from 2022-02-17 01-21-45
> 
> After: Screenshot from 2022-02-17 01-18-37 The width is the same for both side panels.

And the right panel could be probably a bit narrower for the startup. What do you think?
Screenshot from 2022-02-20 08-00-12
I first thought that the initial pane sizes should be the same on both sides but there is an empty space on the right then...

veroandreo:
> And the right panel could be probably a bit narrower for the startup. What do you think? Screenshot from 2022-02-20 08-00-12 I first thought that the initial pane sizes should be the same on both sides but there is an empty space on the right then...

It looks good to me, yes. Esp. since the focus of the right panel at startup is on the Tools tab. Then, it can be resized when there's an output in the Console tab or if the user wants to use the Python terminal tab.

petrasovaa:
This doesn't seem to differ a lot from the original state. I would make the right pane wider because often one of the first things I do is to resize it. Also when I tested it looks like the right pane doesn't have a minimal width, while the left does, was that the intention? I don't have a strong opinion whether they should be both the same width, although I think it would make things simpler. Ultimately I think the layout should be saved in settings (out of scope of this PR).

lindakladivova:
> This doesn't seem to differ a lot from the original state. I would make the right pane wider because often one of the first things I do is to resize it. Also when I tested it looks like the right pane doesn't have a minimal width, while the left does, was that the intention? I don't have a strong opinion whether they should be both the same width, although I think it would make things simpler. Ultimately I think the layout should be saved in settings (out of scope of this PR).

We can make the size of both panels the same, I do not have a strong opinion either. The minimum size could be also set up. There is just a little problem that you actually set up the minimum size only for this particular "startup" notebook but when a user creates a different notebook the minimum size is not applied.

petrasovaa (approved):

PR #2218, opened 2022-02-17
t.rast.neighbors: transfer semantic label

metzm:
transfer semantic label from input to output if existing

### DIFF
--- a/temporal/t.rast.neighbors/t.rast.neighbors.py
+++ b/temporal/t.rast.neighbors/t.rast.neighbors.py
@@ -205,10 +205,10 @@ def main():
                 new_map.set_semantic_label(semantic_label)
         elif new_labels == "method":
             if semantic_label is not None:
-                semantic_label = "%s_%s" % (semantic_label, method)
+                semantic_label = f"{semantic_label}_{method}"
             else:
                 semantic_label = method
-                new_map.set_semantic_label(semantic_label)
+            new_map.set_semantic_label(semantic_label)
         new_maps.append(new_map)
 
         mod = copy.deepcopy(neighbor_module)
--- a/temporal/t.rast.neighbors/t.rast.neighbors.py
+++ b/temporal/t.rast.neighbors/t.rast.neighbors.py
@@ -77,6 +77,17 @@
 # %end
 
 # %option
+# % key: semantic_labels
+# % type: string
+# % options: input,method
+# % description: Set semantic labels
+# % descriptions: input;copy semantic labels from input to output;method;append method name to input label if existing, otherwise use method name
+# % answer: input
+# % required: no
+# % multiple: no
+# %end
+
+# %option
 # % key: nprocs
 # % type: integer
 # % description: Number of r.neighbor processes to run in parallel
@@ -120,6 +131,7 @@ def main():
     method = options["method"]
     nprocs = options["nprocs"]
     time_suffix = options["suffix"]
+    new_labels = options["semantic_labels"]
 
     # Make sure the temporal database exists
     tgis.init()
@@ -188,8 +200,15 @@ def main():
             dbif=dbif,
         )
         semantic_label = map.metadata.get_semantic_label()
-        if semantic_label is not None:
-            new_map.set_semantic_label(semantic_label)
+        if new_labels == "input":
+            if semantic_label is not None:
+                new_map.set_semantic_label(semantic_label)
+        elif new_labels == "method":
+            if semantic_label is not None:
+                semantic_label = "%s_%s" % (semantic_label, method)
+            else:
+                semantic_label = method
+                new_map.set_semantic_label(semantic_label)
         new_maps.append(new_map)
 
         mod = copy.deepcopy(neighbor_module)
--- a/temporal/t.rast.neighbors/t.rast.neighbors.py
+++ b/temporal/t.rast.neighbors/t.rast.neighbors.py
@@ -187,6 +187,9 @@ def main():
             overwrite=overwrite,
             dbif=dbif,
         )
+        semantic_label = map.metadata.get_semantic_label()
+        if semantic_label is not None:
+            new_map.set_semantic_label(semantic_label)
         new_maps.append(new_map)
 
         mod = copy.deepcopy(neighbor_module)
--- a/temporal/t.rast.neighbors/t.rast.neighbors.html
+++ b/temporal/t.rast.neighbors/t.rast.neighbors.html
@@ -16,8 +16,10 @@ processing using a SQL WHERE statement. The number of CPU's to be used
 for parallel processing can be specified with the <em>nprocs</em>
 option to speedup the computation on multi-core system.
 <p>
-Semantic labels present in the input dataset will be transferred to
-the output dataset.
+Semantic labels are needed to relate output raster maps to input raster
+maps. E.g. with <em>method=stddev</em>, the user needs to know the
+spatial extent, the time stamp and the semantic label to determine
+which stddev map corresponds to which input map.
 
 <h2>EXAMPLE</h2>
 
--- a/temporal/t.rast.neighbors/t.rast.neighbors.html
+++ b/temporal/t.rast.neighbors/t.rast.neighbors.html
@@ -15,6 +15,9 @@ The user can select a subset of the input space time raster dataset for
 processing using a SQL WHERE statement. The number of CPU's to be used
 for parallel processing can be specified with the <em>nprocs</em>
 option to speedup the computation on multi-core system.
+<p>
+Semantic labels present in the input dataset will be transferred to
+the output dataset.
 
 <h2>EXAMPLE</h2>
 


wenzeslaus (changes requested):
Can you please add user documentation for this? Perhaps also reasoning in the final/merge commit why it makes sense like you did in the other PR if it is not clear from the user doc.

metzm:
An example is application of a neighborhood modifier to a STRDS with e.g. Landsat or Sentinel bands where semantic labels must be preserved.

metzm:
Explanation for my recent changes to the handling of semantic labels in GRASS: I try to make GRASS compatible with [openEO](https://openeo.org/).

wenzeslaus:
Thanks for the context. Since r.neighbors can do all kinds of operations on all kinds of data, I would like to avoid overspecializing the code for one particular case.

I agree that it makes sense to preserve it in some cases, but always? What if you run with `method=range` or `method=stddev`?

metzm:
> I agree that it makes sense to preserve it in some cases, but always? What if you run with `method=range` or `method=stddev`?

`t.rast.neighbors` processes each raster map separately. Why should the semantic label be dismissed in case of e.g. `method=range` or `method=stddev`? The result is still specific to each raster band with its semantic label.

ninsbl:
> If you have a STRDS with several different semantic labels, you need to know to which input e.g. a stddev map belongs. If you discard semantic labels, the stddev maps become meaningless: stddev of what?
> 
> If you are unhappy with the original semantic labels, e.g. `red`, `green`, `nir`, you could afterwards rename them to e.g. `red_stddev`, `green_stddev`, `nir_stddev`.

Fully agreed. Is it simple / straight forward for a user to change the semantic label for a set of maps afterwards? If not, that may be a reason to have `semantic_label` as a non-required option where users can modify it right away if desired? Applies to some extent to `t.rast.mapcalc` too, and such an option could be reused in `t.rast.ndvi` as well, no?)... Not to make things complicated, just making sure that use cases and user experience are considered throughout...

veroandreo:
> > If you have a STRDS with several different semantic labels, you need to know to which input e.g. a stddev map belongs. If you discard semantic labels, the stddev maps become meaningless: stddev of what?
> > If you are unhappy with the original semantic labels, e.g. `red`, `green`, `nir`, you could afterwards rename them to e.g. `red_stddev`, `green_stddev`, `nir_stddev`.
> 
> Fully agreed. Is it simple / straight forward for a user to change the semantic label for a set of maps afterwards? 

Within the temporal framework, AFAIU, the user would need to re-register all maps in the new strds with the desired semantic label, i.e., create the text file with map names, dates and edited labels. 

I am not sure that using r.support over all maps already within a strds will have any effect. This is the case for r.timestamp for example. The date of a map with a timestamp set using r.timestamp, will be read when registering it for the first time within a strds. However to edit that date, the user needs to use t.register, as there's no communication between the C and python interfaces, i.e., if a map with timestamp is already registered in a strds, changing its time stamp with r.timestamp won't have any effect within the strds. My guess is the same will happen with r.support.

> If not, that may be a reason to have `semantic_label` as a non-required option where users can modify it right away if desired? Applies to some extent to `t.rast.mapcalc` too, and such an option could be reused in `t.rast.ndvi` as well, no?)... Not to make things complicated, just making sure that use cases and user experience are considered throughout...

It's probably not too difficult, but kind of annoying to have to rename afterwards and have no control of what the module will do with semantic labels.

metzm:
In case of `t.rast.neighbors` and `t.rast.mapcalc` I prefer to transfer semantic labels as they are without modification. This is particularly important if there are several different semantic labels in a STRDS.

Regarding renaming, I have already something like `t.rast.renamelabels` which takes a list of old labels and a list of new labels and updates the TGIS db accordingly.

veroandreo:
> In case of `t.rast.neighbors` and `t.rast.mapcalc` I prefer to transfer semantic labels as they are without modification. This is particularly important if there are several different semantic labels in a STRDS.

Can this behavior be made optional at least with a flag or so? Even if there will be a module to rename semantic labels (this could be used internally indeed), if the newly created STRDS is a different variable, why force users to have to rename semantic labels afterwards?
 
> Regarding renaming, I have already something like `t.rast.renamelabels` which takes a list of old labels and a list of new labels and updates the TGIS db accordingly.

This sounds good :)

metzm:
> 
> [...] if the newly created STRDS is a different variable, why force users to have to rename semantic labels afterwards?

Users don't **have to** rename labels, the **can** rename them if they are unhappy with the names. The primary purpose of these labels is to distinguish between different bands, or more generally variables, in a STRDS. These labels can be used, but they can just as well be ignored.

My main point is that not transferring these labels is a serious loss of information and thus a bug.

wenzeslaus:
> In this case you can rename the semantic label.

Rename is fine if I don't want the default, but I'm saying that the default should not be wrong and saying that stddev of red is still red is generally wrong.

The "copy label" behavior might correct in some case, but not in general, so it should not be the default. Flag or option can drive that. 

> The semantic label `red` for stddev of a red band still tells you where the stddev is coming from and helps you to distinguish stddev of a `red` band from stddev of a `green` band. Thus the label is correct because it holds information about the lineage.

This shifts or narrows the meaning of semantic label to lineage where `red` means red something derived from red. However, semantic label generally says what is the meaning of the values, e.g., `landsat_8_4` is band 4 of Landsat 8, so user can assume that this data can be used anywhere where band 4 of Landsat 8 is expected. Assigning `landsat_8_4` to something derived from `landsat_8_4` just because it was derived from it breaks that assumption, so it can't be the default, although it might be what user wants in some cases (e.g. for smoothing or depending how broad the definition of `landsat_8_4` is, e.g., in terms of value ranges).

> My problem with auto-generated semantic labels for `t.rast.neighbors` (or `t.rast.mapcalc`) is that they are different from the input labels. This makes it unnecessarily complicated to track lineage in a longer workflow.

If you are using semantic labels to track lineage, then that's a specific application of semantic labels with your specific definition of what data with one semantic label have in common. Providing your own semantic label throughout the workflow seems like one way to implement it (i.e., modules take it as parameter) or another way is to explicitly request to copy them (so like this PR, but opt-in).

> You would need to know and duplicate the mechanism to auto-generate semantic labels of a module in your workflow. Otherwise queries and selections based on semantic label will fail.

The best way how to achieve this is to be able to ask the module explicitly what you want.

> Therefore I heavily prefer users renaming labels themselves, then they still know exactly the lineage of different raster maps in a STRDS.

However, we don't want to create misleading labels.

> As long as they are not secretly dropped or modified.

You can make the label behavior more explicit with an option (instead of a flag or some in-code only default), for example, `output_semantic_labels=copy` (your desired behavior AFAIU), `output_semantic_labels=auto` (gives `prefix_method`), `output_semantic_labels=none` (drops the labels). This is extensible to applying other labeling schemes in the future such more complex auto-generated ones as what @huhabla suggested for t.rast.mapcalc or using existing 3rd party dictionaries/glossaries/ontologies.

wenzeslaus:
> Let's assume that the purpose of a semantic label is to distinguish different raster maps in a STRDS by their different labels. In this case, appending the method name to different semantic labels provides no information at all (in the sense of information theory) because all input labels get the same method name appended. Keep It Sufficiently Simple.

> For me, the purpose of a semantic label is less to provide an exact description of a raster map, but more to distinguish different raster maps in a STRDS. I would also be happy with meaningless but different semantic labels like `label1`, `label2`, `label3`, `label4`... in a STRDS, kept throughout a longer workflow.

Both original band references with the associated metadata and later semantic labels applied in context of classification do provide information about what the data is. "This is something I can apply Landsat 8 band 4 metadata to." "This is red, so I can apply to it signatures I previously derived from red." The name semantic label suggests that, too.

The implementation gives you the freedom to use them just to number the bands or to use them to track the original source of the data, but that's not their only purpose. The default behavior of the modules should respect semantic labels carry meaning (semantic) and that they are not just numbering the bands.

metzm:
I agree, that "semantic label generally says what is the meaning of the values", but the meaning and interpretation of a label is up to the user. Most importantly, the semantic label is not the only source of information of a raster map. The history of the raster map and the specific workflow used to generate the STRDS with this raster map are also important sources of information.

I object against abusing the semantic label to be the sole source of information about a raster map.

The user who defined a workflow knows best what to do with a STRDS with different semantic labels. But this requires that the user has a chance to know which semantic labels are present in a STRDS and how they relate to previous steps in the workflow. The user knows best what the label `label2` means. This requires that labels are not secretly dropped or modified.

The default behavior of the modules should respect semantic labels in order to allow the user to distinguish between maps with different semantic labels. The emphasis here is: there are different groups of raster maps in a STRDS that need group-specific treatment. The user knows best about the group-specific treatment.

metzm:
Uuh, `t.rast.ndvi` is a bad example because it has been specifically written as an implementation of the openEO process [ndvi](https://processes.openeo.org/#ndvi). That is the reason for the mechanism to auto-detect sensor and band name which I would rather like to avoid. Usually, I would make the two input semantic labels a requirement, the user needs to provide these semantic labels to calculate a normalized difference. The name ndvi is a bit misleading because any band combination (any two semantic labels) can be provided to calculate any normalized difference, not just NDVI.

If a user provides a STRDS with only stddev values to `t.rast.ndvi`, this is neither correct or wrong but the user's decision. The user as the creator of the STRDS with only stddev values knows its contents and the semantic labels, if any, present in this STRDS.

`t.rast.ndvi` uses one STRDS with two different semantic labels to create a new STRDS, or to add the normalized difference with the provided new semantic label to the input STRDS. This is the official openEO definition. In any case, two semantic labels are used to create something new, thus copying one of the input semantic labels does not make sense here. The user decides on a new semantic label. In openEO terminology, this is a reduction of the dimension "bands". `t.rast.series` would be a reduction of the dimension "time".

`t.rast.neighbors` does not combine semantic labels, instead each semantic label is processed independently. Therefore it makes sense to copy input semantic labels to output. It does not make sense to add the method, e.g. stddev, to the semantic labels because the resultant STRDS contains only one single method. That is, you know the output STRDS contains (only!) stddev and if you have the input semantic labels, you know the meaning of each raster map in the STRDS.

I could add a new option `semantic_label` to `t.rast.neighbors` with the options none (do not assign a semantic label), copy (copy from input to output), method (append method to input semantic labels or use method as semantic label if there are no input semantic labels). Obviously, I'm saying that `semantic_label=copy` should be the default in order to automatically trace the lineage of stddev raster maps and keep semantic labels consistent in a longer workflow.

BTW, openEO is the reason why semantic labels, formerly band names, were implemented in GRASS at all.

wenzeslaus:
> I could add a new option `semantic_label` to `t.rast.neighbors` with the options none (do not assign a semantic label), copy (copy from input to output), method (append method to input semantic labels or use method as semantic label if there are no input semantic labels). Obviously, I'm saying that `semantic_label=copy` should be the default in order to automatically trace the lineage of stddev raster maps and keep semantic labels consistent in a longer workflow.

That's what I'm suggesting except that 1) I think there should be a different default, 2) you don't need to implement `semantic_label=method` when you don't need it, and 3) I'm not sure if `semantic_label` is a good name choice here in case user wants to set the label themselves as it was suggested elsewhere (would another option be needed and if yes, how to name it?).

The default value issue can be side-stepped by having no value as the default and ask user to set that explicitly if they care one way or another (which will give us an option to decide the proper default behavior later).

metzm:
Some background on the reason for the existence of semantic labels, formerly band names, in GRASS:

Band names where introduced to GRASS to add support for the openEO dimension "bands", which should be maintained with the renaming of band names to semantic labels.

In openEO. all dimensions are transferred from input to output without modifications, unless a specific dimension is modified by a module. That means semantic labels (the "bands" dimension) are (have to be) transferred without modification unless a semantic label is added or dropped.

Since support for the openEO dimension "bands" was the original reason to implement band names in GRASS, I want to keep support for openEO as the default. That also means making support for the openEO "bands" dimension not the default but optional needs strong arguments which were not provided yet. Consequently, my latest commit to this PR makes dropping support for the openEO "bands" dimension optional.

The argument `semantic_labels=method` is useful if you want to perform different modifications to pixel values of a STRDS with different processes and later on merge all these STRDS's into a single STRDS where you need to be able to distinguish between different raster maps by their semantic label.

temporal/t.rast.neighbors/t.rast.neighbors.py
```diff
@@ -187,6 +199,16 @@ def main():
             overwrite=overwrite,
             dbif=dbif,
         )
+        semantic_label = map.metadata.get_semantic_label()
+        if new_labels == "input":
+            if semantic_label is not None:
+                new_map.set_semantic_label(semantic_label)
+        elif new_labels == "method":
+            if semantic_label is not None:
+                semantic_label = "%s_%s" % (semantic_label, method)
```
wenzeslaus, line 208: You can just use an f-string here. Introduced in Python 3.6. Preferred way of formatting now for simple strings.

```python
semantic_label = f"{semantic_label}_{method}"
```

temporal/t.rast.neighbors/t.rast.neighbors.py
```diff
@@ -187,6 +199,16 @@ def main():
...
+            else:
+                semantic_label = method
+                new_map.set_semantic_label(semantic_label)
```
wenzeslaus, line 211: One indent level less here.

wenzeslaus:
Thank you for the last changes which allow users can get expected behavior (whatever is expected in each case) when they use `semantic_label` explicitly. For any scripted/automated workflow which relies on a specific behavior, I strongly advise to use `semantic_label` explicitly instead of relying on a default value.

> ...In openEO. all dimensions are transferred from input to output without modifications, unless a specific dimension is modified by a module.
> ...
> Since support for the openEO dimension "bands" was the original reason to implement band names in GRASS, I want to keep support for openEO as the default.

Nothing from this was discussed in #1868 (which is the closest thing to actual specification of what semantic labels are) or in the original #63 as far as I can tell. Given what we have reached in #1868 and with changes to classification signatures, I can't give this an approving review at this point. (I have dismissed my previous review which is now outdated.)

I have left some comments to fix/improve the current code in the PR.

metzm:
Thanks for looking at the new code, I applied your suggested changes.

Regarding #1868, I welcomed the change away from "band_reference" because this term was too strict. My only (already fulfilled) requirement was that users can set any arbitrary label and are not restricted to the old `i.bands`. I did / do have no strong opinion on what "band_reference" is called instead. Just one remark: "semantic_label" is dangerously similar to the already existing "[semantic_type](https://github.com/OSGeo/grass/blob/main/python/grass/temporal/abstract_space_time_dataset.py#L186)".

Of course in automated workflows options should be set explicitly because default answers/behaviours are not always desired. For example, the `suffix` option in temporal modules does not always work with the default answer `gran`, causing a fatal error.

veroandreo:
I'm in favor of merging this, now that different potential use cases are considered and integrated. Thanks @metzm for the option and the explanations ;)

Unfortunately this discussion happens only now and not in the original PR proposing the feature #63 (I take my part of responsibility). It was not clear then which was the roadmap of changes, nor the implications (cascade effect) for all other (raster) temporal modules. I hence wonder now, should other t.* modules get this option as well? Say, `t.rast.series`, `t.rast.aggregate`, `t.rast.accumulate`, `t.rast.algebra`, `t.rast.mapcalc`. Some will change the temporal dimension, others, not necessarily. How to deal with those? Do you have plans to tackle other t.* modules (apart from #2219)?

metzm:
Right, some more background information in #63 would have been helpful.

I might want to tackle other t modules, but most probably only if I can add information (transfer semantic labels from input) without modifying the results, as in this PR for `t.rast.neighbors` or in PR #2219 for `t.rast.mapcalc`.

A special case is `t.rast.series` where it would make sense to calculate temporal statistics for each band separately. It does not really make sense to e.g. calculate temporal averages for all red, green, blue, ... bands together, as is done now.

Currently I am looping over all semantic labels in a STRDS, extract each label to a temporary STRDS, perform the original module, assign the current input label to the temporary output STRDS and merge all results to one final output STRDS with input semantic labels preserved. Lots of scripting to avoid information loss.

metzm:
This PR transfers semantic information from input to output.

Without this semantic information, it is not possible to interpret the results and relate output to input.

PR #2220, opened 2022-02-18
r.buildvrt: check if all input maps are NULL-only

### LINKED ISSUE #2217: [Bug] r.buildvrt: Missleading error message when tiles contain only NoData
**Describe the bug**
When patching tiles that by chance only contain NoData, `r.buildvrt` fails with an error message like:

```
WARNING: Illegal filename <?A??U>. Character <?> not allowed.
ERROR: Raster map <?A??U> not found in mapset <?A??U>
```

**To Reproduce**

```
r.mapcalc expression="a=null()"
r.mapcalc expression="b=null()"
r.buildvrt input=a,b output=c
```

**Expected behavior**
Ideally, r.buildvrt creates an empty map as other modules do. That way at least a map is produced and processes don`t fail here. 
If that requires more effort, the error message in this case should be rather something like:
```
ERROR: Only NULL pixels in input raster map
```

- Ubuntu 18.04
- GRASS GIS 8.1.dev

neteler:
Do not continue if all input raster maps only contain no-data.
Feel free to suggest a better error message.

**To Reproduce**

```
r.mapcalc expression="a=null()"
r.mapcalc expression="b=null()"
r.buildvrt input=a,b output=c
```

Old behaviour:
```
GRASS nc_spm_08_grass7/user1:grass_main > r.buildvrt input=a,b output=c --verbose
Input map <a@user1> is all NULL, skipping
Input map <b@user1> is all NULL, skipping
WARNING: Illegal filename <���>. Character <�> not allowed.
ERROR: Raster map <���> not found in mapset <���>
```

New error message:
```
GRASS nc_spm_08_grass7/user1:grass_main > r.buildvrt input=a,b output=c --verbose
Input map <a@user1> is all NULL, skipping
Input map <b@user1> is all NULL, skipping
ERROR: All input maps only contain NULL, cannot continue
```

Fixes #2217

### DIFF
--- a/raster/r.buildvrt/main.c
+++ b/raster/r.buildvrt/main.c
@@ -263,7 +263,7 @@ int main(int argc, char *argv[])
 	if (strcmp(nsresstr, tnsresstr) != 0)
 	    G_warning(_("Input ns resolutions are different"));
 	if (strcmp(ewresstr, tewresstr) != 0)
-	    G_warning(_("Input ns resolutions are different"));
+	    G_warning(_("Input ew resolutions are different"));
 
 	if (cellhd.north < p->cellhd.north)
 	    cellhd.north = p->cellhd.north;
--- a/raster/r.buildvrt/main.c
+++ b/raster/r.buildvrt/main.c
@@ -219,6 +219,8 @@ int main(int argc, char *argv[])
     	}
 	num_inputs = j;
     }
+    if (num_inputs == 0)
+	G_fatal_error(_("All input maps only contain NULL, cannot continue"));
 
     qsort(inputs, num_inputs, sizeof(struct input), cmp_wnd);
 


ninsbl (approved):
Great. Thanks for the quick fix!

metzm (approved):
Thanks for the fix!

neteler:
> While you are at it, you may probably also change the warning message here:
> 
> https://github.com/OSGeo/grass/blob/b9c78120ce3e272ce34f419161ad073a190ad84d/raster/r.buildvrt/main.c#L266
> 
> which should read:
> `"Input ew resolutions are different"` (and not "Input ns resolutions are different", note the direction ew/ns).

Sure, done in e193fc2f32b18b7dff29966198b7c11421f55c86

neteler:
> (Backport once 8.0.1 is out)

done

Issue #1901, opened 2021-09-28
[Bug] 


ar-siddiqui:
**Describe the bug**
Using r.watershed inside QGIS, I am getting no data value along some streams for flow direction.

I tried the r.fill.dir on the same raster and it does not give any no-data value for the same cells (see the image below, r.watershed is on top of r.fill.dir).

![image](https://i.stack.imgur.com/lmnbf.png)

I tried both sink-filled and non-sink-filled DEMs.

Here are my inputs for r.watershed 

![image](https://user-images.githubusercontent.com/53625184/135150123-04d4238e-d719-426c-9253-ed59661c1ff3.png)

**To Reproduce**
Steps to reproduce the behaviour:

1. Run r.watershed on the attached DEM
2. Run r.fill.dir on the attached DEM
3. Compare the result.
4. There will be no-data values in r.watershed results

**Expected behavior**
The output of r.fill.dir and r.watershed be same / no no-data values in r.watershed results

[inflated_dem.zip](https://github.com/OSGeo/grass/files/7246712/inflated_dem.zip)

**System description (please complete the following information):**

- Operating System: [e.g. Windows]
- GRASS GIS version [e.g. 7.8.4]

ar-siddiqui:
I investigated more to find out flow weight (amount of overland flow per cell) is the one causing the issue. Not sure why, because flow direction has nothing to do with weight.

HuidaeCho:
I don't think r.watershed handles null cells in the flow raster properly. It just assigns `-nan` to null flow cells instead of setting them to 0. See #1949.

Even with this PR, with and without `flow=` generate slightly different drainage outputs. Based on a visual check, without `flow=` seems to generate an incorrect output.

* black text: `flow=` (all 1s)
* blue text: DEM
* green arrows: `r.watershed elev=dem flow=hsg drain=drain` 219 => 210, not => 208?
* red arrows: `r.watershed elev=dem drain=drain_wo`
![image](https://user-images.githubusercontent.com/7456117/137611027-d255fffe-a358-49ea-89b9-5b958ed83ac1.png)

PR #1949, opened 2021-10-17
r.watershed: Handle null flow amount as 0

HuidaeCho:
This PR partly addresses #1901 by assigning 0 flow amount to null flow cells.

PR #2221, opened 2022-02-18
r.watershed: Null accumulation outside elevation, 0 flow from null flow

HuidaeCho:
This PR sets accumulation to null where there is no data in elevation. This behavior is consistent with the disk swap -m flag. It also partly addresses https://github.com/OSGeo/grass/issues/1901 by assigning 0 flow amount to null flow cells.


HuidaeCho:
Merged with https://github.com/OSGeo/grass/pull/1949.

https://github.com/OSGeo/grass/pull/1949 partly addresses https://github.com/OSGeo/grass/issues/1901 by assigning 0 flow amount to null flow cells.

PR #2225, opened 2022-02-20
configure: update to latest config.guess and config.sub

nilason:
This updates to latest `config.guess` and `config.sub`. Not much of a change, but doesn't hurt to update this regularly.

### DIFF
--- a/config.guess
+++ b/config.guess
@@ -1,14 +1,14 @@
 #! /bin/sh
 # Attempt to guess a canonical system name.
-#   Copyright 1992-2021 Free Software Foundation, Inc.
+#   Copyright 1992-2022 Free Software Foundation, Inc.
 
 # shellcheck disable=SC2006,SC2268 # see below for rationale
 
-timestamp='2021-06-03'
+timestamp='2022-01-09'
 
 # This file is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by
-# the Free Software Foundation; either version 3 of the License, or
+# the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
 #
 # This program is distributed in the hope that it will be useful, but
@@ -60,7 +60,7 @@ version="\
 GNU config.guess ($timestamp)
 
 Originally written by Per Bothner.
-Copyright 1992-2021 Free Software Foundation, Inc.
+Copyright 1992-2022 Free Software Foundation, Inc.
 
 This is free software; see the source for copying conditions.  There is NO
 warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE."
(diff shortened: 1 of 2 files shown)


nilason:
Should be backported to 8.0 and 7.8 branch.

neteler:
This usually happens pre RC1:
https://github.com/OSGeo/grass/blob/main/doc/howto_release.md#update-of-configure-base-files

I had considered it then skipped it - if we go for a 8.0.1RC2 then it could be backported but if not urgent I'd prefer 8.0.2 and 7.8.8.

neteler (changes requested):
The `autoconf2.69` step is missing, see
https://github.com/OSGeo/grass/blob/main/doc/howto_release.md#update-of-configure-base-files

nilason:
> This usually happens pre RC1: https://github.com/OSGeo/grass/blob/main/doc/howto_release.md#update-of-configure-base-files
> 
> I had considered it then skipped it - if we go for a 8.0.1RC2 then it could be backported but if not urgent I'd prefer 8.0.2 and 7.8.8.

Perfectly fine for me, not at all urgent.

nilason:
> The `autoconf2.69` step is missing, see
> https://github.com/OSGeo/grass/blob/main/doc/howto_release.md#update-of-configure-base-files

I'm not quite sure what you mean, running `autoconf` doesn't make any difference.

neteler:
> I'm not quite sure what you mean, running `autoconf` doesn't make any difference.

I see - mostly it does but apparently it doesn't affect `configure` in this case. Sorry for the confusion.

nilason:
> > I'm not quite sure what you mean, running `autoconf` doesn't make any difference.
> 
> I see - mostly it does but apparently it doesn't affect `configure` in this case. Sorry for the confusion.

No worries :-) As far as I understand it, the config.guess and config.sub are read/run by configure, but not altering the configure file.

neteler:
Looking at the [history](https://github.com/OSGeo/grass/commits/main/configure) of `configure` it appears that sometimes also this file is affected (e.g., d57f40906f6353bd7d0d305eb4b8649b04deb172, 177467461eaf6fefd27825f176585a9eeefd5749, 
fcffbb5546a72c9f1e3402e129b7415be2e5ca31), i.e. to be updated with `autoconf`.

nilason:
> Looking at the [history](https://github.com/OSGeo/grass/commits/main/configure) of `configure` it appears that sometimes also this file is affected (e.g., [d57f409](https://github.com/OSGeo/grass/commit/d57f40906f6353bd7d0d305eb4b8649b04deb172), [1774674](https://github.com/OSGeo/grass/commit/177467461eaf6fefd27825f176585a9eeefd5749),
> [fcffbb5](https://github.com/OSGeo/grass/commit/fcffbb5546a72c9f1e3402e129b7415be2e5ca31)), i.e. to be updated with `autoconf`.

Non of those changes to configure are caused by updating config.guess or config.sub.
The first two are patches (the same applied twice) to autoconf 2.13:
```
for ac_declaration in \
    ''\
    '#include <stdlib.h>' \
    'extern "C" void std::exit (int) throw (); using std::exit;' \
    'extern "C" void std::exit (int); using std::exit;' \
    'extern "C" void exit (int) throw ();' \
    'extern "C" void exit (int);' \
    'void exit (int);'
 do
...
```
(see eg. https://git.rockylinux.org/staging/rpms/autoconf213/-/blob/r8/SOURCES/autoconf-2.13-c++exit.patch or https://bugzilla.redhat.com/show_bug.cgi?id=18829)

The third is a caused by a previous change in `aclocal.m4` without updating configure.

neteler:
Agreed. But just for the future procedure: shall we treat these updates (config.* vs autoconf) separately? Then the release procedure needs an update.

nilason:
> Agreed. But just for the future procedure: shall we treat these updates (config.* vs autoconf) separately? Then the release procedure needs an update.

I think so, the config.* files should/may be updated before RC release.
In addition, it doesn't hurt running autoconf, to see if there are any patches, or other inconsistencies (like the above mentioned case of aclocal.m4). And if there are any patches, they should be added to `configure.in` too.

neteler (approved):

PR #2226, opened 2022-02-20
r.geomorphon: combine profile parser rules to allow usage as pygrass …

### LINKED ISSUE #2029: Test for r.geomorphon fails due to optional with default confusion

ninsbl:
Backport of #2154
Fixes #2029

* remove defaults and add back parser rules

* adjust manual

* reactivate r.geomorphon test

Co-authored by: Denis Ovsienko <denis@ovsienko.info>

### DIFF
--- a/.gunittest.cfg
+++ b/.gunittest.cfg
@@ -17,7 +17,6 @@ exclude =
     python/grass/temporal/testsuite/unittests_temporal_raster_algebra_equal_ts.py
     python/grass/temporal/testsuite/unittests_temporal_raster_conditionals_complement_else.py
     raster/r.contour/testsuite/testrc.py
-    raster/r.geomorphon/testsuite/test_r_geom.py
     raster/r.in.gdal/testsuite/test_r_in_gdal.py
     raster/r.in.lidar/testsuite/test_base_resolution.sh
     raster/r.in.lidar/testsuite/test_base_resolution.sh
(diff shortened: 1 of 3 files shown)


ninsbl:
@neteler OK to merge into release branch or wait for 8.0.2?

neteler:
> @neteler OK to merge into release branch or wait for 8.0.2?

Well, in the worst case just `r.geomorphon` will be broken... you decide.

ninsbl:
Thanks. I feel pretty save as this change actually brings back tests for r.geomorphon, so it is unlikely that the module would be broken...

PR #2233, opened 2022-02-23
shapelib: compiler warning stringop-overflow

BadAssassin:
GCC complains about correct code, but this is marginally faster and avoids the warning.


neteler:
It might be worth to sync our copy in GRASS GIS to the upstream code in
https://github.com/OSGeo/shapelib/blob/master/dbfopen.c

wenzeslaus (changes requested):
Isn't the operator precedence such that the comparison `<` gets evaluated before assignment `=`?

https://en.cppreference.com/w/c/language/operator_precedence

In many cases, compilers/linters suggest `()` around `=` in `if`.

metzm:
The code in question has been changed in upstream. I suggest to revert the changes to [lib/external/shapelib/dbfopen.c](https://github.com/OSGeo/grass/pull/2233/files#diff-d41d8de9bd21e49ae33d0e16c6ef03099dfc2be2279c5e8b8c08461d6658e5c0), sync to upstream in a separate PR and check again. I would be surprised if there are still any compiler warnings.

PR #2241, opened 2022-02-28
Increase minimum Python version to 3.7

wenzeslaus:
This does not make the code incompatible with Python 3.6, but does not check for Python 3.6 compatibility anymore which makes it possible to include Python >=3.7 only code.

To be clear, the minimum required Python version is now Python 3.7, so using features which are only in 3.7 and above is perfectly fine.

This removes Ubuntu 18.04 test from the CI because 18.04 uses Python 3.6 and therefore 18.04 in its default configuration is now no longer supported.


nilason (approved):
Perfectly fine by me. 👍

neteler (approved):
Thanks!

PR #2245, opened 2022-03-03
CI: Test also without OpenMP and pthreads

wenzeslaus:
This adds an additional build configuration to the Ubuntu workflow which builds code without OpenMP and pthreads and runs tests.
Notably, modules using nprocs parameter should accept nprocs > 1 even when they are not able to parallelize due to a missing dependency.
This check tests that behavior.

Other selected optional dependencies can be added in the future. The current goal is to test code without dependencies which don't influence the primary actions the code performs.


ninsbl:
Would that be needed for MSWindows / OSGeo4W as well?

PR #2247, opened 2022-03-04
grass.jupyter: Test init and session

wenzeslaus:
Adds test for init, session, and finish using pytest. Uses temporary script to test automatic (implicit) cleanup with weakref.finalize.

Depends on #2246.

### DIFF
--- a/python/grass/jupyter/tests/grass_jupyter_session_test.py
+++ b/python/grass/jupyter/tests/grass_jupyter_session_test.py
@@ -4,9 +4,6 @@ import subprocess
 import os
 import sys
 
-import grass.script as gs
-import grass.jupyter as gj
-
 
 # All init tests change the global environment, but we use a separate process
 # only when it is necessary.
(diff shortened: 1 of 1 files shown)


python/grass/jupyter/tests/grass_jupyter_session_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def test_init_finish(tmp_path):
+    """Check that init function works with an explicit session finish"""
+    location = "test"
+    gs.core._create_location_xy(tmp_path, location)  # pylint: disable=protected-access
+    session = gj.init(tmp_path / location)
+    gs.run_command("g.region", flags="p")
+    session_file = os.environ["GISRC"]
+    session.finish()
```
wenzeslaus, line 33: This currently changes global library settings and doesn't restore it.

wenzeslaus: Switched to using subprocess.

PR #2258, opened 2022-03-10
t.rast.list: Add CSV, JSON, YAML outputs

wenzeslaus:
* This adds CSV, JSON, and YAML output and generalizes the other outputs.
* Method is now meant only for different data gathering methods, not for the output formatting. method=cols and method=comma are now obsolete.
* All formatting is avaiable with all methods.
* Most code is shared between the two main branches of code (simple list versus delta/gran methods code).
* Columns can be now specified for any option (previously only for cols/list and partically for comma/line output).
* The change is backwards compatible both on module interface level and on Python API level.

There is no performance penalty for the more general, step-by-step code (at least with smaller data; several tests with _perf_ give all times around 1.7s with the 6 map test).

[Binder link](https://mybinder.org/v2/gh/wenzeslaus/grass/exchange-formats-for-t_rast_list?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fbasic_example.ipynb)

### DIFF
--- a/python/grass/temporal/list_stds.py
+++ b/python/grass/temporal/list_stds.py
@@ -10,19 +10,25 @@ Usage:
     tgis.register_maps_in_space_time_dataset(type, name, maps)
 
 
-(C) 2012-2016 by the GRASS Development Team
+(C) 2012-2022 by the GRASS Development Team
 This program is free software under the GNU General Public
 License (>=v2). Read the file COPYING that comes with GRASS GIS
 for details.
 
 :authors: Soeren Gebbert
+:authors: Vaclav Petras
 """
-from __future__ import print_function
+
+import os
+from contextlib import contextmanager
+import sys
+
+import grass.script as gs
+
 from .core import get_tgis_message_interface, get_available_temporal_mapsets, init_dbif
 from .datetime_math import time_delta_to_relative_time
 from .factory import dataset_factory
 from .open_stds import open_old_stds
-import grass.script as gscript
 
 ###############################################################################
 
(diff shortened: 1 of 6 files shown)


wenzeslaus:
## New usage: Load JSON

```python
import json

import grass.script as gs

result = json.loads(
    gs.read_command("t.rast.list", method="gran", input="precip_abs0", format="json")
)
names = [item["name"] for item in result["data"]]
dates = [item["start_time"] for item in result["data"]]
print(names)
print(dates)
```

## Classic usage: Custom parsing

```python
import grass.script as gs

rows = gs.read_command("t.rast.list", method="gran", input="precip_abs0").splitlines()

new_rows = [row.split("|") for row in rows]
new_array = [list(row) for row in zip(*new_rows)]

for column in new_array:
    if column[0] == "name":
        names = column[1:]
    if column[0] == "start_time":
        dates = column[1:]
print(names)
print(dates)
```

Both take approximately same time, approx 3.33s, with variations bigger than the difference.

wenzeslaus:
This has tests, it doesn't change default behavior, and doesn't seem to have any performance issues, I plan to merge this soon still to be included to 8.2. Comments welcome.

veroandreo:
I'd like to test it, but I can during the weekend only. Is that still good timing @wenzeslaus ?

wenzeslaus:
That would be great. Thank you!

veroandreo:
I've been testing different combinations of methods, columns and format options. I really like that method refers now to data gathering and format to printing. 

I found however, that not all column possibilities can go with all methods, eg., columns min and max are not available for method gran or delta given the type of gathering I assume. Maybe document that (if not yet there) saying something like: columns can be specified for all methods and formats, but note that depending on the method some columns might not be available.

wenzeslaus:
So, it actually needed fix for the checking, some valid columns were rejected and some other caused a traceback (relevant exception from the underlying function, just not appropriate in the module context). I added the fix, doc, and tests.

PR #1391, opened 2021-02-21
Add list of commits for git blame to ignore

wenzeslaus:
This adds a list of commits which one may want to ignore when doing `git blame` (`git blame file.c --ignore-revs-file .git-blame-ignore-revs`).

Formatting changes of Python (with Black), the initial C formatting commit, and couple of other commits are included.

The ignore list is for investigating 'standard' code in common use cases. It won't work when an error was introduced by the formatting change (e.g., by fixing some additional lines manually) or when dealing with CI checks changed in the same commit. Notably, language syntax is often updated when formatting happens (removal of Python 2 unicode strings literals or replacement of backticks syntax in bourne-shells) and a change of formatting is connected to an update in automated checks if there are already in place.

PR #2272, opened 2022-03-20
Add .clang-format

nilason:
This is a suggestion to replace [GNU indent](https://www.gnu.org/software/indent/) with [clang-format](https://clang.llvm.org/docs/ClangFormat.html).

Attempting to bulk process the code base with `util/grass_indent_ALL.sh` as reported in #1630 (and noticed in #2270) a number of problem occurs. This is caused by limited functionality and lack of active development on GNU indent.
However, bulk processing with clang-format causes no error, thanks to the fact that it is backed by a full C and (!) C++ code parser.
Replacing the GRASS formatting tool to clang-format may be the better bet for the future.

Initially I put this up as a draft, for testing and discussion. I tried to get the settings as close as the present style guide, the one notable deviation (I could discover so far) is with alignment of trailing comments.

The `.clang-format` file is created with clang-format version 13.

Note: **only** the addition of the `.clang-format` is the ultimate goal for this PR. The added commit with all the formatted source files are for illustration only.

To test this locally:

`find -E . -regex '.*\.(cpp|hpp|c|h)' -exec clang-format -style=file -i {} \;`

for testing alternative settings see https://clang.llvm.org/docs/ClangFormatStyleOptions.html


wenzeslaus:
+1 given the advantages of clang-format over GNU Indent (also considering we may want to indent modern C++). My experience is that clang-format is not stable in between versions, so to check with exact version in the CI, we need to use [DoozyX/clang-format-lint-action](https://github.com/DoozyX/clang-format-lint-action). We will also need to update it time to time like we did with Black before it was stable (and still will have to update but with less changes in the source code).

BTW, the actual formatting (formatting only) changes should go to `.git-blame-ignore-revs` (#1391) - add `add to git-blame-ignore-revs` to the PR and after merge, create another PR to update the file.

nilason:
Just updated the `.clang-format` file. The closest ClangFormat style to ours is the one of LLVM, the file now reflects that and only the deviating settings are not commented out.

I also limited the number of formatted example files, for better and quicker overview online.

> +1 given the advantages of clang-format over GNU Indent (also considering we may want to indent modern C++). My experience is that clang-format is not stable in between versions, so to check with exact version in the CI, we need to use [DoozyX/clang-format-lint-action](https://github.com/DoozyX/clang-format-lint-action). We will also need to update it time to time like we did with Black before it was stable (and still will have to update but with less changes in the source code).

Yes, I'm aware of this circumstances with ClangFormat versions, but I think the advantages compared to GNU indent are worth it. You're right, Black is a good parallel.

nilason:
Fixed some more settings to get even closer to present "GRASS" style and updated the demo files accordingly.

GRASS style's `BreakBeforeBraces` is now custom, but to use one of the builtin brace breaking styles the following cases (and only those) need to be changed to the following:

*Mozilla:*

```cpp
class UntypedStream
{

...

 } else {
```


*Stroustrup:*

```cpp
enum rule_type {

...

struct Cell_head {

...

typedef union _dglHeapData {

...

extern "C" {

```
	
*Linux:*
	
```cpp
enum AMI_stream_type {

...

struct Cell_head {

...

union {

...

class AMI_STREAM : public UntypedStream
{

...

 } else {

...

 extern "C" {
 
```

*WebKit:*

```cpp
enum AMI_stream_type {

...

struct Cell_head {

...

union {

...

 } else {

...

extern "C" {

```

Even though I'm perfectly ok with present GRASS style, I must say I personally prefer to convert to Stroustrup brace breaking style.

neteler:
@nilason would you mind to redo the reformatting afresh on top of the current state of "main"?

nilason:
> @nilason would you mind to redo the reformatting afresh on top of the current state of "main"?

Certainly! Hopefully this crazy busy summer will soon reach the end. I look forward to return to a more active contribution. :-)

nilason:
> @nilason would you mind to redo the reformatting afresh on top of the current state of "main"?

Just updated the formatted examples on current main.

PR #2280, opened 2022-03-25
Make: add -rpath to LD_SEARCH_FLAGS

metzm:
Currently, the GDAL-GRASS plugin works only if the directory with GRASS libraries is added to `LD_LIBRARY_PATH`, otherwise GRASS libraries don't find each other. This PR, together with the corresponding [GDAL PR 5503](https://github.com/OSGeo/gdal/pull/5503), avoids the need to set `LD_LIBRARY_PATH` or add an entry to `/etc/ld.so.conf.d/`, instead the GDAL-GRASS plugin works out of the box without further adjustments to the system's paths with libraries in non-standard locations.

The crucial change is in `aclocal.m4`. Running `autoconf` version 2.69 resulted in more changes to `configure` than expected, I am not sure if the added `runstatedir` could cause any harm. My tests succeeded so far.

### How to test

Failure with:
 1. build and install GRASS without this PR
 2. build and install the GDAL-GRASS plugin
 3. with the NC sample dataset, try `ogrinfo -so /path/to/grassdata/nc_spm_08/PERMANENT/vector/boundary_county/head`
    The GDAL error is that some shared GRASS libraries are not found.

Success with:
 1. build and install GRASS with this PR
 2. build and install the GDAL-GRASS plugin with the [GDAL PR 5503](https://github.com/OSGeo/gdal/pull/5503)
 3. with the NC sample dataset, try `ogrinfo -so /path/to/grassdata/nc_spm_08/PERMANENT/vector/boundary_county/head`
    No GDAL errors.

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -562,7 +562,7 @@ AC_DEFUN([SC_CONFIG_CFLAGS], [
 	    SHLIB_SUFFIX=".so"
 	    LDFLAGS="-Wl,--export-dynamic"
 	    #LD_SEARCH_FLAGS='-rpath ${LIB_RUNTIME_DIR}'
-	    LD_SEARCH_FLAGS='-Wl,-rpath-link,${LIB_RUNTIME_DIR}'
+	    LD_SEARCH_FLAGS='-Wl,-rpath-link,${LIB_RUNTIME_DIR} -Wl,-rpath,${INST_DIR}/lib'
 	    # TODO: add optional pthread support with any combination of: 
 	    # CFLAGS="$CFLAGS -pthread"
 	    # LDFLAGS="$LDFLAGS -lpthread"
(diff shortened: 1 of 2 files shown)


nilason:
The GRASS´ `configure` is generated by 2.69. `runstatedir` is added with 2.70. Note, the 2.69 was released in 2012. That's why it has been patched on some distributions.

nilason:
With unpatched autoconf 2.69, the result is:

```diff
--- configure.orig
+++ configure
@@ -4032,7 +4032,7 @@
 	    SHLIB_SUFFIX=".so"
 	    SHLIB_LD="${CC} -shared"
             LDFLAGS="-Wl,--export-dynamic"
-            LD_SEARCH_FLAGS='-Wl,-rpath-link,${LIB_RUNTIME_DIR}'
+            LD_SEARCH_FLAGS='-Wl,-rpath-link,${LIB_RUNTIME_DIR} -Wl,-rpath,${INST_DIR}/lib'
             LD_LIBRARY_PATH_VAR="LD_LIBRARY_PATH"
             ;;
         *-pc-cygwin)
```

The macOS build has, as you may have noticed, a similar set of flags for rpath based linking, so it looks sound in my eyes. Cannot test though.

nilason:
As a general note regarding autoconf version. Presently with GRASS, with a readily generated configure, it might be worth considering bumping to 2.71, as this dependency is only restricted for actual updating of the `configure` file (not building GRASS). I.e., only devs need a autoconf 2.71.

metzm:
> The macOS build has, as you may have noticed, a similar set of flags for rpath based linking, so it looks sound in my eyes. Cannot test though.

Yes, but the BSD builds were missing this rpath setting, in case of netbsd it was wrong, now fixed.

nilason (approved):
I'm good with the changes related to the main issue, but I advice you to remove the runstatedir additions. They will be removed after next regeneration by a non-patched autoconf.

metzm:
> I'm good with the changes related to the main issue, but I advice you to remove the runstatedir additions. They will be removed after next regeneration by a non-patched autoconf.

OK, the runstatedir additions have been removed.

neteler:
@metzm May I suggest to backport this to G8.0.x? Otherwise we'll not see this change for a while, as it takes time for the various distros to pick up the upcoming G8.2.x.

neteler:
Backport done as discussed with @metzm

neteler:
Backport done to 7.8.x as discussed with @metzm

PR #2284, opened 2022-03-29
grass.jupyter: Generate API doc for Jupyter package

wenzeslaus:
- Updates documentation intro to reflect that it will be officially included in v8.2.
- Adds grass.jupyter to the libpython index TOC.
- Reduces depth of the TOC to make it more clear.
- Updates copyright year for generated files.
- Generates API doc for all grass.jupyter.
- Adds Sphix documentation build to Pylint check.

The Pylint check builds the code which Sphinx needs and there is already the pytest-based Pylint check so it could potentially do some pytest-related Sphinx tasks.

The API doc is now generated for all parts of the grass package. This make it possible for the doc to say grass.jupyter rather than just jupyter. This duplicates the generated doc for grass.script and others, but in the future this should replace the per-subpackage generated doc.

### DIFF
--- a/python/grass/jupyter/__init__.py
+++ b/python/grass/jupyter/__init__.py
@@ -5,7 +5,7 @@
 # PURPOSE:   Display classes and setup functions for running GRASS GIS
 #            in Jupyter Notebooks
 #
-# COPYRIGHT: (C) 2021 Caitlin Haedrich, and by the GRASS Development Team
+# COPYRIGHT: (C) 2021-2022 Caitlin Haedrich, and by the GRASS Development Team
 #
 #            This program is free software under the GNU General Public
 #            License (>=v2). Read the file COPYING that comes with GRASS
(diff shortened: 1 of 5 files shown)


wenzeslaus:
@chaedri Please check the new text for the new grass.jupyter intro in `__init__.py`.

wenzeslaus:
The generated HTML is available as an artifact. (Newly, Sphinx doc is build in the CI and an artifact is created, so one can now check the build documentation for any PR.)

Screenshot from 2022-03-30 11-30-45

python/grass/jupyter/__init__.py
```diff
@@ -5,20 +5,38 @@
 # PURPOSE:   Display classes and setup functions for running GRASS GIS
 #            in Jupyter Notebooks
 #
-# COPYRIGHT: (C) 2021 Caitlin Haedrich, and by the GRASS Development Team
+# COPYRIGHT: (C) 2021-2022 Caitlin Haedrich, and by the GRASS Development Team
```
chaedri, line 8: Feel free to add your name here (and @petrasovaa)

chaedri (approved):

python/grass/jupyter/__init__.py
wenzeslaus: Done. Added to AUTHORS section, but it might be better to have authors in the documentation and here keep only the file authors. We don't really have a system for that, but something to consider.

PR #2286, opened 2022-03-29
CI: Update to Black 22.3 with fixed Click

wenzeslaus:
A new version of Click broke Black. The new version of Black fixes the issue.

See also https://github.com/psf/black/issues/2964

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -20,7 +20,7 @@ jobs:
         include:
           - os: ubuntu-20.04
             python-version: "3.10"
-            black-version: "22.1.0"
+            black-version: "22.3.0"
 
     runs-on: ${{ matrix.os }}
 


wenzeslaus:
The Black check now passes. CI-only change. No other files changed. Merging.

neteler:
@wenzeslaus don't we need this also in the relbranch80? I see related CI errors in backports... May I backport it?

wenzeslaus:
Backporting this would require backporting #2212 first which didn't seem as a backport candidate by itself. Alternatively, you can remove the Black check (delete the black.yml file) from the release branch because unlike the others, it does not check any functionality.

wenzeslaus:
Another solution is to pin the Click version which is suggested in the Black issue I think.

neteler:
I see the suggestion `"click <8.1.0"` which sounds good to me but is it to be added somehow as a dependency or `pip install ...` ?

wenzeslaus:
Likely pip install before or with Black will work.

wenzeslaus:
This is in 8.2 and the test was dropped from 8.0 branch in 65091789e6812fada2bf90a251be8b08e12048a1. Removing the backport label.

PR #2297, opened 2022-04-08
CI: Update to Black 22.3 with fixed Click

neteler:
Backport of #2286 to avoid CI errors with black/click


wenzeslaus:
The Black test fails because it is either (1) the Click version only or (2) you need to do the whole two PRs but without Click (no need for Click then). ...or (3) remove the Black check since it is not testing functionality anyway. grass7 branch of grass-addons has the same issue.

neteler:
Ok, then I drop this PR.

PR #2298, opened 2022-04-08
black: drop test in release branch

neteler:
Note: Black check is not testing any functionality, so it can be removed from a release branch if needed.

(replaces #2297)


wenzeslaus (approved):
+1

Please add to the commit message the "policy" used to remove it: _Black check is not testing any functionality, so it can be removed from a release branch if needed._

PR #2299, opened 2022-04-10
Replace hardcoded POSIX path dir separator "/" with cross platform HOST_DIRSEP macro

tmszi:
Replace hardcoded POSIX path dir separator "/" with cross platform HOST_DIRSEP macro.

Fixes https://github.com/OSGeo/grass-addons/issues/289.

### DIFF
--- a/display/d.mon/list.c
+++ b/display/d.mon/list.c
@@ -12,14 +12,16 @@
 char *get_path(const char *name, int fpath)
 {
     char tmpdir[GPATH_MAX];
+    char *buff = NULL;
     
     G_temp_element(tmpdir);
-    strcat(tmpdir, "/");
-    strcat(tmpdir, "MONITORS");
     if (name) {
-        strcat(tmpdir, "/");
-        strcat(tmpdir, name);
+        G_asprintf(&buff, "%cMONITORS%c%s", HOST_DIRSEP, HOST_DIRSEP, name);
+    } else {
+        G_asprintf(&buff, "%cMONITORS", HOST_DIRSEP);
     }
+    strcat(tmpdir, buff);
+    G_free(buff);
 
     if (fpath) {
         char ret[GPATH_MAX];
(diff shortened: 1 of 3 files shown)


wenzeslaus:
Microsoft Windows supports forward slash (`/`) as an alternative directory separator ([system.io.path.altdirectoryseparatorchar](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.altdirectoryseparatorchar?view=net-5.0)), so backslash is only for beauty (consistency when displayed) or in specific contexts (calls of some Windows-specific command line programs).

The motivation here seems to be the issue in grass-addons with "which throws an escape character error in R on windows". However, an escape-related error seems to be more related to use of backlash rather than a lack of it. If `\` would not be used, the error would likely not happen.

As for R, the best practices for R I have seen recently (can't find a link) argued for using forward slashes on all platforms (saying to make them work on macOS and Linux, so they actually seemed written from Windows point of view). This seems to fit with, e.g., [this blog post](https://datacornering.com/how-to-fix-windows-path-to-use-it-in-r/).

Also, you probably noticed, OSGeo4W / windows-2019 build and tests:
- main branch: From them 218 files (81%) were successful and 52 files (19%) failed.
- this branch: From them 157 files (58%) were successful and 113 files (42%) failed.

tmszi:
> Microsoft Windows supports forward slash (`/`) as an alternative directory separator ([system.io.path.altdirectoryseparatorchar](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.altdirectoryseparatorchar?view=net-5.0)), so backslash is only for beauty (consistency when displayed) or in specific contexts (calls of some Windows-specific command line programs).

Yes, you are absolutely right, we don't have to change anything.

> The motivation here seems to be the issue in grass-addons with "which throws an escape character error in R on windows". However, an escape-related error seems to be more related to use of backlash rather than a lack of it. If `\` would not be used, the error would likely not happen.
> 

Only one thing, same dir separator in path (`...\.tmp/unknown\999.3` should be `\.tmp\unknown\999.3`)

e.g. 

```
C:\Users\IEUser>g.tempfile pid=999
C:\Users\IEUser\Documents\grassdata\world_latlong_wgs84\PERMANENT\.tmp/unknown\999.3
```

wenzeslaus:
Based on the discussion above, specifically that `/` works on platfroms, this PR is not needed. The original issue has a different cause.

PR #2306, opened 2022-04-14
Restructure TimeSeriesMap

chaedri:
* Restructure TimeSeriesMap so methods are consistent with other grass.jupyter classes by removing `overlay` and `baselayer` in favor of adding layers to map in order they are called with `__getattr__`  magic.
* Update notebook example

For example:
```
precip_map = gj.TimeSeriesMap(use_region=True)
precip_map.add_raster_series("precip_sum_2010")
precip_map.d_legend(color="black", at=(10,40,2,6)) #Add legend
precip_map.d_vect(map="boundary_county", fill_color="none")
precip_map.d_barscale()
precip_map.show()
precip_map.save("image.gif")
```

[Binder](https://mybinder.org/v2/gh/chaedri/grass/timeseries-restructure?labpath=doc%2Fnotebooks%2Fgrass_jupyter.ipynb)

### DIFF
--- a/python/grass/jupyter/timeseriesmap.py
+++ b/python/grass/jupyter/timeseriesmap.py
@@ -186,6 +186,8 @@ class TimeSeriesMap:
         :param str timeseries: name of space-time dataset
         :param bool fill_gaps: fill empty time steps with data from previous step
         """
+        if self._timeseries_added and self.timeseries != timeseries:
+            raise AttributeError("Cannot add more than one space time dataset")
         self._element_type = "strds"
         check_timeseries_exists(timeseries, self._element_type)
         self.timeseries = timeseries
@@ -206,6 +208,8 @@ class TimeSeriesMap:
         :param str timeseries: name of space-time dataset
         :param bool fill_gaps: fill empty time steps with data from previous step
         """
+        if self._timeseries_added and self.timeseries != timeseries:
+            raise AttributeError("Cannot add more than one space time dataset")
         self._element_type = "stvds"
         check_timeseries_exists(timeseries, self._element_type)
         self.timeseries = timeseries
@@ -311,6 +315,13 @@ class TimeSeriesMap:
         (i.e. time_slider or animate). Can be time-consuming to run with large
         space-time datasets.
         """
+
+        if not self._timeseries_added:
+            raise RuntimeError(
+                "Cannot render space time dataset since none has been added."
+                "Use TimeSeriesMap.add_raster_series() or "
+                "TimeSeriesMap.add_vector_series() to add dataset"
+            )
 
         # Make base image (background and baselayers)
         # Random name needed to avoid potential conflict with layer names
--- a/python/grass/jupyter/timeseriesmap.py
+++ b/python/grass/jupyter/timeseriesmap.py
@@ -154,7 +154,6 @@ class TimeSeriesMap:
         self.timeseries = None
         self._element_type = None
         self._fill_gaps = None
-        self._bgcolor = "white"
         self._legend = None
         self._base_layer_calls = []
         self._overlay_calls = []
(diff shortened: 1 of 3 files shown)


chaedri:
I changed the background function from `set_background_color` to `d_background` so that it matches the style of the other display elements (i.e. `d_rast` adds a raster). An alternative would've been to use `d_erase` through the `__get_attr__` magic but this seems counterintuitive for new GRASS GIS users: `example_map.d_erase(bgcolor="white")`.

If #2282 is merged, I think we can remove this function entirely.

python/grass/jupyter/timeseriesmap.py
```diff
@@ -394,14 +390,14 @@ def view_image(date):
         # Return interact widget with image and slider
         widgets.interact(view_image, date=slider)
 
-    def animate(
+    def save(
         self,
+        filename=None,
```
wenzeslaus, line 395: None for filename does not longer make sense, no? Also the tests fails in relation to that.

python/grass/jupyter/timeseriesmap.py
```diff
@@ -137,8 +120,8 @@ class TimeSeriesMap:
 
     >>> img = TimeSeriesMap("series_name")
     >>> img.d_legend()  # Add legend
-    >>> img.time_slider()  # Create TimeSlider
-    >>> img.animate()
+    >>> img.show()  # Create TimeSlider
+    >>> img.save()
```
wenzeslaus, line 124: Unlike animate(), save() does not make sense without parameter because it would go to a temporary directory.

python/grass/jupyter/timeseriesmap.py
```diff
@@ -410,25 +406,26 @@ def animate(
...
-        import IPython.display  # pylint: disable=import-outside-toplevel
 
         # Render images if they have not been already
         if not self._layers_rendered:
             self.render()
 
         # filepath to output GIF
-        if not filename:
+        if filename:
+            assert filename.endswith(".gif"), "filename must end in '.gif'"
```
wenzeslaus, line 427: This is possible, but asserts are usually not used for input checking in Python because they can be optimized out.

python/grass/jupyter/timeseriesmap.py
```diff
@@ -410,25 +406,26 @@ def animate(
...
+        else:
```
wenzeslaus, line 428: Creating the file in temporary dir does not make sense anymore for _save_.

python/grass/jupyter/timeseriesmap.py
```diff
@@ -101,32 +101,15 @@ def collect_layers(timeseries, element_type, fill_gaps):
...
-        if not shutil.which(grass_module):
-            raise AttributeError(_("Cannot find GRASS module {}").format(grass_module))
-
-        def wrapper(**kwargs):
-            self.calls.append((grass_module, kwargs))
-
-        return wrapper
+def check_timeseries_exists(timeseries):
+    """Check that timeseries is time space dataset"""
+    test = gs.read_command("t.list", where=f"name='{timeseries}'")
```
wenzeslaus, line 106: Now it can pass if you ask for vector, but have raster of the same name.

python/grass/jupyter/timeseriesmap.py
```diff
@@ -101,32 +101,15 @@ def collect_layers(timeseries, element_type, fill_gaps):
...
-        grass_module = name.replace("_", ".")
-        # Assert module exists
-        if not shutil.which(grass_module):
-            raise AttributeError(_("Cannot find GRASS module {}").format(grass_module))
-
-        def wrapper(**kwargs):
-            self.calls.append((grass_module, kwargs))
-
-        return wrapper
+def check_timeseries_exists(timeseries):
```
wenzeslaus, line 104: It could take name and element_type and return boolean. The calling function could then give a message with the right type mentioned.

Alternatively, the message can be `temporal dataset {name} of type {element_type}` or something like that.

wenzeslaus (changes requested):
Looking good in general. The failing test should be resolved by _save_ with a required parameter I guess. The dataset existence check improvement can be left for later.

python/grass/jupyter/timeseriesmap.py
chaedri: Good catch.

wenzeslaus (changes requested):
I get an error in the example notebook. The rest looks good.

Screenshot from 2022-04-15 21-38-27

...I know, we really need to test the notebooks in the CI.

wenzeslaus (approved):

PR #2308, opened 2022-04-15
jupyter notebooks: use gs.parse_command() for g.region

neteler:
This PR replaces `gs.run_command()` with `gs.parse_command()` in order to print `´g.region` output into jupyter notebook instead of shell terminal.

### DIFF
--- a/doc/notebooks/grass_jupyter.ipynb
+++ b/doc/notebooks/grass_jupyter.ipynb
@@ -43,7 +43,7 @@
     "session = gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
     "\n",
     "# Set computational region to the elevation raster.\n",
-    "gs.parse_command(\"g.region\", raster=\"elevation\")"
+    "gs.parse_command(\"g.region\", raster=\"elevation\", flags=\"pg\"))"
    ]
   },
   {
@@ -195,7 +195,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "gs.parse_command(\"g.region\", save=\"myregion\", n=224000, s=222000, w=633500, e=637300)\n",
+    "gs.run_command(\"g.region\", save=\"myregion\", n=224000, s=222000, w=633500, e=637300)\n",
     "myregion_map = gj.Map(saved_region=\"myregion\")\n",
     "myregion_map.d_rast(map=\"elevation\")\n",
     "myregion_map.d_rast(map=\"lakes\")\n",
--- a/doc/notebooks/grass_jupyter.ipynb
+++ b/doc/notebooks/grass_jupyter.ipynb
@@ -43,7 +43,7 @@
     "session = gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
     "\n",
     "# Set computational region to the elevation raster.\n",
-    "gs.run_command(\"g.region\", raster=\"elevation\")"
+    "gs.parse_command(\"g.region\", raster=\"elevation\")"
    ]
   },
   {
@@ -195,7 +195,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "gs.run_command(\"g.region\", save=\"myregion\", n=224000, s=222000, w=633500, e=637300)\n",
+    "gs.parse_command(\"g.region\", save=\"myregion\", n=224000, s=222000, w=633500, e=637300)\n",
     "myregion_map = gj.Map(saved_region=\"myregion\")\n",
     "myregion_map.d_rast(map=\"elevation\")\n",
     "myregion_map.d_rast(map=\"lakes\")\n",
--- a/doc/notebooks/grass_jupyter.ipynb
+++ b/doc/notebooks/grass_jupyter.ipynb
@@ -43,7 +43,7 @@
     "session = gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
     "\n",
     "# Set computational region to the elevation raster.\n",
-    "gs.parse_command(\"g.region\", raster=\"elevation\", flags=\"pg\"))"
+    "gs.parse_command(\"g.region\", raster=\"elevation\", flags=\"pg\")"
    ]
   },
   {
--- a/doc/notebooks/solar_potential.ipynb
+++ b/doc/notebooks/solar_potential.ipynb
@@ -45,7 +45,7 @@
     "session = gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
     "\n",
     "# Set computational region to elevation raster\n",
-    "gs.run_command(\"g.region\", raster=\"elevation@PERMANENT\", flags=\"pg\")"
+    "gs.parse_command(\"g.region\", raster=\"elevation@PERMANENT\", flags=\"pg\")"
    ]
   },
   {
--- a/doc/notebooks/hydrology.ipynb
+++ b/doc/notebooks/hydrology.ipynb
@@ -46,7 +46,7 @@
     "session = gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
     "\n",
     "# Set computational region to elevation raster\n",
-    "gs.run_command(\"g.region\", raster=\"elevation\", flags=\"pg\")"
+    "gs.parse_command(\"g.region\", raster=\"elevation\", flags=\"pg\")"
    ]
   },
   {
--- a/doc/notebooks/temporal.ipynb
+++ b/doc/notebooks/temporal.ipynb
@@ -74,7 +74,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "gs.run_command(\"g.region\", raster=\"elev_state_500m\")"
+    "gs.parse_command(\"g.region\", raster=\"elev_state_500m\")"
    ]
   },
   {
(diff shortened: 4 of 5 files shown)


chaedri:
Here's a [Binder link](https://mybinder.org/v2/gh/neteler/grass/j_notebooks_gregion_parse?labpath=doc%2Fnotebooks%2Fgrass_jupyter.ipynb) for testing.

doc/notebooks/grass_jupyter.ipynb
```diff
@@ -43,7 +43,7 @@
     "session = gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
     "\n",
     "# Set computational region to the elevation raster.\n",
-    "gs.run_command(\"g.region\", raster=\"elevation\")"
+    "gs.parse_command(\"g.region\", raster=\"elevation\")"
    ]
   },
```
chaedri, line 48: It's nice that it prevents it from printing to the shell but it also prints an empty dictionary as the cell output. Perhaps we could add the print flag so at least the dictionary is full?

chaedri (commented):
Here's a [Binder link](https://mybinder.org/v2/gh/neteler/grass/j_notebooks_gregion_parse?labpath=doc%2Fnotebooks%2Fgrass_jupyter.ipynb) for testing.

doc/notebooks/solar_potential.ipynb
```diff
@@ -45,7 +45,7 @@
     "session = gj.init(\"../../data/grassdata\", \"nc_basic_spm_grass7\", \"user1\")\n",
     "\n",
     "# Set computational region to elevation raster\n",
-    "gs.run_command(\"g.region\", raster=\"elevation@PERMANENT\", flags=\"pg\")"
+    "gs.parse_command(\"g.region\", raster=\"elevation@PERMANENT\", flags=\"pg\")"
```
wenzeslaus, line 48: The change is only relevant here and in viewshed_analysis.ipynb. Both are probably copy-pastes from some other stuff.

wenzeslaus (changes requested):
I think there is no need to print in the first place. Then there is no need for parse command.

If you want people to check the region, add a separate code cell and a Markdown cell to explain what you want them to see.

neteler:
> I think there is no need to print in the first place. Then there is no need for parse command.

Well, if you look at https://gis.stackexchange.com/ or elsewhere, one of the most common user mistakes is to have the computational region setting wrong. Hence it is in my experience a good idea to also print the region settings when using `g.region`.
I agree that the parse command is somewhat disruptive in the flow - I would prefer to use a more pythonic way here than calling commands (use pygrass?).

doc/notebooks/grass_jupyter.ipynb
neteler: Done in 396de52283dc39053831bed48fee0be94477b593

wenzeslaus: Screenshot from 2022-04-15 19-30-16

wenzeslaus (changes requested):

wenzeslaus:
> ...one of the most common user mistakes is to have the computational region setting wrong. Hence it is in my experience a good idea to also print the region settings when using `g.region`.

I don't know how this is going to improve the situation. Won't the user just see a bunch of numbers?

Screenshot from 2022-04-15 19-29-43

Notably, the above starts with `projection: '99'` and `zone: '0'`. Then it contains strange things like `ewres`. I don't think that pretty/plain text output would particular improve the situation.

> ...the parse command is somewhat disruptive in the flow...

Back to what I was saying before, if you think it is worth it to interrupt the flow and teach people check the region, then add a separate code cell with getting the region preceded by a Markdown cell explaining what and why you want to do and what you want them to see (that can be possibly split to the following Markdown cell).

wenzeslaus:
> ...parse command is somewhat disruptive in the flow - I would prefer to use a more pythonic way here than calling commands (use pygrass?).

Me too and I'm not sure what that is :-((.

There is `gs.region()` which gives a dict which already has numbers as numbers. I think the bad part there is that in terms of API usage it is specific just to _g.region_ and not really covering any other case.

doc/notebooks/grass_jupyter.ipynb
wenzeslaus: Note that the notebooks are not tested in the CI. Only manually locally or in Binder.

It is possible to do that and also to generate HTMLs from them. We need to test them, that's for sure, although propagating errors is challenging, so manual check in Binder may still be needed.

Maybe generating the HTMLs is needed too. ...If done also on a release server, maybe we can have better links for the documentation. ...Totally unrelated to this PR.

neteler: > Note that the notebooks are not tested in the CI. Only manually locally or in Binder.

Ops, syntax fixed.

neteler:
> I don't think that pretty/plain text output would particular improve the situation.

Ok, agreed. But in my view the user should not remain in the dark about the amount of pixels currently used.

Background why this PR attempt:
Before opening this PR I had tried with own data and discovered that the current region was (apparently) way too large, i.e. too many raster `cells` to be run in jupyter. That's why I added the `g.region` output in the first place. Without knowing the amount of raster cells using in the computation one would just wait "forever", not knowing why the notebook doesn't reach the next step: in my tests `d.rast` hardly worked but folium didn't come up. Then I eventually discovered the `g.region` output which had been printed into the terminal and not jupyter: `cells=1224000000` which is apparently too much.

petrasovaa:
> > I don't think that pretty/plain text output would particular improve the situation.
> 
> Ok, agreed. But in my view the user should not remain in the dark about the amount of pixels currently used.
> 
> Background why this PR attempt: Before opening this PR I had tried with own data and discovered that the current region was (apparently) way too large, i.e. too many raster `cells` to be run in jupyter. That's why I added the `g.region` output in the first place. Without knowing the amount of raster cells using in the computation one would just wait "forever", not knowing why the notebook doesn't reach the next step: in my tests `d.rast` hardly worked but folium didn't come up. Then I eventually discovered the `g.region` output which had been printed into the terminal and not jupyter: `cells=1224000000` which is apparently too much.

By default, the jupyter rendering API doesn't take the region into account, it's more similar to how GUI works. You would have to specify use_region=True. One of the reasons for that is so that users can display data without needing to deal first with region.

wenzeslaus:
> ...the user should not remain in the dark about the amount of pixels currently used.

Will `'cells': '10000'` somewhere in the output be any help in finding _number of **pixels**_?

I see these two options:

1. Either just rely on the region being correct and minimize information presented to beginners (It is set to "the right thing" in each notebook. Nothing more is needed. The notebooks is an example which runs with given sample data.)
2. or checking current region settings just after explicitly setting it is very important and, at the same time, needs to be discussed in (all or one of) these notebooks, then have a separate setting and getting steps with explanation of the getting (checking) step in the surrounding text.

neteler:
> By default, the jupyter rendering API doesn't take the region into account, it's more similar to how GUI works.

I see - but which is the region used? Say, I have a mapset with a large raster map, computational region set to that map. Now I start a new jupyter session with/in that mapset and want to take a look at this raster map. What would be the best practice here? Sorry for slow and happy to take RTFM pointers.

petrasovaa:
> > By default, the jupyter rendering API doesn't take the region into account, it's more similar to how GUI works.
> 
> I see - but which is the region used? Say, I have a mapset with a large raster map, computational region set to that map. Now I start a new jupyter session with/in that mapset and want to take a look at this raster map. What would be the best practice here? Sorry for slow and happy to take RTFM pointers.

There is not a really good explanation anywhere, there is only a [function documentation](https://grass.osgeo.org/grass81/manuals/libpython/grass.jupyter.html#grass.jupyter.region.RegionManagerFor2D.set_region_from_command) available.

But basically, if you don't explicitly specify you want to use current region (`use_region=True`) or saved region (`saved_region=myregion`) the region is set to match the first raster layer (by default). In your case, the computational region is set already to that raster, so it doesn't matter if you use `use_region=True/False`, the result will be the same. The behavior is optimized so that you can display the data no matter what region you have. So, for example in a workshop, you can start by displaying data and then later explain how region works.

You can also adjust the size with `width` or `height` parameters (you can specify both or one of them or none), that will change the size of the resulting figure.

wenzeslaus:
> There is not a really good explanation anywhere, there is only a function documentation available.

The documentation is in the notebooks. See [grass_jupyter.ipynb](https://github.com/OSGeo/grass/blob/main/doc/notebooks/grass_jupyter.ipynb). Screenshot:

Screenshot from 2022-04-18 20-33-48

Documentation for how to use the API in notebooks is in notebooks because notebooks are for documentation and that's why we have the new code which works well in notebooks and it is documented in notebooks... (and no, it is not like [this](https://youtu.be/pQHX-SjgQvQ?t=133))

wenzeslaus:
I'm not sure how to move forward here. I suggest either close this or add a separate _Computational region_ section to the [basic_example_grass_jupyter.ipynb](https://github.com/OSGeo/grass/blob/main/doc/notebooks/basic_example_grass_jupyter.ipynb) notebook after the _Vector buffer_ section (or something, I think as a first section might be counter productive, or switch the order and start with vector...).

There is a _Computational region_ section in [this notebook](
https://github.com/ncsu-geoforall-lab/grass-gis-workshop-FOSS4G-2021/blob/main/workshop_part_1.ipynb) by @petrasovaa, but it is way too long for an short intro and it is using Bash.

wenzeslaus:
Please, follow up in issue #2320 replaces this PR.

PR #2309, opened 2022-04-17
Update test_module_assertions.py

adinayyu:
"%%python" added to work in jupyterlab binder

### DIFF
--- a/python/grass/gunittest/testsuite/test_module_assertions.py
+++ b/python/grass/gunittest/testsuite/test_module_assertions.py
@@ -1,3 +1,4 @@
+%%python
 import copy
 import subprocess
 


python/grass/gunittest/testsuite/test_module_assertions.py
```diff
@@ -1,3 +1,4 @@
+%%python
```
wenzeslaus, line 1: This is IPython magic. Useful for running content of this file in a Jupyter notebook. Unrelated to running the test.

wenzeslaus (changes requested):
When the test is running okay, it needs to be removed from the  [.gunittest.cfg](https://github.com/OSGeo/grass/blob/main/.gunittest.cfg) file (in the root directory of the repo) which will enable it in the CI.

wenzeslaus:
Not needed and not Python syntax.

PR #2310, opened 2022-04-18
Update test_module_assertions.py

adinayyu:
Added some comments for understanding gunittest methods. It runs successfully on jupyterlab binder with %%python

### DIFF
--- a/python/grass/gunittest/testsuite/test_module_assertions.py
+++ b/python/grass/gunittest/testsuite/test_module_assertions.py
@@ -1,3 +1,4 @@
+# use %%python here to run this code in jupyterlab binder
 import copy
 import subprocess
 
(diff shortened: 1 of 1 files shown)


python/grass/gunittest/testsuite/test_module_assertions.py
```diff
@@ -1,3 +1,4 @@
+# use %%python here to run this code in jupyterlab binder
```
wenzeslaus, line 1: Please remove this line. This applies to every Python script and there is plenty of other ways how to run this file in JupyterLab locally or in Binder and the other ways don't include copy-pasting the file content to cell, so I don't want to encourage `%%python` (see my posts at grass-dev).

python/grass/gunittest/testsuite/test_module_assertions.py
```diff
@@ -30,18 +31,28 @@ def setUp(self):
 
     def test_runModule(self):
         """Correct and incorrect Module used in runModule()"""
+        """Run PyGRASS module. Runs the module and raises an exception if the module ends with non-zero return code. 
+        Usually, this is the same as testing the return code and raising exception but by using this method, you give testing framework more control over the execution, error handling and storing of output.
+        In terms of testing framework, this function causes a common error, not a test failure."""
```
wenzeslaus, line 36: The comment is a copy-paste of documentation of the runModule method. No reason to repeat it here every time the method is called.

https://grass.osgeo.org/grass80/manuals/libpython/gunittest.html#gunittest.case.TestCase.assertModule

This would be a place to describe behavior of the test code.

wenzeslaus (changes requested):
At this point, the patch is not adding anything new. These descriptions are already in the documentation which one should refer to when in doubt about the called method. No reason to repeat function's documentation every time a function is called.

The `%%python` magic is not general enough method for running tests to be mentioned here and the method applies to every file, so it would have to be in every file which would not make much sense.

I suggest to close this and focus on fixing failing and disabled tests or adding new pytest tests.

wenzeslaus (commented):
As I have already mentioned, there is no need for putting the text output of a text as a comment into the test file. Test check or are supposed to check these things themselves when it is relevant.

wenzeslaus:
The update in #2311 replaces this PR, so closing this one. For GSoC test tasks, some actual fixes or updates of tests would be needed.

PR #2311, opened 2022-04-18
Update .gunittest.cfg

adinayyu:
Removed test_module_assertions.py


wenzeslaus (commented):
Tests are running now, so let's see if the test is in the output and passes and then we can merge.

wenzeslaus (approved):

wenzeslaus:
Congrats, @adinayyu! The test runs and passes in CI, so I merged the PR.

PR #2312, opened 2022-04-18
Update __init__.py to only necessary modules

chaedri:
* Remove reprojection_renderer and utils from __init__.py

These modules are not for direct by users of grass.jupyter. They are imported by other modules as needed. This will also shorted the API documentation page to only include modules intended for usage.

### DIFF
--- a/python/grass/jupyter/__init__.py
+++ b/python/grass/jupyter/__init__.py
@@ -44,7 +44,5 @@ The objects in submodules and names of submodules may change in the future.
 from .interactivemap import *
 from .map import *
 from .map3d import *
-from .reprojection_renderer import *
 from .setup import *
 from .timeseriesmap import *
-from .utils import *


wenzeslaus (commented):
This is a good start, but:

> These modules are not for direct by users of grass.jupyter. They are imported by other modules as needed.

There are also functions or classes here and there which are internal, no? I think it would be best to avoid the start imports and import whatever we think should be here. (That's at least one technique how to do that.)

> This will also shorted the API documentation page to only include modules intended for usage.

This would be great. However, I don't see that in the output, for example reprojection_renderer (I'm looking at the Sphinx doc artifact generated by the Pylint workflow). This is PR still worth doing even without this part.

wenzeslaus:
Changed to include only public API objects. Notebooks work for me.

PR #2313, opened 2022-04-18
libpython core docs: explain setting of debug level

neteler:
Add to `script.core.debug()` ([manual](https://grass.osgeo.org/grass80/manuals/libpython/script.html?highlight=grass_verbose#script.core.debug)) that DEBUG level is set with `g.gisenv` (see [manual](https://grass.osgeo.org/grass80/manuals/g.gisenv.html#grass-debugging)).

### DIFF
--- a/python/grass/script/core.py
+++ b/python/grass/script/core.py
@@ -740,7 +740,8 @@ def debug(msg, debug=1):
     """Display a debugging message using `g.message -d`
 
     :param str msg: debugging message to be displayed
-    :param str debug: debug level (0-5)
+    :param str debug: debug level (0-5) which is controlled by
+                      the ``DEBUG`` GRASS gisenv variable (set with `g.gisenv`)
     """
     if debug_level() >= debug:
         # TODO: quite a random hack here, do we need it somewhere else too?


python/grass/script/core.py
```diff
@@ -740,7 +740,8 @@ def debug(msg, debug=1):
     """Display a debugging message using `g.message -d`
 
     :param str msg: debugging message to be displayed
-    :param str debug: debug level (0-5)
+    :param str debug: debug level (0-5) which is controlled by
+                      the ``DEBUG`` GRASS gisenv variable (set with `g.gisenv`)
```
wenzeslaus, line 744: This seems misleading. Why it is set a parameter when it is controlled by DEBUG?

wenzeslaus (changes requested):
The current text mixes setting debug level for the message and setting debug level to control which debug messages are displayed.

Perhaps split the text into a short (one line) description of the parameter and one or two sentences in a separate paragraph providing details about the debug level (such as g.gisenv).

python/grass/script/core.py
neteler: Yes, right, the `g.gisenv` debug messages are unrelated to this. I'll fix this PR.

wenzeslaus (changes requested):
I think the syntax does not work that way. This is what is rendered:

Screenshot from 2022-04-19 09-51-53

The code block is another paragraph, not part of the parameter description.

Keeping the parameter description to just one line is the safe way syntactically, but also in terms of complexity of the text. If it is so complex it needs more than few words, you better explain it in the text with proper sentences and paragraphs. After all, this is how Python documents most functions. (NumPy is different but that's also using different syntax in the docstrings which would be a good discussion for another time.)

The code block is assumed to be Python, so verbatim would need something else. This is more a table, but perhaps it should be a sentence or two. The assignment of recommended meaning to levels lacks context and it is misleading in terms of what is recommend, the numbers or their meanings? Recommend for who?

Perhaps something like "Use 1 for messages generated once of few times, 2 for messages generated for each raster row or vector line, ..."

How does 0 make sense as a parameter here anyway?

wenzeslaus:
> Here some examples how the DEBUG message level management generally works:

Maybe an example in g.gisenv man page? But this seems to be yet another candidate for a tutorial/topic/overview/intro page.

BadAssassin:
Just reviewed the entire thread...

Keep in mind that the debugging routines were designed to debug the core C code. 0 is there to disable debugging (this code generally gets removed for release builds). It is okay to add debug levels for python (ie. 2) without affecting how debugging works. Just document the ones that are python specific and avoid changing the levels that already exist.

wenzeslaus:
Hm, doc/debugging.txt, never seen that. It is little too C-only, but otherwise I guess that's exactly the page I had in mind. I guess first step would be to turn this to Markdown. The second is adding some of these Markdown files to HTML doc.

wenzeslaus:
> ...the debugging routines were designed to debug the core C code. ...add debug levels for python...

See the implementation, currently, the Python function _debug_ which is in question just calls _g.message_ if debug level is greater than zero. (The sync to gisenv is done in _debug_level_.) Python code does not add any levels. Is that what you are thinking?

neteler:
> I think the syntax does not work that way

Ok, fixed, now it looks like this:

image

neteler:
@wenzeslaus ok for merge?

PR #2315, opened 2022-04-19
i18n: Sync from transifex

HuidaeCho:
Weird... Why are previously translated messages deleted?


HuidaeCho:
Just found https://github.com/OSGeo/grass/pull/2271.

PR #2326, opened 2022-04-26
releasebranch_8_2: Update grass80 to grass82

neteler:
Update various mentions of grass80 and grass81 -> grass82

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,34 +1,10 @@
-CHANGES in GRASS GIS 8.x compared to GRASS GIS 7.x
-
-Efforts have concentrated on making the user experience even better,
-providing many new useful additional functionalities to modules and
-further improving the graphical user interface. The new graphical user
-interface with entirely rewritten startup sequence.
+CHANGES in GRASS GIS 8.x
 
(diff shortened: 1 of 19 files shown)


wenzeslaus (approved):
In the scope of version eighty and eighty one to eighty two, this is good.

I have found some potential candidates with `grep -IrnE "[^0-9]8\.[01][^0-9]"` in INSTALL, `g.version.html`, `rpm/grass.spec`, and Docker readme, but only the last one really needs an update; no sure about the rpm.

neteler:
> I have found some potential candidates with `grep -IrnE "[^0-9]8\.[01][^0-9]"` in INSTALL, `g.version.html`, `rpm/grass.spec`, and Docker readme, but only the last one really needs an update; no sure about the rpm.

Thanks, I could catch a few more items to be updated, done in 02248ff.

neteler:
(I missed to update https://github.com/OSGeo/grass/blob/releasebranch_8_2/include/VERSION, done separately)

neteler:
For the record: Additional Windows related changes in #2338

PR #2330, opened 2022-04-26
main: Update grass80/grass81 to grass83

neteler:
Update various mentions of grass80 and grass81 -> grass83

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Build a docker image using the downloaded source code (run this in the directory
 containing the source code):
 
 ```
-docker build -t grassgis80 .
+docker build -t grassgis83 .
 ```
 
 A test run (assuming you have the existing GRASS GIS test location; it can be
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -42,7 +42,7 @@ Build a docker image using the downloaded source code (run this in the directory
 containing the source code):
 
 ```
-docker build -t grassgis80 .
+docker build -t grassgis83 .
```
wenzeslaus, line 45: When using this guide, this will be likely the only image of GRASS GIS you are building, no? So, maybe we don't need the version here, so just `grassgis` or something.

man/sphinx/conf.py
```diff
@@ -161,7 +161,7 @@
 # html_file_suffix = None
 
 # Output file base name for HTML help builder.
-htmlhelp_basename = "grass81Documentationdoc"
+htmlhelp_basename = "grass83Documentationdoc"
```
wenzeslaus, line 164: These, too, look like good candidates for switch to version-less names, but perhaps for a different commit with some testing if these pieces even work.

wenzeslaus (changes requested):
I have found 8.1 style in INSTALL and Docker readme.

https://github.com/neteler/grass/blob/relbranch82_updates/INSTALL#L57
https://github.com/neteler/grass/blob/relbranch82_updates/docker/README.md?plain=1#L7

README.md
neteler: Yes, make sense, will do.

man/sphinx/conf.py
neteler: I'd like to keep the version in the page titles, otherwise we'll have (SEO) confusion...

rpm/grass.spec
```diff
@@ -1,8 +1,8 @@
-%global shortver 80
+%global shortver 82
```
wenzeslaus, line 1: This one and few others around say 8.2? Intention?

wenzeslaus (approved):

PR #2334, opened 2022-04-27
version: Start 8.3.dev

wenzeslaus:
* Update version file to 8.3 (missed in the first version changing commit).
* Update Windows CI (was failing).
* Update pytest and pylint workflows (not failing).


neteler:
I am afraid that the title of this PR doesn't match the content (in case it comes from the new script).

wenzeslaus:
Good catch, but it's copy-pasted, writing the code now. :-)

PR #2342, opened 2022-04-28
r3.out.netcdf: Use current major version in history attribute

wenzeslaus:
r3.out.netcdf creates a netCDF file with attribute history which contains software. This removes the hard-coded software version information and uses the current major version instead for the history text. It uses fairly general macros to encapsulate stringification to avoid complex code needed to use snprintf.

The documentation contains the output, so the version should be there, but I removed it from the examples to avoid the need for updates (making the documentation little less precise).

### DIFF
--- a/raster3d/r3.out.netcdf/main.c
+++ b/raster3d/r3.out.netcdf/main.c
@@ -55,8 +55,9 @@
 #define UNITS           "units"
 #define DEGREES_EAST    "degrees_east"
 #define DEGREES_NORTH   "degrees_north"
-#define HISTORY_TEXT    "GRASS GIS " STR(GRASS_VERSION_MAJOR) " netCDF export of r3.out.netcdf"
-#define CF_SUPPORT      "CF-1.5"
+#define HISTORY_TEXT \
+    "GRASS GIS " STR(GRASS_VERSION_MAJOR) " netCDF export of r3.out.netcdf"
+#define CF_SUPPORT "CF-1.5"
 
 #define ERR(e)                      \
     {                               \
--- a/raster3d/r3.out.netcdf/main.c
+++ b/raster3d/r3.out.netcdf/main.c
@@ -140,7 +140,6 @@ static void write_netcdf_header(int ncid, RASTER3D_Region *region, int *varid,
     if ((retval = nc_put_att_text(ncid, NC_GLOBAL, "Conventions",
                                   strlen(CF_SUPPORT), CF_SUPPORT)))
         ERR(retval);
-
     if ((retval = nc_put_att_text(ncid, NC_GLOBAL, "history",
                                   strlen(HISTORY_TEXT), HISTORY_TEXT)))
         ERR(retval);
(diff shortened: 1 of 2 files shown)


wenzeslaus:
This is ready for review and would be good to get merged before v8.3 because it removes some outdated references to v7 in the code.

The question is, is removing the GRASS GIS version number from the metadata desired?

neteler:
> The question is, is removing the GRASS GIS version number from the metadata desired?

As it is actually `HISTORY_TEXT` being set, I'd suggest to replace the hardcoded version with the respective VERSION variable containing the GRASS GIS version used to write out the file.

wenzeslaus:
It now uses `GRASS_VERSION_MAJOR`. The string manipulation code is somewhat complicated (necessary?).

I did not update the documentation, so the PR is removing the version number from there right now. I'm not sure how to make it simple to maintain while keeping it absolutely precise.

I could also use `GRASS_VERSION_STRING` which is already a string, but it contains year (`"@GRASS_VERSION_NUMBER@ (@GRASS_VERSION_DATE@)"`). Alternatively, we could also define `GRASS_VERSION_NUMBER` in C in `version.h`. In any case, this would make the documentation harder to update if we opt for including the version number there.

raster3d/r3.out.netcdf/main.c
```diff
@@ -135,8 +135,14 @@ static void write_netcdf_header(int ncid, RASTER3D_Region *region, int *varid,
     if ((retval = nc_put_att_text(ncid, NC_GLOBAL, "Conventions",
                                   strlen(CF_SUPPORT), CF_SUPPORT)))
         ERR(retval);
+
+    // Works up to version 9999.
+    // 2 in original, 2 more, \0 is in the original string which sizeof counts.
+    size_t history_size = sizeof(HISTORY_TEXT) / sizeof(char) + 2;
+    char history_string[history_size];
```
nilason, line 142: Seems to me a bit overcomplicated.

What about using macros:

```c
#define STR_HELPER(x)         #x
#define STR(x)                STR_HELPER(x)
#define HISTORY_TEXT \
    "GRASS GIS " STR(GRASS_VERSION_MAJOR) " netCDF export of r3.out.netcdf"
```

then you can use `HISTORY_TEXT` as before?

wenzeslaus: Thank you. That's much better!

nilason:
You will need to clang-format this.

raster3d/r3.out.netcdf/main.c
```diff
@@ -135,6 +140,7 @@ static void write_netcdf_header(int ncid, RASTER3D_Region *region, int *varid,
     if ((retval = nc_put_att_text(ncid, NC_GLOBAL, "Conventions",
                                   strlen(CF_SUPPORT), CF_SUPPORT)))
         ERR(retval);
+
```
wenzeslaus, line 143:
```suggestion
```

wenzeslaus:
Thanks for feedback @neteler and @nilason. I'm ready to merge this.

nilason (approved):
Looks good to me!

PR #2343, opened 2022-04-28
modules: Always use current major version in VTK outputs (r.out.vtk, r3.out.vtk, v.out.vtk)

wenzeslaus:
r.out.vtk, r3.out.vtk, and v.out.vtk include software name in the VTK header line. This removes the major version from the string.

Alternative solution is more complex: Determine the current version in the code and then develop a proper test to avoid the need for the version in the reference data.


wenzeslaus:
This removes outdated references to v7 from the generated VTK files. With this PR, no further updates will be needed because the version is simply not there.

This is similar to #2342. The any mention of the version is removed completely, so the question is if it is okay.

PR #2357, opened 2022-05-03
RFC: Version Numbering

### LINKED ISSUE #2335: [Feat] Let the main branch have x.y.0 version

The version on the release branches always includes the third digit (aka micro or patch), so development versions are then 8.0.1dev.

However, the main branch with the development-only odd minor version numbers never gets a micro, e.g., 8.1dev, version which is inconsistent.

The missing third number also open room for inconsistencies in representing the version representation. It is 8.3dev or 8.3.dev?

The main branch should add 0 to the version, i.e., 8.3.0dev. The 0 (and dev) will be always there.

1. Keep the current state, but be consistent in 8.3dev versus 8.3.dev. (Not realistic and does not address inconsistencies with release branches.)
2. Replace dev by 0, i.e., 8.3.0. The odd minor says that it is a development version. (In odds common versioning schemes which add some type of suffix.) 
3. Drop the version idea completely from main and somehow have it version-less with commit has used everywhere. (Too difficult to implement.)
4. The main branch has the next version to be branched from it, so now it would be 8.3.0dev or 8.4.0dev. This includes either not using the odd-even distinction and threading odd minor versions (main is 8.3.0dev) the same as even ones or skipping odd minor versions completely (main is 8.4.0dev) to keep the practice just for releases. (Although this seems like a big change, it is only change in procedure and it actually greatly simplifies the numbering procedure. It may also allow us to use some existing tooling to automate things and, e.g., improve the suffixes.)
5. Release previews from the main branch, so keep on main whatever is the next version to be released from that code base, but keeping the odd-even distinction. The micro version than increases as needed. (The would need an easy release procedure to be feasible, but releasing unstable previews directly from main is not incompatible with the preferred solution of just including 0 right now.)

**Additional context**

We had several consistency issues recently and the missing micro release complicates the procedure (now captured in a script in #2331).

This issue is partially based on my previous comment: https://github.com/OSGeo/grass/issues/1583#issuecomment-1004982301.

wenzeslaus:
This PR is a RFC for a new version numbering system. The main point is not making any distinction between even and odd version numbers. After 8.2.z, 8.3.0 would follow.

Notably, this RFC does not follow the current RFC procedure (if any), but creates a PR as suggested in the past as a possible improvement of the procedure.

Closes #2335.

### DIFF
--- a/doc/development/rfc/version_numbering.md
+++ b/doc/development/rfc/version_numbering.md
@@ -7,6 +7,7 @@ Status: Draft
 ## Summary
 
 Use version number in a format _major.minor.micro_ where
+
 - _major_ is for large additions or backwards incompatible changes,
 - _minor_ is for additions, and
 - _micro_ is for fixes.
@@ -14,10 +15,11 @@ Use version number in a format _major.minor.micro_ where
 These numbers are assigned to releases. Development versions, pre-release,
 and builds are marked by additional labels as suffixes.
 
-Once a version number is assigned to a release, the associated source code cannot change,
-or in other words, version number is assigned to a particular state of the source code.
-This is sometimes called a hard freeze (or code freeze). Unlike soft freeze
-(or feature freeze) and branching, code freeze does not allow any changes of the code.
+Once a version number is assigned to a release, the associated source code
+cannot change, or in other words, version number is assigned to a particular
+state of the source code. This is sometimes called a hard freeze (or code
+freeze). Unlike soft freeze (or feature freeze) and branching, code freeze
+does not allow any changes of the code.
 
 All major and minor version numbers are used for releases,
 i.e., there is no distinction between even and odd numbers.
@@ -41,9 +43,9 @@ version 5 and 6 series.
 This odd-even practice followed the numbering scheme of the Linux kernel
 which abandoned the practice since then.
 
-At the time of version 5, the odd-even practice replaced a system where multiple numbered beta
-versions were released. Version 5.0beta10 was the last beta release before the first
-pre-release of 5.0.0.
+At the time of version 5, the odd-even practice replaced a system where multiple
+numbered beta versions were released. Version 5.0beta10 was the last beta
+release before the first pre-release of 5.0.0.
 
 The beta version practice was shortly picked up again for 7.0.0 which had four
 beta releases before the first release candidate of 7.0.0.
@@ -88,13 +90,15 @@ any backwards incompatible change in the API which includes
 both interface and behavior changes (API also defines what happens,
 not just names and signatures).
 
-Additionally, major feature additions which would require only a minor release in terms of API stability,
-are strongly suggested to trigger a major release as well.
-This in turn helps to address the issue of outdated tutorials and small, but breaking, changes.
+Additionally, major feature additions which would require only a minor release
+in terms of API stability, are strongly suggested to trigger a major release
+as well. This in turn helps to address the issue of outdated tutorials and
+small, but breaking, changes.
 
 Some major features, such as changes in the GUI, are fully backwards compatible
-(GUI API is not a public API in versions 7 and 8), but major features may heavily influence tutorials and
-other teaching materials. In that case, increasing a major version should be considered given that,
+(GUI API is not a public API in versions 7 and 8), but major features may
+heavily influence tutorials and other teaching materials. In that case,
+increasing a major version should be considered given that,
 in a sense, behavior linked to a particular interface is changed.
 
 There is always a list of many small changes which are not backwards compatible,
@@ -114,8 +118,8 @@ which are not bug fixes.
 If only backwards compatible bug fixes, i.e., fixes of incorrect behavior, are applied,
 a micro version can be released.
 
-The micro version is also known as _patch_ (which is what Semantic Versioning is using)
-and _point_ (which is what was used in some GRASS GIS documents).
+The micro version is also known as _patch_ (which is what Semantic Versioning
+is using) and _point_ (which is what was used in some GRASS GIS documents).
 The word _micro_, rather than patch, is used
 to avoid collision with patch referring to an individual changeset or fix
 (a release contains one or more of these changes).
@@ -194,26 +198,35 @@ In documentation, the version is often really needed, but don't say
 This way the version number in the sentence will always be valid because
 the version when the change was introduced stays the same and does not change.
 
-When a version is part of an output or displayed to the user, the version number should
-be determined dynamically, not hardcoded, even if it is just the major version
-number.
+When a version is part of an output or displayed to the user, the version
+number should be determined dynamically, not hardcoded, even if it is just
+the major version number.
 
 Don't include the version number where it is not needed, for example, text for links
 in a release announcement doesn't need a version because version is already given
 by the context.
 
-In short comments, references using GN where N is major version number are not common in general.
-When the version is important to mention, use vN which is a common practice.
+In short comments, references using GN where N is major version number are not
+common in general. When the version is important to mention, use vN which is
+a common practice.
 
 ## Relation to Other Documents
 
-* [RFC 4: Release Procedure](https://trac.osgeo.org/grass/wiki/RFC/4_ReleaseProcedure): This RFC describes changes to the numbering. RFC 4 describes the release procedure.
-* [Release Schedule](https://trac.osgeo.org/grass/wiki/Release/Schedule) (at Trac wiki under Release): The Release Schedule document describes schedule, branching, release maintenance, and numbering. The numbering is changed, specifically the use of odd version numbers for development.
-* [Semantic Versioning](https://semver.org/) (version 2.0.0 at the time of writing): Semantic Versioning treatment of _major_, _minor_, and _micro_ numbers should be respected. The labeling of other versions does not comply with Semantic Versioning, but it is a desired state for the future.
+- [RFC 4: Release Procedure](https://trac.osgeo.org/grass/wiki/RFC/4_ReleaseProcedure):
+  This RFC describes changes to the numbering. RFC 4 describes the release procedure.
+- [Release Schedule](https://trac.osgeo.org/grass/wiki/Release/Schedule)
+  (at Trac wiki under Release): The Release Schedule document describes
+  schedule, branching, release maintenance, and numbering. The numbering is
+  changed, specifically the use of odd version numbers for development.
+- [Semantic Versioning](https://semver.org/) (version 2.0.0 at the time of
+  writing): Semantic Versioning treatment of _major_, _minor_, and _micro_
+  numbers should be respected. The labeling of other versions does not comply
+  with Semantic Versioning, but it is a desired state for the future.
 
 ## Other Projects
 
-GDAL and PROJ follow the Semantic versioning. QGIS does as well, but in combination with odd numbers marking the development versions.
+GDAL and PROJ follow the Semantic versioning. QGIS does as well, but in
+combination with odd numbers marking the development versions.
 
 Ubuntu and Black lock their version numbering with the release schedule.
 Black, after transitioning from beta, releases a major release yearly in January
@@ -221,5 +234,8 @@ using the last two digits of year as major version and month as minor release.
 
 ## Historical Documents
 
-* Glynn Clements (2007). GRASS-dev GRASS 6.3.0 release preparation. Aug. 12 18:12:32 EDT 2007. <https://lists.osgeo.org/pipermail/grass-dev/2007-August/032705.html>
-* Neteler, Markus (2001). Towards a stable open source GIS: Status and future directions in GRASS development. Second Italian GRASS Users Meeting, University of Trento, Feb. 1-2 2001. <https://www.academia.edu/download/5140572/10.1.1.16.8991.pdf>
+- Glynn Clements (2007). GRASS-dev GRASS 6.3.0 release preparation.
+  Aug. 12 18:12:32 EDT 2007. <https://lists.osgeo.org/pipermail/grass-dev/2007-August/032705.html>
+- Neteler, Markus (2001). Towards a stable open source GIS: Status and future
+  directions in GRASS development. Second Italian GRASS Users Meeting,
+  University of Trento, Feb. 1-2 2001. <https://www.academia.edu/download/5140572/10.1.1.16.8991.pdf>
--- a/doc/development/rfc/version_numbering.md
+++ b/doc/development/rfc/version_numbering.md
@@ -167,7 +167,7 @@ control system, but builds are managed separately.
 ## Relation to Other Documents
 
 * [RFC 4: Release Procedure](https://trac.osgeo.org/grass/wiki/RFC/4_ReleaseProcedure): No changes to release procedure, only the numbering is changed.
-* [Release Schedule](https://trac.osgeo.org/grass/wiki/Release/Schedule#Externalreleaseschedules) (at Trac wiki under Release): The Release Schedule document describes schedule, branching, release maintenance, and numbering. The numbering is changed, specifically the use of odd version numbers for development.
+* [Release Schedule](https://trac.osgeo.org/grass/wiki/Release/Schedule) (at Trac wiki under Release): The Release Schedule document describes schedule, branching, release maintenance, and numbering. The numbering is changed, specifically the use of odd version numbers for development.
 * [Semantic Versioning](https://semver.org/) (version 2.0.0 at the time of writing): Semantic Versioning treatment of _major_, _minor_, and _micro_ numbers should be respected. The labeling of other versions does not comply with Semantic Versioning, but it is a desired state for the future.
 
 ## Other Projects
--- a/doc/development/rfc/version_numbering.md
+++ b/doc/development/rfc/version_numbering.md
@@ -52,10 +52,10 @@ beta releases before the first release candidate of 7.0.0.
 
 Explaining the version numbering should be as easy as possible.
 Ideally, it would not need any explanation at all.
-When all version numbers refer to releases, previews (alphas, betas, daily builds),
+When all version numbers refer to releases, previews (alphas, betas, daily builds)
 always need additional indication of the version being a preview.
-However, using odd version to mark the development versions requires
-still additional explanation, for example, download for 8.1 on the website
+However, using odd version to mark the development versions still requires
+additional explanation, for example, download for 8.1 on the website
 said _preview_ anyway because the odd number does not indicate a
 development version by itself, i.e., it's not self-explanatory.
 Hence, odd numbers for development versions do not bring any advantage.
@@ -66,9 +66,8 @@ a minor version is created, the branch needs to change all the mentions
 of the minor version number to the next even number and, at the same time,
 the _main_ branch needs to change to the next odd number.
 Without the even and odd distinction, the branch keeps the version from
-the main branch while the main branch advanced to the next upcoming version
-resulting only in one operation which is the same whether the current version
-is even or odd.
+the main branch while the main branch advances to the next upcoming version
+resulting only in one operation.
 
 The Semantic Versioning uses labels after the version number to indicate development
 versions, so using that system or a similar one should be sufficient to mark the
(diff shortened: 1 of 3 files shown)


doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,146 @@
...
+## Background
+
+The increasing versions was always following the same ideas as in
+the Semantic Versioning, i.e., _major_, _minor_, and _patch_
+with major for breaking changes, minor for backwards compatible
+features, and patch for fixes.
+
+For the versions 7.0 till 8.2, there was a distinction between even and odd minor versions.
+Even minor versions were released, while odd minor version marked development version.
+This odd versions never had a patch number assigned and were never released.
```
neteler, line 27: Note: The history to this was the adoption of the Linux kernel versioning system:

https://en.wikipedia.org/wiki/Software_versioning#Odd-numbered_versions_for_development_releases

Indeed, the Linux kernel project abandoned it time ago.

wenzeslaus: Any idea when it was first applied to GRASS GIS? Just to have the history complete there.

Also interesting: Linux kernel moved from 2 to 3 _not_ because of a breaking updates. What are the rules now, I don't know.

neteler: Here a relevant snippet (not fully history):
https://lists.osgeo.org/pipermail/grass-dev/2007-August/032705.html

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,146 @@
...
+This in turn helps to address the issue of outdated tutorials and small, but breaking, changes.
+
+Some major features such as changes in the GUI are fully backwards compatible
+(GUI API is not a public API in versions 7 and 8), but are heavily influencing tutorials and
+other teaching materials. In that sense, behavior linked to a particular interface is changed,
+so a major version is required.
+
+There is always a list of many small changes which are not backwards compatible,
+for example, a cleanup of depreciated functions. None of these changes alone does
+not seem worth a major release, but because it would require one, it is never done.
```
veroandreo, line 75:
```suggestion
for example, a cleanup of depreciated functions. None of these changes alone 
seems worth a major release, but because it would require one, it is never done.
```

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,146 @@
...
+Minor version increment is strongly recommend for all new functionality or improvements
+which are not bug fixes.
+
+### Micro
+
+If only backwards compatible bug fixes, i.e., fixes of incorrect behavior, are applied,
+a micro version can be released.
+
+The micro version is also known as _patch_ which is what Semantic Versioning is using
+and _point_ which is what was used in some GRASS GIS documents. The work _micro_ is used
```
veroandreo, line 91:
```suggestion
and _point_ which is what was used in some GRASS GIS documents. The word _micro_ is used
```

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,146 @@
...
+to avoid collision with patch in the meaning of individual changeset or fix
+(release contain one or more of these changes). Point release (or _dot_ release)
+can generally apply to anything after the first dot, so minor or micro release.
+
```
veroandreo, line 95:
```suggestion
can generally apply to anything after the first dot, so minor or micro release
is preferred.

```

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,146 @@
...
+Release candidates (RCs) are pre-releases marked by appending RC and
+a release candidate number to the version, e.g., 3.5.1RC2.
+
+Transition to the Semantic Versioning style for pre-releases is strongly recommend.
+Version number is followed by a dash (hyphen) followed by a dot-separated identifier
+which consists of identification of pre-release type, i.e., rc or RC, optional dot,
+and a release candidate number, e.g., 3.5.1-rc.2, 3.5.1-RC.2, 3.5.1-rc2, or 3.5.1-RC2.
+
+Daily builds and other builds of development versions other than release candidates
+are may use additional dashes to specify the actual source version, e.g.,
```
veroandreo, line 116:
```suggestion
may use additional dashes to specify the actual source version, e.g.,
```

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,146 @@
...
+won't parse the version correctly. A suggested workaround is to use the dash
+only in a file name and elsewhere use plus.
+
+Any version parsing or build system, must support the form with dash
+and the form with plus.
+
+## Relation to Other Documents
+
+* [RFC 4: Release Procedure](https://trac.osgeo.org/grass/wiki/RFC/4_ReleaseProcedure): No changes to release procedure, only the numbering is changed.
+* [Release Schedule](https://trac.osgeo.org/grass/wiki/Release/Schedule#Externalreleaseschedules) (from  at Trac wiki under Release): The Release Schedule document describes schedule, branching, release maintenance, and numbering. The numbering is changed, specifically the use of odd version numbers for development.
```
veroandreo, line 137:
```suggestion
* [Release Schedule](https://trac.osgeo.org/grass/wiki/Release/Schedule#Externalreleaseschedules) (from Trac wiki under Release): The Release Schedule document describes schedule, branching, release maintenance, and numbering. The numbering is changed, specifically the use of odd version numbers for development.
```

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Background
+
+The increasing versions was always following the same ideas as in
+the Semantic Versioning, i.e., _major_, _minor_, and _patch_
+with major for breaking changes, minor for backwards compatible
+features, and patch for fixes.
+
+For the versions 7.0 till 8.2, there was a distinction between even and odd minor versions.
+Even minor versions were released, while odd minor version marked development version.
```
veroandreo, line 26:
```suggestion
Even minor versions were released, while odd minor versions marked development versions.
```

doc/development/rfc/version_numbering.md
veroandreo, line 27:
```suggestion
These odd versions never had a patch number assigned and were never released.
```

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,146 @@
...
+## Motivation
+
+Explaining the version numbering should be as easy as possible.
+Ideally, it would not need any explanation at all.
+When all version numbers refer to releases, previews (alphas, betas, daily builds),
+always need additional indication of the version being a preview.
+However, using odd version to mark the development versions requires
+still additional explanation, for example, download for 8.1 on the website
+said _preview_ anyway because the odd number does not indicate the
+development version.
```
veroandreo, line 38:
```suggestion
said _preview_ anyway because the odd number does not indicate a
development version by itself, i.e., it's not self-explained.
```

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,146 @@
...
+Hence, odd numbers for development versions do not bring any advantage.
+
+The branching and release procedure with even minor numbers for releases
+and odd numbers for development versions require that when new branch for
```
veroandreo, line 42:
```suggestion
and odd numbers for development versions require that when a new branch for
```

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,146 @@
...
+both interface and behavior changes (API also defines what happens,
+not just names and signatures).
+
+Additionally, major feature additions which would require only minor release in terms of API stability,
+are strongly suggested to trigger major release as well.
+This in turn helps to address the issue of outdated tutorials and small, but breaking, changes.
+
+Some major features such as changes in the GUI are fully backwards compatible
+(GUI API is not a public API in versions 7 and 8), but are heavily influencing tutorials and
+other teaching materials. In that sense, behavior linked to a particular interface is changed,
```
veroandreo, line 70:
```suggestion
other teaching materials. In that sense, the behavior linked to a particular interface is changed,
```

doc/development/rfc/version_numbering.md
wenzeslaus: The original wording was unclear. I added more explanation there.

doc/development/rfc/version_numbering.md
wenzeslaus: Thanks. Great material. I updated the document based on these two although I don't reference them in the text.

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,161 @@
...
+
+There is always a list of many small changes which are not backwards compatible,
+for example, a cleanup of depreciated functions. None of these changes alone 
+seems worth a major release, but because it would require one, it is never done.
+With more common major releases, small changes can happen more often.
+
+### Minor
+
+Minor version must be incremented if new functionality is added
+or if existing functionality is marked as deprecated.
```
neteler, line 92: Just nitpicking: deprecated or depreciated as above?

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,161 @@
...
+- _micro_ is for fixes.
+These numbers are assigned to releases. Development versions, pre-release,
+and builds are marked by additional labels as suffixes.
+All major and minor version numbers are used for releases,
+i.e., there is no distinction between even and odd numbers.
+
+## Background
+
+The increasing versions was always following the same ideas as in
+the Semantic Versioning, i.e., _major_, _minor_, and _patch_
```
neteler, line 21:
```suggestion
the [Semantic Versioning](https://semver.org/), i.e., _major_, _minor_, and _patch_
```

Link added as a guide for newcomers

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,161 @@
...
+Even minor versions were released, while odd minor versions marked development versions.
+These odd versions never had a patch number assigned and were never released.
+
+The practice of odd minor version denoting development versions and even minor versions
+denoting releases was introduced with 5.0.0 and was followed in various ways in the
+version 5 and 6 series.
+This odd-even practice followed the numbering scheme of the Linux kernel
+which abandoned the practice since then.
+
+The odd-even practice replaced a system where multiple numbered beta
```
neteler, line 36:
```suggestion
Historically, the odd-even practice replaced a system where multiple numbered beta
```

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,161 @@
...
+and _point_ which is what was used in some GRASS GIS documents. The word _micro_ is used
+to avoid collision with patch in the meaning of individual changeset or fix
+(release contain one or more of these changes). Point release (or _dot_ release)
+can generally apply to anything after the first dot, i.e., minor or micro release.
+Using minor and micro does not have that ambiguity.
+Minor and micro have potential for confusion between these two due to the
+similarity of the names, but minor is an established term in this context
+and minor is sometimes used in this context and other contexts it has
+the right connotations.
+### Development Versions
```
neteler, line 111:
```suggestion
the right connotations.

### Development Versions
```

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,161 @@
...
+
+The micro version is also known as _patch_ which is what Semantic Versioning is using
+and _point_ which is what was used in some GRASS GIS documents. The word _micro_ is used
+to avoid collision with patch in the meaning of individual changeset or fix
+(release contain one or more of these changes). Point release (or _dot_ release)
+can generally apply to anything after the first dot, i.e., minor or micro release.
+Using minor and micro does not have that ambiguity.
+Minor and micro have potential for confusion between these two due to the
+similarity of the names, but minor is an established term in this context
+and minor is sometimes used in this context and other contexts it has
```
neteler, line 109:
```suggestion
and minor is sometimes used in this context and other contexts it has
```

This sentence isn't entirely clear to me. Maybe "... while ..." ?

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,161 @@
...
+3.5.1-dev-05e5df2e7.
+
+Any version parsing or build system, must support both systems, the one without
+a dash and the one with dash.
+
+### Build Information
+
+When build information is captured in the version number, Semantic Versioning
+prescribed plus sign as a separator, e.g., 3.5.1+1 or 3.5.1-RC1+1.
+However, filenames with a plus sign my not work well, so a dash as a separator
```
neteler, line 141:
```suggestion
However, filenames with a plus sign may not work well, so a dash as a separator
```

doc/development/rfc/version_numbering.md
wenzeslaus: I agree that the fact that this is talking about history needs to be emphasizes here. I made it explicit.

```
[-The-]{+At the time of version 5, the+} odd-even practice replaced a system where multiple numbered beta
versions were released. Version 5.0beta10 was a last beta release before the first
pre-release of 5.0.0.
```

doc/development/rfc/version_numbering.md
wenzeslaus: Done for the first mention of Semantic Versioning. It is also linked in the related documents section.

```
The increasing versions was always following the same ideas as in
the [-Semantic Versioning,-]{+[Semantic Versioning](https://semver.org/),+}
i.e., _major_, _minor_, and [-_patch_-]{+_patch_,+}
with major for breaking changes, minor for backwards compatible
features, and patch for fixes.
```

doc/development/rfc/version_numbering.md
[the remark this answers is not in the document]
wenzeslaus: Hm, introduced when using the _Commit suggestion_ feature online; I did a bigger edit using that.

doc/development/rfc/version_numbering.md
wenzeslaus: I improved the whole paragraph. The last sentence about confusing minor and micro might have been confusing also because I confused minor and micro.

```
The micro version is also known as _patch_ which is what Semantic Versioning is using
and _point_ which is what was used in some GRASS GIS documents.
The word [-_micro_-]{+_micro_, rather than patch,+} is used
to avoid collision with patch [-in the meaning of-]{+referring to an+} individual changeset or fix
[-(release contain-]{+(a release contains+} one or more of these changes).
[-Point-]{+A point+} release (or {+a+} _dot_ release) can generally apply to anything after
the first dot, i.e., minor or micro [-release.-]
[-Using-]{+release, while+}
{+using+} minor and micro does not have that ambiguity.
[-Minor-]{+Although minor+} and micro have potential for confusion[-between these two-] due to the
similarity of [-the-]{+their+} names,[-but-] minor is an established term in this context
and [-minor-]{+micro+} is sometimes used in this context and {+in+} other [-contexts-]{+contexts,+}
it has the right connotations.
```

doc/development/rfc/version_numbering.md
wenzeslaus: Fixed.

wenzeslaus:
Thanks @neteler for the comments. The minor versus micro is clearly an issue.

neteler (approved):
IMHO ready for PSC voting which should happen before the upcoming 8.3.0 release.

wenzeslaus:
I identified some todos, but perhaps they are for next version of the RFC document:

* [ ] discourage 82 and 820 versions, recommend 8.2 and 8.2.0
* [ ] include recommendation for dates
* [ ] specifically discourage linking name and version
  * Example: don't use "GRASS GIS 8 includes foo and bar." when you simply mean "the most recent (current) version includes."
  * Example bad: "The best version was GRASS GIS 7." Example good: "The best version of GRASS GIS was 7."
  * "winGRASS 8.2.0/standalone" in release announcement of 8.2.0 does not make much sense. Obviously, the Windows binary is for the corresponding version.  
  *  In short comments: G8 is unclear, when the version is important, using v8 is common practice. 
  * If version is really needed "The default database driver used by GRASS GIS 7 is SQLite." (as opposed to v6), say "The default database driver used by GRASS GIS is SQLite (since version 7)." so that the version number is always valid - the version of introduction stays the same and does not change.
* [ ] consider clarifying that soft freeze is the branching (aka feature freeze) while code freeze is release RFC version (this came up during PSC meeting)

wenzeslaus:
Perhaps discussion in #2575 is an opportunity to reflect again on what versioning and branching scheme we want to adopt. Perhaps what is now major.minor is just major, i.e., branch is major, not major.minor, and we drop micro.

BadAssassin:
The micro number has not been terribly relevant for some time. GRASS 
generally does not do bug fix releases and address version at the same time.

I do like using odd minor numbers for development, but I think we could 
release with major.minor and kill micro. I'm not sure just using major 
would provide enough granularity between stable and unstable releases.

On 10/20/2022 6:01 PM, Vaclav Petras wrote:
>
> Perhaps discussion in #2575 
> <https://github.com/OSGeo/grass/issues/2575> is an opportunity to 
> reflect again on what versioning and branching scheme we want to 
> adopt. Perhaps what is now major.minor is just major, i.e., branch is 
> major, not major.minor, and we drop micro.
>
> Message ID: ***@***.***>
>
-- 
Best Regards,
-Brad

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,184 @@
...
+Author: Vaclav Petras
+
+Status: Draft
+
+## Summary
+
+Use version number in a format _major.minor.micro_ where
+- _major_ is for large additions or backwards incompatible changes,
+- _minor_ is for additions, and
+- _micro_ is for fixes.
```
neteler, line 12: I wonder if we 'd better follow SemVer with "patch" rather than "micro"?

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,184 @@
...
+
+Any version parsing or build system, must support the form with dash
+and the form with plus.
+
+While the assignment of a version is typically done by tagging in a version
+control system, but builds are managed separately.
+
+## Relation to Other Documents
+
+* [RFC 4: Release Procedure](https://trac.osgeo.org/grass/wiki/RFC/4_ReleaseProcedure): No changes to release procedure, only the numbering is changed.
```
neteler, line 169: (URL to be updated after #2815)

doc/development/rfc/version_numbering.md
```diff
@@ -0,0 +1,184 @@
...
+* [Release Schedule](https://trac.osgeo.org/grass/wiki/Release/Schedule#Externalreleaseschedules) (at Trac wiki under Release): The Release Schedule document describes schedule, branching, release maintenance, and numbering. The numbering is changed, specifically the use of odd version numbers for development.
```
neteler, line 170: Where shall we maintain this schedule in future?

wenzeslaus: Python has release schedules for each minor version as a PEP. Our equivalent would be then a living RFC updated as needed. (The RFC updates are not resolved yet besides the fact that RFC is a misnomer.)

Note: Given the description, the link is meant for the whole schedule, not just the external schedules.

doc/development/rfc/version_numbering.md
wenzeslaus: I'm open to it.

I think the idea for micro was to emphasize that minor is not that small. Patch also has a different meaning like diff or commit although that seems to be diminishing given the spread of PRs/MRs. In the old days, we also had plenty of new features in the "patch" releases than just patches, so micro just stays neutral in that regard.

wenzeslaus: Keeping micro for now.

doc/development/rfc/version_numbering.md
wenzeslaus: I mentioned this in #2815.

doc/development/rfc/version_numbering.md
wenzeslaus: Link fixed. Leaving the schedule question for later.

wenzeslaus:
I also implemented the minimal changes required by this RFC. Updating version numbers on main is now the same as on release branch, so the script which was capturing the intricacies of even-odd version distinction is now simpler.

As a consequence, the version on the main branch will no longer be in format X.Y.dev and X.Ydev, but always X.Y.Zdev like the release branches. In other words, with this RFC, there is distinction between even and odd in terms of meaning, there is no distinction in numbering different development branches.

The main branch is simply the upcoming release and the (non-described) branching procedure updates only the version on main, while the branch keeps the original version from main. (Currently both need to be updated.)

The above is a example how this RFC simplifies the branching and release procedure which is an important aspect of this RFC.

The RFC suggests an additional change to the dev labeling (among other things), X.Y.Zdev should become X.Y.Z-dev because that's Semantic Versioning. However, I suggest to implement that separately, perhaps after branching or releasing 8.3.

wenzeslaus:
I guess a number is needed for this RFC. How does that work?

wenzeslaus:
Is accepted as is, it fixes #2335.

utils/update_version.md
```diff
@@ -54,20 +54,22 @@ echo $VERSION
 
 ### Updating Minor Version
 
-Let's say we are at development-only version 3.1.dev and just created
+Let's say were are at the main branch, version 3.2.0dev, and just created
```
wenzeslaus, line 57:
```suggestion
Let's say we are at the main branch, version 3.2.0dev, and just created
```

hellik (approved):

jmckenna:
Nice addition, +1.  (by the way, here was my own definition of versioning, for the MapServer users : https://github.com/MapServer/MapServer/blob/main/SECURITY.md#version-numbering-explained )

neteler:
As soon as the PSC adopts this RFC (voting is ongoing in https://lists.osgeo.org/pipermail/grass-psc/), the state will be changed to "Approved".

Issue #2344, opened 2022-04-28
[Bug] v.profile does not report all crossings of a single line


marisn:
If a profile line crosses the same line more than once, only one of crossing points is listed in the output.

Reported by Frank David on user mailinglist.

PR #2358, opened 2022-05-06
Vlib: Return all intersection points from Vect_line_get_intersections

marisn:
Initial implementation (by me in 2008 https://trac.osgeo.org/grass/changeset/34306) of Vect_line_get_intersections and Vect_line_get_intersections2 returns just one of intersection points between two lines. This is wrong as per bug #2344 users do expect to get back all intersection points. This PR fixes both functions and adds tests to Vlib, v.profile.

Might be a candidate for a backport.

### DIFF
--- a/lib/vector/Vlib/intersect.c
+++ b/lib/vector/Vlib/intersect.c
@@ -56,7 +56,7 @@
    of limited number of decimal places and for different order of
    coordinates, the results would be different)
 
-   (C) 2001-2009 by the GRASS Development Team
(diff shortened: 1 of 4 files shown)


metzm (approved):
Looks good to me.

Do you mind including the changes of PR #2359 here? This might avoid conflicts between these two PRs and I would delete PR #2359.

marisn:
> Do you mind including the changes of PR #2359 here? This might avoid conflicts between these two PRs and I would delete PR #2359.

Yes, a good idea. Done :-) Thanks for your work on #2359

PR #2359, opened 2022-05-06
lib/vector/Vlib: update line intersection documentation

### LINKED ISSUE #2345: [Feat] Document differences between Vect_line_intersection versions
There are two similarly named functions Vect_line_intersection and Vect_line_intersection2 but both of them have identical descriptions. One can only assume that there must be some kind of difference otherwise only one would exist but the documentation does not reveal anything.
Difference between both functions should be documented along recommendations or caveats in their use (if there are some).

metzm:
* update documentation for Vect_line_intersection(), Vect_line_intersection2(), Vect_line_check_intersection(), Vect_line_check_intersection2()

Fixes #2345


marisn (approved):

marisn:
Merged as a part of PR #2358

PR #2365, opened 2022-05-09
t.rast.algebra tests: split into two files

neteler:
Since the `testsuite/test_raster_algebra.py` comes with many tests which overall may take more than 300s of runtime, this PR separates out the `test_temporal*()` tests into an own file. Thus, in the ideal case, a CI timeout is no longer reached in this case.

The runtimes on my laptop are now:

```
GRASS nc_spm_08_grass7/user3:testsuite >

python test_raster_algebra.py 
...
Ran 20 tests in 119.363s
OK


python test_raster_algebra_temporal.py
...
Ran 20 tests in 106.023s
OK
```

Even on a slower machine (e.g., GHA Ubuntu CI machines) which may still avoid the currently defined 300s timeout.

Partially addresses #2185


wenzeslaus (commented):
I'm not sure if this will have a desired effect. As far as I can tell, the runtime of `testsuite/test_raster_algebra.py` doesn't have anything to do with the test hitting the limit. The test hits the limit because _it gets stuck_, not because it needs more time to complete. All other temporal tests are timing out too. With this PR, there will be one more test which can get stuck and thus more tests which can time out and fail the checks.

As for the code itself, the test code is longer, but not too long to absolutely require splitting. The PR introduces duplication in the test setup, so that's unfortunate.

wenzeslaus:
To make this clear, I hit the re-run button couple times on the Ubuntu jobs which were successful. One of the two Ubuntu jobs failed on the third attempt with:

```
Running ./temporal/t.rast.what/testsuite/test_what.py...
......========================================================================
FAILED ./temporal/t.rast.what/testsuite/test_what.py - Timeout >300.0s
```

So, not _t.rast.algebra_, but that's also part of my point. Any of the temporal tests can time out, not just the ones which run longer than the average since the observed timeouts are not a slight fluctuation in execution time. In the past, the tests were timing out after roughly 6 hours which is the GitHub Actions job time limit back when there is was no timeout for the individual tests.

neteler:
Ok, it was an attempt, closing.

wenzeslaus:
Hm, yeah, sorry, and I don't see much reasons to split the test just for code organization purposes.

PR #2371, opened 2022-05-12
grass.jupyter: Add save PNG method to Map

wenzeslaus:
Map from grass.jupyter can now save PNG images. This is easier than using the filename property and it is consistent with TimeSeriesMap.

### DIFF
--- a/python/grass/jupyter/testsuite/map_test.py
+++ b/python/grass/jupyter/testsuite/map_test.py
@@ -101,7 +101,7 @@ class TestMap(TestCase):
         # Make sure image was created
         self.assertFileExists(grass_renderer.filename)
 
-    def test_filename_property(self):
+    def test_safe_file(self):
         """Test saving of file"""
         grass_renderer = gj.Map()
         # Add a vector and a raster to the map
--- a/python/grass/jupyter/testsuite/map_test.py
+++ b/python/grass/jupyter/testsuite/map_test.py
@@ -82,7 +82,7 @@ class TestMap(TestCase):
     def test_filename(self):
         """Test that Map creates maps with unique filenames."""
         # Create map with unique filename
-        custom_filename = "test_filename.png"
+        custom_filename = "test_filename_provided.png"
         grass_renderer = gj.Map(filename=custom_filename)
         # Add files to self for cleanup later
         self.files.append(custom_filename)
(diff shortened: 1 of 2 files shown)


python/grass/jupyter/testsuite/map_test.py
```diff
@@ -101,6 +101,18 @@ def test_filename_property(self):
         # Make sure image was created
         self.assertFileExists(grass_renderer.filename)
 
+    def test_safe_file(self):
```
chaedri, line 104: Is this meant to be called `test_save_file` instead of `test_safe_file`?

chaedri (approved):
It makes sense that we have `save` method - consistent with the other classes. And this looks good. 

One note I have though: I think if you use the filename parameter in the __init__(), you don't need to use save. I'm not sure that's a problem though... the filename parameter still needs to be there for the read_file parameter.

python/grass/jupyter/testsuite/map_test.py
wenzeslaus: Right, although it would be great if it is safe too :-)

wenzeslaus:
Thanks @chaedri! I take it as an oversight that we didn't add that already after the change of the API and naming, so that's why the backport label. Here is some additional reasoning for the change partially based on your thoughts.

> ...if you use the filename parameter in the `__init__()`, you don't need to use save.

Right and additionally there is the filename property which gives you name of the file if it is temporary file, but `save("...")` is good for consistency, but it makes the API worse from point of view of Zen of Python ([PEP 20](https://peps.python.org/pep-0020/)): _There should be one-- and preferably only one --obvious way to do it._

Now we have three ways, but each has a specific use case. 1) `__init__` parameter: You know you want that in the file or you want to use an existing file. 2) `save` method: Saving is perhaps an afterthought in your workflow or you want to be really explicit about what you are doing (both great for tutorials! but generally too). 3) `filename` property: You want to use the file further, but it is temporary and you don't want to take care of the file. This is available mostly because the implementation makes it easy to offer and it has potential to make certain workflows faster. You pick the one which suites you or, if you don't care, you pick the one from tutorial which will likely be `save` and that's perfectly fine.

> ...the filename parameter still needs to be there for the read_file parameter.

Agreed, it is just needed even if we would strongly encourage other ways. Leaving out `read_file` and having it default to True when `filename` would make sense, but I think it is better to keep the mapping of parameters to behavior more explicit.

python/grass/jupyter/testsuite/map_test.py
wenzeslaus: I have renamed the temporary files too because they could conflict - these tests are not written in 100% isolated way which I forgot.

PR #2377, opened 2022-05-17
init: Use VACUUM for large, modified SQLite files only

wenzeslaus:
For SQLite database, VACUUM is used when the session is ending. There is a cost in running the task even when the database is small and no vacuuming is needed. This change adds a check for the size of the database and ignores databases under 100MB.

The message is now only verbose and printed when the task is running.

Unlike the original implementation, the vacuum task is now synchronous so that there is no running GRASS process after the session has ended.

The original implementation comes from Trac ticket 3697 (https://trac.osgeo.org/grass/ticket/3697)
and is best visible in a bulk backport commit 73736 (f790d68470cf8c3eba0de8feb2d6937c4f9c7493).

### DIFF
--- a/lib/init/grass.py
+++ b/lib/init/grass.py
@@ -42,6 +42,7 @@ import sys
 import os
 import errno
 import atexit
+import datetime
 import gettext
 import shutil
 import signal
@@ -2111,11 +2112,11 @@ def clean_temp():
     nul.close()
 
 
-def clean_all():
+def clean_all(*, start_time):
     from grass.script import setup as gsetup
 
     # clean default sqlite db
-    gsetup.clean_default_db()
+    gsetup.clean_default_db(modified_after=start_time)
     # remove leftover temp files
     clean_temp()
     # save 'last used' GISRC after removing variables which shouldn't
@@ -2648,6 +2649,8 @@ def main():
         fatal(e.args[0])
         sys.exit(_("Exiting..."))
 
+    start_time = datetime.datetime.now(datetime.timezone.utc)
+
     # unlock the mapset which is current at the time of turning off
     # in case mapset was changed
     atexit.register(lambda: unlock_gisrc_mapset(gisrc, gisrcrc))
@@ -2667,12 +2670,12 @@ def main():
     # only non-error, interactive version continues from here
     if params.batch_job:
         returncode = run_batch_job(params.batch_job)
-        clean_all()
+        clean_all(start_time=start_time)
         sys.exit(returncode)
     elif params.exit_grass:
         # clean always at exit, cleans whatever is current mapset based on
         # the GISRC env variable
-        clean_all()
+        clean_all(start_time=start_time)
         sys.exit(0)
     else:
         if use_shell:
@@ -2727,7 +2730,7 @@ def main():
         close_gui()
 
         # here we are at the end of grass session
-        clean_all()
+        clean_all(start_time=start_time)
         mapset_settings = load_gisrc(gisrc, gisrcrc=gisrcrc)
         if not params.tmp_location or (
             params.tmp_location and mapset_settings.gisdbase != os.environ["TMPDIR"]
--- a/python/grass/script/setup.py
+++ b/python/grass/script/setup.py
@@ -80,6 +80,7 @@ for details.
 # then this could even do locking
 
 from pathlib import Path
+import datetime
 import os
 import shutil
 import subprocess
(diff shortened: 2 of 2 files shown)


wenzeslaus:
Or a simpler alternative: #2449 removes the automatic vacuum completely.

metzm:
Looks good to me. I prefer this approach over #2449 because the initial motivation for this automatic vacuum was that users may not be aware that they can save potentially GBs of disk space with `db.execute sql=VACUUM`. Back then we found it to be quite convenient for users that automatic cleaning is performed at the end of a session, even if it might take some time.

lib/init/grass.py
```diff
@@ -2110,11 +2111,11 @@ def clean_temp():
     nul.close()
 
 
-def clean_all():
+def clean_all(*, start_time):
```
metzm, line 2114: I understand you like this new syntax `(*, keyword)`, but is it really needed here? Golden rule: Keep code sufficiently simple.

wenzeslaus:
Thanks for the explanation of the original motivation.

> ...we found it to be quite convenient for users that automatic cleaning is performed at the end of a session, even if it might take some time.

The issue is that it takes significant time (0.5s for 25MB, 4s for 380MB) to complete. With the original code, there was no visible slowdown because the session ended without waiting for the vacuum process, so with interactive session, one did not feel the cost of the vacuum back then.

With the PR, in extreme example, you pay 4s or more for deleting one row from the database. Disk space is now usually not that costly. For example, Git makes the assumption that you have plenty of disk space. We may work with larger data, but perhaps for us, too, time is more valuable than disk space (this would be different from raster compression where we save I/O time and disk space, right?).

> ...the initial motivation for this automatic vacuum was that users may not be aware that they can save potentially GBs of disk space with `db.execute sql=VACUUM`.

Maybe, the same can be addressed by a new module which does `db.execute sql=VACUUM` and/or turns on SQLite auto-vacuum mentioned by @tmszi in #2449. It somewhat promotes the issue and its solution, but it is opt-in (as with the default SQLite behavior) and it is managed separately (as with PostgreSQL).

metzm:
I approve the change that running time for cleaning up the db is now apparent to the user (no longer running in the background).

It is true that disk space is not that costly, but how many users are able and allowed to change the drive in their laptop for a larger one? IMHO, disk space is still an issue, as is time, and this PR is a good compromise.

wenzeslaus:
OSGeo4W jobs are failing now in general, but before OSGeo4W check passed.

python/grass/script/setup.py
```diff
@@ -412,31 +414,50 @@ def finish(self):
...
+        return
+    if modified_after:
+        modified_time = datetime.datetime.fromtimestamp(
+            file_stat.st_mtime, tz=datetime.timezone.utc
+        )
+        if modified_after >= modified_time:
+            return
+    process = gs.start_command("db.execute", sql="VACUUM")
+    gs.verbose(_("Cleaning up default SQLite database..."))
+    process.wait()
```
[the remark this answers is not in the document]
wenzeslaus: Almost the same. The message job runs in parallel. Because the message is a subprocess, it takes some tiny but measurable time to execute, so I tried to shave off some time by showing it in parallel while the vacuum is running. I tried to clarify that now in the comment.

With the greater threshold for vacuum I introduced later, the practical difference is likely negligible, but doesn't do any harm. It breaks the promise of no multiple processes, but not in a data processing context, so likely permissible. Anyway, I can remove it if it seems too strange.

lib/init/grass.py
wenzeslaus: I don't think it is necessary in this particular function or PR, but it contributes to both documentation and enforcement of correct use and it seems as something which is becoming standard Python now (also judging from introduction of the `/` to mark positional-only).

Python is using it (at least [os.path in 3.10](https://docs.python.org/3.10/library/os.path.html#os.path.realpath): `os.path.realpath(path, *, strict=False)`) and [scipy](
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.sampling.DiscreteAliasUrn.html#scipy.stats.sampling.DiscreteAliasUrn) does too: `scipy.stats.sampling.DiscreteAliasUrn(dist, *, domain=None, urn_factor=1, random_state=None)`.

As documentation, it is not needed for running the code, but it seems to be increasingly used for documenting Python.

As an enforcement of using `start_time` as only a keyword argument, `*` is the minimal code needed to specify that rule.

`*` says "I may add more parameters in the future and I don't know how I will order them." This is quite fitting here and many other places in the Python API.

wenzeslaus:
I'm ready to merge this. It improves the situation significantly when you don't edit the database. (Any edit, e.g., adding a vector map with a table, triggers the vacuum when the file is larger than MEMORYMB.)

PR #2378, opened 2022-05-17
lib/cairo: Add cairo status to error message

wenzeslaus:
The original message about failing to initialize a Cairo surface does not include any reason. This adds size and status from Cairo.

### Original

```
ERROR: Failed to initialize Cairo surface
```

### New

```
ERROR: Failed to initialize Cairo surface (width: 100000, height: 100000):
       invalid value (typically too big) for the size of the input
       (surface, pattern, etc.)
```

The result depends on what message is generated by Cairo.

## In action

```
> d.mon stop=cairo; d.mon start=cairo width=10000 height=10000 --o; d.rast map=dsm; d.mon stop=cairo
WARNING: File <map.png> already exists and will be overwritten
Output file: /home/vpetras/Projects/grass/code/grass/map.png
 100%
> d.mon start=cairo width=100000 height=10000 --o; d.rast map=dsm; d.mon stop=cairo
Current region rows: 1991, cols: 2027
ERROR: G_malloc: unable to allocate 18446744073414584320 bytes of memory at
       lib/cairodriver/graph.c:210
> d.mon start=cairo width=100000 height=100000 --o; d.rast map=dsm; d.mon stop=cairo
ERROR: Failed to initialize Cairo surface (width: 100000, height: 100000):
       invalid value (typically too big) for the size of the input
       (surface, pattern, etc.)
```

### DIFF
--- a/lib/cairodriver/graph.c
+++ b/lib/cairodriver/graph.c
@@ -373,7 +373,10 @@ static void init_cairo(void)
     }
 
     if (cairo_surface_status(surface) != CAIRO_STATUS_SUCCESS)
-	G_fatal_error(_("Failed to initialize Cairo surface"));
+        G_fatal_error(_("Failed to initialize Cairo surface"
+                        " (width: %d, height: %d): %s"),
+                      ca.width, ca.height,
+                      cairo_status_to_string(cairo_surface_status(surface)));
 
     cairo = cairo_create(surface);
 }


neteler:
This is a nice improvement, I just tested it locally:

image

which looks much better than before (compare https://github.com/OSGeo/grass/pull/2376#issuecomment-1128968607).

wenzeslaus:
Thanks @neteler. Given the context of your test, I take it you vote for backporting this to 8.2. The cairo function is available since 1.0 and otherwise it is minimal and after-error code only, so backporting to 8.2 is reasonable.

Issue #2257, opened 2022-03-09
[Bug] g.region -n  Flag broken


joneir:
**Describe the bug**

**To Reproduce**
Steps to reproduce the behavior - my session:

 ________________________________________________________________________________ 
| ~ @ uppsetnings-iMac (jeir) 
| => /Applications/GRASS-8.0.app/Contents/MacOS/Grass.sh
Rebuilding Addon HTML manual pages index...
cp: directory /Users/jeir/Library/GRASS/8.0/Modules/docs/html does not exist
cp: directory /Users/jeir/Library/GRASS/8.0/Modules/docs/html does not exist
cp: directory /Users/jeir/Library/GRASS/8.0/Modules/docs/html does not exist
/Applications/GRASS-8.0.app/Contents/MacOS/build_html_user_index.sh: line 104: cd: /Users/jeir/Library/GRASS/8.0/Modules/docs/html: No such file or directory
Rebuilding Addon menu...
/Applications/GRASS-8.0.app/Contents/MacOS/build_gui_user_menu.sh: line 58: cd: /Users/jeir/Library/GRASS/8.0/Modules/bin: No such file or directory
Default locale settings are missing. GRASS running with C locale.
Starting GRASS GIS...
Cleaning up temporary files...

          __________  ___   __________    _______________
         / ____/ __ \/   | / ___/ ___/   / ____/  _/ ___/
        / / __/ /_/ / /| | \__ \\_  \   / / __ / / \__ \
       / /_/ / _, _/ ___ |___/ /__/ /  / /_/ // / ___/ /
       \____/_/ |_/_/  |_/____/____/   \____/___//____/

Welcome to GRASS GIS 8.0.0
GRASS GIS homepage:                      https://grass.osgeo.org
This version running through:            Bash Shell (/bin/bash)
Help is available with the command:      g.manual -i
See the licence terms with:              g.version -c
See citation options with:               g.version -x
If required, restart the GUI with:       g.gui wxpython
When ready to quit enter:                exit

Launching <wxpython> GUI in the background, please wait...

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.

GRASS nc_spm_08_grass7/user1:~ > g.region -p
projection: 99 (Lambert Conformal Conic)
zone:       0
datum:      nad83
ellipsoid:  a=6378137 es=0.006694380022900787
north:      220750
south:      220000
west:       638300
[...]

PR #2394, opened 2022-05-23
libgproj: Use SRID for PJ object creation when possible

HuidaeCho:
This PR tries to address #2257 by using SRID, not GRASS key-value pairs, for creating an PJ object. Here is [a related Q&A](https://geopandas.org/en/stable/docs/user_guide/projections.html#i-get-a-bound-crs).


lib/proj/get_proj.c
```diff
@@ -249,7 +249,8 @@ int pj_get_kv(struct pj_info *info, const struct Key_Value *in_proj_keys,
     alloc_options("type=crs");
 #endif
     pjc = proj_context_create();
-    if (!(pj = proj_create_argv(pjc, nopt, opt_in))) {
+    if ((info->srid && !(pj = proj_create(pjc, info->srid))) ||
+        (!info->srid && !(pj = proj_create_argv(pjc, nopt, opt_in)))) {
```
metzm, line 253: I recommend to first create a PJ object and then recycle the code at https://github.com/OSGeo/grass/blob/main/lib/proj/do_proj.c#L332 to check if it is a `PJ_TYPE_BOUND_CRS`, in which case the source crs should be retrieved.

Applies not only to `pj_get_kv()`, but also to  `pj_get_string()`

metzm (changes requested):
The idea is correct, but this PR does not fix cases where no SRID is available and a bound CRS is provided. Should be easy to add.

HuidaeCho:
@metzm Do we ever hit this PROJ error when `pj` is already not `NULL`?
https://github.com/OSGeo/grass/blob/830f31d215038c1a67112aa350619f652e04e70b/lib/proj/get_proj.c#L275-L278

HuidaeCho:
> @metzm Do we ever hit this PROJ error when `pj` is already not `NULL`?
> 
> https://github.com/OSGeo/grass/blob/830f31d215038c1a67112aa350619f652e04e70b/lib/proj/get_proj.c#L275-L278

I think we can remove this check.

https://proj.org/development/reference/functions.html#c.proj_create_argv

> If creation of the transformation object fails, the function returns 0 and the PROJ error number is updated.

metzm:
Apparently `proj_errno(pj)` can still return an error code, if `pj` is not NULL. At least this is used in various places in PROJ itself.

metzm (approved):
Thanks for the fix! This will not be the only one regarding the evolving PROJ lib I guess...

PR #2405, opened 2022-05-25
contributing: Add separate section for OSGeo server uploads

wenzeslaus:
* Get the tag using Git assuming another person is doing the change, so variables need to be created and update version script gives a different status.
* Move changelog step to the server section because it is needed only for that and the file is needed on the computer.
* Move winGRASS steps into a separate section (copied as is).
* Fix syntax error in the tagging part of the doc.


neteler (approved):
Thanks!

PR #2425, opened 2022-06-10
Imagery: add support of original class value in sig file

marisn:
Signature files (generated by i.gensig) contain class labels but do not contain original values from map used to generate signatures. Thus map with predicted values by i.maxlik will contain different values for same classes in comparison to training map.
This PR adds an optional ability to store original class values and apply them on to classified map to restore a complete match between training and predicted class values.

Signature files written after merging this PR are incompatible with older GRASS versions (8.0, 8.2), but older signature files (8.0, 8.2) still can be read.


veroandreo:
I pulled these changes and tested by digitizing a vector map with 2 classes (43 and 65, just because), converted to raster using the class number, ran i.gensig and then i.maxlik. While i.gensig uses the right number for classes, i.maxlik still renames them to 1 and 2.

include/grass/defs/imagery.h
```diff
@@ -160,7 +160,7 @@ void I_free_signatures_list(int, char ***);
 int I_init_signatures(struct Signature *, int);
 int I_new_signature(struct Signature *);
 int I_free_signatures(struct Signature *);
-int I_read_one_signature(FILE *, struct Signature *);
+int I_read_one_signature(FILE *, int, struct Signature *);
```
wenzeslaus, line 163: Seems like an API change. Probably not serious in practice since nobody picked up the new signature API yet. It still should be clarified in the doc or something.

include/grass/imagery.h
```diff
@@ -51,20 +51,22 @@ struct Control_Points
...
     int have_color;
+    int oclass;         /* original class value */
 };
 
 struct Signature
 {
     int nbands;         /* band (imagery group member) count */
     char **semantic_labels;    /* list of semantic labels */
     int nsigs;          /* signature count */
+    int have_oclass;     /* is One_Sig.oclass filled with values? */
```
wenzeslaus, line 69: Booleans should be `bool` at this point.

lib/imagery/sig.c
```diff
@@ -81,7 +82,7 @@ int I_free_signatures(struct Signature *S)
     return 0;
 }
 
-int I_read_one_signature(FILE * fd, struct Signature *S)
+int I_read_one_signature(FILE * fd, int have_oclass, struct Signature *S)
```
wenzeslaus, line 85: Unlike the previous version which was relatively clear, this new version really needs Doxygen doc.

lib/imagery/sig.c
```diff
@@ -204,7 +208,13 @@ int I_read_signatures(FILE * fd, struct Signature *S)
         return -1;
     }
 
-    while ((n = I_read_one_signature(fd, S)) == 1) ;
+    /* Read marker of original class value presence */
+    if (ver == 2 && fscanf(fd, "%d", &(S->have_oclass)) != 1) {
```
wenzeslaus, line 212: Unless the v3 format will be completely different, `ver >= 2` is probably the right code for saying "process something introduced in version 2" like testing for Python 3 behavior with `== 3` versus `>= 3`.

wenzeslaus (commented):
Although this has a test for C functions which is great, there is no higher level test for i.gensig and i.maxlik or related modules. i.maxlik has a partial example in the manual page which requires an example from i.cluster. I would be more confident if we have at least some smoke test to see that the workflows really run. Sorry, I was not following this closely, but having a test for a workflow which didn't work before and now works and also for a workflow which worked and continues to work would be great showing both what was improved and also that nothing was damaged.

include/grass/defs/imagery.h
marisn: I_read_one_signature should not be a part of public API as it is just an implementation detail of I_read_signatures.

lib/imagery/sig.c
marisn: Done.

include/grass/imagery.h
marisn: So be it. Only now it will rise a question why other bool properties in the struct are int but this one is bool.

include/grass/defs/imagery.h
wenzeslaus: If the function cannot be static in a file, than the method used elsewhere to mark it as internal/private/protected is double underscore after the prefix, i.e., `I__read_one_signature`. I think that works pretty well, although it is probably not well documented. Unfortunately, Doxygen does not have a good way of marking internal functions, `@protected` and `@private` might be somewhat close to the right thing.

include/grass/imagery.h
wenzeslaus: I think we shifted from purely preferring consistency to adding modern code even when surrounding code is in legacy style. At least I did. :-)

marisn: Boolean in this case is not worth the trouble – there is no printf/scanf formatting type for bool. Not an issue for printf as it promotes anything smaller than int to int, but scanf is more problematic as an intermediate variable has to be used to read into int and then cast back to bool.

wenzeslaus: I wouldn't say it is a showstopper in general, but it does weight against using bool here.

include/grass/defs/imagery.h
marisn: I'm not so keen on making large changes as imagery code predates 3.2... I just added a docstring with a note.

include/grass/imagery.h
marisn: Sorry, I'm too lazy to work around bool as int works just fine & matches the code style.

marisn:
I added two tests for i.maxlik and one for i.gensig. All three tests just test successful case, there are no tests for various modes of failure.
Having a test was a really good idea as I had made an one-off error in i.maxlik code :(

@veroandreo could you give this PR another shot?

imagery/i.gensig/testsuite/test_i_gensig.py
```diff
@@ -0,0 +1,160 @@
...
+
+class SuccessTest(TestCase):
+    """Test successful generation of a sig file"""
+
+    @classmethod
+    def setUpClass(cls):
+        """Ensures expected computational region and generated data"""
+        cls.use_temp_region()
+        cls.runModule("g.region", n=5, s=0, e=5, w=0, res=1)
+        cls.data_dir = os.path.join(pathlib.Path(__file__).parent.absolute(), "data")
```
wenzeslaus, line 47: This should be relative path (to the current dir), so simple `"data"` is sufficient. The [documentation](https://grass.osgeo.org/grass83/manuals/libpython/gunittest_testing.html#data-specific-to-one-test) says:

> Data specific to one test
>
> If the data required by the test are not part of standard location and cannot be part of the test file itself, this data should be stored in files in data subdirectory of testsuite directory. The test should access the data using a relative path from its location, i.e. all data will be accessed using `data/...`. This data directory might be used directly when running test file directly in the directory in the source code or might be copied to the test current working directory when running tests by the main test invoking tool.

This is also similar to how some pytest plugins work, so I think it is unlikely we will change that.

lib/imagery/sig.c
```diff
@@ -204,6 +211,12 @@ int I_read_signatures(FILE * fd, struct Signature *S)
         return -1;
     }
 
+    /* Read marker of original class value presence */
+    if (ver >= 2 && fscanf(fd, "%d", &S->have_oclass) != 1) {
+        G_warning(_("Invalid signature file"));
```
wenzeslaus, line 216: The message is too generic while there is context available. Please add info on what went wrong, e.g., `Invalid signature file: Original class value presence not readable`

wenzeslaus (changes requested):

imagery/i.gensig/testsuite/test_keyvalue_result.txt
```diff
@@ -0,0 +1,12 @@
+time=1.103s
```
wenzeslaus, line 1: Committed by mistake. `git add {directory}` usually goes wrong. Possibly needs to be added to `.gitignore` and ideally not generated in these cases.

wenzeslaus (changes requested):
The new things look good.

imagery/i.gensig/testsuite/test_keyvalue_result.txt
marisn: I usually was running tests with absolute path and thus test_keyvalue file wasn't an issue :-(

Adding this to the .gitignore is the right thing to do. Could you create a PR?

wenzeslaus (approved):
This is good to go as far as the changed code and new tests go. I'm still little uneasy about the i.cluster -> i.maxlik workflow which is not tested anywhere.

imagery/i.gensig/testsuite/test_keyvalue_result.txt
wenzeslaus: .gitignore done in #2445.

veroandreo (approved):
Just tested again. It works as expected, class numbers are now preserved in the result of i.maxlik

marisn:
> This is good to go as far as the changed code and new tests go. I'm still little uneasy about the i.cluster -> i.maxlik workflow which is not tested anywhere.

You mean i.cluster -> sig file. Writing and reading sig file is tested in lib tests and sig file -> i.maxlik is tested in i.maxlik tests.

PR #2434, opened 2022-06-14
v.info: Add vector name to no db connection message

wenzeslaus:
The error messegage about missing database connection for a given layer does not include vector name which may not be clear when scripting and an undelying function calls v.info.


neteler:
Is anything still missing here (as it is a draft PR)?

wenzeslaus:
The code does not compile, but the conceptual issue here is that similar message is at many other places, but there it is not possible to get vector map name so easily.

wenzeslaus:
The code issue is fixed now, but I'm not sure if to introduce this message when we have similar messages but without that additional info. Maybe better more info here is better than consistency with less info.

PR #2447, opened 2022-06-17
t.unregister: add support to unregister maps from other mapsets

anikaweinmann:
By the try to unregister raster maps from a STRDS an error accured:
```
t.unregister input=test_strds@test_mapest maps=elevation@PERMANENT
Unregister maps
ERROR: Unable to execute sql statement. You have no permission to access
mapset <PERMANENT>, or mapset <PERMANENT> has no temporal database.
Accessible mapsets are: <test_mapset>
```
This changes fixes this bug and you can unregister raster maps from other mapset in the STRDS.

### DIFF
--- a/temporal/t.unregister/t.unregister.py
+++ b/temporal/t.unregister/t.unregister.py
@@ -50,7 +50,6 @@
 
 import grass.script as grass
 
-
 # lazy imports at the end of the file
 
 ############################################################################
@@ -76,6 +75,13 @@ def main():
 
     dbif = tgis.SQLDatabaseInterfaceConnection()
     dbif.connect()
+
+    # create new stds only in the current mapset
+    # remove all connections to any other mapsets
+    # ugly hack !
+    currcon = {}
+    currcon[mapset] = dbif.connections[mapset]
+    dbif.connections = currcon
 
     # In case a space time dataset is specified
     if input:
@@ -129,7 +135,7 @@ def main():
         map = tgis.dataset_factory(type, mapid)
 
         # Unregister map if in database
-        if map.is_in_db(dbif):
+        if map.is_in_db(dbif, mapset=mapset):
             # Unregister from a single dataset
             if input:
                 # Collect SQL statements
--- a/temporal/t.unregister/t.unregister.py
+++ b/temporal/t.unregister/t.unregister.py
@@ -76,7 +76,7 @@ def main():
     dbif = tgis.SQLDatabaseInterfaceConnection()
     dbif.connect()
 
-    # create new stds only in the current mapset
+    # modify a stds only if it is in the current mapset
     # remove all connections to any other mapsets
     # ugly hack !
     currcon = {}


temporal/t.unregister/t.unregister.py
```diff
@@ -77,6 +76,13 @@ def main():
     dbif = tgis.SQLDatabaseInterfaceConnection()
     dbif.connect()
 
+    # create new stds only in the current mapset
```
metzm, line 79:
```suggestion
    # modify a stds only if it is in the current mapset
```

metzm:
Just for clarification: this PR makes sure that `t.unregister` only modifies the TGIS db and thus a STDS in the current mapset. Actual maps to be unregistered can be located in different mapsets. The same has already been implemented for `t.register`.

metzm (approved):
Thanks for the fix!

wenzeslaus:
Thanks for the clarification @metzm!

temporal/t.unregister/t.unregister.py
```diff
@@ -77,6 +76,13 @@ def main():
     dbif = tgis.SQLDatabaseInterfaceConnection()
     dbif.connect()
 
+    # modify a stds only if it is in the current mapset
+    # remove all connections to any other mapsets
+    # ugly hack !
+    currcon = {}
+    currcon[mapset] = dbif.connections[mapset]
+    dbif.connections = currcon
```
wenzeslaus, line 84: The way to do this in a non-ugly way is to wrap whatever behavior is needed in the library, e.g., by adding function _keep_only_current_mapset_connection_ and then document that `map.is_in_db(dbif, mapset=current_mapset)` can only be used when the above was called. ....or something like that.

metzm: True, should be easy to add with a simple boolean when initiating a new `SQLDatabaseInterfaceConnection`. But I would leave this for a separate PR.

metzm: See https://github.com/OSGeo/grass/pull/2448

wenzeslaus: With the PR already open, I'm quite comfortable merging this as is.

metzm: About `map.is_in_db(dbif, mapset=mapset)`: If you are in mapset A, working with a space-time dataset in mapset B which contains maps from mapset C, the temporal lib must use the TGIS db of the space-time dataset, i.e. of mapset B and ignore the (different or non-existing) TGIS dbs in mapsets A and C. This is the reason for the mapset parameter in `is_in_db()`.

wenzeslaus: Thanks for the explanation.

BTW, the pytest-based tests might be a better tool for testing these inter-mapset things in temporal modules than the gunittest-based tests because you just manage the locations (and mapsets) yourself.

PR #2449, opened 2022-06-18
init: Remove automatic vacuum

wenzeslaus:
This removes automatic vacuum of SQLite database, i.e., reverts the addition of automatic vacuum when a session ends.

Doing vacuum can take long time which the current implementation hides by running the vacuum process in the background. The typical time is much longer than the 0.1 sec suggested by the sleep call.

Doing vacuum only sometimes would be less predictable for the users and doing it only for large database would mean that it will take really long time (with a lot of disk or memory usage) when it actually happens suggesting that that's something user may want to trigger manually.

Notably, SQLite doesn't do the vacuums automatically, although PostgreSQL has autovacuum on by default.

This is an alternative to #2377 which uses VACUUM for large, modified files only.

### DIFF
--- a/lib/init/grass.py
+++ b/lib/init/grass.py
@@ -2111,10 +2111,6 @@ def clean_temp():
 
 
 def clean_all():
-    from grass.script import setup as gsetup
-
-    # clean default sqlite db
-    gsetup.clean_default_db()
     # remove leftover temp files
     clean_temp()
     # save 'last used' GISRC after removing variables which shouldn't
--- a/python/grass/script/setup.py
+++ b/python/grass/script/setup.py
@@ -415,30 +415,6 @@ class SessionHandle:
         finish()
 
 
-# clean-up functions when terminating a GRASS session
-# these fns can only be called within a valid GRASS session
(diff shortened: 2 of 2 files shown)


tmszi:
> Notably, SQLite doesn't do the vacuums automatically, although PostgreSQL has autovacuum on by default.

Introduced in SQLite 3.1, [auto_vacuum](https://www.sqlite.org/pragma.html#pragma_auto_vacuum) is an optional feature that automatically vacuums the database to minimize the size of the database file. With this feature, you don't have to manually run the [VACUUM statement](https://www.sqlite.org/lang_vacuum.html). By default, the auto_vacuum feature is disabled.

wenzeslaus:
This works well with #2462 which supports user in doing vacuum when needed and enable auto-vacuum (only draft now).

wenzeslaus:
The #2449 (vacuum for large and modified only) is merged, so closing this one. However,  removing vacuum completely, i.e., leaving it up to the user and database would still make it faster for many users. Open a new issue or PR if you want that change.

PR #2453, opened 2022-06-20
grass.jupyter: Support pathlib.Path in TimeSeriesMap.save

wenzeslaus:
* API: TimeSeriesMap.save now supports pathlib.Path objects (and strings as before).
* tests: Use tmp_path, not current dir. Files generated by the tests should use tmp_path (or other pytest tmp dir ways), not the current directory.


chaedri (approved):
Looks good to me! Does writing to the current directory cause tests to fail?

I tried to figure out why the pytest check is failing (`AttributeError: 'PosixPath' object has no attribute 'endswith'`) but I'm not sure.

wenzeslaus:
The code should actually support both string and Path objects, but it supported only strings, so the new text code broke it. This PR now fixes the code.

Writing to the the current works (most of the time), but a file is left in that directory while using pytest tmp dir, the files are managed by pytest and in one place. (They are accessible after the test finishes and are deleted time to time.)

chaedri (approved):
I understand now. Thanks! Looks good to me.

PR #2457, opened 2022-06-21
init: Decrease verbosity of --exec

wenzeslaus:
* If --exec or -e (exit with -c) is used, do not show the 'Starting' message. Only one operation will happen, so no need to emphasize there is a session starting.
* Do not show 'Creating' message for --tmp-location. It is temporary, no need to know it is being created when that's what was requested and it will be deleted at the end.
* Show 'Executing' and 'Execution...finished' messages only in verbose mode. By default, there is no need to echo the command which was received.
* Clarify interface of run_batch_job and fix its documentation. Remove unused env var related code.

### DIFF
--- a/lib/init/grass.py
+++ b/lib/init/grass.py
@@ -1143,11 +1143,13 @@ def set_mapset(
                     # create new location based on the provided EPSG/...
                     if not geofile:
                         fatal(_("Provide CRS to create a location"))
-                    message(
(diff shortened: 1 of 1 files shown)


wenzeslaus:
This fixes documentation, includes comments, and less verbose `--exec` was requested in the past. If tests pass, I'll merge this.

PR #2459, opened 2022-06-21
v.to.rast: Use verbose for no area message, not warning

wenzeslaus:
When there are no areas in the vector map, v.to.rast would give warning, 'No areas selected'. However, no areas is a perfectly fine state when converting a vector map with points or lines only, so the warning is misleading.

To cover the case where a user may need to know about the number of areas, a verbose message is now added reporting the number replacing both the warning and a debug message with the number.

### DIFF
--- a/vector/v.to.rast/vect2rast.c
+++ b/vector/v.to.rast/vect2rast.c
@@ -138,11 +138,9 @@ int vect_to_rast(const char *vector_map, const char *raster_map, const char *fie
     Points = Vect_new_line_struct();
 
     if (use != USE_Z && use != USE_D && (ftype & GV_AREA)) {
-	if ((nareas = sort_areas(&Map, Points, field, cat_list)) == 0)
-	    G_warning(_("No areas selected from vector map <%s>"),
-			  vector_map);
-
-	G_debug(1, "%d areas sorted", nareas);
+        nareas = sort_areas(&Map, Points, field, cat_list);
+        G_verbose_message(_("Number of areas selected from vector map <%s>: %d"),
+                          vector_map, nareas);
     }
     if (nareas > 0 && dense) {
 	G_warning(_("Area conversion and line densification are mutually exclusive, disabling line densification."));


wenzeslaus:
This PR is trying to still address whatever was the motivation for the warning by giving a more detailed verbose message. If the user is wondering what is happening, `--verbose` will give the clear information about areas.

There is a series of issues with the current behavior:

## 1. False alarm

When you rasterize points or lines, you get a message `WARNING: No areas selected from vector map <my_points_only_map>` which is misleading. You didn't explicitly ask for areas, your data does not have any areas, and you don't expect any areas, yet you get a warning.

Without this PR:

```
$ v.to.rast input=bridges@PERMANENT output=bridges use=val            
WARNING: No areas selected from vector map <bridges@PERMANENT>
Pass 1 of 5:
Reading features...
...
v.to.rast complete.
(Thu Jun 23 15:44:47 2022) Command finished (4 sec) 
```

With the PR:
```
$ v.to.rast input=bridges@PERMANENT output=bridges use=val                        
Pass 1 of 5:
Reading features...
...
v.to.rast complete.
```

## 2. Inconsistency between types

The behavior is not consistent for areas-only map where you get no warning about no lines or points (or whatever else is accepted here for type). The module uses the areas without warnings `WARNING: No points selected from vector map <my_areas>` and `WARNING: No lines selected from vector map <my_areas>`.

```
$ v.to.rast input=nc_state@PERMANENT output=boundary use=val          
Pass 1 of 5:
Reading areas...
...
v.to.rast complete.
```

## 3. Inconsistency with other modules

The warning for no features of a given type is not consistent with other modules which take type and have default which contains multiple types. For example, v.extract has `type` with default `point,line,boundary,centroid,area,face`, but does not issue a warning when input misses one of those. Same for d.vect which is perhaps closer to v.to.rast than v.extract.

## 4. Empty selection is generally a valid result

Even when type option is explicitly set to a single type, v.extract does not give a warning, it is simply an empty selection like anything else. d.vect does not give any warning either for `d.vect nc_state type=point` or `d.vect bridges type=area`. This is consistent with `d.vect bridges where="YEAR_BUILT > 2025"` which results in an empty selection, but gives no warning.

wenzeslaus:
Merged. This removes the false alarm, warning inconsistency in behavior between types, and warning inconsistency with other modules. Empty selection is generally a valid result, so here it also gives no warning.

The alternative of always giving warning when no features of a specific type are present, generates a lot of warnings (see #2498). Another alternative would be to completely rework of how the type option works, e.g., the default would be empty and warnings would be present if the type was explicit, but that still would be somewhat inconsistent with other modules and does not follow that an empty selection is generally a valid result.

PR #2469, opened 2022-06-28
wxGUI: implement workspace recent files submenu for single/multiple window mode

tmszi:
Implement workspace recent files submenu for single/multiple window mode (same feature has wxGUI component Simple Python Editor).

**Screenshots:**

wxgui_single_win_mode_main_menu_workspace_recent_files_submenu


petrasovaa (commented):
I like the idea and will try to test it, but I was wondering if the functionality should be rather in the WorkspaceManager than as a mixin? I think that would result in fewer changes.

tmszi:
> I like the idea and will try to test it, but I was wondering if the functionality should be rather in the WorkspaceManager than as a mixin? I think that would result in fewer changes.

I agree with you, I implemented it in the `WorkspaceManager` class commit no. https://github.com/OSGeo/grass/pull/2469/commits/3f8c65b622b7f228677c11bbaa56e0e2aa7ed6e5.

petrasovaa (approved):
Looks great!

PR #2479, opened 2022-07-12
g.extension: Add details to URL error message

wenzeslaus:
The URL validation introduced in 45ff0ead5acfbcb574a4326b54538b28821a2aea does not include the original error message details in the user-visible error message. This adds the details and also fixes the formatting which was done on the regular string literal instead of the translated string.


ninsbl (approved):
Looks good to me.

PR #2486, opened 2022-07-20
v.fill.holes: Add new tool to remove inner isles, keep outer boundary

wenzeslaus:
### Input

Screenshot from 2022-07-20 13-02-31

### Output

Screenshot from 2022-07-20 13-02-57


wenzeslaus:
I would like for this to go to core rather than the grass-addons repo because it seems like basic functionality, but I have seen some topology issues with complex data (will post on that later).

wenzeslaus:
What I'm missing here? I have an issue with the current implementation. If there are touching areas, it does not create a valid topology: some of the areas are not build. The same happens to be me with real-word large data and small hand-drawn data. Any comments or questions welcome.

## Failing example

### Input

Screenshot from 2022-07-22 15-48-44

### Output

Screenshot from 2022-07-22 15-48-51

### Output: boundaries and centroids

Screenshot from 2022-07-22 15-50-29

### Output: vector digitizer rendering


Screenshot from 2022-07-22 15-50-56


### v.build output

```
Number of centroids exceeds number of areas: 4 > 2
Number of incorrect boundaries: 2
Number of centroids outside area: 2 
```

### Data

GitHub does not take the GRASS pack format even with other file extensions I tried, so it is a zipped `.pack` file: [test2.zip](https://github.com/OSGeo/grass/files/9170371/test2.zip)

neteler:
Does @metzm have a suggestion here on the topology issue?

vector/v.fill.holes/main.c
```diff
@@ -0,0 +1,224 @@
...
+    /* Create and initialize struct's where to store points/lines and categories */
+    struct line_pnts *points = Vect_new_line_struct();
+    struct line_cats *area_cats = Vect_new_cats_struct();
+    struct line_cats *empty_cats = Vect_new_cats_struct();
+
+    struct ilist *all_cats = Vect_new_list();
+    struct ilist *field_cats = Vect_new_list();
+
+    plus_t num_areas = Vect_get_num_areas(&input);
+
```
metzm, line 129: create an index for lines already copied to the output, needed to write out boundaries only once, initialized to zero
```suggestion
plus_t num_lines = Vect_get_num_lines(&input);
int *line_written_out = G_calloc(num_lines + 1, sizeof(int));
```

vector/v.fill.holes/main.c
```diff
@@ -0,0 +1,224 @@
...
+                                     options.cats->answer);
+
+    /* Create and initialize struct's where to store points/lines and categories */
+    struct line_pnts *points = Vect_new_line_struct();
+    struct line_cats *area_cats = Vect_new_cats_struct();
+    struct line_cats *empty_cats = Vect_new_cats_struct();
+
+    struct ilist *all_cats = Vect_new_list();
+    struct ilist *field_cats = Vect_new_list();
+
```
metzm, line 127:
```suggestion
    struct ilist *area_bounds = Vect_new_list();
```

vector/v.fill.holes/main.c
```diff
@@ -0,0 +1,224 @@
...
+        db_close_database_shutdown_driver(driver);
+    }
+
+    Vect_destroy_list(field_cats);
+    Vect_destroy_list(all_cats);
+
+    Vect_build(&output);
+
+    Vect_set_release_support(&input);
+    Vect_set_release_support(&output);
```
metzm, line 215: Not needed, the operating system does it better and faster.

vector/v.fill.holes/main.c
```diff
@@ -0,0 +1,224 @@
...
+        Vect_read_line(&input, points, area_cats, centroid);
+
+        if (constraint_cat_list &&
+            !Vect_cats_in_constraint(area_cats, field, constraint_cat_list)) {
+            continue;
+        }
+        Vect_write_line(&output, GV_CENTROID, points, area_cats);
+
+        Vect_get_area_points(&input, area, points);
+        Vect_write_line(&output, GV_BOUNDARY, points, empty_cats);
```
metzm, line 148:
```suggestion
```

wenzeslaus:
The new version works! Thanks @metzm for the pointers!

The updated version contains documentation including images and notebook to generate them. I included some notes on topology because it was not what I expected. It as pytest-based tests.

Some features this could have, but I do _not_ plan to implement:

* maximum (or minimum) size to remove
* keep features of other vector types (e.g. keep points)
* keep boundaries with categories (may represent lines which are at the same time boundaries)
* copy all attribute tables with layer=-1 (Now it copies only relevant rows of the attribute table for the selected layer. This usually should be the whole table because all areas with centroids are kept, so wanting to simply preserve everything makes sense.)

From my perspective, this is ready to merged. Reviews welcome.

vector/v.fill.holes/main.c
wenzeslaus: Thanks. This was my main misunderstanding. The topology for areas is build automatically from a pile of centroids and boundaries, but I need to handle those directly instead of just dealing with areas.

vector/v.fill.holes/main.c
wenzeslaus: My other misunderstanding was that `Vect_get_area_boundaries` is a list of outer boundary lines, not the outer ring plus the different inner rings.

vector/v.fill.holes/main.c
wenzeslaus: I tested the performance on a vector with 186,940 areas (180,015 after the operation). Here are its stats:

```
nodes=227263
boundaries=352424
centroids=166520
areas=186940
islands=61779
primitives=518944
```

I used perf stat for time and valgrind for the leaked memory. Here are the results:

| memory release | perf stat -r 10 seconds time elapsed (3x10 runs) | valgrind LEAK SUMMARY still reachable; definitively lost |
| --- | --- | --- |
| enabled | 24.39, stddev 0.18 (24.181, 24.488, 24.513) | 8,297,834 bytes in 145 blocks; 191,032 bytes in 184,529 blocks |
| disabled | 23.97, stddev 0.36 (23.557, 24.212, 24.151) | 248,489,855 bytes in 6,580,166 blocks blocks; 6,579 bytes in 76 blocks |
| difference (enabled minus disabled) | 0.42 (1.7%) | -240,192,021 (regression) ; 177,950 (improvement)

I don't think the results here support the idea that OS will be "much faster" as the old [doc/debugging.txt](https://github.com/OSGeo/grass/blob/main/doc/debugging.txt) claims in a note from 2005 (https://github.com/OSGeo/grass-legacy/commit/fb92c0fd2163474f18daf0413d62e89e28249fec). The difference of averages is smaller than 2*stddev.

However, given the overall lack of memory handling and a policy to tackle that, I'm fine leaving that out for the small performance gain (for now).

vector/v.fill.holes/main.c
```diff
@@ -0,0 +1,212 @@
+
+/****************************************************************
+ *
+ * MODULE:     v.fill.holes
+ *
+ * AUTHOR:     Vaclav Petras
+ *
+ * PURPOSE:    Fill holes in an area, i.e., preserve only its outer boundary
+ *
+ * COPYRIGHT:  (C) 2022 by Vaclav Petras and the GRASS Development Team
```
petrasovaa, line 10: 2023

vector/v.fill.holes/v.fill.holes.html
```diff
@@ -0,0 +1,119 @@
...
+    <li>
+        <em><a href="v.clean.html">v.clean</a></em>
+        for removing topological issues,
+    </li>
+    <li>
+        <em><a href="r.fillnulls.html">r.fillnulls</a></em>
+        for filling empty spaces in raster maps using interpolation,
+    </li>
+    <li>
+        <em><a href="r.fillnulls.html">r.fill.stats</a></em>
```
petrasovaa, line 112:
```suggestion
        <em><a href="r.fill.stats.html">r.fill.stats</a></em>
```

wenzeslaus: Fixed. Thanks.

PR #2488, opened 2022-07-21
wxGUI: do not crash on localized GUIs

marisn:
PR #2469 (908c5e23bf8fc23f5a0602dd1cc03cab6bd4ad37) introduced a code that uses English UI strings to identify UI elements. In the case of a translated UI it fails to get element and then proceeds to manipulate with not existing element leading to an assert in C++ wx code.


tmszi (approved):
Thanks!

Issue #2484, opened 2022-07-18
[Bug] Issue in C code any Python environment interplay


marisn:
When running i.maxlik from GUI, its UI works fine. When running from command line (terminal not GUI one), module UI fails to start with an error on system being not initalized (G_gisinit() not called).
The issue can be fixed with plain `from grass.lib import gis` in gselect.py but it seems to be a hack as code should work in its current form.
https://github.com/OSGeo/grass/blob/1961472afeb7633c9b744b0a60c923fb9b1d4411/gui/wxpython/gui_core/gselect.py#L2825

git bisect indicates on 8e66baaf5e752dc2e9050bc1d888bb3391f1aaad (PR #2202) being to blame but it more seems to just uncover underlying issue than being the cause itself.

marisn:
I do not see any problem of intermingling C calls in GUI *in broad terms*. If environment initialization happens properly, there shouldn't be any problem with it.
In this particular case – I do agree – a module managing signature files would be good. Although question if it should be a separate module or new functionality of already existing g.* modules is not closed yet.
I managed to get down to the root of this issue. pygrass modules have calls to G_gisinit("") and thus plain import performs initialization. In case of parser starting GUI, there might be no calls to G_gisinit from python been made. Thus a simple solution is just to call G_gisinit when forms.py is run standalone.

wenzeslaus:
> I do not see any problem of intermingling C calls in GUI _in broad terms_.

The library is currently following the idea that it is used primarily in processing tools, not persistent applications. So, the question is if we want to use it in GUI and then try to catch up in the library or try to avoid using it and improve the library first.

> If environment initialization happens properly, there shouldn't be any problem with it.

That is an issue in the GUI since it is hard to determine what is the initialization order due to functions being called based on user action and due to multiple executable files being involved. Currently, there is no provider or manager object in the API which would ensure things are happening in the right way.

> Although question if it should be a separate module or new functionality of already existing g.* modules is not closed yet.

I would go with separate module(s). Signature files have things specific just to them which would complicate aligning them with e.g. raster maps in g.list. In other words, I would add add them to existing modules only if it is clear that this will not complicate interface or code of the existing modules. A standalone prototype might be a good way of figuring it out anyway.

rajdeepdas2000:
I am interested to work on this issue.

wenzeslaus:
[...]

PR #2489, opened 2022-07-21
GUI: call G_gisinit when forms are run standalone (e.g. from parser)

marisn:
Repeated calls to G_gisinit are safe; When forms.py are loaded as a part of other code, one can assume that
G_gisinit has already been (or will be) called.

See #2484 for discussion.


gui/wxpython/gui_core/forms.py
```diff
@@ -79,6 +79,9 @@
 
 # needed when started from command line and for testing
 if __name__ == "__main__":
+    from grass.lib.gis import G_gisinit
```
wenzeslaus, line 82: The import here is too soon. The other GRASS-related imports are only later. (This needs a fix with GRASS Python libs just being installed in the system and rest of the code can just assume they are there, but the current situation is that the code can't assume that.)

All ctypes imports need to be in `try-except ImportError` block as e.g. in `gui/wxpython/nviz/wxnviz.py`. The idea is that the GUI components start and work even if ctypes are broken for some reason. Only the parts which need to work can fail, ideally gracefully. (Standalone NVIZ does not have to work, but main GUI should.)

wenzeslaus (changes requested):

gui/wxpython/gui_core/forms.py
marisn: I made a better version – import is done only before use. As G_gisinit("") is safe to call after initialization has been already done, calling it one more time wouldn't harm.
When  a separate C module for signature file handling will be made, this code will have to change. In the meantime – it is better than failing GUI.

wenzeslaus (approved):
Great. You could also do that in `_append_mapset_signatures` because after calling that function, `items` can be already empty anyway. Interfacing with C would be then limited to `_append_mapset_signatures` only.

marisn:
Is it better now?

wenzeslaus (approved):
Thank you for addressing the review! I merged the main branch just to run the test with latest code. (The last run was old and failed for unrelated reason.)

PR #2499, opened 2022-07-26
tests: Add timeout to pytest

wenzeslaus:
Now the individual pytest tests are terminated after 300 seconds just like the gunittest ones. New dependency pytest-timeout is needed for testing to use the timeout (or avoid the warning about unused configuration). In the future, individual tests can change the time if needed using the pytest.mark.timeout decorator.


wenzeslaus:
pytest checks pass (OSGeo4W is broken). Merging.

Let's see if some individual tests timeout. This should prevent CI from waiting 6 hours for stuck tests.

PR #2468, opened 2022-06-28
CI: ugly hack to debug Windows builds in CI fail

neteler:
Attempt to see more (error) messages.

Related to bug #2454

PR #2508, opened 2022-08-02
CI: introduce setup-msys2@v2 setup action

ninsbl:
For using MSYS in CI (esp. github) it is recommended to use [msys2/setup-msys2@v2](https://github.com/msys2/setup-msys2) action. See also: https://www.msys2.org/docs/ci/

This PR introduces that action and should make the build pass again without errors.

Replaces #2468 

Should also speedup the test a few minutes and maybe simplifies to explore other possibilities (e.g. using UCRT)

### DIFF
--- a/.github/workflows/osgeo4w.yml
+++ b/.github/workflows/osgeo4w.yml
@@ -27,6 +27,11 @@ jobs:
           git config --global core.autocrlf false
           git config --global core.eol lf
       - uses: actions/checkout@v2
+      - uses: msys2/setup-msys2@v2
+        with:
+          msystem: MINGW64
+          update: true
+          install: tar libintl make bison flex diffutils git dos2unix zip mingw-w64-x86_64-toolchain mingw-w64-x86_64-cairo mingw-w64-x86_64-fftw mingw-w64-x86_64-lapack mingw-w64-x86_64-pkgconf mingw-w64-x86_64-gcc mingw-w64-x86_64-ccache mingw-w64-x86_64-zlib mingw-w64-x86_64-libiconv mingw-w64-x86_64-bzip2 mingw-w64-x86_64-gettext mingw-w64-x86_64-libsystre mingw-w64-x86_64-libtre-git mingw-w64-x86_64-libwinpthread-git mingw-w64-x86_64-libpng mingw-w64-x86_64-pcre mingw-w64-x86_64-python3-six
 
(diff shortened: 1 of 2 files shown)


ninsbl:
Hmmm.... This seem to resurface #1697 as compilation now fails with:  

```
D:\a\grass\grass\lib\cairodriver/text.c:207: undefined reference to `cairo_ft_font_face_create_for_pattern'
collect2.exe: error: ld returned 1 exit status
```

@nilason any idea?

hellik (commented):
The yml changes are looking good.

Regarding the cairo issue, some more checks why it pops up again may be good.

ninsbl:
Working now...

Since OSGeo4W checks succeed for #2506 too, it seems the issue was a bug in MSYS2. So, I guess checks would now pass also without these changes. Seems to be messed up OSGeo4W vs. MSYS2 cairo...

Still, the PR makes the workflow a bit cleaner and avoids mixing OSGeo4W and MSYS2 cairo library. And using `msys2/setup-msys2@v2` seems to be an advantage anyway, so I thing the changes are still an improvement...

BTW: cairo-devel from OSGeo4W does not seem to provide a cairo.pc, thus the warning about pkg-config...

neteler (approved):
Great work - I cannot judge in detail but the fact that the CI job finally compiles again is a good sign :)

hellik (approved):
Looks good! Thanks the PR!

nilason:
> Hmmm.... This seem to resurface #1697 as compilation now fails with:
> 
> ```
> D:\a\grass\grass\lib\cairodriver/text.c:207: undefined reference to `cairo_ft_font_face_create_for_pattern'
> collect2.exe: error: ld returned 1 exit status
> ```
> 
> @nilason any idea?

I did take a look at the Cairo code and couldn't find any related changes there. I suppose(d) it was build set up related somehow... Great you found the way!

ninsbl:
Backported in #2510

wenzeslaus:
FYI, the OSGeo4W check passed [four days ago on August 1](https://github.com/OSGeo/grass/runs/7611369095?check_suite_focus=true) for c0a83a39655ece962595b06c2dbe6b6c87954f5c. That's before this was merged in 53b69fbdadacd2f9f3a9005f9efd5bd2de7e6d32. See the recent commits on main:

https://github.com/OSGeo/grass/commits/main

PR #2513, opened 2022-08-05
wxGUI/gui_core: fix import Rasterlite DB raster

tmszi:
**Describe the bug**
Import Rasterlite DB raster via wxGUI Import raster data dialog doesn't work.

**To Reproduce**

1. Launch wxGUI
2. Export some raster map into Rasterlite DB e.g. `r.out.gdal --overwrite input=elevation@PERMANENT output=/tmp/elevation.db format=Rasterlite`
3. From the Data catalog toolbar activate Import raster data tool
4. From the Import raster data dialog Source type choose Database RadioButton widget
5. From the Source input Format ComboBox widget choose Rasterlite
6. Set Name TextCtrl widget Rasterlite DB path `/tmp/elevation.db`
7. List of raster layers ListCtrl widget is empty, no raster layers are loaded.

**Expected behavior**
Loading Rasterlite DB raster layer and importing via wxGUI Import raster data dialog  should be work.

**Screenshots**

Current behavior:

wxgui_import_raster_data_rasterlite_db_raster_current

Expected behavior:

wxgui_import_raster_data_rasterlite_db_raster_exp

- Operating System: all
- GRASS GIS version: all


tmszi:
Rebase https://github.com/OSGeo/grass/commit/e0377def61f3a8ed1a0d00b857fe8e29ff59f6c3.

gui/wxpython/gui_core/gselect.py
```diff
@@ -2646,6 +2657,128 @@ def _getPGDBRasters(self, dsn):
...
+            raster_tables,
+        )
+        if table_names_without_suffix:
+            from collections import Counter
+
+            # [:-1] filter out last separator
+            raster_tables_count = Counter(
+                table_names_without_suffix[:-1].split(raster_table_sep)
+            )
+            # Filter out other tables which is not raster tables (has only
```
neteler, line 2765:
```suggestion
            # Filter out other tables which are not raster tables (has only
```

tmszi:
Rebase https://github.com/OSGeo/grass/commit/c0668a97ccad9129785252ed63404f8ec8bee3e3.

gui/wxpython/gui_core/gselect.py
```diff
@@ -2370,6 +2370,17 @@ def getProjMatchCaption(projectionMatch):
                     )
                     data.append((layerId, raster, int(projectionMatch), grassName))
                     layerId += 1
+            elif self.dbWidgets["format"].GetStringSelection() == "Rasterlite":
+                rasters = self._getRasterliteDBRasters(dsn)
```
petrasovaa, line 2374: I don't know anything about rasterlite, but that's a LOT of code just to list rasters. Can't GDAL be used for that? If it's really needed, maybe we should consider moving this somewhere else.

tmszi: > I don't know anything about rasterlite, but that's a LOT of code just to list rasters. Can't GDAL be used for that? If it's really needed, maybe we should consider moving this somewhere else.

That was the first solution I looked at, but for some reason I decided not to use `gdalinfo`. I looked again today and yes you are right the command `gdalinfo -json elevation.sqlite` is sufficient.

tmszi:
With https://github.com/OSGeo/grass/pull/2513/commits/3ef684ebfec1945b0f2bab37c9a4c28cf17d599d is used `gdalinfo` for getting list of raster names of Rasterlite DB. Sucessfully tested with importing one or multiple rasters from Rasterlite DB.

gui/wxpython/gui_core/gselect.py
petrasovaa: Would `r.external` give you sufficient information? GDAL tools are not required dependency. If there is no tool in GRASS that would work, I am for using gdalinfo, but there would have to be better error handling in case users don't have gdal tools installed.

neteler: Maybe @rouault has a suggestion here?

rouault: > Maybe @rouault has a suggestion here?

if you can use the GDAL python bindings:
```
ds = gdal.Open("my.rasterlite")
ds.GetSubDatasets()
```
cf https://gis.stackexchange.com/questions/72178/how-to-extract-subdataset-from-hdf-raster

tmszi: > Would `r.external` give you sufficient information? GDAL tools are not required dependency. If there is no tool in GRASS that would work, I am for using gdalinfo, but there would have to be better error handling in case users don't have gdal tools installed.

`r.external` is not enough. Better error handling if Python `gdal` module is missing is fixed with https://github.com/OSGeo/grass/pull/2513/commits/da15c4eda7bb80ce22e7f7806d8e11a592ebd76c.

tmszi: > if you can use the GDAL python bindings:

Yes of course. Thanks.

petrasovaa: Looks like there is not a great solution here since neither gdal tools nor gdal python bindings are a GRASS dependency. I don't think python bindings are necessarily better solution than using gdal tools. Should we perhaps look at why we can't get the needed info from the existing grass tools that use gdal library? Maybe we can implement something there. Or maybe not, but I can't help with the decision without understanding this little more. 
Generally, it would be helpful and more effective to include some background information and reasoning behind the changes.

tmszi: What kind of information we need:

1. Get raster or rasters names from the Rasterlite DB.

I didn't found any GRASS GIS module (import or raster group) which was useful, that's was my first step before created this PR and  that's why I originally decided to use `db.select` module with SQL https://github.com/OSGeo/grass/pull/2513/commits/20b283052c3934be08166acb6ccf32e2d8e583ae.

gui/wxpython/gui_core/gselect.py
```diff
@@ -2652,6 +2663,27 @@ def _getPGDBRasters(self, dsn):
...
+        :param str dsn: Rasterlite DB data source name
+
+        :return list: list of Rasterlite DB rasters
+        """
+        try:
+            from osgeo import gdal
+        except ImportError:
+            GError(
+                parent=self,
+                message=_("The GDAL library is missing. Please install it."),
```
petrasovaa, line 2678:
```suggestion
                message=_("The Python osgeo package is missing. Please install it."),
```

gui/wxpython/modules/import_export.py
```diff
@@ -470,8 +470,31 @@ def OnRun(self, event):
...
+                        )
+                        return
+                    if dataset_info.stdout:
+                        import json
+
+                        dataset_info = json.loads(
+                            grass.decode(dataset_info.stdout),
+                        )
+                        # Rasterlite DB
+                        if "Rasterlite" in dataset_info["driverShortName"]:
```
petrasovaa, line 496: Wouldn't `self.dsnInput.GetFormat` be enough here instead of calling gdalinfo?

tmszi: > Wouldn't `self.dsnInput.GetFormat` be enough here instead of calling gdalinfo?

`self.dsnInput.GetFormat()` return empty string.

gui/wxpython/gui_core/gselect.py
tmszi: Fixed with https://github.com/OSGeo/grass/pull/2513/commits/0f2fccaa1efe22293793f4d22bfd787757fd980f.

gui/wxpython/modules/import_export.py
tmszi: The solution would be to use the Python Osgeo GDAL package again to get the driver name https://github.com/OSGeo/grass/pull/2513/commits/0f2fccaa1efe22293793f4d22bfd787757fd980f.

petrasovaa: > > Wouldn't `self.dsnInput.GetFormat` be enough here instead of calling gdalinfo?
> 
> `self.dsnInput.GetFormat()` return empty string.

Any idea why?

tmszi: > > > Wouldn't `self.dsnInput.GetFormat` be enough here instead of calling gdalinfo?
> > 
> > 
> > `self.dsnInput.GetFormat()` return empty string.
> 
> Any idea why?

`self.dsnInput.GetFormat()` return format in the case when you try import raster from Directory radio button widget and you choose format from the Format combo box widget.

gui/wxpython/gui_core/gselect.py
wenzeslaus: Confusingly, the package is called GDAL as afar as I can tell: https://pypi.org/project/GDAL/

tmszi: Fixed with https://github.com/OSGeo/grass/pull/2513/commits/685cb7af6288604fd4acd84549f5befc6d9402a6.

gui/wxpython/modules/import_export.py
```diff
@@ -470,8 +470,24 @@ def OnRun(self, event):
...
                 if "PG:" in dsn:
                     idsn = f"{dsn} table={layer}"
+                elif os.path.exists(idsn):
+                    try:
+                        from osgeo import gdal
+                    except ImportError:
+                        GError(
+                            parent=self,
+                            message=_(
+                                "The Python osgeo package is missing."
```
petrasovaa, line 483: It should be GDAL package as above.

petrasovaa (changes requested):
This can be merged, there is just the package name to fix.

gui/wxpython/modules/import_export.py
petrasovaa, line 483:
```suggestion
                                "The Python GDAL package is missing."
```

petrasovaa (approved):

PR #2519, opened 2022-08-10
libproj: simple check for area of use

metzm:
This PR adds a simple check if reprojection would fall outside the intended area of use of the target CRS.

I would prefer to refuse reprojection in these extreme cases, but this might be too restrictive. If subsequent errors occur because reprojected coordinates are invalid, this can be changed accordingly to actually refuse reprojection which is the current behaviour, but without explanation why.

### DIFF
--- a/lib/proj/do_proj.c
+++ b/lib/proj/do_proj.c
@@ -167,6 +167,20 @@ int get_pj_area(const struct pj_info *iproj, double *xmin, double *xmax,
                 *ymax = y[i];
         }
 
+        /* detect an attempt to reproject outside the intended area of use */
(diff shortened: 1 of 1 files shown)


wenzeslaus:
I have resolved the indentation-related conflicts here.

neteler:
@metzm shall I bump this PR to 8.4.0?

metzm:
This test is not working as expected. PROJ has its own mechanism to check for the area of use with its spatial criterions `PROJ_SPATIAL_CRITERION_STRICT_CONTAINMENT` and `PROJ_SPATIAL_CRITERION_PARTIAL_INTERSECTION`, the latter is used by default in GDAL.

PR #2524, opened 2022-08-14
utils: Remove HTMLParseError

wenzeslaus:
HTMLParseError is just an alias for Exception. There is no HTMLParseError in Python 3. The error reporting fails because of access to attributes.

This clears up so far the error reporting, but it needs complete removal of HTMLParseError. Further considerations about strict and non-strict parsing may be needed.


wenzeslaus:
Given that this fixes the error which happens when accessing the attributes in case of an exception, this is useful even without any other improvements to the rest of the code. It should be ready to merge, please review.

tmszi (approved):
It looks good to me. I tried to simulate the error by adding an unknown HTML element `<test>Test</test>` tag to the module manual HTML file and the compilation fail and the error was displayed as expected.

wenzeslaus:
Thanks @tmszi!

About the attributes included in HTMLParseError, at this point I think there is nothing to report, so this should be completely solved. (I indicated that in the commit message, too.)

PR #2525, opened 2022-08-14
wxGUI/vdigit: fix map window focus after starting edits

tmszi:
**Describe the bug**
After start editing some vector map, Map window doesn't have automatic focus, keyboard shortcut for activating Digitizer Point/Line tool doesn't work.  First, it is necessary to focus Map window with the left mouse button.

**To Reproduce**

1. Launch wxGUI
2. Activate Vector Digitizer from Map window toolbar ComboBox widget, choose Vector digitizer item
3. From Vector Digitizer toolbar ComboBox choose New vector map
4. Set new vector map e.g. vdigit on the Create new vector map dialog and hit OK button
5. Vector Attribute Table Manager window is opened
6. Close Vector Attribute Table Manager window
7. Try activate Vector Digitizer Point tool via keyboard shortcut Ctrl + P or Line tool Ctrl + L
8. Tool is not activated, because Map window doesn't have automatic focus (before you activate tool via keyboard shortcut you have to mouse left click on Map window to get focus)

or 

2. Add some vector map e.g. `d.vect census`
3. Activate Vector Digitizer from the Layers tree, right mouse click on the vector layer census to invoke layer menu and choose Start editing item
4. Try activate Vector Digitizer Point tool via keyboard shortcut Ctrl + P or Line tool Ctrl + L
5. Tool is not activated, because Map window doesn't have automatic focus (before you activate tool via keyboard shortcut you have to mouse left click on Map window to get focus)

**Expected behavior**
After start editing some vector map, Map window should be have automatic focus and keyboard shortcut for activating Digitizer Point/Line tool work immediately.

**System description:**

- Operating System: all
- GRASS GIS version: all


petrasovaa (changes requested):
The part in OnCloseWindow would set focus on window in every case, even when you are not using digitizer, so that doesn't seem to be a good solution, it's not particularly elegant anyway. You could implement some sort of callback that would be registered only for cases when db manager is called from digitizer, but that seems unnecessarily complex for the benefit. 
Unless you have some idea what to do with this, I suggest to leave only the change in toolbars.py.

petrasovaa (approved):

PR #2529, opened 2022-08-20
man: allow register all addons keywords in main Keywords Index

tmszi:
Implement https://github.com/OSGeo/grass-addons/issues/776.


neteler (approved):
Thanks, works fine for https://github.com/OSGeo/grass-addons/issues/776

PR #2532, opened 2022-08-20
init: Fix sudo message on Ubuntu

### LINKED ISSUE #753: [Feat] Remove run as administrator from startup message in CLI
The message after starting GRASS GIS in CLI contains the following on Linux with Bash:

```
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.
```

This is caused by a file `.sudo_as_admin_successful` missing from the "fake" home directory GRASS GIS uses to start the shell in `lib/init/grass.py`.

One solution would be to reduce the tricks done in grass.py which would than pick up the `~/.sudo_as_admin_successful` file. Another much simple is to create this file in the "fake" home.

The "fake" home is the mapset, so maybe probably creating and deleting it later is desired or perhaps some "try create" is the best.

The simpler solution with just creating the file makes this actually a good first issue. No reworking of "fake" home tricks in grass.py is needed to resolve this issue.

susi:
Fixes #753 by creating `<mapset>/.sudo_as_admin_successful` when `$HOME/.sudo_as_admin_successful` exists

### DIFF
--- a/lib/init/grass.py
+++ b/lib/init/grass.py
@@ -2032,7 +2032,7 @@ def sh_like_startup(location, location_name, grass_env_file, sh):
     """.format(
             sh_history=sh_history
         )
-        # Ubuntu sudo creates a file .sudo_as_admin_successful and cbash checks
+        # Ubuntu sudo creates a file .sudo_as_admin_successful and bash checks
         # for this file in the home directory from /etc/bash.bashrc and prints a
         # message if it's not detected. This can be suppressed with either
         # creating the file ~/.sudo_as_admin_successful (it's always empty) or
--- a/lib/init/grass.py
+++ b/lib/init/grass.py
@@ -2039,6 +2039,12 @@ def sh_like_startup(location, location_name, grass_env_file, sh):
     """.format(
             sh_history=sh_history
         )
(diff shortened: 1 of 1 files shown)


wenzeslaus (changes requested):
Can you please add a comment to the code explaining what it is about, so that one does not have to go all the way to the GitHub issue to see reasoning for this?

Advantages of new file versus copying the file is not clear to me. Either one works at this point I guess.

The Black reformat is unfortunate, but given your plans to restructure grass.py, I think it is fine to leave it here although it unrelated.

susi:
Added a massive explanation :stuck_out_tongue:

lib/init/grass.py
```diff
@@ -2039,6 +2032,21 @@ def sh_like_startup(location, location_name, grass_env_file, sh):
     """.format(
             sh_history=sh_history
         )
+        # Ubuntu sudo creates a file .sudo_as_admin_successful and cbash checks
```
wenzeslaus, line 2035: cbash, typo or a thing?

susi: It's a typo...

wenzeslaus (approved):

PR #2537, opened 2022-08-26
v.random: do not transfer cat from area if attributes are coppied too

### LINKED ISSUE #2475: [Bug] v.to.rast use=attr does not convert all points
For some reason v.to.rast fails to convert all vector points if as a value source is set attribute table. Although attribute column contains 2000 not null values, only 113 raster points are generated. If value source is set to val, all 2000 points are present in output raster.

A further investigation needs to be done to determine why cat values in do_line L40 coming back from Vect_read_line are botched.

marisn:
Current code creates two cat entries for each point on the same level -
one from area point is in and one matching new entry in the database.
As entries in db have new cat values, old cat value points to a random
or non-existing entry in the db thus causing a mess.
Fixes #2475.


wenzeslaus:
I resolved the conflict in the GitHub web interface by using the new line.

PR #2547, opened 2022-08-27
HTML header charset changed from ISO-8859-1 to UTF-8

landam:
+ mkhtml.py simplified (py2-specific code removed)


wenzeslaus (approved):
I think the files are already in UTF-8, so this should work.

wenzeslaus:
I resolved the conflict by combining the two lines in the GitHub web interface.

PR #2548, opened 2022-08-27
version: Update GRASS version to 8.2.1dev

landam:
Self-explanatory


neteler (approved):
(... how could this happen...)

wenzeslaus:
Procedure updated in #2551.

PR #2549, opened 2022-08-27
lib/db: Fix formatting of MySQL error message

wenzeslaus:
One of the three %s placeholders was escaped. Now the translatable message contains the placeholders.


wenzeslaus:
The issue was identified by GitHub syntax coloring.

PR #2568, opened 2022-09-02
wxGUI: Fix behaviour of top Single-Window GUI toolbars

lindakladivova:
Fixes the non-standard behavior of top Single-Window GUI toolbars.

**Bug:**
- bigger icons when undocking and docking the toolbar (takes into consideration that one does not drop a toolbar during that process)
bigger icons
after redocking all toolbars were resized

- no toolbar name displayed after undocking
no-name

**After fix:**
Now it inherits from AuiToolbar. This approach corresponds to the usage of aui agw library which is used for drawing the main window frame.

After conversion to AuiToolbar:

- sizes of toolbars are still the same
toolbar

- toolbar name is displayed after undock
worrkspace_toolbar

Potentially this PR could also fix some other related bugs  (probably different depending on OS) caused by inheriting from the inappropriate wx.Toolbar class.

### DIFF
--- a/gui/wxpython/gui_core/toolbars.py
+++ b/gui/wxpython/gui_core/toolbars.py
@@ -304,8 +304,8 @@ class AuiToolbarController(ToolbarController):
     """Controller specialized for wx.lib.agw.aui.auibar.AuiToolBar subclass"""
 
     def _defineTool(self, name=None, icon=None, handler=None, item=wx.ITEM_NORMAL):
-        """Define tool.
-        Position is not needed since wx.lib.agw.aui.auibar.AuiToolBar does not have InsertTool method."""
+        """Define tool. Position is not needed since wx.lib.agw.aui.auibar.AuiToolBar
+        does not have InsertTool method."""
         if name:
             return (
                 name,
--- a/gui/wxpython/gui_core/toolbars.py
+++ b/gui/wxpython/gui_core/toolbars.py
@@ -20,6 +20,7 @@ import platform
 import os
 
 import wx
+import wx.lib.agw.aui as aui
 
 from core import globalvar
 from core.debug import Debug
@@ -29,11 +30,6 @@ from core.globalvar import IMGDIR
 from gui_core.wrap import ToolBar, Menu, BitmapButton, NewId
 
 from grass.pydispatch.signal import Signal
-
-try:
-    from agw import aui
-except ImportError:
-    import wx.lib.agw.aui as aui
 
 
 BaseIcons = {
@@ -103,13 +99,14 @@ class ToolbarController:
         self.parent = parent
         self.toolSwitcher = toolSwitcher
         self.handlers = {}
+        self.data = None
 
     def InitToolbar(self, toolData):
         """Initialize toolbar, add tools to the toolbar"""
         for tool in toolData:
             self.CreateTool(*tool)
 
-        self.widget.setToolData(toolData)
+        self.data = toolData
 
     def CreateTool(self, label, bitmap, kind, shortHelp, longHelp, handler, pos=-1):
         """Add tool to the toolbar
@@ -131,7 +128,7 @@ class ToolbarController:
                 3, "CreateTool(): tool=%d, label=%s bitmap=%s" % (tool, label, bitmap)
             )
             if pos < 0:
-                toolWin = self.classObject.AddLabelTool(
+                toolWin = self.classObject.AddTool(
                     self.widget,
                     tool,
                     label,
@@ -142,7 +139,7 @@ class ToolbarController:
                     longHelp,
                 )
             else:
-                toolWin = self.classObject.InsertLabelTool(
+                toolWin = self.classObject.InsertTool(
                     self.widget,
                     pos,
                     tool,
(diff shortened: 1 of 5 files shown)


gui/wxpython/gui_core/toolbars.py
```diff
@@ -30,6 +30,11 @@
 
 from grass.pydispatch.signal import Signal
 
+try:
+    from agw import aui
+except ImportError:
+    import wx.lib.agw.aui as aui
+
```
petrasovaa, line 37: Please check current import style and update if needed

gui/wxpython/gui_core/toolbars.py
```diff
@@ -304,6 +300,221 @@ def CreateSelectionButton(self, tooltip=_("Select graphics tool")):
         return button
 
 
+class AuiToolbarController(ToolbarController):
+    """Provides handling for aui.AuiToolBar widget"""
```
petrasovaa, line 304: agw?

gui/wxpython/gui_core/toolbars.py
```diff
@@ -304,6 +300,221 @@ def CreateSelectionButton(self, tooltip=_("Select graphics tool")):
...
+
+    def _defineTool(self, name=None, icon=None, handler=None, item=wx.ITEM_NORMAL):
+        """Define tool"""
```
petrasovaa, line 307: Explain why it needs to be different

gui/wxpython/gui_core/toolbars.py
```diff
@@ -172,7 +166,7 @@ def EnableLongHelp(self, enable=True):
 
         :param enable: True for enable otherwise disable
         """
-        for tool in self._data:
+        for tool in self.widget._data:
```
petrasovaa, line 169: Accessing private attribute

gui/wxpython/gui_core/toolbars.py
```diff
@@ -304,6 +300,221 @@ def CreateSelectionButton(self, tooltip=_("Select graphics tool")):
...
+        """@copydoc ToolbarController:_getToolbarData()"""
+        return self.controller.GetToolbarData(data)
+
+    def _onMenu(self, data):
+        """@copydoc ToolbarController:_onMenu()"""
+        self.controller.OnMenu(data)
+
+    def InitToolbar(self, toolData):
+        """@copydoc ToolbarController:InitToolbar()"""
+        self.controller.InitToolbar(toolData)
```
petrasovaa, line 404:
```
    def __getattr__(self, name):
        return getattr(self.controller, name)
```

gui/wxpython/gui_core/toolbars.py
lindakladivova: I checked in wxPython Demo version 4.1.1. ale it's used like that

gui/wxpython/gui_core/toolbars.py
lindakladivova: The problem is that AUI does not have InsertTool.. As I get it right, it is quite a new method in wx which is probably the reason. Look at https://www.geeksforgeeks.org/wxpython-inserttool-function-in-wx-toolbar/.

petrasovaa:
I will merge this to get it tested more, could @nilason or @tmszi test it on Mac and Windows? The basic test is if the GUI launches at all, then it's worth testing undocking and docking the main toolbar, and switching to 3D and back.

It should be backported, but only after sufficient testing.

petrasovaa (approved):

tmszi:
> I will merge this to get it tested more, could @nilason or @tmszi test it on Mac and Windows? The basic test is if the GUI launches at all, then it's worth testing undocking and docking the main toolbar, and switching to 3D and back.

Successfully tested on the MS Windows.

nilason:
Tested on Mac with:
Python 3.10.7 (v3.10.7:6cc6b13308, Sep  5 2022, 14:02:52) [Clang 13.0.0 (clang-1300.0.29.30)]
4.2.0 osx-cocoa (phoenix) wxWidgets 3.2.0


Most things works smoothly and as expected. Actually a very pleasant experience, good work!!

But switching to and fro 3D mode causes a glitch:







Secondly, initially the main buttons are not visible:



...resizing the panel frame causes a redraw, which fixes that:



Finally, the Layer tool bar, starts displaced, and you also see a Map Display tab (it is also to be found where it should be in the middle pane):



after a window resize (just enough to initialise a redraw) they come to the right place but the tool bar seems too high:

tmszi:
> Secondly, initially the main buttons are not visible:
> 
> 
> 
> ...resizing the panel frame causes a redraw, which fixes that:
> 
> 

The same on MS Windows OS. I missed it the first time I tested it.

petrasovaa:
Thanks! Any idea if these issues were there before this was merged?

lindakladivova:
> I will merge this to get it tested more, could @nilason or @tmszi test it on Mac and Windows? The basic test is if the GUI launches at all, then it's worth testing undocking and docking the main toolbar, and switching to 3D and back.
> 
> It should be backported, but only after sufficient testing.

@petrasovaa Anna, thank you for your testing and looking into this PR more deeply. I did not manage to react earlier, I am sorry.
Thank also @nilason and @tmszi for testing!

PR #2573, opened 2022-09-09
r.kappa: Fix failures, garbage output, fallback to category values

marisn:
r.kappa has some strange design decisions and produces strange or outright wrong output if input data doesn't match its expectations.
This PR fixes:
* printing garbage instead of NA for the first raster category for commision, ommision and kappa values;
* printing raster category values if labels are missing for matrix-only output mode;

### DIFF
--- a/raster/r.kappa/prt2csv_mat.c
+++ b/raster/r.kappa/prt2csv_mat.c
@@ -97,7 +97,7 @@ void prn2csv_error_mat(int out_cols, int hdr)
             if (cl)
                 G_strip(cl);
             if (cl == NULL || *cl == 0)
-                fprintf(fd, "%d\t", cats[j]);
+                fprintf(fd, "%ld\t", cats[j]);
             else
                 fprintf(fd, "%s\t", cl);
         }
@@ -113,7 +113,7 @@ void prn2csv_error_mat(int out_cols, int hdr)
             if (cl)
                 G_strip(cl);
             if (cl == NULL || *cl == 0)
-                fprintf(fd, "%d\t", cats[rndx]);
+                fprintf(fd, "%ld\t", cats[rndx]);
             else
                 fprintf(fd, "%s\t", cl);
             /* entries */
--- a/raster/r.kappa/prt2csv_mat.c
+++ b/raster/r.kappa/prt2csv_mat.c
@@ -97,8 +97,9 @@ void prn2csv_error_mat(int out_cols, int hdr)
             if (cl)
                 G_strip(cl);
             if (cl == NULL || *cl == 0)
-                cl = "NULL";
-            fprintf(fd, "%s\t", cl);
+                fprintf(fd, "%d\t", cats[j]);
+            else
+                fprintf(fd, "%s\t", cl);
         }
         /*for (cndx = first_col; cndx < last_col; cndx++) */
         /*    fprintf(fd, "%ld\t", rlst[cndx]); */
@@ -112,8 +113,9 @@ void prn2csv_error_mat(int out_cols, int hdr)
             if (cl)
                 G_strip(cl);
             if (cl == NULL || *cl == 0)
-                cl = "NULL";
-            fprintf(fd, "%s\t", cl);
+                fprintf(fd, "%d\t", cats[rndx]);
+            else
+                fprintf(fd, "%s\t", cl);
             /* entries */
             for (cndx = first_col; cndx < last_col; cndx++) {
                 thisone = (ncat * rndx) + cndx;
--- a/raster/r.kappa/testsuite/test_r_kappa.py
+++ b/raster/r.kappa/testsuite/test_r_kappa.py
@@ -75,8 +75,9 @@ ColSum\t7\t3\t4\t0\t1\t3\t18
         self.assertEqual(out, ref)
 
 
-class CalculationCorrectnessTest(TestCase):
-    """Test correctness of values in printout mode"""
+class CalculationCorrectness1Test(TestCase):
+    """Test correctness of values in printout mode
+    First version of reference data"""
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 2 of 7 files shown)


raster/r.kappa/prt2csv_mat.c
```diff
@@ -97,8 +97,9 @@
             if (cl)
                 G_strip(cl);
             if (cl == NULL || *cl == 0)
-                cl = "NULL";
-            fprintf(fd, "%s\t", cl);
+                fprintf(fd, "%d\t", cats[j]);
```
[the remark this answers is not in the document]
wenzeslaus: GCC reports in CI:

```
2022-09-09T15:04:03.2858422Z gcc  -std=gnu99 -fPIC -Werror  -I/home/runner/work/grass/grass/dist.x86_64-pc-linux-gnu/include -I/home/runner/work/grass/grass/dist.x86_64-pc-linux-gnu/include    -DPACKAGE=\""grassmods"\"   -I/home/runner/work/grass/grass/dist.x86_64-pc-linux-gnu/include -I/home/runner/work/grass/grass/dist.x86_64-pc-linux-gnu/include -DRELDIR=\"raster/r.kappa\" -o OBJ.x86_64-pc-linux-gnu/prt2csv_mat.o -c prt2csv_mat.c
2022-09-09T15:04:03.3238001Z prt2csv_mat.c: In function ‘prn2csv_error_mat’:
2022-09-09T15:04:03.3238991Z prt2csv_mat.c:100:31: error: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘long int’ [-Werror=format=]
2022-09-09T15:04:03.3239423Z   100 |                 fprintf(fd, "%d\t", cats[j]);
2022-09-09T15:04:03.3239718Z       |                              ~^     ~~~~~~~
2022-09-09T15:04:03.3239969Z       |                               |         |
2022-09-09T15:04:03.3240242Z       |                               int       long int
2022-09-09T15:04:03.3240504Z       |                              %ld
2022-09-09T15:04:03.3241073Z prt2csv_mat.c:116:31: error: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘long int’ [-Werror=format=]
2022-09-09T15:04:03.3241529Z   116 |                 fprintf(fd, "%d\t", cats[rndx]);
2022-09-09T15:04:03.3241845Z       |                              ~^     ~~~~~~~~~~
2022-09-09T15:04:03.3242115Z       |                               |         |
2022-09-09T15:04:03.3242373Z       |                               int       long int
2022-09-09T15:04:03.3242636Z       |                              %ld
2022-09-09T15:04:03.3403299Z cc1: all warnings being treated as errors
2022-09-09T15:04:03.3422870Z make[4]: *** [../../include/Make/Compile.make:32: OBJ.x86_64-pc-linux-gnu/prt2csv_mat.o] Error 1
```

wenzeslaus (commented):
GCC and CodeQL are not happy. Otherwise no idea without testing or tests, sorry.

Probably another candidate for JSON output, but that's out of scope for this PR.

marisn:
> GCC and CodeQL are not happy. Otherwise no idea without testing or tests, sorry.

There are more issues as some values are calculated incorrectly if input data is not sequential numbers starting at 0. Thus just a draft as more work is needed.

> Probably another candidate for JSON output, but that's out of scope for this PR.

I was thinking of shell script style with "-g" flag, but going for JSON also is an option. In separate PR, of course, after this PR is merged as this PR will have to be backported.

raster/r.kappa/testsuite/test_r_kappa.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def test_standard_output(self):
+        out = read_command(
+            "r.kappa",
+            reference=self.ref_1,
+            classification=self.class_1,
+            flags="w",
+            quiet=True,
+        )
+        rows = out.split("\n")
```
wenzeslaus, line 160: Use `splitlines()`.

raster/r.kappa/testsuite/test_r_kappa.py
```diff
@@ -0,0 +1,296 @@
...
+
+        # Per class coefficients
+        for c in range(len(self.per_class["producer"])):
+            vals = rows[20 + c].split(sep=None)
+            self.assertTrue(self.match(vals[1], self.per_class["user"][c]))
+            self.assertTrue(self.match(vals[2], self.per_class["producer"][c]))
+            self.assertTrue(self.match(vals[3], self.per_class["cond.kappa"][c]))
+
+        # Kappa value
+        vals = rows[28].split(sep=None)
```
wenzeslaus, line 285: Given the examples in documentation and tutorials, I think it is generally understood that `split()` operates on whitespace (or at least spaces) and/or that it is the same as as `split(sep=None)`.

The [str.split](https://docs.python.org/3/library/stdtypes.html?highlight=split#str.split) documentation says:

> If sep is not specified or is None, a different splitting algorithm is applied: runs of consecutive whitespace are regarded as a single separator, and the result will contain no empty strings at the start or end if the string has leading or trailing whitespace. Consequently, splitting an empty string or a string consisting of just whitespace with a None separator returns [].

Using `"a\tb".split()` gives `['a', 'b']`.

wenzeslaus (approved):
The code changes make sense to me in light of the individual commit messages (please include these descriptions into the final commit message, you could add pieces of them as comments too).

I executed the tests locally with the original code and in the one case where it actually runs and gives results, I get the same values as the ones in the tests, so the outputs which are working were not broken.

Thank you for the extra effort with the correctness test! I was really just thinking some "current values" check to see the before and after difference (or no difference).

I only noticed sub-optimal use of split in Python in tests which would be nice to have before merge.

raster/r.kappa/testsuite/test_r_kappa.py
marisn: Thanks.

raster/r.kappa/testsuite/test_r_kappa.py
marisn: Thanks.

PR #2574, opened 2022-09-10
man: use relative path for core modules man HTML page

tmszi:
Fixes https://github.com/OSGeo/grass-addons/issues/791.

### DIFF
--- a/man/build_keywords.py
+++ b/man/build_keywords.py
@@ -67,7 +67,7 @@ def get_module_man_html_file_path(module):
             ".",
         )
     else:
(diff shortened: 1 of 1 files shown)


tmszi:
https://github.com/OSGeo/grass/commit/051c85d294c163cd37a54eb32240f58cb6f4a491 same relative path style `./` as for addons.

neteler:
This solves the problem, the `grass83/manuals/...` like fine now.

All recent fixes/updates would need to be backported to grass82 as well.

PR #1474, opened 2021-03-21
Add zones in t.rast.univar

NikosAlexandris:
This is NOT a real pull request.  Rather a prototype idea.
Not extensively tested, no real tests written.

PR #2588, opened 2022-09-20
t.rast.univar / t.rast3d.univar: Add support for zones

ninsbl:
Inspired by #1474, this PR adds support for a zones map in t.rast.univar and t.rast3d.univar, covered by tests.

Maybe also `semantic_labels` should be added to the output (if the STRDS contains some)? What do you think. Happy to add it if you consider it useful.

Should I add a check if the ` zones` map is of type CELL?

Of course, I don` expect you all to review, but I would be happy about some feedback, esp. with regards to the two questions above.


ninsbl:
P.S.: Like #1474, this implements only a single, static raster map for zones. That is probably the most common use case (at least it is for me). However, using another STRDS with a defined temporal relation could be interesting as well, but would be a different PR...

veroandreo:
> Inspired by #1474, this PR adds support for a zones map in t.rast.univar and t.rast3d.univar, covered by tests.
> 
> Maybe also `semantic_labels` should be added to the output (if the STRDS contains some)? What do you think. Happy to add it if you consider it useful.

I think it would be useful, especially for STRDS of multiple bands. However, we faced a problem recently when teaching at FOSS4G, if you estimate NDVI with t.rast.mapcalc using a STRDS of S2 b4 and b8, the inherited semantic label is b8 (Section 10 [here](https://github.com/veroandreo/foss4g2022_grass4rs/blob/main/grassgis4rs.ipynb))... if we then ask for univar stats, this will also inherit an incorrect semantic label. In any case, it would be a nice addition.

> Should I add a check if the ` zones` map is of type CELL?

Sounds reasonable to me. Are there use cases in which zones might be float?

Time series of zones would be cool too...

ninsbl:
Thanks for all the testing you did (also on addons!).
The semantic_labels and time series for zones are likely issues for different PRs.
Will address the CELL-check ASAP.

ninsbl:
STRDS as input for zones could probably be implemented like https://github.com/OSGeo/grass/blob/releasebranch_8_2/temporal/t.rast.aggregate.d

However, a more general question then is if this should be implemented as a separate module (e.g. `t.rast.univar.ds`) or as part of `t.rast.univar`. Both options have pros and cons, but I am leaning towards implementing that as a new module (if there isn't an add on already)...

ninsbl:
> Sounds reasonable to me. Are there use cases in which zones might be float?

The point of such a check would be mostly to catch a user error, as `r.univar` would refuse ro work.

ninsbl:
Does anyone have an idea why the **CentOS** build fails to import `RasterRow`?

The error occurs when the manual is build. The `RasterRow` module seems rarely used in the core code-base outside of tests. So maybe this is another, more fundamental OS/distro specific issue surfacing here, as tests for CentOS only cover build and execution of the `grass` command?

nilason:
According to the log the build first fails at:


```
  File "/github/home/anaconda3/lib/python3.7/ctypes/__init__.py", line 97, in CFUNCTYPE
    return _c_functype_cache[(restype, argtypes, flags)]
KeyError: (<class 'ctypes.c_int'>, (<class 'grass.lib.ctypes_preamble.String'>,), 1)

  File "/__w/grass/grass/dist.x86_64-pc-linux-gnu/scripts/t.rast.univar", line 69, in <module>
    from grass.pygrass.raster import RasterRow
  File "/__w/grass/grass/dist.x86_64-pc-linux-gnu/etc/python/grass/pygrass/raster/__init__.py", line 19, in <module>
  File "/__w/grass/grass/dist.x86_64-pc-linux-gnu/etc/python/grass/lib/gis.py", line 566, in <module>
    ('checker', CFUNCTYPE(UNCHECKED(c_int), String)),
  File "/github/home/anaconda3/lib/python3.7/ctypes/__init__.py", line 99, in CFUNCTYPE
    class CFunctionType(_CFuncPtr):
TypeError: item 1 in _argtypes_ passes a union by value, which is unsupported.
make[3]: *** [t.rast.univar.tmp.html] Error 1
```

also, as far as I can tell from the log, it is using Python 3.7.6:

```
...
The following NEW packages will be INSTALLED:
...
python             pkgs/main/linux-64::python-3.7.6-h0371630_2
...
```

which is infamously bad with ctypes (eg. https://trac.osgeo.org/grass/ticket/4018 or https://trac.osgeo.org/grass/ticket/3883).

ninsbl:
Thanks @nilason for your swift reply! Now I changed the code so that `RasterRow` is no longer used. The CentOS issue remains, but does not surface here any more.

veroandreo:
> STRDS as input for zones could probably be implemented like https://github.com/OSGeo/grass/blob/releasebranch_8_2/temporal/t.rast.aggregate.d
> 
> However, a more general question then is if this should be implemented as a separate module (e.g. `t.rast.univar.ds`) or as part of `t.rast.univar`. Both options have pros and cons, but I am leaning towards implementing that as a new module (if there isn't an add on already)...

Well, yes, perhaps a separate module is the easiest. IIUC, t.rast.aggregate.ds aggregates a STRDS with the granularity of another STRDS, so it's indeed a good starting point for the temporal sampling part (the input STRDS would need to be sampled with the granularity of the zones STRDS), the next step would be the spatial aggregation to get the stats for the zones in each time.. There will be then methods for temporal and methods for spatial aggregation... Yeah, better in a separate module

python/grass/temporal/univar_statistics.py
```diff
@@ -109,19 +124,34 @@ def print_gridded_dataset_univar_statistics(
...
+                    zone = str(stats["zone"])
+                    eol = "\n"
+                else:
+                    eol = ""
+            string += fs + str(stats["mean"]) + fs + str(stats["min"])
+            string += fs + str(stats["max"]) + fs + str(stats["mean_of_abs"])
+            string += fs + str(stats["stddev"]) + fs + str(stats["variance"])
+            string += fs + str(stats["coeff_var"]) + fs + str(stats["sum"])
+            string += fs + str(stats["null_cells"]) + fs + str(stats["n"])
+            string += fs + str(int(stats["n"]))
```
wenzeslaus, line 148: When rewriting portions of code like this, use f-strings where you can. No reason to be consistent with outdated code (it was never best practice AFAIK and now is probably flagged by linters).

The `string +=` you probably need to keep to minimize changes, but I think `strings.append(...)` and `"".join(strings)` is the solution there (theoretically faster, practically I did not test).

temporal/t.rast.univar/testsuite/test_t_rast_univar.py
```diff
@@ -8,6 +8,7 @@
 @author Soeren Gebbert
 """
 
+import grass.script as gscript
```
wenzeslaus, line 11: Perhaps unused now, but if it should stay, use the `...as gs` practice.

temporal/t.rast.univar/t.rast.univar.py
```diff
@@ -87,11 +93,24 @@ def main():
...
+        no_header,
+        separator,
+        rast_region,
+        zones,
     )
 
 
 if __name__ == "__main__":
-    options, flags = grass.parser()
+    options, flags = gs.parser()
```
wenzeslaus, line 115: Since you are changing it anyway, can you move this to main to avoid an unnecessary global variable.

temporal/t.rast3d.univar/t.rast3d.univar.py
```diff
@@ -83,11 +91,16 @@ def main():
...
         output = None
 
+    # Check if zones map exists and is of type CELL
+    if zones:
+        if gs.raster.raster_info(zones)["datatype"] != "CELL":
+            gs.fatal(_("Zoning raster must be of type CELL"))
+
     tgis.print_gridded_dataset_univar_statistics(
-        "str3ds", input, output, where, extended, no_header, separator
+        "str3ds", input, output, where, extended, no_header, separator, False, zones
```
wenzeslaus, line 100: `False` definitively needs keyword argument to provide context.

temporal/t.rast3d.univar/testsuite/test_t_rast3d_univar.py
```diff
@@ -162,6 +169,43 @@ def test_6_error_handling_no_input(self):
         # No input
         self.assertModuleFail("t.rast3d.univar", output="out.txt")
 
+    def test_7(self):
```
wenzeslaus, line 172: Let's not continue the bad practice of numbering the functions instead of naming them.

wenzeslaus (changes requested):
Just some small style changes. I did not check the logic.

temporal/t.rast3d.univar/t.rast3d.univar.py
ninsbl: I agree

temporal/t.rast3d.univar/testsuite/test_t_rast3d_univar.py
ninsbl: Right.

ninsbl:
Thanks for the thorough review and feedback. Will address this later today.

temporal/t.rast3d.univar/testsuite/test_t_rast3d_univar.py
ninsbl: Adressed in [5e8c340](https://github.com/OSGeo/grass/pull/2588/commits/5e8c340160734e60a8f506f2563f25fd882f06ca) and [f6b9729](https://github.com/OSGeo/grass/pull/2588/commits/f6b972962c678a042d8152cc5b92075a4fcf92f4)

temporal/t.rast3d.univar/t.rast3d.univar.py
ninsbl: Adressed in [5ba0978](https://github.com/OSGeo/grass/pull/2588/commits/5ba09782ef9c3e7df0c9d0613e51a8624ada1b42) and [086b431](https://github.com/OSGeo/grass/pull/2588/commits/086b4317bf07f921035fd45bd4c515b0d257ed9f)

temporal/t.rast.univar/t.rast.univar.py
ninsbl: Adressed in [5ba0978](https://github.com/OSGeo/grass/pull/2588/commits/5ba09782ef9c3e7df0c9d0613e51a8624ada1b42) and [086b431](https://github.com/OSGeo/grass/pull/2588/commits/086b4317bf07f921035fd45bd4c515b0d257ed9f)

ninsbl:
Hm... Any Ctypes import seems to break the _CentOS 7_ build...
Is it worth keeping _CentOS 7_ tests in the light of that? If I understand correctly _CentOS_ transitioned to _CentOS Stream_. Unfortunately, newer CentOS images, like Stream, are not available in GitHub CI out of the box...

ninsbl:
Tests are passing for ubuntu: https://github.com/OSGeo/grass/pull/2588/checks#step:10:216 and https://github.com/OSGeo/grass/pull/2588/checks#step:10:222
However, as mentioned CetnOS build tests fail when importing tgis: https://github.com/OSGeo/grass/actions/runs/3206937539/jobs/5241279653#step:11:9830
That worked before, so failing tests may be due to the order of when the import occurs?

temporal/t.rast.univar/t.rast.univar.py
```diff
@@ -71,13 +76,15 @@ def main():
     import grass.temporal as tgis
 
     # Get the options
+    options, flags = gs.parser()
```
wenzeslaus, line 79: This should be in main, yes, but in the same order as when outside of it, so before the import of grass.temporal.

wenzeslaus:
> Any Ctypes import seems to break the _CentOS 7_ build... Is it worth keeping _CentOS 7_ tests in the light of that?

I would be reluctant to remove CentOS just because one module failure to build when the other modules work.

> Unfortunately, newer CentOS images, like Stream, are not available in GitHub CI out of the box...

Neither is CentOS 7, we are running it using a Docker image. Are there no Docker images for Stream?

With this said, the CentOS was always on shaky ground. It is partially using conda to get acceptable dependencies, so it is a poor test of CentOS. I was hoping it will move to more pure CentOS, but that did not happen yet. However, it does catch things which are related to robustness in sub-optimal environments.

neteler:
> Neither is CentOS 7, we are running it using a Docker image. Are there no Docker images for Stream?

We could use AlmaLinux:
https://hub.docker.com/_/almalinux

temporal/t.rast.univar/t.rast.univar.py
ninsbl: Thanks, that fixed the failing test.

ninsbl:
> I would be reluctant to remove CentOS just because one module failure to build when the other modules work.

Fair enough. And with your suggestion, tests are no longer failing. So I agree, best to keep it. Maybe improve the CentOS test in future, e.g. with the docker containers @neteler suggested...

ninsbl:
OK to merge? With this in, I can take on parallelization...

wenzeslaus (approved):
Thank you for addressing my concerns. The code is agreeable in general and it has tests. I did not review the functionality or logic.

If at least one of the other suggested reviewers reviews the logic or tests, that would be great. @lucadelu @neteler @NikosAlexandris @veroandreo

ninsbl:
@lucadelu @neteler @NikosAlexandris @veroandreo if one of you has a chance to test, that would be very much appreciated? Otherwise, we can trust CI and merge the next days...

lucadelu:
@ninsbl sorry but in this period I have really no time to do this

ninsbl:
@lucadelu , fully understand. Given that the module is covered by tests and results in tests are unchanged, I merge the change.

PR #2596, opened 2022-10-12
gpde: fix error handing parameters

BadAssassin:
Fixes usage of Rast3d_fatal_error()

### DIFF
--- a/lib/gpde/n_arrays_io.c
+++ b/lib/gpde/n_arrays_io.c
@@ -363,7 +363,7 @@ N_array_3d *N_read_rast3d_to_array_3d(char *name, N_array_3d * array,
 
     /* Close files and exit */
     if (!Rast3d_close(map))
-        Rast3d_fatal_error(_("Error closing g3d file"));
+        Rast3d_fatal_error(_("Error closing g3d file <%s>"), name);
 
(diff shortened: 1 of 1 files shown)


lib/gpde/n_arrays_io.c
```diff
@@ -363,7 +363,7 @@ N_array_3d *N_read_rast3d_to_array_3d(char *name, N_array_3d * array,
 
     /* Close files and exit */
     if (!Rast3d_close(map))
-        Rast3d_fatal_error(map, NULL, 0, _("Error closing g3d file"));
+        Rast3d_fatal_error(_("Error closing g3d file"));
```
wenzeslaus, line 366: `name` seems to be available here (with the right content). Can you please change it as you did in #2597?

wenzeslaus (changes requested):
Thank you. If it is possible, please change it to show the name.

lib/gpde/n_arrays_io.c
nilason: I took the liberty to add this.

wenzeslaus (approved):
Looks good. Thanks @BadAssassin and @nilason!

PR #2602, opened 2022-10-20
docker Ubuntu: update to Ubuntu 22.04, update PDAL and laz-perf

neteler:
- `docker/ubuntu/Dockerfile`
    - update to Ubuntu 22.04 base image
    - update to latest PDAL (to be compiled since laz-perf missing from Ubuntu)
    - update to latest laz-perf
- `Dockerfile`
    - sync Dockerfile in main directory to `docker/ubuntu/Dockerfile` (avoids current confusion, fixes #2575)

Issue #2575, opened 2022-09-10
[Bug] Remove outdated Dockerfile


neteler:
https://github.com/OSGeo/grass/blob/6b61d1087ba9cf7429df8830c636a827bc5c3f5d/Dockerfile#L1

This file should be removed as the latest ones are in `docker/`.

wenzeslaus:
Nice thing about Dockerfile in the top directory was that `docker build` worked with GitHub repo URL and things such as `docker build -t grassgis .` from the readme work too. Is that still a nice feature to have even with the Docker directory to create reusable Docker images? Seems friendlier for experiments and it is short. Is there an easy way to point docker build to a specific file?

neteler:
Conserving this outdated dockerfile:

- in GRASS GIS 8.2.x we cannot remove the outdated file (as that would be a breaking change) but we could sync it to docker/ubuntu/Dockerfile.

- in main (= upcoming G 8.3.0) we could delete the file unless the comment above is prevailing. Then still the fact that the file is badly outdated remains. No idea if a redirect to a subdirectory is possible in git.

wenzeslaus:
> ...in GRASS GIS 8.2.x we cannot remove the outdated file (as that would be a breaking change)...in main (= upcoming G 8.3.0) we could delete the file unless...

Breaking change between 8.2. and 8.3 would break expectations in semantic versioning. As the [RFC Version Numbering draft](https://github.com/OSGeo/grass/pull/2357) says:

> Use version number in a format _major.minor.micro_ where
> - _major_ is for large additions or backwards incompatible changes,
> - _minor_ is for additions, and
> - _micro_ is for fixes.

While the semantic versioning is expected even now, the RFC is a draft, so we can change things there if needed.

PR #2604, opened 2022-10-21
contributing: Enforce that duplicate files are the same

wenzeslaus:
Sometimes two files in the repo need to be the same. Various Dockerfiles are in docker directory, but having a (the one) Dockerfile in the main directly makes sense, too. A new check enforces that the file in synced with the other copy.

### DIFF
--- a/.github/workflows/additional_checks.yml
+++ b/.github/workflows/additional_checks.yml
@@ -38,6 +38,10 @@ jobs:
         with:
           python-version: '3.10'
 
+      - name: Check that files with the same content are the same
+        run: |
+          diff Dockerfile docker/ubuntu/Dockerfile
+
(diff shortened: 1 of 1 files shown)


wenzeslaus:
This is trying to resolve issues raised in #2575 (needs two files, but files need to be in sync) and is waiting for #2602 (files are actually in sync).

wenzeslaus:
The two tested files `Dockerfile` and `docker/ubuntu/Dockerfile` differ. So this PR works great!

However, is the difference intentional? I guess #2771 needs to be done for `Dockerfile`. @neteler ?

neteler:
> However, is the difference intentional? I guess #2771 needs to be done for `Dockerfile`. @neteler ?

Gnarf. Addressed in #2820

PR #2605, opened 2022-10-21
nsis installer changed SID of grass82.py

dnewcomb:
In theory, this should fix [2603](https://github.com/OSGeo/grass/issues/2603). On Windows 10 Professional , the permissions are correct.  Will need to test on Windows 10 Enterprise


ninsbl (approved):
It is a small change that seems fine to me. But I have not tested and I do not think this is covered by CI. So is should be tested on the build server.
But it should be fairly easy to change (back) if needed... We should wait for one successfull installer build and test before backporting though...

hellik (approved):
it seems ok, as @ninsbl suggested wait for the next build for merging PR

PR #2611, opened 2022-10-26
lib config: create dynamic library of C++ libraries

nilason:
Up till now linking C++ based libraries (at this point only the **iostream** library) was forced into creating a static library. This PR enables the possibility to create a dynamic library with the new configure flag 'SHLIB_LDX', which could use CXX instead of CC.

`IS_CXX = yes` must be set before including the 'Lib.make' file in the library's Makefile.


nilason:
GH doesn't recognise additional commits... ??? Closing this and opening a new one.

PR #2616, opened 2022-10-28
CI: Address Windows specific issues in testsuite

ninsbl:
Replaces: #1683

-----

July 2026 AI analysis of what is possibly still relevant:

- [x] ~r.mfilter: the NamedTemporaryFile → gs.tempfile fix is a textbook Windows bug (can't reopen an open NamedTemporaryFile) and directly explains the 6 xfail'd tests there.~ https://github.com/OSGeo/grass/pull/7878
- [ ] grassdb/test_manage.py: the os.sep-aware assertions map one-to-one onto the 4 xfail'd tests.
- [ ] test_parser_json.py: uses raw subprocess.Popen (not grass.script), so the shell= on win32 fix is still relevant; all 3 tests are xfail'd.
- [ ] test_checkers.py / test_assertions.py: the newline="" fixes for MD5-checked files still apply, though test_assertions.py now writes via Path.write_text(), so the fix needs rewriting.
- [ ] case.py: normalizing/stripping the reference string in assertLooksLike/assertModuleKeyValue (main only normalizes actual).
- [ ] toolboxes.sh (python→python3) and v.edit/select_all_flag.sh (tr -d '\r'): both files are still on the exclusion list in .github/workflows/osgeo4w_gunittest.cfg, so the fixes are still applicable.

These would be better as individual PRs.

### DIFF
--- a/gui/wxpython/core/testsuite/toolboxes.sh
+++ b/gui/wxpython/core/testsuite/toolboxes.sh
@@ -3,5 +3,5 @@
 # run make in gui/wxpython before the test
 # run test using sh -e
 
-python $GISBASE/gui/wxpython/core/toolboxes.py doctest
-python $GISBASE/gui/wxpython/core/toolboxes.py test
+python3 $GISBASE/gui/wxpython/core/toolboxes.py doctest
+python3 $GISBASE/gui/wxpython/core/toolboxes.py test
--- a/gui/wxpython/core/toolboxes.py
+++ b/gui/wxpython/core/toolboxes.py
@@ -791,6 +791,8 @@ def do_doctest_gettext_workaround():
 
     if sys.version_info[0] == 2:
         import __builtin__
+    elif locals().get("__builtin__", globals().get("__builtin__")) is None:
+        import builtins as __builtin__
 
     __builtin__._ = new_translator
 
--- a/gui/wxpython/core/toolboxes.py
+++ b/gui/wxpython/core/toolboxes.py
@@ -792,7 +792,7 @@ def do_doctest_gettext_workaround():
     if sys.version_info[0] == 2:
         import __builtin__
 
-        __builtin__._ = new_translator
+    __builtin__._ = new_translator
 
 
 def doc_test():
--- a/gui/wxpython/core/toolboxes.py
+++ b/gui/wxpython/core/toolboxes.py
@@ -794,6 +794,7 @@ def do_doctest_gettext_workaround():
 
         __builtin__._ = new_translator
 
+
 def doc_test():
     """Tests the module using doctest
 
--- a/gui/wxpython/core/toolboxes.py
+++ b/gui/wxpython/core/toolboxes.py
@@ -27,11 +27,6 @@ if hasattr(etree, "ParseError"):
 else:
     ETREE_EXCEPTIONS = expat.ExpatError
 
-if sys.version_info[0:2] > (2, 6):
-    has_xpath = True
-else:
-    has_xpath = False
-
 import grass.script.task as gtask
 import grass.script.core as gcore
 from grass.script.utils import try_remove, decode
@@ -430,18 +425,7 @@ def _expandToolboxes(node, toolboxes):
             items = n.find("./items")
             idx = list(items).index(subtoolbox)
 
-            if has_xpath:
-                toolbox = toolboxes.find(
-                    './/toolbox[@name="%s"]' % subtoolbox.get("name")
-                )
-            else:
-                toolbox = None
-                potentialToolboxes = toolboxes.findall(".//toolbox")
-                sName = subtoolbox.get("name")
-                for pToolbox in potentialToolboxes:
-                    if pToolbox.get("name") == sName:
-                        toolbox = pToolbox
-                        break
+            toolbox = toolboxes.find('.//toolbox[@name="%s"]' % subtoolbox.get("name"))
 
             if toolbox is None:  # not in file
                 continue
@@ -576,15 +560,7 @@ def _expandItems(node, items, itemTag):
     """
     for moduleItem in node.findall(".//" + itemTag):
         itemName = moduleItem.get("name")
-        if has_xpath:
-            moduleNode = items.find('.//%s[@name="%s"]' % (itemTag, itemName))
-        else:
-            moduleNode = None
-            potentialModuleNodes = items.findall(".//%s" % itemTag)
-            for mNode in potentialModuleNodes:
-                if mNode.get("name") == itemName:
-                    moduleNode = mNode
-                    break
+        moduleNode = items.find('.//%s[@name="%s"]' % (itemTag, itemName))
 
         if moduleNode is None:  # module not available in dist
             continue
@@ -789,12 +765,9 @@ def do_doctest_gettext_workaround():
     sys.displayhook = new_displayhook
     sys.__displayhook__ = new_displayhook
 
-    if sys.version_info[0] == 2:
-        import __builtin__
-    elif locals().get("__builtin__", globals().get("__builtin__")) is None:
-        import builtins as __builtin__
+    import builtins as _builtins
 
-    __builtin__._ = new_translator
+    _builtins.__dict__["_"] = new_translator
 
 
 def doc_test():
--- a/gui/wxpython/core/toolboxes.py
+++ b/gui/wxpython/core/toolboxes.py
@@ -789,10 +789,10 @@ def do_doctest_gettext_workaround():
     sys.displayhook = new_displayhook
     sys.__displayhook__ = new_displayhook
 
-    import __builtin__
+    if sys.version_info[0] == 2:
+        import __builtin__
 
-    __builtin__._ = new_translator
-
+        __builtin__._ = new_translator
 
 def doc_test():
     """Tests the module using doctest
--- a/lib/gis/testsuite/test_parser_json.py
+++ b/lib/gis/testsuite/test_parser_json.py
@@ -83,7 +83,9 @@ class TestParserJson(TestCase):
             }
         ]
 
-        stdout, stderr = subprocess.Popen(args, stdout=subprocess.PIPE).communicate()
+        stdout, stderr = subprocess.Popen(
+            args, stdout=subprocess.PIPE, shell=sys.platform == "win32"
+        ).communicate()
         print(stdout)
         json_code = json.loads(decode(stdout))
         self.assertEqual(json_code["module"], "v.out.ascii")
--- a/lib/gis/testsuite/test_parser_json.py
+++ b/lib/gis/testsuite/test_parser_json.py
@@ -8,10 +8,12 @@ for details.
 @author Soeren Gebbert
 """
 
+import json
 import subprocess
+import sys
+
 from grass.gunittest.case import TestCase
 from grass.script import decode
-import json
 
 
 class TestParserJson(TestCase):
@@ -50,7 +52,9 @@ class TestParserJson(TestCase):
             },
         ]
 
-        stdout, stderr = subprocess.Popen(args, stdout=subprocess.PIPE).communicate()
+        stdout, stderr = subprocess.Popen(
+            args, stdout=subprocess.PIPE, shell=sys.platform == "win32"
+        ).communicate()
         print(stdout)
         json_code = json.loads(decode(stdout))
         self.assertEqual(json_code["module"], "r.slope.aspect")
@@ -101,7 +105,9 @@ class TestParserJson(TestCase):
             {"param": "layer", "value": "1"},
         ]
 
-        stdout, stderr = subprocess.Popen(args, stdout=subprocess.PIPE).communicate()
+        stdout, stderr = subprocess.Popen(
+            args, stdout=subprocess.PIPE, shell=sys.platform == "win32"
+        ).communicate()
         print(stdout)
         json_code = json.loads(decode(stdout))
         self.assertEqual(json_code["module"], "v.info")
--- a/lib/init/testsuite/test_grass_tmp_mapset.py
+++ b/lib/init/testsuite/test_grass_tmp_mapset.py
@@ -30,7 +30,7 @@ class TestTmpMapset(unittest.TestCase):
     """Tests --tmp-mapset option of grass command"""
 
     # TODO: here we need a name of or path to the main GRASS GIS executable
-    executable = "grass"
+    executable = shutil.which("grass")
     # an arbitrary, but identifiable and fairly unique name
     location = "test_tmp_mapset_xy"
 
--- a/lib/init/testsuite/test_grass_tmp_mapset.py
+++ b/lib/init/testsuite/test_grass_tmp_mapset.py
@@ -18,7 +18,7 @@ import unittest
 import os
 import shutil
 import subprocess
-
+import sys
 
 # Note that unlike rest of GRASS GIS, here we are using unittest package
 # directly. The grass.gunittest machinery for mapsets is not needed here.
@@ -46,7 +46,8 @@ class TestTmpMapset(unittest.TestCase):
     def test_command_runs(self):
         """Check that correct parameters are accepted"""
         return_code = subprocess.call(
-            [self.executable, "--tmp-mapset", self.location, "--exec", "g.proj", "-g"]
+            [self.executable, "--tmp-mapset", self.location, "--exec", "g.proj", "-g"],
+            shell=sys.platform == "win32",
         )
         self.assertEqual(
             return_code,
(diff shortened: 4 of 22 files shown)


ninsbl:
With this PR success-rate of unit tests on MS Windows is up to 87% (from 80%), with 243 tests passing (almost as many as on Linux). There are a couple of tests left (e.g. `i.maxlik` or  `v.what` tests), where I suspect that either tests (or assertions in the testsuite (e.g. with `v.what`) are not really cross platform. For the rest we may start looking at the module code to check if the test failure actually is a bug. 
An example, where I here fixed the test, but test failure actually may be a bug is the `v.in.lidar` module. That module seems to behave differently on MS Windows and Linux with regards to projection handling and that may be actually a bug...

Anyway, I am happy about all sorts of feedback on this PR.

gui/wxpython/core/testsuite/toolboxes.sh
```diff
@@ -3,5 +3,5 @@
 # run make in gui/wxpython before the test
 # run test using sh -e
 
-python $GISBASE/gui/wxpython/core/toolboxes.py doctest
-python $GISBASE/gui/wxpython/core/toolboxes.py test
+python3 $GISBASE/gui/wxpython/core/toolboxes.py doctest
+python3 $GISBASE/gui/wxpython/core/toolboxes.py test
```
wenzeslaus, line 7: Do we have GRASS_PYTHON defined here or not? If yes, that might be the best choice, although I would prefer solution where plain python just works.

gui/wxpython/core/toolboxes.py
```diff
@@ -789,9 +789,10 @@ def new_translator(string):
     sys.displayhook = new_displayhook
     sys.__displayhook__ = new_displayhook
 
-    import __builtin__
+    if sys.version_info[0] == 2:
+        import __builtin__
 
-    __builtin__._ = new_translator
+        __builtin__._ = new_translator
```
wenzeslaus, line 795: The "underscore" issue should still apply, so I would be surprised if doing it just for 2 is enough. Besides that, if there is 2-only code somewhere, it should be deleted. (But I think that's not the case here.)

lib/init/testsuite/test_grass_tmp_mapset.py
```diff
@@ -30,13 +30,16 @@ class TestTmpMapset(unittest.TestCase):
     """Tests --tmp-mapset option of grass command"""
 
     # TODO: here we need a name of or path to the main GRASS GIS executable
-    executable = "grass"
+    executable = shutil.which("grass")
```
wenzeslaus, line 33: I would think only shutil.which or shell=True are needed for Windows, not both. I would prefer shutil.which or some equivalent of sys.executable in the _grass_ package (e.g., grass.script.setup.executable).

python/grass/gunittest/gmodules.py
```diff
@@ -138,4 +139,11 @@ def call_module(
     returncode = process.poll()
     if returncode:
         raise CalledModuleError(module, kwargs, returncode, errors)
-    return decode(output) if output else None
+    output = decode(output) if output else None
+    # Make sure that universal newlines are returned
+    output = (
+        output.replace(os.linesep, "\n")
+        if os.linesep != "\n" and type(output) == str
```
wenzeslaus, line 146: `type(output) is str`?

raster/r.mfilter/testsuite/test_r_mfilter.py
```diff
@@ -1,4 +1,4 @@
-from tempfile import NamedTemporaryFile
+import grass.script as gscript
```
wenzeslaus, line 1: `import grass.script as gs`

raster/r.mfilter/testsuite/test_r_mfilter.py
```diff
@@ -184,7 +184,7 @@ class TestNeighbors(TestCase):
 
     def create_filter(self, options):
         """Create a temporary filter file with the given name and options."""
-        f = NamedTemporaryFile()
+        f = gscript.tempfile(create=False)
```
wenzeslaus, line 187: Returns string, the original is an IO object.

scripts/g.extension/testsuite/test_addons_download.py
```diff
@@ -120,7 +120,9 @@ def test_github_install_official(self):
 
         for file in files:
             self.assertFileExists(file)
-            if file.suffix != ".html":
+            if file.suffix != ".html" and not ms_windows:
+                # Prebuild C-modules for MS Windows are unlikely to run
+                # on fresh compiled GRASS GIS
```
wenzeslaus, line 125: I thought this should fail at g.extension step already.

vector/v.edit/testsuite/select_all_flag.sh
```diff
@@ -15,7 +16,7 @@ v.edit -r map=vedit_test tool=catadd layer=2 cats=10
 
 expected="10
 10"
-out=$(v.category option=print layer=2 input=vedit_test)
+out=$(v.category option=print layer=2 input=vedit_test| tr -d '\r')
```
wenzeslaus, line 19: BTW, skipping all .sh-based tests on Windows would be a fair solution I think.

ninsbl: I have no strong opinion here. MSYS has been dropped, so it is very unlikely that users will run shell scripts in real world cases. And I am not sure this is a real MS Windows test in that sense.
However, most of the shell tests work. So maybe it does not hurt to have them running for MS Windows as well...

scripts/g.extension/testsuite/test_addons_download.py
ninsbl: Addons are pre-build for MS Windows also for the development version:
https://wingrass.fsv.cvut.cz/grass83/addons/grass-8.3.dev/logs/
However, getting the addon-help fails because build versions differ...

gui/wxpython/core/toolboxes.py
ninsbl: Right. We do not support Python 2 anymore, so it is probably even better to remove Python 2 specific code to cleanup (also other places where Python 2 code pops up)?

lib/init/testsuite/test_grass_tmp_mapset.py
ninsbl: I see. Unfortunately, both shutil.which and shell do not seem to help. Maybe better to switch to subprocess.run?

gui/wxpython/core/toolboxes.py
```diff
@@ -789,7 +789,10 @@ def new_translator(string):
     sys.displayhook = new_displayhook
     sys.__displayhook__ = new_displayhook
 
-    import __builtin__
+    if sys.version_info[0] == 2:
+        import __builtin__
+    elif locals().get("__builtin__", globals().get("__builtin__")) is None:
+        import builtins as __builtin__
 
     __builtin__._ = new_translator
```
wenzeslaus, line 797: Just drop the Python 2 part from here. The rest can be simpler, `python/grass/init.py` uses:

```python
import builtins as _builtins
_builtins.__dict__["_"] = _translate
```

lib/init/testsuite/test_grass_tmp_mapset.py
wenzeslaus: subprocess.run is just an interface to Popen like all the others. I don't expect it would help. shutil.which needs `grass` to be on path to work which is likely not fulfilled.

This test is now unique - we don't have other tests which test the executable - we we may have more in the future or even write such code, so perhaps ignoring it now and adding `grass.script.setup.executable` later is the way to go.

raster/r.mfilter/testsuite/test_r_mfilter.py
```diff
@@ -184,10 +184,10 @@ class TestNeighbors(TestCase):
...
     def create_filter(self, options):
         """Create a temporary filter file with the given name and options."""
-        f = NamedTemporaryFile()
-        f.write(options)
-        f.flush()
-        return f
+        grass_tempfile = gs.tempfile(create=False)
+        with open(grass_tempfile, "wb") as f:
+            f.write(options)
+        return grass_tempfile
```
wenzeslaus, line 190: I still don't get what's wrong with NamedTemporaryFile.

scripts/g.extension/testsuite/test_addons_download.py
wenzeslaus: Oh, I missed what test_github_install_official is. That makes sense. Mention the different build versions in the comment.

raster/r.mfilter/testsuite/test_r_mfilter.py
ninsbl: I am a bit puzzled by that as well. `NamedTemporaryFile` leads to errors in the OSGeo4W tests. See:
https://github.com/OSGeo/grass/actions/runs/3368434482/jobs/5586972317#step:9:3031

I guess there is some _cygwin_-path-wrangling going on, maybe. At least the `C:\\Users\\RUNNER~1` short form in the path string is not working. If the GRASS GIS DB is in the users home directory, the grass.tempfile may fail as well with the same issue... It is jus a try (as I am not coding on Windows)... If that fails too, we may revert and try to resolve the Path in a way that it is not shortened....

gui/wxpython/core/toolboxes.py
ninsbl: I droped all Python 2 from that test now...

ninsbl:
> subprocess.run is just an interface to Popen like all the others. I don't expect it would help. shutil.which needs grass to be on path to work which is likely not fulfilled.

> This test is now unique - we don't have other tests which test the executable - we we may have more in the future or even write such code, so perhaps ignoring it now and adding grass.script.setup.executable later is the way to go.

I am fine with that.

wenzeslaus (commented):
I don't get the `"does\\not\\exist" if ms_windows else "does/not/exist"` changes. Windows supports `/` and our code should support it on Windows, too, so we should test for it (and fix it if needed). Adding additional tests for Windows for support of `\` that would be different.

ninsbl:
> I don't get the `"does\\not\\exist" if ms_windows else "does/not/exist"` changes. Windows supports `/` and our code should support it on Windows, too, so we should test for it (and fix it if needed). Adding additional tests for Windows for support of `\` that would be different.

Yes, but at least in the environment we run the tests in, paths are put together with OS specific separators, so tests fail with OSGeo4W

ninsbl:
> I'm just working from what is in the PR, but it seems to me that if tests fail in this case, the issue is in the actual code, not tests. The current tests need to work because we need to support `/` as a separator everywhere and we need to support paths with mixed separators (which is what likely happens in the current test cases).

Path-handling with pathlib across plattforms is unfortunately still not trivial as some functions like `resolve()` behave differently on OSes, in certain cases e.g. if paths do not exist in the file system (see: https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573). I guess that affects the tests too.
However, from my understanding, the main problem with the `test_manage.py` test is that we compare string representations of paths in some `assertEqual` functions to strings with universal separators (`/`). When transforming Path (which use `/`) to string objects, the pathlib uses platform-specific separators. So in ` str(Path().home())` the separator `/` is turned into `\` on MS Windows. Thus a string representation of a WindowsPath will not be `does/not/exist` but `does\not\exist`. To fix this we could replace `/` with `os.sep` in the assertation as tried in the latest commit.

> In any case, this looks like a material for a separate PR (and this PR is getting too long anyway).

Agreed. So I wrap up here. I am fine with reverting any change you suggest dropping for now... I can also split the PR...
And thanks for your patience and willingness to review repeatedly!

wenzeslaus:
> Path-handling with pathlib across platforms is unfortunately still not trivial as some functions like `resolve()` behave differently on OSes, in certain cases e.g. if paths do not exist in the file system...

I put a note about that to _grass.grassdb.create.resolve_mapset_path_ and that's one place where GRASS GIS needs to deal with that. If the behavior is platform-dependent because of the libraries, so be it. We just need to document that and test on each platform what makes sense there.

> ...To fix this we could replace `/` with `os.sep` in the assertation as tried in the latest commit.

Fixing the assertions and just those which need that sounds great.

> Agreed. So I wrap up here. I am fine with reverting any change you suggest dropping for now... I can also split the PR...

Smaller PRs are likely easier to review here. Generally, I think it is easier for people to approve or at least +1 thing related to one piece of code they know then >20 files. The *.in.lidar stuff could even go in without review given that it will be dropped in the future.  You could also separate out anything which is controversial to get the other stuff in quickly.

> And thanks for your patience and willingness to review repeatedly!

You are welcome. Ha ha, yes, this saturates all my current review time, but it is very important to get this fixed, so I'm happy to see that happening. Thank you!

raster/r.mfilter/testsuite/test_r_mfilter.py
wenzeslaus: Putting together the [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) doc and [this SE](https://stackoverflow.com/questions/18903069/how-can-i-read-namedtemporaryfile-in-python), I think that we need to use delete=False, write to the file, close it, and then delete it manually.

lib/init/testsuite/test_grass_tmp_mapset.py
wenzeslaus: This is one the of things for a separate PR. Getting the _grass.script.setup.executable_ piece working would be nice to have. [get_install_path](https://github.com/OSGeo/grass/blob/main/python/grass/script/setup.py#L115) is perhaps a good starting point (Python package path is known, path to installation is not). Lazy-loading seems appropriate, but there are still no properties for modules yet (?), but there is [__getattr__](https://peps.python.org/pep-0562/).

lib/gis/testsuite/test_parser_json.py
```diff
@@ -50,7 +52,9 @@ def test_r_slope_aspect_json(self):
             },
         ]
 
-        stdout, stderr = subprocess.Popen(args, stdout=subprocess.PIPE).communicate()
+        stdout, stderr = subprocess.Popen(
+            args, stdout=subprocess.PIPE, shell=sys.platform == "win32"
+        ).communicate()
```
wenzeslaus, line 57: This would probably make more sense with direct support for this in grass.script, but to limit the solution only to here, would shutil.which work here at line 22?

vector/v.edit/testsuite/select_all_flag.sh
wenzeslaus: Okay, please add comment that the \r part is for Windows, so we know when to remove it.

However, I think the role of Bash/sh/... tests is to provide option for "quick and dirty" tests and for very specific cases, so supporting only Linux or unix is fine. Something to keep in mind.

echoix:
Do you still think this PR is the way forward? It has conflicts, but for instance the builtins import issue has need fixed in another PR in January, and there was a PR recently that changed a temporary file/NamedTemporaryFile, and it was the only usage in the code at the time.

echoix:
I've skimmed through very quickly, and I think cutting out some line endings in the tests only for Windows is maybe not the right approach. It is something that we should probably fix for real instead, and to not care too much about io with Python, as it is handled way more transparently than on Python 2

echoix:
Is there still something useful and relevant here that should be kept, 3 years later? Our testing situation is a bit better now, our Python code quality is way up since more than a year and a half, grass start up and initialization, especially for usage as a Python library is recently better, etc.

petrasovaa:
AI analysis of this PR, what is possibly still relevant:

- r.mfilter: the NamedTemporaryFile → gs.tempfile fix is a textbook Windows bug (can't reopen an open NamedTemporaryFile) and directly explains the 6 xfail'd tests there.
- grassdb/test_manage.py: the os.sep-aware assertions map one-to-one onto the 4 xfail'd tests.
- test_parser_json.py: uses raw subprocess.Popen (not grass.script), so the shell= on win32 fix is still relevant; all 3 tests are xfail'd.
- test_checkers.py / test_assertions.py: the newline="" fixes for MD5-checked files still apply, though test_assertions.py now writes via Path.write_text(), so the fix needs rewriting.
- case.py: normalizing/stripping the reference string in assertLooksLike/assertModuleKeyValue (main only normalizes actual).
- toolboxes.sh (python→python3) and v.edit/select_all_flag.sh (tr -d '\r'): both files are still on the exclusion list in .github/workflows/osgeo4w_gunittest.cfg, so the fixes are still applicable.

These would be better as individual PRs.

ninsbl:
> AI analysis of this PR, what is possibly still relevant:

* t.rast.import / t.vect.import is fixed in #7742

We can keep the PR open for reference or open an issue, close this PR and point to the PR in the issue...  Both is fine with me. I agree it should be implemented in separate, smaller PRs...

PR #2617, opened 2022-10-29
wxGUI: fix display of all single window mode main toolbar tools

tmszi:
Current main toolbar tools layout (some tools are hidden): 

wxgui_main_toolbars_tool_current

Expected main toolbar tools layout:

wxgui_main_toolbars_tool_expected

- Operating System: MS Windows, macOS
- GRASS GIS version: 8.3.dev

**Additional context**

Another solution is to set margins after this line: 
https://github.com/OSGeo/grass/blob/c492ae50bcb2f349bd64393ef85624b5b80c9e03/gui/wxpython/gui_core/toolbars.py#L392

```python
if platform.system() in ("Windows", "Darwin"):
    self.SetMargins(8, 8)
```

#2568

### DIFF
--- a/gui/wxpython/gui_core/toolbars.py
+++ b/gui/wxpython/gui_core/toolbars.py
@@ -386,7 +386,7 @@ class AuiToolbar(aui.AuiToolBar):
         parent,
         toolSwitcher=None,
         style=wx.NO_BORDER | wx.TB_HORIZONTAL,
-        agwStyle=aui.AUI_TB_PLAIN_BACKGROUND,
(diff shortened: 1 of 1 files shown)


petrasovaa (approved):
Thanks for looking into it, it's rather weird issue. This solution seems fine.

tmszi:
Another solution is use agwStyle `wx.aui.AUI_TB_DEFAULT_STYLE` only.

nilason (approved):
Works fine for me. Thank you!

PR #2619, opened 2022-11-01
wxGUI: fix setting of user defined window position and size for single window mode

### LINKED ISSUE #2600: Set initial window size in single-window mode - currently default window is too large.
**Description**
The initial window size in "single-window" mode is too large - the window width covers both of my two-screen setup.

**To Reproduce**
1. Launch GRASS.
2. In Settings, choose "single-window" mode.
3. Restart GRASS.
4. See that the GRASS window upon open occupies the full screen (or screens).

**Expected behavior**
It would be great if the initial window size would be remembered between sessions, or allow the user to set the window size in preferences.

- Operating System: Linux CentOS 7
- GRASS GIS version 8.2.0

**Additional context**

I have tried changing the default settings in `.grass8/wx.json` as below, but these do not seem to have any effect.

```
  "general": {
    "defWindowPos": {
      "enabled": true,
      "dim": "1600,100,1280,1024"
    },
```

tmszi:
Fixes #2600.

How to test this PR:

1. Launch wxGUI with single window mode
2. Change and remember visually wxGUI window size and position 
3. Open wxGUI settings via menu Settings -> Preferences
4. On the General page (tab) from the Workspace setting **check Save current windows layout as default**
5. Hit Preferences dialog Save button
6. Close wxGUI
7. Launch wxGUI and check window position and size

If you want to change the wxGUI position and size again, you need to open the wxGUI preferences dialog again and press the Save button again.

### DIFF
--- a/gui/wxpython/main_window/frame.py
+++ b/gui/wxpython/main_window/frame.py
@@ -189,6 +189,8 @@ class GMFrame(wx.Frame):
             except Exception:
                 pass
             self.Layout()
+            if (w, h) <= globalvar.GM_WINDOW_SIZE:
+                self.Fit()
         else:
             self.Layout()
             self.Fit()
--- a/gui/wxpython/main_window/frame.py
+++ b/gui/wxpython/main_window/frame.py
@@ -189,7 +189,7 @@ class GMFrame(wx.Frame):
             except Exception:
                 pass
             self.Layout()
-            if (w, h) <= globalvar.GM_WINDOW_SIZE:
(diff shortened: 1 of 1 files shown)


petrasovaa (commented):
Works fine, thank you! I don't think the tuple comparison works well here, it's not the same thing as `if w <= GM_WINDOW_SIZE[0] or h  <= GM_WINDOW_SIZE[1]` if that's what you meant.

tmszi:
> I don't think the tuple comparison works well here, it's not the same thing as `if w <= GM_WINDOW_SIZE[0] or h <= GM_WINDOW_SIZE[1]` if that's what you meant.

Yes you are right. Thank you.

petrasovaa (approved):

PR #2630, opened 2022-11-05
v.in.ogr: skip columns with unsupported data type instead of failing to import

ninsbl:
Before this PR, vector data failed to import if it contained columns in the attribute table that are of a data type that is not supported by GRASS\` s DBMI - like binary data (BLOB). This PR changes the behavior of `v.in.ogr` to drop such columns from import instead of failing to import the data in total.

### DIFF
--- a/vector/v.in.ogr/main.c
+++ b/vector/v.in.ogr/main.c
@@ -380,12 +380,6 @@ int main(int argc, char *argv[])
     flag.notab = G_define_standard_flag(G_FLG_V_TABLE);
     flag.notab->guisection = _("Attributes");
 
-    flag.drop = G_define_flag();
-    flag.drop->guisection = _("Attributes");
-    flag.drop->key = 'd';
-    flag.drop->description =
-        _("Drop columns with unsupported data type from import");
-
     flag.over = G_define_flag();
     flag.over->key = 'o';
     flag.over->label =
--- a/vector/v.in.ogr/main.c
+++ b/vector/v.in.ogr/main.c
@@ -1053,12 +1053,19 @@ int main(int argc, char *argv[])
                 if (key_idx[layer] > -1 && key_idx[layer] == i)
                     continue; /* skip defined key (FID column) */
 
-		i_out++;
-                
 		Ogr_field = OGR_FD_GetFieldDefn(Ogr_featuredefn, i);
 		Ogr_ftype = OGR_Fld_GetType(Ogr_field);
 
 		G_debug(3, "Ogr_ftype: %i", Ogr_ftype);	/* look up below */
+
+                /* skip columns with unsupported data type */
+                if (Ogr_ftype == OFTBinary) {
+                    G_warning(_("Datatype of column <%s> is not supported. "
+                    "Omitting that column."), OGR_Fld_GetNameRef(Ogr_field));
+                    ncols_out = ncols_out - 1;
+                    continue; /* skip defined key (FID column) */
+				}
+		i_out++;
 
 		if (i < ncnames - 1) {
 		    Ogr_fieldname = G_store(param.cnames->answers[i + 1]);
--- a/vector/v.in.ogr/main.c
+++ b/vector/v.in.ogr/main.c
@@ -1176,17 +1176,13 @@ int main(int argc, char *argv[])
 		}
 		else {
                 /* handle columns of unsupported data type */
-                G_warning(_("Column type (Ogr_ftype: %d) not supported (Ogr_fieldname: %s)\n"),
-                          Ogr_ftype, Ogr_fieldname);
-                if (flag.drop->answer) {
-                    col_info[i_out] = col_info[i_out + 1];
-                    i_out--;
-                    ncols_out--;
-                }
-                else {
-                    buf[0] = 0;
-                    col_info[i_out].type = G_store(buf);
-                }
+                G_warning(_("Column <%s> is of unsupported data type (Ogr_ftype: %d)\n"
+                           "It is omitted from import\n"), Ogr_fieldname,
+                          Ogr_ftype);
+                buf[0] = 0;
+                col_info[i_out].type = G_store(buf);
+                i_out--;
+                ncols_out--;
             }
             G_free(Ogr_fieldname);
         }
@@ -1395,10 +1391,6 @@ int main(int argc, char *argv[])
 			    db_set_string(&strval, (char *)Ogr_fstring);
 			    db_double_quote_string(&strval);
 			    G_rasprintf(&sqlbuf, &sqlbufsize, ", '%s'", db_get_string(&strval));
-			}
-			else {
-			    /* column type not supported */
-			    G_rasprintf(&sqlbuf, &sqlbufsize, "%c", '\0');
 			}
 		    }
 		    else {
--- a/vector/v.in.ogr/main.c
+++ b/vector/v.in.ogr/main.c
@@ -137,7 +137,7 @@ int main(int argc, char *argv[])
         struct Option *snap, *type, *outloc, *cnames, *encoding, *key, *geom;
     } param;
     struct _flag {
-	struct Flag *list, *no_clean, *force2d, *notab,
+	struct Flag *list, *no_clean, *force2d, *notab, *drop,
 	    *region, *over, *extend, *formats, *tolower, *no_import,
             *proj;
     } flag;
@@ -379,6 +379,12 @@ int main(int argc, char *argv[])
 
     flag.notab = G_define_standard_flag(G_FLG_V_TABLE);
     flag.notab->guisection = _("Attributes");
+
+    flag.drop = G_define_flag();
+    flag.drop->guisection = _("Attributes");
+    flag.drop->key = 'd';
+    flag.drop->description =
+        _("Drop columns with unsupported data type from import");
 
     flag.over = G_define_flag();
     flag.over->key = 'o';
@@ -1053,19 +1059,12 @@ int main(int argc, char *argv[])
                 if (key_idx[layer] > -1 && key_idx[layer] == i)
                     continue; /* skip defined key (FID column) */
 
+        i_out++;
+
 		Ogr_field = OGR_FD_GetFieldDefn(Ogr_featuredefn, i);
 		Ogr_ftype = OGR_Fld_GetType(Ogr_field);
 
 		G_debug(3, "Ogr_ftype: %i", Ogr_ftype);	/* look up below */
-
-            /* skip columns with unsupported data type */
-            if (Ogr_ftype == OFTBinary) {
-                G_warning(_("Column <%s> is omitted, binary data type is not supported."),
-                          OGR_Fld_GetNameRef(Ogr_field));
-                ncols_out = ncols_out - 1;
-                continue;
-            }
-            i_out++;
 
 		if (i < ncnames - 1) {
 		    Ogr_fieldname = G_store(param.cnames->answers[i + 1]);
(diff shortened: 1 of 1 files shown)


vector/v.in.ogr/main.c
```diff
@@ -1053,13 +1053,20 @@ int main(int argc, char *argv[])
...
-		i_out++;
-                
 		Ogr_field = OGR_FD_GetFieldDefn(Ogr_featuredefn, i);
 		Ogr_ftype = OGR_Fld_GetType(Ogr_field);
 
 		G_debug(3, "Ogr_ftype: %i", Ogr_ftype);	/* look up below */
 
+                /* skip columns with unsupported data type */
+                if (Ogr_ftype == OFTBinary) {
+                    G_warning(_("Datatype of column <%s> is not supported. "
```
nilason, line 1063: Perhaps the message could be changed to something like:

`Column <%s> is omitted, binary data type is not supported.`

vector/v.in.ogr/main.c
```diff
@@ -1053,13 +1053,20 @@ int main(int argc, char *argv[])
...
+                    "Omitting that column."), OGR_Fld_GetNameRef(Ogr_field));
+                    ncols_out = ncols_out - 1;
+                    continue;
+				}
```
nilason, line 1067: Please correct the indent.

nilason (changes requested):
Looks good to me. Just had some minor remarks.

vector/v.in.ogr/main.c
ninsbl: OK, I changed the warning message and tired to follow:
https://trac.osgeo.org/grass/wiki/Submitting/C#Indentation
regarding indent.
However, runing
```
bash utils/grass_indent.sh vector/v.in.ogr/main.c
```
gave me:
```
indent: vector/v.in.ogr/main.c:1464: Error:Stmt nesting error.
indent: vector/v.in.ogr/main.c:1786: Error:Stmt nesting error.
indent: vector/v.in.ogr/main.c:2049: Error:Stmt nesting error.
```
So, I kept only the changed indent for the block I added. I hope that is according to the guidelines. If other indent changes are needed I would appreciate a pointer to what should be done specifically...

metzm:
Instead of adding a special case for `Ogr_ftype == OFTBinary`, handling of unsupported field types should be fixed in 
https://github.com/OSGeo/grass/blob/main/vector/v.in.ogr/main.c#L1177

vector/v.in.ogr/main.c
```diff
@@ -1053,13 +1053,20 @@ int main(int argc, char *argv[])
...
-		i_out++;
-                
 		Ogr_field = OGR_FD_GetFieldDefn(Ogr_featuredefn, i);
 		Ogr_ftype = OGR_Fld_GetType(Ogr_field);
 
 		G_debug(3, "Ogr_ftype: %i", Ogr_ftype);	/* look up below */
 
+            /* skip columns with unsupported data type */
+            if (Ogr_ftype == OFTBinary) {
+                G_warning(_("Column <%s> is omitted, binary data type is not supported."),
```
metzm, line 1063: Not supported data types are handled below at L1184

nilason: Good point! And there all columns with unsupported data types may be omitted. However, that would be a break with present behaviour (keep column, but omit rows). Perhaps a new flag for “omit columns with unsupported data types”?

vector/v.in.ogr/main.c
nilason: GNU indent can handle that file.
I was first thinking if the bad indent on line 1067, but keeping grass_indents changes for only added code is good in this case.

vector/v.in.ogr/main.c
ninsbl: OK, but isn't the binary column then included in the output table creation struct `grass_col_info` already in L1091: 
https://github.com/OSGeo/grass/blob/5d64de22416511b4a32e9b1546b5ae869cd0d48e/vector/v.in.ogr/main.c#L1091

So, I guess if we fix it in L1184, it would have to be removed from `col_info`?
How would I do that? A web-search unfortunately did not give me a straight forward solution...

vector/v.in.ogr/main.c
wenzeslaus: While most code for re-indented and fixed to indent without errors, the v.in.ogr nesting error still remains (https://github.com/OSGeo/grass/issues/1630#issuecomment-1295442400).

nilason: > GNU indent can handle that file.

I meant to say "GNU indent **can't** handle that file."

vector/v.in.ogr/main.c
```diff
@@ -1175,13 +1181,23 @@ int main(int argc, char *argv[])
...
-	    }
+                /* handle columns of unsupported data type */
+                G_warning(_("Column type (Ogr_ftype: %d) not supported (Ogr_fieldname: %s)\n"),
+                          Ogr_ftype, Ogr_fieldname);
+                if (flag.drop->answer) {
+                    col_info[i_out] = col_info[i_out + 1];
+                    i_out--;
+                    ncols_out--;
+                    G_warning(_("Column <%s> will be omitted from import"),
+                              Ogr_fieldname);
```
wenzeslaus, line 1192: I'm not sure about the need to keep current broken behavior, but if the flag is explicitly used, then why warning? At max a verbose message, I would say, is what best reflects the user input.

The other G_warning is not needed (at least not as a warning) either for the same reason in case the flag is active.

wenzeslaus (changes requested):

metzm:
There is a deeper bug in `v.in.ogr` regarding not supported column types: the behaviour of creating columns and filling columns does not match if there is a not supported column type. See following suggestions to this PR.

vector/v.in.ogr/main.c
ninsbl: Now (in [47b4efd](https://github.com/OSGeo/grass/pull/2630/commits/47b4efd8760ea141183bbac639b79f0ec3ea3de0)) I prepared a solution with a new flag (`-d`) to drop unsupported columns from import. Without the `-d` -flag, import of a map with a binary column breaks the import.

I have no strong opinions about keeping the old behavior, yet, a break in the import may be a good thing, as it makes users aware of the issue. Also, other unsupported types that can be represented as text (not sure if there are (m)any though), may import good enough with the current behavior that redefines them as text. So columns are not dropped unless requested.

I removed the second warning.

Thanks, @nilason, @metzm and @wenzeslaus for the suggestions. Would be great if you could have a second look at the updated code...

vector/v.in.ogr/main.c
```diff
@@ -380,6 +380,12 @@ int main(int argc, char *argv[])
     flag.notab = G_define_standard_flag(G_FLG_V_TABLE);
     flag.notab->guisection = _("Attributes");
 
+    flag.drop = G_define_flag();
+    flag.drop->guisection = _("Attributes");
+    flag.drop->key = 'd';
+    flag.drop->description =
+        _("Drop columns with unsupported data type from import");
+
```
metzm, line 388:
```suggestion

```
Unsupported data types can not be imported, there is no choice.

vector/v.in.ogr/main.c
```diff
@@ -1175,13 +1181,23 @@ int main(int argc, char *argv[])
 			      Ogr_fieldname, OFTIntegerListlength);
 		}
 		else {
-		    G_warning(_("Column type (Ogr_ftype: %d) not supported (Ogr_fieldname: %s)"),
-			      Ogr_ftype, Ogr_fieldname);
-		    buf[0] = 0;
-		    col_info[i_out].type = G_store(buf);
```
metzm, line 1181: Keep the current behaviour, add
```
i_out--;
```
remove
```
col_info[i_out].type = G_store(buf);
```

vector/v.in.ogr/main.c
```diff
@@ -1175,13 +1181,23 @@ int main(int argc, char *argv[])
...
-		}
-		G_free(Ogr_fieldname);
-	    }
+                /* handle columns of unsupported data type */
+                G_warning(_("Column type (Ogr_ftype: %d) not supported (Ogr_fieldname: %s)\n"),
+                          Ogr_ftype, Ogr_fieldname);
+                if (flag.drop->answer) {
```
metzm, line 1187: As above, the drop flag does not make sense: unsupported column types can not be transferred in any reasonable way.

metzm:
When filling attributes, the code block at  needs to be removed:
```
			else {
			    /* column type not supported */
			    G_rasprintf(&sqlbuf, &sqlbufsize, "%c", '\0');
			}
```
this is the hidden bug: should be `", %c"` to create an empty entry for a not supported column type, but dropping any not supported column type avoids confusion later on.

ninsbl:
Thanks again. Suggestions implemented in [5cf030a](https://github.com/OSGeo/grass/pull/2630/commits/5cf030ad054c00f6c30ae1e689bb788698d46f74). I hope I got those right...
At least my testcase as well as the unit-test work fine...

ninsbl:
P.S.: I modified the warning message a bit in an attempt to make it a bit more clear what is happening. Happy to change it back if preferred...

nilason:
> P.S.: I modified the warning message a bit in an attempt to make it a bit more clear what is happening. Happy to change it back if preferred...

With latest changes, may I suggest the following warning message:

```c
                G_warning(_("Column <%s> is of unsupported \"%s\" data type, "
                            "it is omitted from import."),
		          Ogr_fieldname, OGR_GetFieldTypeName(Ogr_ftype));
```

which will result in a message like:

```
WARNING: Column <blob> is of unsupported "Binary" data type, it is omitted
         from import.
```

General note: G_warning will automatically reflow the printed message to fit in a 80 col terminal, no need to break `\n` (not at the end either).

vector/v.in.ogr/main.c
```diff
@@ -1053,8 +1053,8 @@ int main(int argc, char *argv[])
                 if (key_idx[layer] > -1 && key_idx[layer] == i)
                     continue; /* skip defined key (FID column) */
 
-		i_out++;
-                
+        i_out++;
```
nilason, line 1056: Line 1056 has bad indention, you may revert this change.

vector/v.in.ogr/main.c
```diff
@@ -1175,13 +1175,17 @@ int main(int argc, char *argv[])
...
-		    col_info[i_out].type = G_store(buf);
-		}
-		G_free(Ogr_fieldname);
-	    }
+                /* handle columns of unsupported data type */
+                G_warning(_("Column <%s> is of unsupported data type (Ogr_ftype: %d)\n"
+                           "It is omitted from import\n"), Ogr_fieldname,
+                          Ogr_ftype);
+                buf[0] = 0;
+                col_info[i_out].type = G_store(buf);
```
nilason, line 1183: As @metzm mentioned, Line 1183 (`col_info[i_out].type = G_store(buf);`)  may be removed.

ninsbl: L1182 should then be removed too, no?

nilason: > L1182 should then be removed too, no?

That's right. Not needed.

ninsbl: Thanks! PR updated in [cd67af7](https://github.com/OSGeo/grass/pull/2630/commits/cd67af7292f9f6edb05238616529042d53259a62) accordingly.

wenzeslaus:
@ninsbl Can you please update the PR description to reflect the current intention and changes done? (will be needed for the merge/final commit anyway)

ninsbl:
> @ninsbl Can you please update the PR description to reflect the current intention and changes done? (will be needed for the merge/final commit anyway)

Done.

nilason (approved):
Great!

vector/v.in.ogr/main.c
```diff
@@ -1175,13 +1175,15 @@ int main(int argc, char *argv[])
...
-		    G_warning(_("Column type (Ogr_ftype: %d) not supported (Ogr_fieldname: %s)"),
-			      Ogr_ftype, Ogr_fieldname);
-		    buf[0] = 0;
-		    col_info[i_out].type = G_store(buf);
-		}
-		G_free(Ogr_fieldname);
-	    }
+                /* handle columns of unsupported data type */
+                G_warning(_("Column <%s> is of unsupported \"%s\" data type, "
+                            "it is omitted from import."),
```
wenzeslaus, line 1180: The message can be simpler something like:

```
"Column <%s> of unsupported data type \"%s\" is omitted from import"
```

ninsbl: Warning simplified as suggested in [ba86d84](https://github.com/OSGeo/grass/pull/2630/commits/ba86d84c01556fa6f0107e6bf9bf0b266ea0efd7)

wenzeslaus (approved):
That looks good! I'm wondering if the line is not too long, but formatting in this file needs fixing anyway, so maybe not a big deal as long as we are planning to fix it soon.

Issue #2610, opened 2022-10-25
[Bug] Version check in G__gisinit & G__no_gisinit is too strict for non-git builds


sebastic:
**Describe the bug**

Every time GRASS is rebuilt, even when there are no changes, the timestamp recorded in `GRASS_HEADERS_VERSION` changes which breaks the gdal-grass plugins:
```
$ ogrinfo -so -al ~/grassdata/nc_basic_spm_grass7/PERMANENT/vector/zipcodes/head
Warning 1: GRASS warning: GISBASE environment variable was not set, using:
/usr/lib/grass82
ERROR: Module built against version 2022-06-11T10:48:36+00:00 but trying to
       use version 2022-10-25T05:34:10+00:00. You need to rebuild GRASS GIS
       or untangle multiple installations.
```
The timestamp is not used for git builds, the commit hash for the last commit affecting the `include` directory is used.

The error should not be triggered when there were no changes to the GRASS source tree that actually require rebuilding the gdal-grass plugins.

The GDAL version check only requires that the gdal-grass plugins are rebuilt when the major and or minor version changes which suggest ABI breakages.

Non-git builds of GRASS should use something like `GRASS_VERSION_NUMBER` for `GRASS_HEADERS_VERSION` which doesn't change when GRASS is rebuilt.

**To Reproduce**

1. Rebuild GRASS from source tarball
2. Execute `ogrinfo` on a GRASS dataset

**Expected behavior**

No error is trigggered.

**System description**

- Operating System: Debian unstable
- GRASS GIS version: 8.2.0

**Additional context**

Maintaining gdal-grass packages is very painful, because they need to be rebuilt every time GRASS is.

sebastic:
Something like this may suffice:
```diff
--- a/configure.ac
+++ b/configure.ac
@@ -139,7 +139,7 @@ changequote([,])
 GRASS_VERSION_GIT="exported"
 # get git short hash + date of last change in GRASS headers
 # (and anything else in include)
-GRASS_HEADERS_GIT_HASH=`date -u +%FT%T%z | sed 's/\(..\)$/:\1/'`
+GRASS_HEADERS_GIT_HASH="${GRASS_VERSION_NUMBER}"
 GRASS_HEADERS_GIT_DATE=`date -u +%FT%T%z | sed 's/\(..\)$/:\1/'`
 AC_PATH_PROG(GIT, git, no)
 if test "$GIT" != "no" ; then
```
Which results in a `version.h` like this:
```
[...]

PR #2636, opened 2022-11-08
Use GRASS version for non-git builds.

sebastic:
As reported in #2610, rebuilding the GRASS sources (from the release tarballs) with no changes break the gdal-grass plugins because the `GIS_H_VERSION` value changes.

Non-git builds use the current timestamp in `GRASS_HEADERS_GIT_HASH` which is also used for `GRASS_HEADERS_DATE`.

This PR uses `GRASS_VERSION_NUMBER` for `GRASS_HEADERS_GIT_HASH` as a better value for non-git builds, that value won't change for simple rebuilds of the source tree.

### DIFF
--- a/configure
+++ b/configure
@@ -3845,7 +3845,7 @@ if test "$GIT" != "no" ; then
    fi
    GRASS_HEADERS_GIT_HASH=`$GIT log -1 --pretty=format:"%h" -- "${SRCDIR}/include" 2>/dev/null`
    if test -z "$GRASS_HEADERS_GIT_HASH"; then
-      GRASS_HEADERS_GIT_HASH=`date -u +%FT%T%z | sed 's/\(..\)$/:\1/'`
+      GRASS_HEADERS_GIT_HASH="${GRASS_VERSION_NUMBER}"
    fi
    GRASS_HEADERS_GIT_DATE=`date -d $($GIT log -1 --pretty=format:"%cI" -- "${SRCDIR}/include") -u +%FT%T%z | sed 's/\(..\)$/:\1/'` 2>/dev/null
    if test -z "$GRASS_HEADERS_GIT_DATE"; then
--- a/configure
+++ b/configure
@@ -3798,7 +3798,7 @@ LIB_VER=`echo ${GRASS_VERSION_NUMBER} | sed 's/^\([0-9.]*\).*$/\1/'`
 GRASS_VERSION_GIT="exported"
 # get git short hash + date of last change in GRASS headers
 # (and anything else in include)
-GRASS_HEADERS_GIT_HASH=`date -u +%FT%T%z | sed 's/\(..\)$/:\1/'`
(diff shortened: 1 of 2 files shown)


wenzeslaus (approved):
Assuming a release when not using Git sounds good to me.

sebastic:
> Assuming a release when not using Git sounds good to me.

When `git` is available the commit hash and date are used:
```bash
# get git short hash + date of last change in GRASS headers
# (and anything else in include)
GRASS_HEADERS_GIT_HASH="${GRASS_VERSION_NUMBER}"
GRASS_HEADERS_GIT_DATE=`date -u +%FT%T%z | sed 's/\(..\)$/:\1/'`
AC_PATH_PROG(GIT, git, no)
if test "$GIT" != "no" ; then
   GRASS_VERSION_GIT=`$GIT rev-parse --short HEAD 2>/dev/null`
   if test -z "$GRASS_VERSION_GIT"; then
      GRASS_VERSION_GIT="exported"
   fi
   GRASS_HEADERS_GIT_HASH=`$GIT log -1 --pretty=format:"%h" -- "${SRCDIR}/include" 2>/dev/null`
   if test -z "$GRASS_HEADERS_GIT_HASH"; then
      GRASS_HEADERS_GIT_HASH=`date -u +%FT%T%z | sed 's/\(..\)$/:\1/'`
   fi
   GRASS_HEADERS_GIT_DATE=`date -d $($GIT log -1 --pretty=format:"%cI" -- "${SRCDIR}/include") -u +%FT%T%z | sed 's/\(..\)$/:\1/'` 2>/dev/null
   if test -z "$GRASS_HEADERS_GIT_DATE"; then
      GRASS_HEADERS_GIT_DATE=`date -u +%FT%T%z | sed 's/\(..\)$/:\1/'`
   fi
fi
```

nilason:
You need to run `autoconf269` to update the `configure` file.

sebastic:
Generated files should not be committed.

nilason:
> Generated files should not be committed.

The generated file has been there in 20 years, since version 5. As long as it is there, the two should be in sync, no?
That being said, there is no reason -- as I see it -- this (i.e. generated file being part of source) have to  remain the case indefinitely...

nilason:
If `git` as a program is available, but building is made from a non-git repo (e.g. release tarball), the `GRASS_HEADERS_GIT_HASH` will still be reset with date on L152:

https://github.com/OSGeo/grass/blob/5184a91f1fb6c4e0791f5c551a5cc94010084f12/configure.ac#L144-L158

Now, the question is whether L152 should change to:
```
      GRASS_HEADERS_GIT_HASH="${GRASS_VERSION_NUMBER}"
```
too, as a fallback if test on L150 fails?

sebastic:
Since there is no `--without git` configure option, that might be good.

nilason (approved):
Looks good to me.
@wenzeslaus Please go ahead and merge if you still approve after latest changes.
@sebastic Thanks !

PR #2642, opened 2022-11-10
cdhc lib: add statlib cdh fortran code to lib docs

nilason:
This PR aim to add the Fortran source code, which was a base for cdhc library, for documentation purposes in the form of a Markdown file.

This is the source the library's author Darrell McCauley mention in "Goodness of Fit Tests" ([lib/cdhc/docs/goodness.pdf](https://github.com/OSGeo/grass/blob/main/lib/cdhc/doc/goodness.pdf)):

> Each function in the library returns a pointer to static double. The cdhc library was inspired by Johnson’s STATLIB collection of FORTRAN routines for testing distribution assumptions [22]. Some functions in cdhc are loosely based on Johnson’s work (they have been completely rewritten, reducing memory requirements and number of computations and fixing a few bugs). Others are based on algorithms found in Applied Statistics, Technometrics, and other related journals.

Limited amount of documentation remains, links mentioned in https://lists.osgeo.org/pipermail/grass-user/1995-July/023710.html are no longer available, thus this addition might be of help in the future.


echoix (approved):
Looks ok to me, it’s a doc-only change, and is justified to be here for posterity.

neteler:
> it’s a doc-only change

In this case, with only a markdown file change, I would expect all CI build jobs being skipped. But indeed they all run (effectively not generating any insights):

image

I ask because of the meanwhile closed issue #2174.

nilason:
> > it’s a doc-only change
> 
> In this case, with only a markdown file change, I would expect all CI build jobs being skipped. But indeed they all run (effectively not generating any insights):
> 

> I ask because of the meanwhile closed issue #2174.

Technically adding `[skip ci]` to commit description work by default on GH to avoid CI checks. The rebase commit message ("Merge branch 'main' into cdhc_add_fortran") didn't contain that tag. For that to work now it had to be eg. "Merge branch 'main' into cdhc_add_fortran [skip ci]".

echoix:
We have some checks that apply to markdown, docs and more, so it still makes sense to have some run.

PR #2650, opened 2022-11-14
Fix -Wunused-but-set-variable compiler warnings, part 2

nilason:
Continuation of #2648.

As reported with #2156.

Following lib parts are affected:
- arraystats
- cairodriver
- cdhc
- db/dbmi_base
- dspf
- external/shapelib
- gis
- gpde
- imagery
- lidar
- ogsf
- proj
- raster
- rst/interp_float
- vector/Vlib
- vector/dglib
- vector/neta


nilason:
@marisn Thank you for looking at these!!

PR #2655, opened 2022-11-16
toolboxes.xml: add r.path

pesekon2:
So far, it was impossible to find module `r.path` through GUI. This PR fixes it.

### DIFF
--- a/gui/wxpython/xml/toolboxes.xml
+++ b/gui/wxpython/xml/toolboxes.xml
@@ -807,6 +807,9 @@
       <module-item name="r.cost">
         <label>Cost surface</label>
       </module-item>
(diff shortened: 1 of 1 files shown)


landam (approved):
Section "Terrain analysis" seems to be a right place for 'r.path'.
Screenshot from 2022-11-20 22-35-15

landam:
@neteler To my understanding, this fix should be backported also to releasebranch_8_2 and 8_0. Changing milestone to 8.3.0 means no backport, right?

neteler:
I always set the version to where it currently is. Once backported, I update the version label. Overall, I strongly suggest that we revisit the backport labels and introduce more clear ones.

PR #2658, opened 2022-11-17
v.in.ogr: handle skipped, not empty columns

ninsbl:
There was a glitch in #2630 that values from skipped but not empty columns of unsupported data types were still tried to insert into the table in the GRASS DB.
This PR skips also to INSERT values from those columns, using the same logic as for the columns themself.


nilason (approved):
Looks good to me, testing worked fine.

Just a small typo -> "skipped".

ninsbl:
> Looks good to me, testing worked fine.
> 
> Just a small typo -> "skipped".

Thanks for both checking and pointing out. Fixed now.

PR #2660, opened 2022-11-21
r.terraflow: use default

nilason:
Force compiler create constructors and assignment operators, fixes following warnings:

- `-Wclass-memaccess`
- `-Wdeprecated-declaration`


nilason:
If there are no objections, I’ll merge this shortly.

PR #2661, opened 2022-11-23
CI: add macOS build

nilason:
Add a macOS build with tests on the CI.


neteler:
"macOS / macOS build (pull_request) Successful in 17m"

Superb :100:  (also compared to the Ubuntu CI "speed" here)

wenzeslaus:
> ...compared to the Ubuntu CI "speed" here

Notably, this is currently without tests unlike the Ubuntu CI. For speed, compare with _GCC C/C++ standards check_ instead.

.github/workflows/macos_dependencies.txt
```diff
@@ -0,0 +1,167 @@
+# This file may be used to create an environment using:
+# $ conda create --name <env> --file <this file>
+# platform: osx-64
+@EXPLICIT
+https://conda.anaconda.org/conda-forge/osx-64/pandoc-2.19.2-h694c41f_1.tar.bz2
```
wenzeslaus, line 5: I'm not familiar with this type of file, so I'm wondering why this file and not an environment.yml file?

nilason:
> > ...compared to the Ubuntu CI "speed" here
> 
> Notably, this is currently without tests unlike the Ubuntu CI. For speed, compare with _GCC C/C++ standards check_ instead.

Compile time is decent, nothing spectacular. Testing is **not** particularly fast.
It would have been good to have an Apple ARM based platform. GitHub doesn't have any unfortunately. Maybe https://cirrus-ci.org could be an alternative?

nilason:
I have now opened this for review.

Following tests fails:

**r.in.pdal**
- test_r_in_pdal_binning
- test_r_in_pdal_selection

**r.sun**
- test_rsun

Testreport available as artefact.

nilason:
Now added using cache for deps, saving ~3 min.

wenzeslaus:
As an alternative to test success percentage, you could also disable specific tests, but at this point you would have to duplicate the whole test configuration in the `.gunittest.cfg` file because the system can't combine two configurations.

nilason:
Oops, this installs grass into environment before caching, thus includes initial build of grass into cache. Not so good. Either remove or different install dir...

wenzeslaus:
I don't quickly see why is that, but the 3 mins may not be worth the trouble given the total time (the time fluctuation of GCC seems about 3 mins or so).

nilason:
> I don't quickly see why is that, but the 3 mins may not be worth the trouble given the total time (the time fluctuation of GCC seems about 3 mins or so).

See log extract:
```
ERROR: Module built against version 87ba3f2 but trying to use version
[1806](https://github.com/OSGeo/grass/actions/runs/3567806784/jobs/5995944634#step:8:1807)
       771e1f0. You need to rebuild GRASS GIS or untangle multiple
[1807](https://github.com/OSGeo/grass/actions/runs/3567806784/jobs/5995944634#step:8:1808)
rm test.raster3d.lib.tmp.html
[1808](https://github.com/OSGeo/grass/actions/runs/3567806784/jobs/5995944634#step:8:1809)
       installations.
[1809](https://github.com/OSGeo/grass/actions/runs/3567806784/jobs/5995944634#step:8:1810)
/usr/bin/install -c  -m 644 N_gwflow.h /Users/runner/work/grass/grass/dist.x86_64-apple-darwin13.4.0/include/grass/N_gwflow.h
[1810](https://github.com/OSGeo/grass/actions/runs/3567806784/jobs/5995944634#step:8:1811)
make[3]: *** [test.raster3d.lib.tmp.html] Error 1
[1811](https://github.com/OSGeo/grass/actions/runs/3567806784/jobs/5995944634#step:8:1812)
```

Now I added separate install dir. Let's see how it works out. (No trouble on my side :)

nilason:
> > How would I duplicate the `.gunittest.cfg` file (where to and to what name)? And how to reference to it for the tests on Mac?
> 
> The test executor has a `--config` parameter:
> 
> ```
> > python -m grass.gunittest.main --help
> ...
>   --config CONFIG       Path to a configuration file (default: .gunittest.cfg)
> ```
> 
> The file could be in `.github/workflows/` like the CI scripts or in `macosx/` if that seems to be general limitation on macOS right now, named, e.g., `gunittest_macos.cfg`.

Aha, thanks!

nilason:
The `v.db.select` test exits with:

```
ERROR: testGroup (__main__.SelectTest.testGroup)
Testing v.db.select with group option
  File "vector/v.db.select/testsuite/test_v_db_select.py", line 210, in testGroup
    sel.run()
  File "etc/python/grass/pygrass/modules/interface/module.py", line 823, in run
    self.wait()
  File "etc/python/grass/pygrass/modules/interface/module.py", line 844, in wait
    raise CalledModuleError(
grass.exceptions.CalledModuleError: Module run `v.db.select map=geology layer=1 columns=GEO_NAME group=GEO_NAME format=plain -c` ended with an error.
The subprocess ended with a non-zero return code: -11. See errors above the traceback or in the error output.

```

"code: -11" is likely a segmentation fault.

nilason:
This now runs with `-Wall -Werror` on Clang (now version 14.0.6), e.g. it fails if new compiler warnings are detected.
If there are no objections, I'll merge this shortly.

PR #2663, opened 2022-11-24
lib/imagery: fix -Wfortify-source warning

nilason:
This addresses following warning message:
> warning: 'fscanf' may overflow; destination buffer in argument 3 has size 1024, but the corresponding specifier may require size 1025

Found working on #2661.


marisn (approved):
Did I got it right – it is because title could be 1023 chars + \0 ?

nilason:
> Did I got it right – it is because title could be 1023 chars + \0 ?

Yes.

PR #2666, opened 2022-11-24
Add json output option to r.kappa

marisn:
Current output of r.kappa is not machine readable. This PR adds a new flag to print out all data in a json format for greater reusability.


raster/r.kappa/calc_metrics.c
```diff
@@ -0,0 +1,203 @@
...
+    for (i = 0; i < ncat2; i++)
+        rlst[ncat1 + i] = clst[i];
+    qsort(rlst, ncat1 + ncat2, sizeof(long), longcomp);
+    ncat = collapse(rlst, ncat1 + ncat2);
+    rlst = (long *)G_realloc(rlst, ncat * sizeof(long));
+    G_free(clst);
+
+    /* allocate matrix and fill in with cats' value */
+    metrics = (METRICS *) G_malloc(sizeof(METRICS));
+    metrics->matrix = (long *)G_malloc(ncat * ncat * sizeof(long));
```
nilason, line 48:
```suggestion
    metrics->matrix = (long *)G_malloc((size_t)ncat * ncat * sizeof(long));
```
should be enough.

marisn: Thank you for your hint. I am not good at C :-(

raster/r.kappa/calc_metrics.c
```diff
@@ -0,0 +1,213 @@
...
+void calc_metrics(void)
+{
+    int i, j, k;
+    long *clst;
+    int ncat1, ncat2;
+    int cndx;
+    double *pi, *pj, *pii;
+    double p0 = 0, pC = 0;
+    double inter1 = 0, inter2 = 0;
+    int a_i, b_i;
```
nilason, line 20: Use `0.0` for double and initialise a_i and b_i (`int a_i = 0, b_i = 0;`).

raster/r.kappa/prt_json.c
```diff
@@ -0,0 +1,94 @@
...
+
+    if (output != NULL)
+        fd = fopen(output, "w");
+    else
+        fd = stdout;
+
+    if (fd == NULL) {
+        G_fatal_error(_("Cannot open file <%s> to write JSON output"),
+                      output);
+        return;
```
nilason, line 21: No need for `return;` (and thus no need for curly braces).

raster/r.kappa/prt_kappa.c
```diff
@@ -0,0 +1,57 @@
...
+{
+    int i;
+    FILE *fd;
+
+    if (output == NULL)
+        fd = stdout;
+    else if ((fd = fopen(output, "a")) == NULL) {
+        G_fatal_error(_("Cannot open file <%s> to write kappa and relevant parameters"),
+                      output);
+        return;
```
nilason, line 18: Again, no return.

raster/r.kappa/prt_kappa.c
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /* print out the comission and omission accuracy, and conditional kappa */
+    fprintf(fd, "\nCats\t%% Comission\t%% Omission\tEstimated Kappa\n");
+    for (i = 0; i < ncat; i++) {
+        fprintf(fd, "%ld\t", rlst[i]);
+        if (metrics->user_acc[i] == -999)
```
nilason, line 25: Perhaps make a constant for `-999` in `kappa.h`?

raster/r.kappa/prt_mat.c
```diff
@@ -41,51 +32,11 @@ void prn_error_mat(int out_cols, int hdr)
         return;
```
nilason, line 32: No need for return.

nilason (changes requested):
Looks fine to me, tests succeeds and I found no regressions (tested with example in manual).

I have some minor suggestions I noted here and there, I attach patch here: 
[PR2666_fixes.diff.txt](https://github.com/OSGeo/grass/files/10135110/PR2666_fixes.diff.txt).

With that patch the source code for r.kappa passes `-Wall -Wextra` on clang without warnings, and `cppcheck` and `scan-build` are also happy.

I do miss your name in the head of `main.c`, and while you're there also update (c) date :-).

raster/r.kappa/prt_kappa.c
nilason: I was thinking something like `static const double na_value = -999.0;` not a `#define`.

raster/r.kappa/calc_metrics.c
```diff
@@ -0,0 +1,213 @@
...
+            t_col += metrics->matrix[x];
+            x += ncat;
+            t_row += metrics->matrix[cndx * ncat + k];
+        }
+        metrics->obs += t_row;
+        metrics->colsum[cndx] = t_col;
+        metrics->rowsum[cndx] = t_row;
+    }
+    if (metrics->obs == 0) {
+        metrics->total_acc = 0;
```
nilason, line 92:
```suggestion
        metrics->total_acc = 0.0;
```

raster/r.kappa/testsuite/test_r_kappa.py
```diff
@@ -292,5 +297,194 @@ def test_standard_output(self):
...
+        )
+
+    @classmethod
+    def tearDownClass(cls):
+        """Remove temporary data"""
+        cls.del_temp_region()
+        for ref in cls.refs:
+            cls.runModule("g.remove", flags="f", type="raster", name=ref)
+        for clas in cls.clas:
+            cls.runModule("g.remove", flags="f", type="raster", name=clas)
```
wenzeslaus, line 459: With whole words, this can nicely work with singulars and plurals as `for reference in references` and `for classification in classifications`.

raster/r.kappa/calc_metrics.c
```diff
@@ -0,0 +1,213 @@
...
+    int a_i, b_i;
+
+    metrics = (METRICS *) G_malloc(sizeof(METRICS));
+    if (nstats == 0) {
+        G_warning(_("Both maps have nothing in common. Check the computational region."));
+        metrics->obs = 0;
+        metrics->correct = 0;
+        metrics->total_acc = 0;
+        metrics->kappa = -999;
+        metrics->kappa_var = -999;
```
wenzeslaus, line 29: While NaN is not part of JSON, it is a common extension because it is part of JavaScript. Python [supports that by default](https://docs.python.org/3/library/json.html#json.dump) and db.univar and v.db.univar (#2386) behave that way.

raster/r.kappa/main.c
```diff
@@ -113,9 +115,18 @@ int main(int argc, char **argv)
     flags.m->description = _("Print Matrix only");
     flags.m->guisection = _("Output settings");
 
+    flags.j = G_define_flag();
+    flags.j->key = 'j';
+    flags.j->description = _("Print in JSON");
+    flags.j->guisection = _("Output settings");
```
wenzeslaus, line 121: For all the new JSON outputs we are adding an option right now. While flag is a nice quick hack, it is not scalable. v.db.select had two existing outputs and adding both JSON and CSV made sense there. That would be 3 flags. Now it has format which captures all the output styles. The format option is now also in t.rast.list, db.univar, and v.db.univar. r.geomorphon has profileformat with the same logic. It also fits with tools like r.out.gdal or v.in.ascii.

raster/r.kappa/prt_kappa.c
```diff
@@ -0,0 +1,57 @@
+#include <stdlib.h>
+#include <grass/gis.h>
+#include <grass/glocale.h>
+#include "kappa.h"
+#include "local_proto.h"
+
+
+void prt_kappa(void)
```
wenzeslaus, line 8: While _strč prst skrz krk_ is much more popular Czech and Slovak tongue-twister, this reminded me of _krt prd skrz drn, zprv zhlt hrst zrn_. In other words, I would prefer `print_kappa` here.

wenzeslaus (changes requested):
I'm excited to see another JSON output. I think that especially the JSON output would be better with full words rather than various abbreviations. -999 does not seem needed when we have NaN in the "common" (not standard) JSON. (C99 has [INFINITY](https://en.cppreference.com/w/c/numeric/math/INFINITY), but [NAN](https://en.cppreference.com/w/c/numeric/math/NAN) may not be available in general, but it is a [GNU extension](https://www.gnu.org/software/libc/manual/html_node/Infinity-and-NaN.html).) Can you please sync the names with docs and interface too?

raster/r.kappa/prt_json.c
marisn: You are totally right about return being unreachable. Although I am not a huge fan of no curly braces for a single statement as it is a potential trap in future.

raster/r.kappa/prt_kappa.c
marisn: Good idea.

raster/r.kappa/calc_metrics.c
```diff
@@ -0,0 +1,213 @@
...
+        return;
+    }
+
+    /* get the cat lists */
+    rlst = (long *)G_calloc(nstats * 2, sizeof(long));
+    clst = (long *)G_calloc(nstats, sizeof(long));
+    for (i = 0; i < nstats; i++) {
```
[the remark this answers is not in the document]
marisn: +1 for reducing compiler warnings. Done.

raster/r.kappa/testsuite/test_r_kappa.py
```diff
@@ -230,8 +235,8 @@ def setUpClass(cls):
     def tearDownClass(cls):
         """Remove temporary data"""
         cls.del_temp_region()
-        # cls.runModule("g.remove", flags="f", type="raster", name=cls.ref_1)
-        # cls.runModule("g.remove", flags="f", type="raster", name=cls.class_1)
+        cls.runModule("g.remove", flags="f", type="raster", name=cls.ref_1)
+        cls.runModule("g.remove", flags="f", type="raster", name=cls.class_1)
```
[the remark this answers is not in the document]
marisn: A? M? Huh?

raster/r.kappa/prt_json.c
nilason: > Although I am not a huge fan of no curly braces for a single statement as it is a potential trap in future.

I agree with you on this, 100 per cent. I is currently GRASS "standard" though.

raster/r.kappa/calc_metrics.c
marisn: This is a tricky question. As it does not confirm with JSON spec, not all parsers are happy with NaN, NA, etc. -999 looks ugly but is compatible with current generation of R JSON parsers.
```
>>> json.dumps([1.0, 2.0, float('nan'), 3.0])
'[1.0, 2.0, NaN, 3.0]'
```
```
> fromJSON('[1.0, 2.0, NaN, 3.0]')
Error in fromJSON("[1.0, 2.0, NaN, 3.0]") : 
  not all data was parsed (9 chars were parsed out of a total of 20 chars)
```

raster/r.kappa/calc_metrics.c
```diff
@@ -0,0 +1,214 @@
...
+                    b_i = i;
+            }
+            inter2 +=
+                Gstats[l].count * pow((pi[a_i] + pj[b_i]), 2.) / metrics->obs;
+        }
+    }
+    metrics->kappa_var = (inter1 + pow((1 - p0), 2.) * inter2 -
+                          pow((p0 * pC - 2 * pC + p0), 2.)) / pow((1 - pC),
+                                                                  4.) /
+        metrics->obs;
```
nilason, line 176: GNU indent makes a lousy job here. This is what clang-format would produce:
```c
    metrics->kappa_var = (inter1 + pow((1 - p0), 2.) * inter2 -
                          pow((p0 * pC - 2 * pC + p0), 2.)) /
                         pow((1 - pC), 4.) / metrics->obs;
```

(just a little push for #2272).

raster/r.kappa/main.c
marisn: +1 for unification. Done.

raster/r.kappa/testsuite/test_r_kappa.py
marisn: Come on, this code saves a few bytes of disk space! Besides, don't you love keeping as close to Fortran as possible? ;-)

raster/r.kappa/calc_metrics.c
marisn: Yeah, clang-format version looks better. Pity that #2272 contains both – .clang-format and code formatting as merging just configuration file would be easier.

nilason: As I mentioned in the PR description:

> Note: only the addition of the .clang-format is the ultimate goal for this PR. The added commit with all the formatted source files are for illustration only.

The issue has been stalled somehow, see also my question https://github.com/OSGeo/grass/pull/2272#issuecomment-1075676212 . Should probably bring the question to dev-ML.

raster/r.kappa/prt_kappa.c
marisn: All functions and files are renamed.

raster/r.kappa/calc_metrics.c
marisn: We could go with `null` although it is also not without downsides:
```
> fromJSON('[1.0, 2.0, -999.0, 3.0]')
[1]    1    2 -999    3
> fromJSON('[1.0, 2.0, null, 3.0]')
[[1]]
[1] 1

[[2]]
[1] 2

[[3]]
NULL

[[4]]
[1] 3
```

raster/r.kappa/calc_metrics.c
marisn: Sorry for missing the note.
Yes, you should write to dev ML with clean statement – if no objections are made till XX date, we'll go with Stroustrup.

nilason (approved):
Great work, looks good to me!

Re.: -999, it is probably the safest solution.

wenzeslaus:
> Re.: -999, it is probably the safest solution.

Output - JSON: Generally, if people write in Python, chances are that JSON outputs will contain NaN without even making a decision to actually do that. A lot of reading will likely happen in Python where NaN will be accepted (read as `float("nan")`). Strangely, JavaScript is the one (?) place which insists on standardized JSON and one needs to use 3rd party lib to read JSON with NaNs.

Output - consistency: Using -999 for NaN would mean that also other places which output that should use the same number, but that's of course neither ensured nor feasible. ...the old NULL/no-data debate.

Code: #2681 now introduces NAN to the C code base, but I'm more concerned with outputs.

Possibly, infinity may work everywhere?

Issue #1748, opened 2021-07-19
[Feat] Single-Window GUI: Implement a button for undocking a AuiNotebook tab to wx.Frame and conversely


lindakladivova:
**Is your feature request related to a problem? Please describe.**
We could allow anybody to make the Multi-Window GUI. Each AuiNotebook Map Display tab could include a special button at the top part. This button could enable to "Undock" a tab to an independent frame. On the other side, the created frame could have a special button "Dock" which inserts a frame back to an AuiNotebook.

Thanks to @petrasovaa for this great idea and some suggestions for implementation :-)

PR #2667, opened 2022-11-27
wxGUI: adding a button for undocking an AuiNotebook tab to wx.Frame (Single-Window GUI)

lindakladivova:
The implementation of the ability to undock and dock a map window is needed for those of you who would like to still use GRASS as a Multi-Window in the future. It is also very convenient for example when using g.gui.tangible module.

The map panel can be undocked and then docked back into AuiNotebook using the "un/dock" button:

Screenshot from 2022-11-27 13-39-53

After undocking, the independent map display frame is created. This frame can be moved around as a regular frame (you can have it on the second monitor for example).
Screenshot from 2022-11-27 13-49-23

This PR is still under process.

Things that need to be issued:

- [x] undocked and docked functionality implemented via a new "un/dock" button
- [x] resizing undocked frame to the size of the original MapPanel
- [x] when docking, focus on the newly added AuiNotebook page
- [x] closing the undocked wx.Frame needs to close the associated Layers tab
- [x] closing the Layers tab must close the associated wx.Frame
- [x] when closing GRASS, undocked frames and Aui.Notebook pages need to be also closed.
- [x] fix adding a new Map Notebook page (the AuiNotebook page indexing is id being changed by undocking)

- [x] fix switching to 3D/2D in undocked frame (meantime found the bug https://github.com/OSGeo/grass/issues/2672)
- [x] when undocked, focus on the associated Layers tab
- [x] check whether Multi-Window still works and adapt
- [x] create nicer icon for Map Display undock/dock
- [x] fix close event on undocked wx.Frame
- [x] fix renaming a page in Layers tab

- [x] add support for workspaces (isDocked parameter will be stored in a workspace file and after loading a workspace file, GRASS decide which map displays should be undocked right away)
- [x] fix error related to closing undocked map frame _wrapped C/C++ object of type MapPageFrame has been deleted_ (appeared when before loading a workspace in the moment when all map displays are being closed)
 
Successful merge of this PR closes the Issue https://github.com/OSGeo/grass/issues/1748.


petrasovaa:
Perhaps an icon may look better? See for example https://github.com/qgis/QGIS/blob/master/images/themes/default/mDockify.svg
maybe we could tweak it.

lindakladivova:
> Perhaps an icon may look better? See for example https://github.com/qgis/QGIS/blob/master/images/themes/default/mDockify.svg maybe we could tweak it.

@petrasovaa  What you do think about this idea?
The docking functionality is similarly as New Display or Display Settings related to the whole window. I tried to tweak a bit the Map Display Settings icon.
First suggestion:
Screenshot from 2022-12-03 14-10-49

Second suggestion:
Screenshot from 2022-12-03 14-26-16

gui/wxpython/mapdisp/frame.py
```diff
@@ -659,6 +684,18 @@ def RemoveQueryLayer(self):
...
+    def SetDockingCallback(self, function):
+        """Sets docking bound method to dock or undock"""
+        self.dck_callback = function
+
+    def OnDockUndock(self, event):
+        """Dock or undock map display panel to independent MapFrame"""
+        self.dck_callback(self)
+        self.docked = not self.docked
+
+    def GetDockingState(self):
```
petrasovaa, line 696: This should be more like "IsDocked"

gui/wxpython/mapdisp/frame.py
```diff
@@ -659,6 +684,18 @@ def RemoveQueryLayer(self):
         for layer in qlayer:
             self.GetMap().DeleteLayer(layer)
 
+    def SetDockingCallback(self, function):
+        """Sets docking bound method to dock or undock"""
+        self.dck_callback = function
+
+    def OnDockUndock(self, event):
+        """Dock or undock map display panel to independent MapFrame"""
+        self.dck_callback(self)
```
petrasovaa, line 693: It's better to use full words.

gui/wxpython/mapdisp/frame.py
petrasovaa: And then separate function IsDockable.

gui/wxpython/mapdisp/frame.py
```diff
@@ -78,6 +100,7 @@ def __init__(
         lmgr=None,
         Map=None,
         auimgr=None,
+        docked=None,
```
petrasovaa, line 103: dockable?

gui/wxpython/main_window/frame.py
```diff
@@ -961,29 +993,37 @@ def OnCBPageClosing(self, event):
...
+                               is undocked to independent frame
         """
         self.notebookLayers.Unbind(FN.EVT_FLATNOTEBOOK_PAGE_CLOSING)
         self.notebookLayers.DeletePage(pgnum_dict["layers"])
         self.notebookLayers.Bind(
             FN.EVT_FLATNOTEBOOK_PAGE_CLOSING,
             self.OnCBPageClosing,
         )
-        self.mapnotebook.DeletePage(pgnum_dict["mapnotebook"])
+        if is_docked is True:
```
petrasovaa, line 1025:
```suggestion
        if is_docked:
```

gui/wxpython/mapdisp/frame.py
lindakladivova: Makes more sense. Thank you.

gui/wxpython/mapdisp/frame.py
```diff
@@ -113,6 +136,13 @@ def __init__(
         # checks for saving workspace
         self.canCloseDisplayCallback = None
 
+        # distinquishes whether map panel is dockable (Single-Window)
+        self.dockable = dockable
+
+        if self.dockable:
+            # distinguishes whether map panel is docked or not
+            self.docked = True
```
petrasovaa, line 144: You need to instantiate self.docked even when it's not dockable.

gui/wxpython/mapdisp/frame.py
```diff
@@ -659,6 +689,21 @@ def RemoveQueryLayer(self):
...
+        """Sets docking bound method to dock or undock"""
+        self.docking_callback = function
+
+    def OnDockUndock(self, event):
+        """Dock or undock map display panel to independent MapFrame"""
+        self.docking_callback(self)
+        self.docked = not self.docked
+
+    def IsDocked(self):
+        return self.docked
```
petrasovaa, line 702: These could be private variables I assume.

gui/wxpython/main_window/frame.py
```diff
@@ -961,29 +993,37 @@ def OnCBPageClosing(self, event):
 
         maptree = self.notebookLayers.GetPage(event.GetSelection()).maptree
         maptree.GetMapDisplay().CleanUp()
-        self.mapnotebook.DeletePage(self.GetMapDisplayIndex())
+        if self.GetMapDisplay().IsDocked():
+            self.mapnotebook.DeletePage(self.GetMapDisplayIndex())
+        else:
+            self.GetMapDisplay().GetParent().Destroy()
```
petrasovaa, line 999: I think just `self.GetMapDisplay().Destroy()` should be enough thanks to the FrameMixin.

lindakladivova:
The situation after undocking looks as follows:
Screenshot from 2022-12-13 15-58-46

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,141 @@
+"""
+@package frame.notebook
```
petrasovaa, line 2: Should this be main_window.mapnotebook?

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,141 @@
...
+    def __init__(
+        self,
+        parent,
+        agwStyle=aui.AUI_NB_DEFAULT_STYLE | aui.AUI_NB_TAB_EXTERNAL_MOVE | wx.NO_BORDER,
+    ):
+        self.parent = parent
+        super().__init__(parent=self.parent, id=wx.ID_ANY, agwStyle=agwStyle)
+        client_size = self.parent.GetClientSize()
+        self.SetPosition(wx.Point(client_size.x, client_size.y))
+        self.SetSize(430, 200)
```
petrasovaa, line 61: Is this really needed?

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,141 @@
...
+        super().AddPage(*args, **kwargs)
+        self.SetSelection(self.GetPageCount() - 1)
+
+    def SetSelectionToPage(self, page):
+        """Overrides Aui.Notebook SetSelectionToPage method.
+        Decides whether to set selection to a MapNotebook page
+        or an undocked independent frame"""
+        try:
+            super().SetSelection(self.GetPageIndex(page))
+        except Exception:
```
petrasovaa, line 105: Specify which exception exactly this catches.

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,141 @@
...
+        """Overrides Aui.Notebook DeletePage method.
+        Decides whether to destroy a MapNotebook page
+        or an undocked independent frame"""
+        try:
+            super().DeletePage(self.GetPageIndex(page))
+        except Exception:
+            pass
+
+        if not page.IsDocked():
+            page.Destroy()
```
petrasovaa, line 121: I didn't test, but wouldn't this be better?

```suggestion
        if page.IsDocked():
            super().DeletePage(self.GetPageIndex(page))
        else:
            page.Close()
```

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,141 @@
...
+        """Overrides Aui.Notebook SetSelectionToPage method.
+        Decides whether to set selection to a MapNotebook page
+        or an undocked independent frame"""
+        try:
+            super().SetSelection(self.GetPageIndex(page))
+        except Exception:
+            pass
+
+        if not page.IsDocked():
+            wx.CallLater(500, page.GetParent().Raise)
```
petrasovaa, line 109: Would page.Raise() work here? It uses the FrameMixin, no?

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,141 @@
...
+        Decides whether sets title to MapNotebook page
+        or an undocked independent frame"""
+        try:
+            super().SetPageText(page_idx=self.GetPageIndex(page), text=text)
+        except Exception:
+            pass
+
+        if not page.IsDocked():
+            frame = page.GetParent()
+            frame.SetTitle(text)
```
petrasovaa, line 134: I may be confused here, but page object here uses the FrameMixin, no? If yes, SetTitle needs to be added there or it may already be there.

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,141 @@
...
+            wx.CallLater(500, frame.Raise)
+
+    def OnClose(self, event):
+        """Page of map notebook is being closed"""
+        display = self.GetCurrentPage()
+        display.OnCloseWindow(event=None, askIfSaveWorkspace=True)
+        event.Veto()
```
petrasovaa, line 141: event.veto() ??

gui/wxpython/mapdisp/frame.py
```diff
@@ -659,6 +666,21 @@ def RemoveQueryLayer(self):
         for layer in qlayer:
             self.GetMap().DeleteLayer(layer)
 
+    def SetDockingCallback(self, function):
+        """Sets docking bound method to dock or undock"""
+        self.docking_callback = function
+
+    def OnDockUndock(self, event):
+        """Dock or undock map display panel to independent MapFrame"""
+        self.docking_callback(self)
```
petrasovaa, line 675: It looks like the docking_callback may (theoretically) not be defined when this method is called, so I suggest to initialize it to None and test here before calling it.

gui/wxpython/mapdisp/frame.py
```diff
@@ -659,6 +666,21 @@ def RemoveQueryLayer(self):
         for layer in qlayer:
             self.GetMap().DeleteLayer(layer)
 
+    def SetDockingCallback(self, function):
+        """Sets docking bound method to dock or undock"""
+        self.docking_callback = function
```
petrasovaa, line 671: _docking_callback would be more consistent with the rest of _dock...

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,141 @@
...
+
+        if not page.IsDocked():
+            wx.CallLater(500, page.GetParent().Raise)
+
+    def DeletePage(self, page):
+        """Overrides Aui.Notebook DeletePage method.
+        Decides whether to destroy a MapNotebook page
+        or an undocked independent frame"""
+        try:
+            super().DeletePage(self.GetPageIndex(page))
```
petrasovaa, line 116: Since page here is not index, it may be better to not use the same name DeletePage, but create a new method to avoid confusion. Maybe DeleteMapDisplay or something like that.

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,141 @@
...
+        except Exception:
+            pass
+
+        if not page.IsDocked():
+            page.Destroy()
+
+    def SetPageText(self, page, text):
```
petrasovaa, line 123: See the renaming comment.

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,141 @@
...
+            lambda evt: self.GetCurrentPage().onFocus.emit(),
+        )
+        self.Bind(aui.EVT_AUINOTEBOOK_PAGE_CLOSE, self.OnClose)
+
+    def UndockMapDisplay(self, page):
+        """Undock active map display to independent MapFrame object"""
+        idx = self.GetPageIndex(page)
+        text = self.GetPageText(idx)
+        self.RemovePage(idx)
+        fr = MapPageFrame(parent=self.parent, mapdisplay=page, title=text)
```
petrasovaa, line 77: Use frame instead of fr, index instead of idx. Full words are preferable.

gui/wxpython/main_window/frame.py
```diff
@@ -961,29 +943,35 @@ def OnCBPageClosing(self, event):
...
         self.notebookLayers.Unbind(FN.EVT_FLATNOTEBOOK_PAGE_CLOSING)
         self.notebookLayers.DeletePage(pgnum_dict["layers"])
         self.notebookLayers.Bind(
             FN.EVT_FLATNOTEBOOK_PAGE_CLOSING,
             self.OnCBPageClosing,
         )
-        self.mapnotebook.DeletePage(pgnum_dict["mapnotebook"])
+        if is_docked:
+            map_page = self.mapnotebook.GetPage(pgnum_dict["mapnotebook"])
+            self.mapnotebook.DeletePage(map_page)
```
petrasovaa, line 974: If you don't redefine DeletePage, this can stay as it was, no?

gui/wxpython/main_window/notebook.py
lindakladivova: you are right, it was there before this PR but it is not needed

gui/wxpython/main_window/notebook.py
lindakladivova: It does not use FrameMixin. We use FrameMixin only for Multi-Window. We can create the MapDisplay object for Single-Window as well and assume that we will never use Mixin methods on AuiNotebook.

gui/wxpython/main_window/notebook.py
lindakladivova: See above.

gui/wxpython/main_window/notebook.py
lindakladivova: it just a copy paste of the code:

   def OnMapNotebookClose(self, event):
        """Page of map notebook is being closed"""
        display = self.GetMapDisplay(onlyCurrent=True)
        display.OnCloseWindow(event=None, askIfSaveWorkspace=True)
        event.Veto()
I tested it and it does not work without event.Veto()

gui/wxpython/main_window/notebook.py
lindakladivova: It is this exception: No layer tree selected. It occurs when you launch GRASS. I have tried to avoid the "exception: pass" weird handling and defined the exception by "hasattr" if clause.

petrasovaa:
One more thing that needs to be fixed is saving and loading workspace. When loading a workspace (previously saved with one docked and one undocked window) I get:
```
  File "/home/anna/.local/lib/python3.8/site-
packages/wx/core.py", line 2346, in Notify

self.notify()
  File "/home/anna/.local/lib/python3.8/site-
packages/wx/core.py", line 3552, in Notify

self.result = self.callable(*self.args, **self.kwargs)
RuntimeError
:
wrapped C/C++ object of type MapPageFrame has been deleted

```

Ideally, the workspace would save the docking state and use it.

lindakladivova:
> One more thing that needs to be fixed is saving and loading workspace. When loading a workspace (previously saved with one docked and one undocked window) I get:
> 
> ```
>   File "/home/anna/.local/lib/python3.8/site-
> packages/wx/core.py", line 2346, in Notify
> 
> self.notify()
>   File "/home/anna/.local/lib/python3.8/site-
> packages/wx/core.py", line 3552, in Notify
> 
> self.result = self.callable(*self.args, **self.kwargs)
> RuntimeError
> :
> wrapped C/C++ object of type MapPageFrame has been deleted
> ```
> 
> Ideally, the workspace would save the docking state and use it.

Thanks for catching it!

gui/wxpython/mapdisp/frame.py
```diff
@@ -974,7 +1000,14 @@ def OnCloseWindow(self, event, askIfSaveWorkspace=True):
             if pgnum_dict is not None:
                 self.CleanUp()
                 if pgnum_dict["layers"] > -1:
-                    self.closingDisplay.emit(pgnum_dict=pgnum_dict)
+                    if self.IsDockable():
+                        self.closingDisplay.emit(
+                            pgnum_dict=pgnum_dict, is_docked=self.IsDocked()
+                        )
+                        if not self.IsDocked():
+                            self.GetParent().Destroy()
```
lindakladivova, line 1008: I added the support for workspaces but this error is not related to workspaces (it is just a good way how to notice it) .. it is related to a problem in closing MapPageFrame. The problem is specifically on this row. I started to think that it should be fixed by using signals but not sure how to use them. We would need something like "when closing map display panel automatically raise closing map display frame"...

lindakladivova: The problem is probably that when undocking the map frame, I remove the page from notebook by RemovePage method but the pointer still exists on this page...

petrasovaa:
I found one error. Reproduce by opening 2 map windows, detach the second one and then go to layer manager, select the first map window tab and then click at the x on the right:
```
gnu/gui/wxpython/main_window/frame.py", line 928, in
OnCBPageChanged

self.mapnotebook.SetSelectionToPage(self.GetMapDisplay())
gnu/gui/wxpython/main_window/notebook.py", line 105, in
SetSelectionToPage

self.SetSelection(self.GetPageIndex(page))
  File "/home/anna/.local/lib/python3.8/site-
packages/wx/lib/agw/aui/auibook.py", line 4320, in
SetSelection

wnd = self._tabs.GetWindowFromIdx(new_page)
  File "/home/anna/.local/lib/python3.8/site-
packages/wx/lib/agw/aui/auibook.py", line 1229, in
GetWindowFromIdx

return self._pages[idx].window
IndexError
:
list index out of range
```

petrasovaa:
Regarding loading the workspace, the position of the undocked map window is not saved to the workspace.

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,132 @@
...
+        self.SetSelection(self.GetPageCount() - 1)
+
+    def SetSelectionToPage(self, page):
+        """Overrides Aui.Notebook SetSelectionToPage method.
+        Decides whether to set selection to a MapNotebook page
+        or an undocked independent frame"""
+        self.SetSelection(self.GetPageIndex(page))
+
+        if not page.IsDocked():
+            wx.CallLater(500, page.GetParent().Raise)
```
petrasovaa, line 108: This seems to solve the RuntimeError:

```suggestion
            frame = page.GetParent()
            wx.CallLater(500, lambda: frame.Raise if frame else None)
```

gui/wxpython/mapdisp/frame.py
petrasovaa: The problem is in CallLater, the frame does not exist when Raise is eventually called.

lindakladivova: Thanks for finding this, I dived too deep into the problem and then was looking for the problem where no problem existed.

lindakladivova:
> Regarding loading the workspace, the position of the undocked map window is not saved to the workspace.

@petrasovaa I have tried to fix it but I cannot find out the current window position. The positioning behaves differently than the sizing. I still have the same frame position in GetPosition() method. I would need to use the event EVT_MOVING: Process a wxEVT_MOVING event, which is generated while the user is moving the window but this event is available only for Windows.

And event wx.EVT_MOVE is useless in this case since we need to track position changes made by a user, not a program. So, I am not sure how to deal with that.
We actually do not use this positioning currently anywhere. When you @petrasovaa test:
```

        # layer manager
        windowPos = self.lmgr.GetPosition()
        windowSize = self.lmgr.GetSize()

```
in 882 row in workpace/core.py, you store again the static position to a workspace file even if you prior have changed the lmgr frame position.

We basically always store the wrong position. It works for all frame cases - the lmgr frame (M-W, S-W) as well as map displays in M-W.

We can only SetPosition - this is something we do when we create map display. But we cannot track position changes. At least this is how I understand it.

petrasovaa:
Not sure I understand your last comment. It looks like the position problem is fixed now, no?

lindakladivova:
> Not sure I understand your last comment. It looks like the position problem is fixed now, no?

No, it is not fixed. I am afraid we do not have any mechanism for tracking changes on wx.Frame position. So we basically store the position parameter to a workspace file but it is useless because GetPosition always returns the same number. And this problem applies to all wx.Frames (as I have written above we e.g. store the position of the lmgr frame (main frame) and here the problem is the same).

You can try it out - change the position of the main frame (eighter in M-W or S-W, does not matter) and then store a workspace file and load it again. The position will be the default because already when you saved the workspace file, it stored the default position not the current position at that moment.

lindakladivova:
> I found one error. Reproduce by opening 2 map windows, detach the second one and then go to layer manager, select the first map window tab and then click at the x on the right:
> 
> ```
>   File "/home/anna/dev/grass/grass/dist.x86_64-pc-linux-
> gnu/gui/wxpython/main_window/frame.py", line 928, in
> OnCBPageChanged
> 
> self.mapnotebook.SetSelectionToPage(self.GetMapDisplay())
>   File "/home/anna/dev/grass/grass/dist.x86_64-pc-linux-
> gnu/gui/wxpython/main_window/notebook.py", line 105, in
> SetSelectionToPage
> 
> self.SetSelection(self.GetPageIndex(page))
>   File "/home/anna/.local/lib/python3.8/site-
> packages/wx/lib/agw/aui/auibook.py", line 4320, in
> SetSelection
> 
> wnd = self._tabs.GetWindowFromIdx(new_page)
>   File "/home/anna/.local/lib/python3.8/site-
> packages/wx/lib/agw/aui/auibook.py", line 1229, in
> GetWindowFromIdx
> 
> return self._pages[idx].window
> IndexError
> :
> list index out of range
> ```

I fixed this by [6b9bedf](https://github.com/OSGeo/grass/pull/2667/commits/6b9bedf7fa9fcaa1ddf2889fbb73cc30024c3a33)

lindakladivova:
> I found one error. Reproduce by opening 2 map windows, detach the second one and then go to layer manager, select the first map window tab and then click at the x on the right:
> 
> ```
>   File "/home/anna/dev/grass/grass/dist.x86_64-pc-linux-
> gnu/gui/wxpython/main_window/frame.py", line 928, in
> OnCBPageChanged
> 
> self.mapnotebook.SetSelectionToPage(self.GetMapDisplay())
>   File "/home/anna/dev/grass/grass/dist.x86_64-pc-linux-
> gnu/gui/wxpython/main_window/notebook.py", line 105, in
> SetSelectionToPage
> 
> self.SetSelection(self.GetPageIndex(page))
>   File "/home/anna/.local/lib/python3.8/site-
> packages/wx/lib/agw/aui/auibook.py", line 4320, in
> SetSelection
> 
> wnd = self._tabs.GetWindowFromIdx(new_page)
>   File "/home/anna/.local/lib/python3.8/site-
> packages/wx/lib/agw/aui/auibook.py", line 1229, in
> GetWindowFromIdx
> 
> return self._pages[idx].window
> IndexError
> :
> list index out of range
> ```

fixed by f7d35e87a9

lindakladivova:
I tested again all the use cases and I did not come to any bugs. Please, could @petrasovaa test it as well?

petrasovaa:
> No, it is not fixed. I am afraid we do not have any mechanism for tracking changes on wx.Frame position. So we basically store the position parameter to a workspace file but it is useless because GetPosition always returns the same number. And this problem applies to all wx.Frames (as I have written above we e.g. store the position of the lmgr frame (main frame) and here the problem is the same).

Why would you need to track the changes? Whenever the workspace is saved, it gets the current position and saves it there. I don't see a problem here. Your commit 7e316497 is doing that in workspace.py.
> 
> You can try it out - change the position of the main frame (eighter in M-W or S-W, does not matter) and then store a workspace file and load it again. The position will be the default because already when you saved the workspace file, it stored the default position not the current position at that moment.

That's not how it behaves for me, for me it works now as expected.

lindakladivova:
> > No, it is not fixed. I am afraid we do not have any mechanism for tracking changes on wx.Frame position. So we basically store the position parameter to a workspace file but it is useless because GetPosition always returns the same number. And this problem applies to all wx.Frames (as I have written above we e.g. store the position of the lmgr frame (main frame) and here the problem is the same).
> 
> Why would you need to track the changes? Whenever the workspace is saved, it gets the current position and saves it there. I don't see a problem here. Your commit [7e31649](https://github.com/OSGeo/grass/commit/7e3164975b05afca09ce1071db49e99872b8f1b0) is doing that in workspace.py.
> 
> > You can try it out - change the position of the main frame (eighter in M-W or S-W, does not matter) and then store a workspace file and load it again. The position will be the default because already when you saved the workspace file, it stored the default position not the current position at that moment.
> 
> That's not how it behaves for me, for me it works now as expected.

So when you change the position of whatever frame and save the workspace file, you really get a different number for its position in the workspace file?

For me, it does not work like that... I always get the same numbers 26,23 and when I open the workspace file, the pages are loaded to the wrong places (not the places they were before when saving workspace file).. 

I think it is the problem with my system. I have quite recently on my new laptop Python 3.10, wxPython 4.2.0, and Ubuntu 22.04.
I did not originally want to create such an environment but I had problems with wxPython 4.0.7. When I used wxPython 4.0.7 together with Python 3.10 I got this error:

GRASS world_latlong_wgs84/PERMANENT:grass > Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/wx/core.py", line 3282, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
  File "/home/linduska/grass/dist.x86_64-pc-linux-gnu/gui/wxpython/wxgui.py", line 95, in show_main_gui
    mainframe = GMFrame(parent=None, id=wx.ID_ANY, workspace=self.workspaceFile)
  File "/home/linduska/grass/dist.x86_64-pc-linux-gnu/gui/wxpython/main_window/frame.py", line 168, in __init__
    self.BuildPanes()
  File "/home/linduska/grass/dist.x86_64-pc-linux-gnu/gui/wxpython/main_window/frame.py", line 641, in BuildPanes
    self._auimgr.AddPane(
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/framemanager.py", line 4711, in AddPane
    return self.AddPane4(window, arg1, target)
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/framemanager.py", line 4879, in AddPane4
    self.UpdateNotebook()
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/framemanager.py", line 6653, in UpdateNotebook
    notebook.AddPage(pane.window, title, True, pane.icon)
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/auibook.py", line 3575, in AddPage
    return self.InsertPage(self.GetPageCount(), page, caption, select, bitmap, disabled_bitmap, control, tooltip)
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/auibook.py", line 3653, in InsertPage
    self.SetSelectionToWindow(page)
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/auibook.py", line 4410, in SetSelectionToWindow
    self.SetSelection(idx)
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/auibook.py", line 4357, in SetSelection
    ctrl.MakeTabVisible(ctrl_idx, ctrl)
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/auibook.py", line 1843, in MakeTabVisible
    if not self.IsTabVisible(tabPage, self.GetTabOffset(), dc, win):
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/auibook.py", line 1732, in IsTabVisible
    self.Render(dc, wnd)
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/auibook.py", line 1687, in Render
    page.rect, tab_button.rect, x_extent = self._art.DrawTab(dc, wnd, page, rect, tab_button.cur_state)
  File "/usr/lib/python3/dist-packages/wx/lib/agw/aui/tabart.py", line 475, in DrawTab
    r.SetHeight(r.GetHeight()/2)
TypeError: Rect.SetHeight(): argument 1 has unexpected type 'float'

I think that Python 3.10 is probably too new for wxPython 4.2.0 so this is the reason why it does not work... but Python 3.10 is automatically downloaded with dependencies:

sudo apt build-dep grass 

And I am lost in how to force it to install Python 3.9 (which should be stable according to Python docs).. 
Do you have idea how to do that?

neteler:
> I think that Python 3.10 is probably too new for wxPython 4.0.7 so this is the reason why it does not work... but Python 3.10 is automatically downloaded with dependencies:

Here on my Fedora 37 box, I have:

```
python3 -c "import sys, wx; print(sys.version); print(wx.version())"
3.11.1 (main, Jan  6 2023, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]
4.2.0 gtk3 (phoenix) wxWidgets 3.2.1
```

To be honest, I didn't follow much the problem you try to tackle due to lack of time. Shall I test this PR on my environment?

lindakladivova:
My system is this one:
```
python3 -c "import sys, wx; print(sys.version); print(wx.version())"
3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0]
4.2.0 gtk3 (phoenix) wxWidgets 3.2.0
```
seems to me not much obsolete. If you would like to try out the PR, it would be nice but depends on your time.. I will try to play a bit with versions on my system. We will see if it helps.

petrasovaa:
I will try to test it with the newer versions, I have older Python and wx right now. But in general, I think the code is fine, so it would be a bug related to wx, so we might not be able to do anything about it.

neteler:
> Here on my Fedora 37 box, I have:
> 
> ```
> python3 -c "import sys, wx; print(sys.version); print(wx.version())"
> 3.11.1 (main, Jan  6 2023, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]
> 4.2.0 gtk3 (phoenix) wxWidgets 3.2.1
> ```

I applied the current PR and get:

```
  File "/usr/lib64/python3.11/site-packages/wx/core.py", line 3427, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/wxgui.py", line 95, in show_main_gui
    mainframe = GMFrame(parent=None, id=wx.ID_ANY, workspace=self.workspaceFile)
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/main_window/frame.py", line 209, in __init__
    self.NewDisplay(show=False)
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/main_window/frame.py", line 431, in NewDisplay
    self.currentPage.maptree = LayerTree(
                               ^^^^^^^^^^
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/lmgr/layertree.py", line 156, in __init__
    self.mapdisplay = createNewMapDisplay(layertree=self)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/main_window/frame.py", line 410, in CreateNewMapDisplay
    mapdisplay = MapPanel(
                 ^^^^^^^^^
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/mapdisp/frame.py", line 146, in __init__
    self.AddToolbar(toolb)
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/mapdisp/frame.py", line 613, in AddToolbar
    self.toolbars["map"] = MapToolbar(
                           ^^^^^^^^^^^
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/mapdisp/toolbars.py", line 89, in __init__
    self.InitToolbar(self._toolbarData())
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/gui_core/toolbars.py", line 113, in InitToolbar
    self.CreateTool(*tool)
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/gui_core/toolbars.py", line 137, in CreateTool
    toolWin = self.classObject.AddTool(
              ^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: ToolBar.AddTool(): arguments did not match any overloaded call:
  overload 1: argument 2 has unexpected type 'WindowIDRef'
  overload 2: argument 4 has unexpected type 'NoneType'
  overload 3: argument 4 has unexpected type 'NoneType'
```

The GUI looks like this at this stage:

image

Trying to close the GUI leads to:

```
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/main_window/frame.py", line 2275, in OnCloseWindowOrExit
    self._closeWindow(event)
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/main_window/frame.py", line 2285, in _closeWindow
    if not self.workspace_manager.CanClosePage(caption=_("Quit GRASS GUI")):
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/lmgr/workspace.py", line 454, in CanClosePage
    maptree = self._giface.GetLayerTree()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/lmgr/giface.py", line 242, in GetLayerTree
    return self.lmgr.GetLayerTree()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-linux-gnu/gui/wxpython/main_window/frame.py", line 1110, in GetLayerTree
    return self.currentPage.maptree
           ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Panel' object has no attribute 'maptree'
```

Happy to test more.
And wx related bugs we got fixed in the past via Scott, if I recall correctly.

nilason:
Just tested on Mac with
```
3.10.7 (v3.10.7:6cc6b13308, Sep  5 2022, 14:02:52) [Clang 13.0.0 (clang-1300.0.29.30)]
4.2.0 osx-cocoa (phoenix) wxWidgets 3.2.0
```

Clicking the `(Un)dock Map Display` button results in the disappearance of the Map Display panel, but no error messages at all. I have to close GUI and restart to get it back.

petrasovaa:
Tested now with slightly newer configuration on Ubuntu 20.04 and it works fine here.
```
Python 3.10.10 (main, Feb  8 2023, 14:50:01) [GCC 9.4.0]
4.2.0 gtk3 (phoenix) wxWidgets 3.2.0
```

petrasovaa:
> I applied the current PR and get:

That is really strange, I don't see how this PR would do that. Make sure you recompile it in gui/ folder, not just in wxpython/

petrasovaa:
> Just tested on Mac with
> 
> ```
> 3.10.7 (v3.10.7:6cc6b13308, Sep  5 2022, 14:02:52) [Clang 13.0.0 (clang-1300.0.29.30)]
> 4.2.0 osx-cocoa (phoenix) wxWidgets 3.2.0
> ```
> 
> Clicking the `(Un)dock Map Display` button results in the disappearance of the Map Display panel, but no error messages at all. I have to close GUI and restart to get it back.

Just to be clear, no frame with map display appears when you undock it? Could be behind the main window.

nilason:
> > Clicking the `(Un)dock Map Display` button results in the disappearance of the Map Display panel, but no error messages at all. I have to close GUI and restart to get it back.
> 
> Just to be clear, no frame with map display appears when you undock it? Could be behind the main window.

Just disappears, nowhere to be found (even looked under the keyboard :). Strange.

nilason:
Is this to be found in wxdemo?

nilason:
> I’ll give it a try with a different Python - wxPython combo.

The same happens with:
```
3.9.13 | packaged by conda-forge | (main, May 27 2022, 17:00:33) 
[Clang 13.0.1 ]
4.1.1 osx-cocoa (phoenix) wxWidgets 3.1.5
```

lindakladivova:
> Tested now with slightly newer configuration on Ubuntu 20.04 and it works fine here.
> 
> ```
> Python 3.10.10 (main, Feb  8 2023, 14:50:01) [GCC 9.4.0]
> 4.2.0 gtk3 (phoenix) wxWidgets 3.2.0
> ```

That is a very similar version as mine. Have you tested the problem with storing position in workspace file?

lindakladivova:
> Is this to be found in wxdemo?

I have tried to find the methods "Reparent" and "RemovePage" that we use and I have not found them in wxdemo. But both those methods could be simply found in docs - https://docs.wxpython.org/wx.Window.html#wx.Window.Reparent and https://docs.wxpython.org/wx.Window.html. 
"Undock" and "Dock", both are called on wx.Panel in our case. Undock is basically the wx.Panel reparent from wx.lib.agw.aui.Notebook to wx.Frame, and the Dock is the opposite operation. 

I can share the minimum working undock/dock example which just undock and dock a wx.Panel from one wx.Frame to a different wx.Frame.

Could you try that code?
```

import wx

class MainFrame(wx.Frame):

    def __init__(self, parent, id=-1, title="", pos=wx.DefaultPosition,
                 size=wx.DefaultSize, style=wx.DEFAULT_FRAME_STYLE |
                                            wx.SUNKEN_BORDER |
                                            wx.CLIP_CHILDREN):

        wx.Frame.__init__(self, parent, id, title, pos, size, style)

        # tell FrameManager to manage this frame
        self.panel = MyPanel(self)
        self.panel.SetCallback(self.Undock)
        self.sizer = wx.BoxSizer()
        self.sizer.Add(self.panel, 1, wx.EXPAND)
        self.SetSizer(self.sizer)

    def Undock(self, panel):
        fr = SecondFrame(self)
        fr.SetTitle("Panel")
        panel.Reparent(fr)
        panel.SetCallback(self.Dock)
        self.sizer.Detach(panel)
        fr.sizer.Add(panel)
        fr.Show()

    def Dock(self, panel):
        fr = panel.GetParent()
        panel.Reparent(self)
        panel.SetCallback(self.Undock)
        fr.Close()


class MyPanel(wx.Panel):
    def __init__(self, parent):
        wx.Panel.__init__(self, parent=parent)
        self.parent = parent
        self.callback = None
        sizer = wx.BoxSizer(wx.VERTICAL)
        self.btn = wx.Button(self, label="(un)dock")
        sizer.Add(self.btn, proportion=0, flag=wx.EXPAND | wx.ALL, border=5)
        sizer.Add(wx.StaticText(self, label="This panel can be separated to a different wx.Frame"),
                  proportion=0, flag=wx.EXPAND | wx.ALL, border=5)
        self.SetSizerAndFit(sizer)
        self.btn.Bind(wx.EVT_BUTTON, self.OnButton)

    def SetCallback(self, function):
        self.callback = function

    def OnButton(self, event):
        self.callback(self)


class SecondFrame(wx.Frame):
    def __init__(self, parent):
        wx.Frame.__init__(self, parent=parent)
        self.sizer = wx.BoxSizer(wx.HORIZONTAL)
        self.SetSizerAndFit(self.sizer)


if __name__ == '__main__':
    app = wx.App()
    fr = MainFrame(None)
    fr.Show()
    app.MainLoop()

```

nilason:
> Could you try that code?

Thanks! This will probably simplify a lot the tracking down of this issue on Mac. Clicking the button erases the frame in a similar way as in GRASS. Will look into it.

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class MapPageFrame(wx.Frame):
+    """Frame for independent map display window."""
+
+    def __init__(self, parent, mapdisplay, size, pos, title):
+        wx.Frame.__init__(self, parent=parent, size=size, pos=pos, title=title)
+        self.mapdisplay = mapdisplay
+        self.sizer = wx.BoxSizer(wx.HORIZONTAL)
+        self.SetSizerAndFit(self.sizer)
```
nilason, line 35:
```suggestion
        self.SetSizer(self.sizer)
        self.Centre()
```

This solves the problem on Mac, specifically `self.SetSizerAndFit` was the culprit.
`self.Centre()` was needed as the detached panel was placed partly below the macOS main menu.

petrasovaa:
> I tested again all the use cases and I did not come to any bugs. Please, could @petrasovaa test it as well?

Another problem is when you close an undocked window by closing the associated layer tree with the x. The layer tree is removed, but the window stays.

petrasovaa:
> > Tested now with slightly newer configuration on Ubuntu 20.04 and it works fine here.
> > ```
> > Python 3.10.10 (main, Feb  8 2023, 14:50:01) [GCC 9.4.0]
> > 4.2.0 gtk3 (phoenix) wxWidgets 3.2.0
> > ```
> 
> That is a very similar version as mine. Have you tested the problem with storing position in workspace file?

Yes, no problems there.  Not sure what's going on, but maybe focus on the other issues now.

gui/wxpython/main_window/notebook.py
lindakladivova: Nicklas, thank you for testing it!
Unfortunately, on my Ubuntu the result does not look good after that change:
Screenshot from 2023-02-13 17-37-37
The content has been put a bit lower and a status bar is not visible.

So we will probably need to apply different solutions based on OS. It would be nice if someone tested it also on Windows. @hellik could I ask you? We would specifically need to verify that undocking and docking using the new Undock/Dock button works fine here.

nilason: I think I had the same problem. But I was happy to see THAT much ! :-) , a manual window resize and everything good.
I'll see if I can come up with something regarding this detail.

lindakladivova:
> > I tested again all the use cases and I did not come to any bugs. Please, could @petrasovaa test it as well?
> 
> Another problem is when you close an undocked window by closing the associated layer tree with the x. The layer tree is removed, but the window stays.

You are right, the window is only

> > I tested again all the use cases and I did not come to any bugs. Please, could @petrasovaa test it as well?
> 
> Another problem is when you close an undocked window by closing the associated layer tree with the x. The layer tree is removed, but the window stays.

Thanks for catching it. Should be okay now.:)

gui/wxpython/mapdisp/frame.py
```diff
@@ -1005,7 +1005,8 @@ def OnCloseWindow(self, event, askIfSaveWorkspace=True):
                             pgnum_dict=pgnum_dict, is_docked=self.IsDocked()
                         )
                         if not self.IsDocked():
-                            self.GetParent().Destroy()
+                            frame = self.GetParent()
+                            frame.closeFrameNoEvent()
```
petrasovaa, line 1009: This is strange, why is this needed? Based on how closing windows work, the method closeFrameNoEvent should be equivalent to calling Destroy().

lindakladivova: right, it is not needed at all, I have replaced it by Destroy

lindakladivova:
> When I drag a second map display to have 2 displays side by side I get:
> 
> ```
>   File "/home/akratoc/dev/grass/venv/lib/python3.10/site-
> packages/wx/lib/agw/aui/auibook.py", line 5281, in
> OnTabEndDrag
> 
> self.SetSelectionToPage(page_info)
>   File "/home/akratoc/dev/grass/grass_main/dist.x86_64-pc-
> linux-gnu/gui/wxpython/main_window/notebook.py", line 120,
> in SetSelectionToPage
> 
> if not page.IsDocked():
> AttributeError
> :
> 'AuiNotebookPage' object has no attribute 'IsDocked'
> ```

Unfortunately, I have still a problem in the moving operation itself:
https://github.com/OSGeo/grass/issues/2671
so I cannot reproduce it..  I am wondering if not to establish an Issue on wxPython on it...

The problem here is in SetSelectionToPage, we need to define a custom SetSelectionToMapPage because we need the wx.Panel as the input, not AuiNotebookPage. Please could you @petrasovaa test it after the changes I made in c3a161e53e ?

petrasovaa:
> The problem here is in SetSelectionToPage, we need to define a custom SetSelectionToMapPage because we need the wx.Panel as the input, not AuiNotebookPage. Please could you @petrasovaa test it after the changes I made in [c3a161e](https://github.com/OSGeo/grass/commit/c3a161e53e8c62a9a71e689862c4995c5a37f3fe) ?

Works great!

gui/wxpython/main_window/notebook.py
nilason: I have tried everything possible, that I could think of, but was unable to make this work as it should (so far). Only resizing "manually", even with as little as a pixel size change, would re-render the status bar and all content correctly.

petrasovaa: Not sure what you tried, but right now the mapdisplay is added to be managed by the sizer after the frame is created (see the UndockMapDisplay method), so it is not completely surprising this doesn't quite work. I suggest to move the `frame.sizer.Add(...)` call into the MapPageFrame.

lindakladivova: You are right @petrasovaa .. it is not so easy because we have the reparent method here. I have tried several options and the only thing I could think of is putting self.mapdisplay.Reparent(self) directly inside MapPageFrame __init__. SetSizer needs to be called after reparenting otherwise it does not work.

Now it works for me as expected. Please @nilason, could you test it one more time?

nilason: With [e2f1db8](https://github.com/OSGeo/grass/pull/2667/commits/e2f1db8baa65a59cbca22a2f48ad5ac9a39f0414) it works for me too!

I still need a `frame.Centre(wx.BOTH)` for correct positioning though, not only will the undocked window be placed partially under macOS menu:

Screenshot

but with multiple displays, the undocked window will always open in primary display (even if single-window is in secondary. With centre command the undocked window opens centred in the same display.

gui/wxpython/main_window/notebook.py
```diff
@@ -0,0 +1,153 @@
...
+        original_pos = page.GetPosition()
+        self.RemovePage(index)
+        frame = MapPageFrame(
+            parent=self.parent,
+            mapdisplay=page,
+            size=original_size,
+            pos=original_pos,
+            title=text,
+        )
+        frame.SetDockingCallback(self.DockMapDisplay)
```
nilason, line 105: This works for me:

```suggestion
        frame.Centre(wx.BOTH)
        frame.SetDockingCallback(self.DockMapDisplay)
```

gui/wxpython/main_window/notebook.py
wenzeslaus: I'm sorry, I'm not following this fully, but centering is best done on the parent than on the screen. Center on screen does not work well on extremely large displays like [this one](https://doi.org/10.3390/ijgi4020942):

Screenshot from 2023-02-20 09-43-25

gui/wxpython/main_window/notebook.py
lindakladivova: Thank you for testing @nilason! Just from my side, it might be better to put this line to the MapPageFrame init (all layout code is currently in _layout method). Is it still working after [60cd7da](https://github.com/OSGeo/grass/pull/2667/commits/60cd7da7ef39f36b542a1a9bec7047c26c61e7e5)?

lindakladivova: Good. I have used the CentreOnParent. For me there is no change in the results - from [1c4aef9](https://github.com/OSGeo/grass/pull/2667/commits/1c4aef99c22b1162ea89fe5c69001306c7d94b01) to [d5609a3](https://github.com/OSGeo/grass/pull/2667/commits/d5609a3af73d8f8ee79055b28a2f607bb093841a), I experienced the same behavior on Ubuntu.

wenzeslaus: There should be no or little change for a maximized window.

nilason:
All fine with [d5609a3](https://github.com/OSGeo/grass/pull/2667/commits/d5609a3af73d8f8ee79055b28a2f607bb093841a) too!

gui/wxpython/main_window/notebook.py
wenzeslaus: Thanks @lindakladivova for implementing and @nilason for testing. I'm resolving this conversation.

petrasovaa (approved):
Since we have now release branch 8.3, we can merge this. More testing welcome.

PR #2674, opened 2022-11-30
grass.script: replace deprecated use of pipes.quote()

nilason:
`pipes` is deprecated since Python 3.11 and will be removed in version 3.13.
`pipes.quote()` is deprecated since version 2.7.


nilason:
Merged and backported to 8.2.

PR #2680, opened 2022-12-03
Add MCC to r.kappa

marisn:
This PR adds Matthews (Mattheus) Correlation Coefficient as one of indices provided by r.kappa.

Should be applied on the top of PR #2666


marisn:
Ping @wenzeslaus @nilason

nilason (approved):
Looks good to me!

raster/r.kappa/calc_metrics.c
```diff
@@ -213,3 +242,20 @@ static int longcomp(const void *aa, const void *bb)
 
     return (*a > *b);
 }
+
+/* Implements improved Kahan–Babuska algorithm by Neumaier, A. 1974 */
+void update_sum(double *sum, double *c, double value)
+{
+    double tmp;
+
+    tmp = *sum + value;
```
wenzeslaus, line 251: Elsewhere we declare and init at the same line.

raster/r.kappa/calc_metrics.c
```diff
@@ -213,3 +242,20 @@ static int longcomp(const void *aa, const void *bb)
...
+
+    if (fabs(*sum) >= fabs(value))
+        *c = *c + (*sum - tmp) + value;
```
wenzeslaus, line 254: Any reason not to use the `+=` operator? Even with the extra `*`, `*c += (*sum...` may communicate the intention clearly to both Python and C programmers (operator precedence should be clear or is at least suggested by the formatting). Wikipedia version uses `+=` in its pseudocode example.

raster/r.kappa/calc_metrics.c
```diff
@@ -142,10 +148,12 @@ void calc_metrics(void)
...
             metrics->producers_accuracy[i] = 100 * (pii[i] / pj[i]);
         /* theta 1 */
-        p0 += pii[i];
+        update_sum(&p0, &p0c, pii[i]);
         /* theta 2 */
-        pC += pi[i] * pj[i];
+        update_sum(&pC, &pCc, pi[i] * pj[i]);
     }
+    p0 = p0 + p0c;
+    pC = pC + pCc;
```
wenzeslaus, line 156: `+=` expresses the intention better in C (and Python). Same for inter1.

wenzeslaus (changes requested):
The Neumaier's Kahan-Babuska algorithm fits with the version in Wikipedia. I did not check MCC itself.

Given that Kahan-Babuska is applied to existing code too, this may change some results, no? I see it does not change any of the tested code. I think it is the right thing to do and that we can do that between minor versions, but we may need to emphasize that in the release notes (as a separate item in "changed" besides addition to the MCC item).

I'm asking for some code tweaks in the comments.

raster/r.kappa/testsuite/test_r_kappa.py
```diff
@@ -181,6 +181,7 @@ def test_standard_output(self):
         # Kappa value
         vals = rows[28].split()
         self.assertTrue(self.match(vals[0], 0.52091))
+        self.assertTrue(self.match(vals[2], 0.55930))
```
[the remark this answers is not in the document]
marisn: Rossiter does not describe MCC in his publication.

raster/r.kappa/calc_metrics.c
marisn: I think we should move update_sum to main library (gmath) and replace most of += with nkb in all libraries and modules. But that's for another day.

raster/r.kappa/calc_metrics.c
```diff
@@ -213,3 +242,20 @@ static int longcomp(const void *aa, const void *bb)
...
+    tmp = *sum + value;
+
+    if (fabs(*sum) >= fabs(value))
+        *c = *c + (*sum - tmp) + value;
+    else
+        *c = *c + (value - tmp) + *sum;
+
+    *sum = tmp;
+
+    return;
```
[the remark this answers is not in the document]
marisn: Done.

raster/r.kappa/calc_metrics.c
marisn: Merged to a single line.

raster/r.kappa/calc_metrics.c
marisn: No reason. Changed to +=

marisn:
> Given that Kahan-Babuska is applied to existing code too, this may change some results, no? I see it does not change any of the tested code. I think it is the right thing to do and that we can do that between minor versions, but we may need to emphasize that in the release notes (as a separate item in "changed" besides addition to the MCC item).

In most of cases results shouldn't change. NKB will give different outcome only in rare corner cases or when n is large.

raster/r.kappa/calc_metrics.c
wenzeslaus: I agree that's best.

Here, I meant just:
 
```suggestion
    p0 += p0c;
    pC += pCc;
```

Sorry, I was not clear.

raster/r.kappa/testsuite/test_r_kappa.py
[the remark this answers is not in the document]
wenzeslaus: You specify very nicely how the value in the other test was obtained. Can you please mention how this was obtained? One sentence is enough just to distinguish if it is an independent computation (by hand, another software) or the current result you consider correct right now.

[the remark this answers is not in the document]
marisn: I added a comment.

raster/r.kappa/calc_metrics.c
marisn: Is there a some kind of style guideline demanding such change? I do use += for counters and accumulators in loops but prefer to have all terms explicit when summation is neither of them.

wenzeslaus (approved):

wenzeslaus:
> Is there a some kind of style guideline [for using `a += b` instead of `a = a + b`]? I do use += for counters and accumulators in loops but prefer to have all terms explicit when summation is neither of them.

While in math it is common to have the terms explicit, the current programming practices usually aim primarily at documenting intention given that people spend most of the time by reading code (theirs or someone else's), so _a += b_ comes out better because it clearly says add _b_ to _a_.

In the following code:

```
    p0 = p0 + p0c;
    pC = p0 + pCc;
```

did I really mean to do `p0 + pCc` or is it a copy-paste error and I actually mean `pC + pCc`?

When using `+=`, it becomes clear:

```
    p0 += p0c;
    pC += pCc;
```

Clearly I meant to add to the existing value. If I would really mean `p0 + pCc`, I would write: 

```
    p0 += p0c;
    pC = p0 + pCc;
```

marisn:
> did I really mean to do `p0 + pCc` or is it a copy-paste error and I actually mean `pC + pCc`?

This is a good reason. Thank you for pointing out to this aspect of coding style.

wenzeslaus:
You are welcome. I'm sorry I was not more clear before.

PR #2703, opened 2022-12-15
Add packaging status info widget across various GNU/Linux distributions

tmszi:
Add [packaging status info widget](https://repology.org/badge/vertical-allrepos/grass.svg?exclude_unsupported=1&exclude_sources=modules,site&minversion=8.2.0)  (use https://repology.org/project/grass/badges) across various GNU/Linux distributions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,3 +84,7 @@ this issue, clean all the compiled files from the source code:
 ```bash
 make distclean
 ```
+
+## Packages
+
+[![Packaging status](https://repology.org/badge/vertical-allrepos/grass.svg?exclude_unsupported=1&exclude_sources=modules,site&minversion=8.2.0)](https://repology.org/project/grass/versions)
--- a/README.md
+++ b/README.md
@@ -85,6 +85,6 @@ this issue, clean all the compiled files from the source code:
 make distclean
 ```
 
-## Packages
+## Available GNU/Linux and Unix-like OS packages
 
 [![Packaging status](https://repology.org/badge/vertical-allrepos/grass.svg?exclude_unsupported=1&exclude_sources=modules,site&minversion=8.2.0)](https://repology.org/project/grass/versions)


README.md
```diff
@@ -84,3 +84,7 @@ this issue, clean all the compiled files from the source code:
 ```bash
 make distclean
 ```
+
+## Packages
```
neteler, line 88:
```suggestion
## Available Linux packages
```

... just to avoid confusion with Win/Mac/etc.

neteler: Mhh, just discovering "MacPorts" in the list :-)

tmszi: Yes your are right. Better section name will be 'Available GNU/Linux and Unix-like OS packages', because list contains FreeBSD and macOS.

nilason:
I don't think the README, nor in source repo at all is the place for this addition. Incorporated somewhere under https://grass.osgeo.org/download/, however, would make perfect sense.

tmszi:
> I don't think the README, nor in source repo at all is the place for this addition. Incorporated somewhere under https://grass.osgeo.org/download/, however, would make perfect sense.

My source of inspiration was some projects I saw here on GitHub (README file). But I agree that a better place would be https://grass.osgeo.org/download/.

nilason:
Now I remember, there is actually a link to repology at https://grass.osgeo.org/download/linux/, but displaying the badge might be a good addition (but not under "Linux" as there are BSD based system also).

veroandreo:
IMHO, such a long image won't look good in the website either. 

Instead, the "tiny per-repository badges with version" displayed [here](https://repology.org/project/grass/badges) could be added on a per distro basis, i.e. only badges for those distros in the https://grass.osgeo.org/download/linux/ page

nilason:
> IMHO, such a long image won't look good in the website either.

I absolutely agree on that one. I was thinking about that, thought of perhaps a popup image or a toggle button. In both cases the image is shown only "on demand" and might not be so distracting.

> Instead, the "tiny per-repository badges with version" displayed [here](https://repology.org/project/grass/badges) could be added on a per distro basis, i.e. only badges for those distros in the https://grass.osgeo.org/download/linux/ page

These need to be "managed" in some degree, I guess. If that's the case it would mean unnecessary work on our part to keep track etc.

nilason:
Either way, it is also possible to divide it into columns:

https://repology.org/badge/vertical-allrepos/grass.svg?exclude_unsupported=1&exclude_sources=modules,site&minversion=8.2.0&columns=3

veroandreo:
Packaging info is now included in the website with https://github.com/OSGeo/grass-website/pull/354, closing.

PR #2414, opened 2022-06-03
contributing: Optimize commands, text in release procedure

wenzeslaus:
Update release procedure after 8.2.0 (that's right).

- Add cron job update
- Restructure updates to website and wiki
- Remove Launchpad notes (I think the reason was they are not used, at least not by the release manager.)
- Change of procedure for the mailing list announcements
- Improves precision in the text

PR #2707, opened 2022-12-18
doc: update howto_release.md to release_branch

neteler:
Changes of this PR:

**fixes**:
- fixes `git diff` to `git show`
- fix path to `generate_release_notes.py`
- updates SERVER1 URL after recent lxd container migration

TODO: how to enforce the correct release branch in `git push upstream` step?

**enhancements**:
- adds release branch parameter to `git` calls
- changes hardcoded version to `$VERSION`
- adds some comments

TODO after merge: fwd-port fixes in this PR to `main`.


doc/howto_release.md
```diff
@@ -150,6 +151,7 @@ git tag -n --sort=-taggerdate
 Now push the tag upstream - this will trigger the automated workflows linked to tags:
 
 ```bash
+# TODO: how to enforce the right release branch?
```
neteler, line 154: @wenzeslaus How does the tag magic work for `release_branch_8_2`? With `git push upstream` I guess it would land in `main` and not the branch...

doc/howto_release.md
```diff
@@ -86,10 +86,10 @@ git status
 git show
 ```
 
-Push the tag to the upstream repo:
+Push the tag to the upstream repo, release branch:
 
 ```bash
-git push upstream
+git push upstream releasebranch_8_2
```
wenzeslaus, line 92: This is pushing the commits on the release branch. I'm not sure about the tag, probably not. I don't know why this says tag actually. There is no tag created above, so I guess I copy-pasted some other text.

wenzeslaus (changes requested):

wenzeslaus:
This SO question seems to cover the tag and branch topic well: [Do Git tags only apply to the current branch?](https://stackoverflow.com/questions/14613540/do-git-tags-only-apply-to-the-current-branch)

doc/howto_release.md
neteler: This is where I recently messed it up (I tried to use this procedure for the G78 RC3 release but failed).

doc/howto_release.md
neteler: > The tag is made on the branch - a specific commit actually - not pushed to the branch, so what matters is the `git tag` step.

Yes, I see (still learning this part of git). From related the [SO](https://stackoverflow.com/a/27154277/452464) answer which you indicated I read:

_Specifying a branch name as the tag target defaults to that branch's most recent commit; e.g.:_

    git tag v0.1.0  develop   # tags HEAD of 'develop' branch

_(As others have noted, you can also specify a commit ID explicitly as the tag's target.)_

So this should translate into

    git tag v7.8.8RC3 releasebranch_8_2   # tags HEAD of 'releasebranch_8_2' branch

... right? But then `push` to where?

(Importantly, the instructions here shall become fail-safe to also be able to successfully release even when tired ;-) )

neteler: > BTW, do you want to name next release branch release_branch_8_3, not releasebranch_8_3? :-)

Sorry, typo...

neteler:
Needed: partial sync to #2414

neteler:
Replaced by #3030, #3031, #3032.

PR #2718, opened 2022-12-25
wxGUI/console: Adding prompt string to command prompt

lindakladivova:
This PR adds input prompt to Console in order to clarify beginners that they are supposed to write their commands here. Follows the idea given in https://github.com/OSGeo/grass/issues/2585.

Screenshot from 2022-12-27 14-37-09

### DIFF
--- a/gui/wxpython/gui_core/prompt.py
+++ b/gui/wxpython/gui_core/prompt.py
@@ -187,10 +187,10 @@ class GPromptSTC(GPrompt, wx.stc.StyledTextCtrl):
         self.SetUseHorizontalScrollBar(True)
 
         # prompt of the interpreter
-        self.sysps3 = str(sys.ps3)
-        self.write(self.sysps3)
+        self.prompt_string = ">>> "
+        self.write(self.prompt_string)
         self.DocumentEnd()
-        self.promptPosEnd = self.GetCurrentPos()
+        self.prompt_pos_end = self.GetCurrentPos()
 
         # support light and dark mode
         bg_color = wx.SystemSettings().GetColour(wx.SYS_COLOUR_WINDOW)
@@ -221,7 +221,14 @@ class GPromptSTC(GPrompt, wx.stc.StyledTextCtrl):
 
     def CanEdit(self):
         """Return true if editing should succeed."""
-        return self.GetCurrentPos() >= self.promptPosEnd
+        if self.GetSelectionStart() != self.GetSelectionEnd():
+            if self.GetSelectionStart() >= self.prompt_pos_end \
+                   and self.GetSelectionEnd() >= self.prompt_pos_end:
+                return True
+            else:
+                return False
+        else:
+            return self.GetCurrentPos() >= self.prompt_pos_end
 
     def OnTextSelectionChanged(self, event):
         """Copy selected text to clipboard and skip event.
@@ -299,7 +306,7 @@ class GPromptSTC(GPrompt, wx.stc.StyledTextCtrl):
         text += end
 
         self.AddText(compl)
-        pos = len(text) + self.promptPosEnd
+        pos = len(text) + self.prompt_pos_end
         self.SetCurrentPos(pos)
 
         cmd = text.strip().split(" ")[0]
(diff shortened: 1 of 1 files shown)


veroandreo:
Hi Linda! 
I'm wondering: Is there a reason to not use `>` similar to the GRASS terminal and Python console? I find the `<--` a bit confusing... does the user write before or after the arrow?

lindakladivova:
> Hi Linda! I'm wondering: Is there a reason to not use `>` similar to the GRASS terminal and Python console? I find the `<--` a bit confusing... does the user write before or after the arrow?

Hi Vero, I am not sure about '>', do you mean the `>>> ` which is used in Python console? If yes, I agree. I first wanted to use a sys module variable for a prompt string (https://docs.python.org/3/library/sys.html) but for some reason sys.ps1 and sys.ps2 were not defined and only sys.ps3 worked.

tmszi:
Perhaps the `>_` symbol would be more appropriate, since the `>>>` symbol is a Python shell prompt, which the  wxGUI Console is not. It can be misleading.

veroandreo:
> Perhaps the `>_` symbol would be more appropriate, since the `>>>` symbol is a Python shell prompt, which the wxGUI Console is not. It can be misleading.

I tend to agree, either `>_` or simply `>`

lindakladivova:
> > Perhaps the `>_` symbol would be more appropriate, since the `>>>` symbol is a Python shell prompt, which the wxGUI Console is not. It can be misleading.
> 
> I tend to agree, either `>_` or simply `>`
Just realized that for example the Python Shell prompt in Spyder is also interesting - it uses "In [1]:" where the number 1, 2, etc. refers to the order of executed commands.

tmszi:
Or we can distinguish between console prompt symbol according OS:

GNU/Linux and Unix-like (Free/OpenBSD, macOS) terminal uses the prompt symbol `$`.

MS Windows CMD uses the prompt symbol `>`.

lindakladivova:
> Or we can distinguish between console prompt symbol according OS:
> 
> GNU/Linux and Unix-like (Free/OpenBSD, macOS) terminal uses the prompt symbol `$`.
> 
> MS Windows CMD uses the prompt symbol `>`.

I am still not sure if this one sign (may it be `$` or `>`) truly helps beginners to get oriented. What about something like `username:~$ `(for all OS options)?

tmszi:
> I am still not sure if this one sign (may it be `$` or `>`) truly helps beginners to get oriented. What about something like `username:~$ `(for all OS options)?

I'm ok with it, but for GRASS GIS under MS Windows, it would be better to use e.g. ` C:\Users\IEUser>` style used by CMD or Powershell.

lindakladivova:
> > I am still not sure if this one sign (may it be `$` or `>`) truly helps beginners to get oriented. What about something like `username:~$ `(for all OS options)?
> 
> I'm ok with it, but for GRASS GIS under MS Windows, it would be better to use e.g. `C:\Users\IEUser>` style used by CMD or Powershell.

Yes @tmszi, I agree. I looked under Windows and GRASS's independent CMD actually also uses this `C:\Users\IEUser>` for GRASS's CMD:
windows

When I looked under my Ubuntu, it is different, it uses `location@mapset:grass >` style:
Screenshot from 2023-01-02 07-45-16

First I thought to preserve that style exactly as it is. The first style for Windows and the second one for Unix-like systems.

But for Unix-like systems, I am not sure how the Console prompt should behave when a user switches to a different mapset. Should the prompt string be also changed? Probably yes and then it is getting a bit complicated. :-)

So maybe the variant with `username >` for Unix-like systems is the most similar to the current state and we do not need to care of some outer changes. (Just thoughts that when we do that once we would find the solution that would be as user-friendly and natural as it is possible).

tmszi:
> So maybe the variant with `username >` for Unix-like systems is the most similar to the current state and we do not need to care of some outer changes. (Just thoughts that when we do that once we would find the solution that would be as user-friendly and natural as possible).

Yes I agree as you suggest, let's use the `username >` style for the wxGUI Console prompt on all OS.

hellik:
>I was wondering a different solution would work better (maybe not): When there is no focus on the console and it's empty, it would display grayish text (e.g. something like "Type command here and press Enter") and then on focus the text would disappear. When the widget looses focus, it would test if it's empty and if yes, it would show the help text again.

I concur here regarding usability

lindakladivova:
Implemented according to the suggestion https://github.com/OSGeo/grass/pull/2718#issuecomment-1369200501 in https://github.com/OSGeo/grass/pull/2728 .

PR #2724, opened 2023-01-02
i.maxlik: fix crash when classification result is NULL

marisn:
If any input cell from imagery group is NULL, the result of classification is also NULL. The implementation of original cat restoration (79f950095283fc351ecc83fc54b5cd7271101d21) did not check for this corner case and thus caused an out of bounds value access.


marisn:
The erroneous code was present only in the main branch thus no backporting is required.

PR #2727, opened 2023-01-04
Fix missing function prototypes

nilason:
This adds most, but not all, missing function prototypes. 

Aaron Ballman, from llvm, [summarise](https://discourse.llvm.org/t/rfc-enabling-wstrict-prototypes-by-default-in-c/60521):
> Functions without prototypes have never been a recommended practice in any standard version of C. In fact, they were introduced into C89 as already being deprecated (see C89 3.9.4 and 3.9.5). After a 35+ year deprecation period, C2x will be removing support for functions with identifier lists (http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2432.pdf) and will be changing the behavior of prototypeless functions with empty parentheses to match the behavior of C++ ([N 2841: No function declarators without prototypes](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2841.htm)).

It is only a question of time before this **must** be fixed, but for a clearer and safer code I think there is no need to wait.

These fixes are trivial, but there were 4 files needing special attention and will be object for separate issues/pull requests.


nilason:
If there are no objections to this PR, I'll merge it in the next few days.
I have further PRs in line, which builds on this.

PR #2728, opened 2023-01-04
wxGUI/console: Adding hint to command prompt

lindakladivova:
Reimplementation of #2718

According to the simple solution suggested by @petrasovaa in https://github.com/OSGeo/grass/pull/2718#issuecomment-1369200501 (thanks for that, sometimes it is handy to look at the problem from an entirely different perspective). 

Screenshot from 2023-01-05 17-28-40

### DIFF
--- a/gui/wxpython/gui_core/prompt.py
+++ b/gui/wxpython/gui_core/prompt.py
@@ -312,9 +312,8 @@ class GPromptSTC(GPrompt, wx.stc.StyledTextCtrl):
         event.Skip()
 
     def OnSetFocus(self, event):
-        if not self.IsEmpty() and self.GetText() == self.hint:
+        if self.GetText() == self.hint:
             self.ClearAll()
-            self.SetFocus()
         event.Skip()
 
     def SetTextAndFocus(self, text):
--- a/gui/wxpython/gui_core/prompt.py
+++ b/gui/wxpython/gui_core/prompt.py
@@ -185,7 +185,7 @@ class GPromptSTC(GPrompt, wx.stc.StyledTextCtrl):
         self.SetUseTabs(False)
         self.UsePopUp(True)
         self.SetUseHorizontalScrollBar(True)
-        self.SetHint(_("Type command here and press Enter"))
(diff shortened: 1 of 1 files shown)


petrasovaa (changes requested):
I don't see the hint after starting GUI, perhaps the events are not called at that point yet.

lindakladivova:
> I don't see the hint after starting GUI, perhaps the events are not called at that point yet.

You are right, I see it now - previously I tested it only on Multi-Window where this problem does not occur.

lindakladivova:
Screenshot from 2023-01-05 17-28-40

gui/wxpython/gui_core/prompt.py
```diff
@@ -301,10 +313,20 @@ def OnItemSelected(self, event):
...
 
     def OnKillFocus(self, event):
-        """Hides autocomplete"""
+        """Hides autocomplete and writes a hint if a control is currently empty."""
         # hide autocomplete
         if self.AutoCompActive():
             self.AutoCompCancel()
+        if self.IsEmpty():
+            self.StyleSetForeground(0, self.text_color["hint"])
+            self.WriteText(self.hint)
```
petrasovaa, line 323: These could be methods ShowHint and HideHint so you don't need to repeat it upon initialization and the colors can be defined in those function so you won't need self.text_color.

gui/wxpython/gui_core/prompt.py
```diff
@@ -300,11 +304,40 @@ def OnItemSelected(self, event):
             except IOError:
                 self.cmdDesc = None
 
+    def IsHintShown(self):
+        return self._hint_shown
+
+    def SetHintShown(self, shown):
+        self._hint_shown = shown
```
petrasovaa, line 311: Just some minor things. You don't need these methods, they are not supposed to be called from outside, just directly set the attribute. Also methods ShowHint/HideHint could be _showHint/_hideHint since they are used only within the class.

petrasovaa (approved):
Thanks for addressing all the small issues!

PR #2732, opened 2023-01-04
Add clang-format shell script

nilason:
Adds a grass_clang_format.sh script.

Usage:
- If you have *clang-format* in PATH, execute for complete source formatting:
   `./utils/grass_clang_format.sh`

- Setting 'GRASS_CLANG_FORMAT' to explicitly set clang-format binary (its naming can depend on how various
   distributions name it):
   `GRASS_CLANG_FORMAT="clang-format-15" ./utils/grass_clang_format.sh`

- It is also possible to format the content in a (one) given directory:
   `./utils/grass_clang_format.sh ./lib/raster`


utils/grass_clang_format.sh
```diff
@@ -0,0 +1,83 @@
...
+#            License (>=v2). Read the file COPYING that comes with GRASS
+#            for details.
+#
+###############################################################################
+
+# Required clang-format version
+req_cf_v="15"
+
+# No need to continue if the .clang-format file isn't found
+if [ ! -f .clang-format ] && [ ! -f ../.clang-format ]; then
```
neteler, line 37:
```suggestion
if [ ! -f .clang-format ] && [ ! -f ../.clang-format ]; then
```

While we have our official file in place at https://github.com/OSGeo/grass/blob/main/.clang-format (resp. in a local repo copy thru `git clone`) we cannot be sure that `../.clang-format` contains the same.
Perhaps better limit the choice to the file included in the repo?

neteler:
I have tested the script on my Fedora 36 box (in order to update an existing PR) but `find` doesn't find anything yet:

```
# attempting to update PR cmake_prep_infinity_nan
bash -x grass_clang_format.sh
+ '[' -z '' ']'
+ case "$-" in
+ __lmod_vx=x
+ '[' -n x ']'
+ set +x
Shell debugging temporarily silenced: export LMOD_SH_DBG_ON=1 for this output (/usr/share/lmod/lmod/init/bash)
Shell debugging restarted
+ unset __lmod_vx
+ set -eu
+ req_cf_v=15
+ '[' '!' -f .clang-format ']'
+ '[' '!' -f ../.clang-format ']'
+ '[' -z ']'
+ fmt=clang-format
+ clang-format --version
++ clang-format --version
+ clang_version_full='clang-format version 15.0.6 (https://github.com/ssciwr/clang-format-wheel a427ab6a8d0bbc93a80b82a99afe9996c06377b4)'
++ echo 'clang-format version 15.0.6 (https://github.com/ssciwr/clang-format-wheel a427ab6a8d0bbc93a80b82a99afe9996c06377b4)'
++ cut -f3 '-d '
++ cut -f1 -d.
+ clang_version=15
+ '[' 15 -lt 15 ']'
+ case "$(uname)" in
++ uname
+ find=find
+ '[' 0 -eq 1 ']'
+ dir=.
+ find . -type f -iregex '.*\.(cpp|hpp|c|h)' -exec clang-format -i '{}' +

git diff
# ... no changes.

# checking only find:
find . -type f -iregex '.*\.(cpp|hpp|c|h)'
# ... no result.

find --version
find (GNU findutils) 4.9.0
```

It seems the regex needs another tweak for (Linux) GNU findutils.

nilason:
> It seems the regex needs another tweak for (Linux) GNU findutils.

Seems [GNU find uses emacs regex](https://www.gnu.org/software/findutils/manual/html_node/find_html/Regular-Expressions.html) by default. Please try:

```bash
find . -type f -regextype posix-extended -iregex '.*\.(cpp|hpp|c|h)' 
```

utils/grass_clang_format.sh
nilason: You're right. I made a logical mismatch, figured this to be good for rebase/merging old PRs to main, but then you need this script too.... ahhhh. Will update to check only in source dir.

neteler:
Yes, the modified version appears to work with GNU find:

```shell
find . -type f -regextype posix-extended -iregex '.*\.(cpp|hpp|c|h)'  | wc -l
3357
```

nilason:
> Yes, the modified version appears to work with GNU find:
> 
> ```shell
> find . -type f -regextype posix-extended -iregex '.*\.(cpp|hpp|c|h)'  | wc -l
> 3357
> ```

Good! Updated accordingly.

neteler:
Almost there... (GNU find here):

```shell
sh -x grass_clang_format.sh db
+ set -eu
...
+ find db -type f '-regextype posix-extended -iregex' '.*\.(cpp|hpp|c|h)' -exec clang-format -i '{}' +
find: unknown predicate `-regextype posix-extended -iregex'

```

utils/grass_clang_format.sh
```diff
@@ -0,0 +1,86 @@
...
+esac
+
+# One argument, a directory path is allowed
+if [ "$#" -eq 1 ] && [ -d "$1" ]; then
+    dir="$1"
+else
+    dir="."
+fi
+
+${find} "${dir}" -type f "${regex}" '.*\.(cpp|hpp|c|h)' -exec \
```
neteler, line 85:
```suggestion
${find} "${dir}" -type f ${regex} '.*\.(cpp|hpp|c|h)' -exec \
```

neteler:
With un-quoting `${regex}` it works for me.

nilason:
> With un-quoting `${regex}` it works for me.

I have the same problems using `zsh -x ./utils/grass_clang_format.sh`, but it works fine running just `./utils/grass_clang_format.sh`.

Unquoting $regex causes shellcheck to complain:
```bash
shellcheck ./utils/grass_clang_format.sh

In ./utils/grass_clang_format.sh line 85:
${find} "${dir}" -type f ${regex} '.*\.(cpp|hpp|c|h)' -exec \
                         ^------^ SC2086 (info): Double quote to prevent globbing and word splitting.

```

neteler:
> I have the same problems using `zsh -x ./utils/grass_clang_format.sh`, but it works fine running just `./utils/grass_clang_format.sh`.

Here it doesn't:

```
utils/grass_clang_format.sh db
find: unknown predicate `-regextype posix-extended -iregex'
```

nilason:
> > I have the same problems using `zsh -x ./utils/grass_clang_format.sh`, but it works fine running just `./utils/grass_clang_format.sh`.
> 
> Here it doesn't:
> 
> ```
> utils/grass_clang_format.sh db
> find: unknown predicate `-regextype posix-extended -iregex'
> ```

Always these minor differences. Let's use standard find and skip regular expression. Probably the safest way.
Updated.

neteler:
Great, now it works also for me!

Just a glitch I spotted (shall I open a new issue?) - the header formatting is a bit distorted. This happens in the cmake related PR which I am trying to update:

```
diff --git a/db/db.columns/main.c b/db/db.columns/main.c
index 3cdda5384e..9e7400462b 100644
--- a/db/db.columns/main.c
+++ b/db/db.columns/main.c
@@ -3,9 +3,9 @@
  *
  * MODULE:       db.columns
  * AUTHOR(S):    Radim Blazek <radim.blazek gmail.com> (original contributor)
- *               Glynn Clements <glynn gclements.plus.com>, Markus Neteler <neteler itc.it>
- * PURPOSE:      list the column names for a table
- * COPYRIGHT:    (C) 2002-2006 by the GRASS Development Team
+ *               Glynn Clements <glynn gclements.plus.com>, Markus Neteler
+ *<neteler itc.it> PURPOSE:      list the column names for a table COPYRIGHT:
+ *(C) 2002-2006 by the GRASS Development Team
  *
  *               This program is free software under the GNU General Public
  *               License (>=v2). Read the file COPYING that comes with GRASS
@@ -19,17 +19,13 @@
 #include <grass/dbmi.h>
 #include <grass/glocale.h>
```

nilason:
> Great, now it works also for me!

That's good news!

> 
> Just a glitch I spotted (shall I open a new issue?) - the header formatting is a bit distorted. This happens in the cmake related PR which I am trying to update:
> 
> ```
> diff --git a/db/db.columns/main.c b/db/db.columns/main.c
> ...

I went through all files manually (in addition to clang-format) in the big formatting commits, removing remaining tabs in comments and fixed the above mentioned headers (hopefully all) where it was needed. Because of this, `clang-format` isn't always enough, for some old PRs.

If there are no changes to the header in the PR, then you should leave this hunk as is (as it is good in main).
(Out of the hat, I don't know a git-y way to do that, but you can always manually copy that part from the main version before merging/rebasing.)

nilason:
> > Great, now it works also for me!
> 
> That's good news!
> 
> > Just a glitch I spotted (shall I open a new issue?) - the header formatting is a bit distorted. This happens in the cmake related PR which I am trying to update:
> > ```
> > diff --git a/db/db.columns/main.c b/db/db.columns/main.c
> > ...
> > ```
> 
> I went through all files manually (in addition to clang-format) in the big formatting commits, removing remaining tabs in comments and fixed the above mentioned headers (hopefully all) where it was needed. Because of this, `clang-format` isn't always enough, for some old PRs.
> 
> If there are no changes to the header in the PR, then you should leave this hunk as is (as it is good in main). (Out of the hat, I don't know a git-y way to do that, but you can always manually copy that part from the main version before merging/rebasing.)

I should add: It is best to only apply clang-format to the files in a particular PR that are modified (and not to the whole source dir).

neteler:
> I should add: It is best to only apply clang-format to the files in a particular PR that are modified (and not to the whole source dir).

Good point. 

BTW: This would be a trick to easily **list the files changed in a PR** (source: https://stackoverflow.com/a/68682405/452464):

```
# list files changed in PR 2681
gh pr view 2681 --json files --jq '.files.[].path'
```

Only those then would undergo the `clang-format -i <file>` beautification.

wenzeslaus:
Is this still useful with pre-commit? (Just trying to clean up the 8.3.0 milestone.)

nilason:
As long as pre-commit is not a requirement (and/or we use [pre-commit-ci](https://pre-commit.ci)), then this might be useful for those who for whatever reason do not want to install pre-commit (and simplify their lives :). My intention was to replace the now obsolete grass_indent.sh.
I have no strong opinion on this, either we bin it or merge it now, no need to bump milestone on this.

wenzeslaus:
Okay, let's make this available. We want to make it easy, so an alternative is good. Can you please delete the old indent files?

wenzeslaus (approved):

neteler:
> Is this still useful with pre-commit? (Just trying to clean up the 8.3.0 milestone.)

It is useful since pre-commit is, as far as I can see, yet heavily underused (also note to self).

wenzeslaus:
You can run `pre-commit run clang-format --all-files` instead of the script and you need to install something in both cases, but if you Python environment is not in perfect order, installing pre-commit will can be harder than installing clang-format. That's what the slight advantage of a script is. Neither will force people to use that. That's what only CI can do.

PR #2735, opened 2023-01-05
diglib: fix integer overflow

metzm:
This PR fixes an integer overflow when writing out spatial index files (sidx) larger than 2GB. With the integer overflow, the resultant file size is underestimated, a wrong size for writing out file offsets is used and the spatial index file becomes unreadable. Other potentially large vector files (coordinates, topology, category index) are not affected.

### DIFF
--- a/lib/vector/diglib/spindex_rw.c
+++ b/lib/vector/diglib/spindex_rw.c
@@ -64,10 +64,10 @@ int dig_Wr_spidx_head(struct gvfile *fp, struct Plus_head *ptr)
     /* use ptr->off_t_size = 4 if possible */
     if (sizeof(off_t) > 4) {
         size = 145; /* max header size, see below */
-        size += ptr->Node_spidx->n_nodes * ptr->Node_spidx->nodesize;
-        size += ptr->Line_spidx->n_nodes * ptr->Line_spidx->nodesize;
-        size += ptr->Area_spidx->n_nodes * ptr->Area_spidx->nodesize;
-        size += ptr->Isle_spidx->n_nodes * ptr->Isle_spidx->nodesize;
+        size += (size_t)ptr->Node_spidx->n_nodes * ptr->Node_spidx->nodesize;
+        size += (size_t)ptr->Line_spidx->n_nodes * ptr->Line_spidx->nodesize;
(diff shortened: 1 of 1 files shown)


nilason (approved):
Looks good to me!

FYI, this is listed among another 100+ similar overflow issues in the CodeQL CI check.
Out of curiosity: I assume you hit upon this as a bug (segfault or similar) running some code, was it difficult to the track this down?

metzm:
> FYI, this is listed among another 100+ similar overflow issues in the CodeQL CI check.

An integer overflow can happen whenever two integer values are added, subtracted, or multiplied, right? It is practically impossible to cast the operation to the next larger integer type. In GRASS, we should do this for the number of raster cells = rows * columns.

> Out of curiosity: I assume you hit upon this as a bug (segfault or similar) running some code, was it difficult to the track this down?

The error message came from `G_fseek()`, reason being an invalid offset argument. Once I got a sample vector to reproduce this error, it was relatively easy to trace down the bug.

nilason:
> > FYI, this is listed among another 100+ similar overflow issues in the CodeQL CI check.
> 
> An integer overflow can happen whenever two integer values are added, subtracted, or multiplied, right? It is practically impossible to cast the operation to the next larger integer type. In GRASS, we should do this for the number of raster cells = rows * columns.

The issues reported by CodeQL are arithmetics (multiplication) on smaller types which result is stored to a larger type. This reflects the programmers expectations, e.g.:

```c
size_t size;
int a, b, c;

size = a * b; // in this case multiplication precedes the (implicit) cast and its result may be overflowed

size = (size_t)a * b; // the cast precedes the multiplication, no problems

c = a * b; // the programmer don't expect larger result, no problems

```
This is why the CodeQL issues *only* are numbered in the hundreds (and not thousands).
> 
> > Out of curiosity: I assume you hit upon this as a bug (segfault or similar) running some code, was it difficult to the track this down?
> 
> The error message came from `G_fseek()`, reason being an invalid offset argument. Once I got a sample vector to reproduce this error, it was relatively easy to trace down the bug.

Good to hear. Once I, or someone else, manage to fix those relatively simply fixed CodeQL issues, another potential similar bug to this will be avoided.

PR #2746, opened 2023-01-13
wxGUI Console: Button reorganization

lindakladivova:
This PR solves two things:
* reduces the number of Console buttons by removing the Clear prompt button
* removes bounding boxes around buttons
* moves buttons to the bottom of the Console pane

Before:
Screenshot from 2023-01-13 16-00-36

The main problem was that the right part of the button line could be cut off in Single-Window GUI (simply by shrinking the side Console pane):
Screenshot from 2023-01-13 16-00-58

After:
Screenshot from 2023-01-13 15-45-19
The suggestion is very similar to the solution in other panes (Tools, Python).
Follows the suggestion in https://github.com/OSGeo/grass/issues/2585#issuecomment-1333143997.

### DIFF
--- a/gui/wxpython/gui_core/goutput.py
+++ b/gui/wxpython/gui_core/goutput.py
@@ -171,7 +171,6 @@ class GConsoleWindow(wx.SplitterWindow):
                 flag=wx.EXPAND | wx.LEFT | wx.RIGHT | wx.TOP,
                 border=3,
             )
-
             helpText = StaticText(
                 self.panelPrompt,
                 id=wx.ID_ANY,
--- a/gui/wxpython/gui_core/goutput.py
+++ b/gui/wxpython/gui_core/goutput.py
@@ -43,7 +43,7 @@ from core.gconsole import (
 )
 from core.globalvar import CheckWxVersion, wxPythonPhoenix
 from gui_core.prompt import GPromptSTC
-from gui_core.wrap import Button, ClearButton, StaticText, StaticBox
+from gui_core.wrap import Button, ClearButton, StaticText
 from core.settings import UserSettings
 
 
@@ -135,36 +135,22 @@ class GConsoleWindow(wx.SplitterWindow):
         if not self._gcstyle & GC_PROMPT:
             self.cmdPrompt.Hide()
 
-        if self._gcstyle & GC_PROMPT:
-            cmdLabel = _("Command prompt")
-            self.outputBox = StaticBox(
-                parent=self.panelOutput, id=wx.ID_ANY, label=" %s " % _("Output window")
-            )
-
-            self.cmdBox = StaticBox(
-                parent=self.panelOutput, id=wx.ID_ANY, label=" %s " % cmdLabel
-            )
-
         # buttons
-        self.btnOutputClear = ClearButton(parent=self.panelOutput)
+        self.btnOutputClear = ClearButton(parent=self.panelPrompt)
         self.btnOutputClear.SetToolTip(_("Clear output window content"))
-        self.btnCmdClear = ClearButton(parent=self.panelOutput)
-        self.btnCmdClear.SetToolTip(_("Clear command prompt content"))
-        self.btnOutputSave = Button(parent=self.panelOutput, id=wx.ID_SAVE)
+        self.btnOutputSave = Button(parent=self.panelPrompt, id=wx.ID_SAVE)
         self.btnOutputSave.SetToolTip(_("Save output window content to the file"))
         self.btnCmdAbort = Button(parent=self.panelProgress, id=wx.ID_STOP)
         self.btnCmdAbort.SetToolTip(_("Abort running command"))
-        self.btnCmdExportHistory = Button(parent=self.panelOutput, id=wx.ID_ANY)
+        self.btnCmdExportHistory = Button(parent=self.panelPrompt, id=wx.ID_ANY)
         self.btnCmdExportHistory.SetLabel(_("&Export history"))
         self.btnCmdExportHistory.SetToolTip(
             _("Export history of executed commands to a file")
         )
 
         if not self._gcstyle & GC_PROMPT:
-            self.btnCmdClear.Hide()
             self.btnCmdExportHistory.Hide()
 
-        self.btnCmdClear.Bind(wx.EVT_BUTTON, self.cmdPrompt.OnCmdErase)
         self.btnOutputClear.Bind(wx.EVT_BUTTON, self.OnOutputClear)
         self.btnOutputSave.Bind(wx.EVT_BUTTON, self.OnOutputSave)
         self.btnCmdAbort.Bind(wx.EVT_BUTTON, self._gconsole.OnCmdAbort)
@@ -176,13 +162,6 @@ class GConsoleWindow(wx.SplitterWindow):
         """Do layout"""
         self.outputSizer = wx.BoxSizer(wx.VERTICAL)
         progressSizer = wx.BoxSizer(wx.HORIZONTAL)
-        btnSizer = wx.BoxSizer(wx.HORIZONTAL)
-        if self._gcstyle & GC_PROMPT:
-            outBtnSizer = wx.StaticBoxSizer(self.outputBox, wx.HORIZONTAL)
-            cmdBtnSizer = wx.StaticBoxSizer(self.cmdBox, wx.HORIZONTAL)
-        else:
-            outBtnSizer = wx.BoxSizer(wx.HORIZONTAL)
-            cmdBtnSizer = wx.BoxSizer(wx.HORIZONTAL)
 
         if self._gcstyle & GC_PROMPT:
             promptSizer = wx.BoxSizer(wx.VERTICAL)
@@ -192,6 +171,7 @@ class GConsoleWindow(wx.SplitterWindow):
                 flag=wx.EXPAND | wx.LEFT | wx.RIGHT | wx.TOP,
                 border=3,
             )
+
             helpText = StaticText(
                 self.panelPrompt,
                 id=wx.ID_ANY,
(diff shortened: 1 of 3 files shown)


lindakladivova:
I am thinking about the suggestion and I still do not like the way how buttons are organized. All buttons are actually more related to Console Output than to Command Prompt. We could probably preserve Clear button for Command Prompt and put two buttons Save and Clear at the bottom of Console Output and two buttons Export history and Clear to the bottom of Command prompt.. So we would clearly define what button belongs to which widget. Any ideas or disagreeing comments on this? :-)

veroandreo:
> I am thinking about the suggestion and I still do not like the way how buttons are organized. All buttons are actually more related to Console Output than to Command Prompt. We could probably preserve Clear button for Command Prompt and put two buttons Save and Clear to the bottom of Console Output and two buttons Export history and Clear to the bottom of Command prompt.. So we would clearly present what button belongs to which widget.
> 
> Any ideas or disagreeing comments on this? :-)

I like the idea of putting the buttons where they belong to ;)

lindakladivova:
> > I am thinking about the suggestion and I still do not like the way how buttons are organized. All buttons are actually more related to Console Output than to Command Prompt. We could probably preserve Clear button for Command Prompt and put two buttons Save and Clear to the bottom of Console Output and two buttons Export history and Clear to the bottom of Command prompt.. So we would clearly present what button belongs to which widget.
> > Any ideas or disagreeing comments on this? :-)
> 
> I like the idea of putting the buttons where they belong to ;)

Good. I implemented it according to the last suggestion. If I make the right pane really narrow it looks as follows:
Screenshot from 2023-01-16 21-30-40
I think it elegantly solves the problem with button "cut off" in Single-Window GUI.

@petrasovaa, do you have any suggestions on this?

petrasovaa:
> Good. I implemented it according to the last suggestion. If I make the right pane really narrow it looks as follows: Screenshot from 2023-01-16 21-30-40 I think it elegantly solves the problem with button "cut off" in Single-Window GUI.
> 
> @petrasovaa, do you have any suggestions on this?

I don't have a better solution right now. How are the buttons aligned? I would align them all probably to the right, but no strong opinion.

veroandreo:
> I agree with you Vero and Anna, aligning to the right side looks better: Screenshot from 2023-01-19 18-02-21 
Shall I adapt something more or could we merge this?

Now that buttons are placed where they belong, I wonder if the texts `Save output` and `Export history` are really needed... Wouldn't `Save` be enough in both cases? What do you think?

lindakladivova:
> > I agree with you Vero and Anna, aligning to the right side looks better: Screenshot from 2023-01-19 18-02-21
> > Shall I adapt something more or could we merge this?
> 
> Now that buttons are placed where they belong, I wonder if the texts `Save output` and `Export history` are really needed... Wouldn't `Save` be enough in both cases? What do you think?

Well, I think that `Export history` is probably more clear if we consider what it is really doing (especially when we changed the way how it behaves. please Vero look at https://github.com/OSGeo/grass/pull/2682)... and generally speaking, I do not see the reason why not write a bit longer description. It definitely rather helps more than harms. :-) Of course, it would be a different case if we had little space for that but it is not the case now... Just my opinion...

wenzeslaus:
What about considering the whole thing one larger widget? Then there would be one row with three buttons in the bottom:

```
------------------------------------
...output...
------------------------------------
...input...
------------------------------------
Save output | Export history | Clear
------------------------------------
```

Clear would clear both input and output. 

This suggestion avoids two Clear buttons and two rows of buttons.

petrasovaa:
> What about considering the whole thing one larger widget? Then there would be one row with three buttons in the bottom:
> 
> ```
> ------------------------------------
> ...output...
> ------------------------------------
> ...input...
> ------------------------------------
> Save output | Export history | Clear
> ------------------------------------
> ```
> 
> Clear would clear both input and output.
> 
> This suggestion avoids two Clear buttons and two rows of buttons.

I agree, I think the interface will look better.

lindakladivova:
> This PR solves two things:
> 
>     * reduces the number of Console buttons by removing the Clear prompt button
> 
>     * removes bounding boxes around buttons
> 
>     * moves buttons to the bottom of the Console pane
> 
> 
> Before: Screenshot from 2023-01-13 16-00-36
> 
> The main problem was that the right part of the button line could be cut off in Single-Window GUI (simply by shrinking the side Console pane): Screenshot from 2023-01-13 16-00-58
> 
> After: Screenshot from 2023-01-13 15-45-19 The suggestion is very similar to the solution in other panes (Tools, Python). Follows the suggestion in [#2585 (comment)](https://github.com/OSGeo/grass/issues/2585#issuecomment-1333143997).

Well... The layout you are proposing is basically very similar to what I implemented at the beginning of this PR.. :-D I like the idea of merging Clear button into one Clear button but I still think that we do not solve the problem of lack of space... The button widget is too wide and it will not look good with the default settings of the right panel width... but I can go back in the commit history and implement your proposal and we can then decide what is better...

petrasovaa:
> Well... The layout you are proposing is basically very similar to what I implemented at the beginning of this PR.. :-D I like the idea of merging Clear button into one Clear button but I still think that we do not solve the problem of lack of space... The button widget is too wide and it will not look good with the default settings of the right panel width... but I can go back in the commit history and implement your proposal and we can then decide what is better...

Regarding the width, the output of r.info etc. is anyway rather wide, so you end up extending it anyway. And this is still an improvement comparing to the current state. 

Unrelated, it would be nice to make the console space little taller (less narrow), right now the space for typing is rather narrow and long commands don't fit.

lindakladivova:
> > Well... The layout you are proposing is basically very similar to what I implemented at the beginning of this PR.. :-D I like the idea of merging Clear button into one Clear button but I still think that we do not solve the problem of lack of space... The button widget is too wide and it will not look good with the default settings of the right panel width... but I can go back in the commit history and implement your proposal and we can then decide what is better...
> 
> Regarding the width, the output of r.info etc. is anyway rather wide, so you end up extending it anyway. And this is still an improvement comparing to the current state.
> 
> Unrelated, it would be nice to make the console space little taller (less narrow), right now the space for typing is rather narrow and long commands don't fit.

Do you mean the whole right pane make a bit wider or just put a splitter inside the Console pane higher (so that the space for Prompt would be taller)? I made the minimum height of the Prompt greater:

_self.SetMinimumPaneSize(self.btnCmdClear.GetSize()[1] + 100)_

instead of 

_self.SetMinimumPaneSize(self.btnCmdClear.GetSize()[1] + 25)_

lindakladivova:
I would suggest making this change:

         # split window
         if self._gcstyle & GC_PROMPT:
-            self.SplitHorizontally(self.panelOutput, self.panelPrompt, -50)
+            self.SplitHorizontally(self.panelOutput, self.panelPrompt, 0)
         else:
-            self.SplitHorizontally(self.panelOutput, self.panelPrompt, -45)
+            self.SplitHorizontally(self.panelOutput, self.panelPrompt, 0)
             self.Unsplit()
         self.SetMinimumPaneSize(self.btnClear.GetSize()[1] + 100)
 
-        self.SetSashGravity(1.0)
+        self.SetSashGravity(0.5)

It makes the Sash position looser and the size is determined by the size of the inner two widgets. With these settings, the Console Prompt takes about one third of the Console size. What do you think @petrasovaa ?

gui/wxpython/gui_core/goutput.py
```diff
@@ -135,37 +135,23 @@ def __init__(
...
         # buttons
-        self.btnOutputClear = ClearButton(parent=self.panelOutput)
-        self.btnOutputClear.SetToolTip(_("Clear output window content"))
-        self.btnCmdClear = ClearButton(parent=self.panelOutput)
-        self.btnCmdClear.SetToolTip(_("Clear command prompt content"))
-        self.btnOutputSave = Button(parent=self.panelOutput, id=wx.ID_SAVE)
+        self.btnClear = ClearButton(parent=self.panelPrompt)
+        self.btnClear.SetToolTip(_("Clear content of output and command windows"))
+        self.btnOutputSave = Button(parent=self.panelPrompt, id=wx.ID_SAVE)
         self.btnOutputSave.SetToolTip(_("Save output window content to the file"))
```
veroandreo, line 142:
```suggestion
        self.btnOutputSave.SetToolTip(_("Save output window content to a file"))
```

gui/wxpython/gui_core/goutput.py
```diff
@@ -135,37 +135,23 @@ def __init__(
...
-            )
-
         # buttons
-        self.btnOutputClear = ClearButton(parent=self.panelOutput)
-        self.btnOutputClear.SetToolTip(_("Clear output window content"))
-        self.btnCmdClear = ClearButton(parent=self.panelOutput)
-        self.btnCmdClear.SetToolTip(_("Clear command prompt content"))
-        self.btnOutputSave = Button(parent=self.panelOutput, id=wx.ID_SAVE)
+        self.btnClear = ClearButton(parent=self.panelPrompt)
+        self.btnClear.SetToolTip(_("Clear content of output and command windows"))
```
petrasovaa, line 140:
```suggestion
        self.btnClear.SetToolTip(_("Clear prompt and output window"))
```

gui/wxpython/gui_core/goutput.py
petrasovaa: Or even simpler "Save output to a file"

gui/wxpython/gui_core/goutput.py
```diff
@@ -274,14 +226,12 @@ def _layout(self):
...
-        else:
-            self.SplitHorizontally(self.panelOutput, self.panelPrompt, -45)
+        self.SplitHorizontally(self.panelOutput, self.panelPrompt, 0)
+        if not (self._gcstyle & GC_PROMPT):
             self.Unsplit()
-        self.SetMinimumPaneSize(self.btnCmdClear.GetSize()[1] + 25)
+        self.SetMinimumPaneSize(self.btnClear.GetSize()[1] + 100)
 
-        self.SetSashGravity(1.0)
+        self.SetSashGravity(0.5)
```
petrasovaa, line 234: I think 1.0 is better behavior.

petrasovaa (approved):

PR #2749, opened 2023-01-16
configure: remove pdal cppflags

nilason:
GRASS configure now fails with PDAL 2.5, see current CI Mac builds. Upstream pdal-config is outdated returning `-std=c++11`, while C++17 is in fact needed.

From https://github.com/PDAL/PDAL/blob/master/RELEASENOTES.txt under "2.4.0":
> A compiler that supports C++17 is now required to build PDAL or include PDAL headers in your code.

This is the least invasive fix for this I could think of, and may be backported too.

### DIFF
--- a/configure
+++ b/configure
@@ -9400,7 +9400,6 @@ fi
 
   if test "$PDAL_CONFIG" != "" ; then
     PDAL_LIBS=`"$PDAL_CONFIG" --libs`
-    PDAL_CPPFLAGS=`"$PDAL_CONFIG" --cxxflags`
     PDAL_INC=`"$PDAL_CONFIG" --includes`
     USE_PDAL=1
   fi
@@ -9408,10 +9407,8 @@ fi
   PDAL=
   ac_save_libs="$LIBS"
   ac_save_cflags="$CFLAGS"
-  ac_save_cppflags="$CPPFLAGS"
   LIBS="$LIBS $PDAL_LIBS"
   CFLAGS="$CFLAGS $PDAL_CFLAGS"
-  CPPFLAGS="$CPPFLAGS $PDAL_CPPFLAGS $PDAL_INC"
 
 cat confdefs.h - <<_ACEOF >conftest.$ac_ext
 /* end confdefs.h.  */
@@ -9458,7 +9455,6 @@ rm -f core conftest.err conftest.$ac_objext \
     conftest$ac_exeext conftest.$ac_ext
   LIBS=${ac_save_libs}
   CFLAGS=${ac_save_cflags}
-  CPPFLAGS=${ac_save_cppflags}
 
 
 $as_echo "#define HAVE_PDAL 1" >>confdefs.h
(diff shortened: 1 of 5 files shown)


neteler:
Thanks, this PR seems to fix the compilation error on Fedora 38, with PDAL 2.5.0 (reported in #2750).

nilason:
As a general note, compiler support for C++17 has been strong since its introduction, starting with GCC 8, Clang 5 and MSVC 19.15. The compiler flag for standard supported is needed for *building* PDAL, not necessary for linking to it (indirectly, the existence of PDAL 2.4+ on a system implies a compiler supporting C++17) .

Note: the PDAL [pkg-config](https://github.com/PDAL/PDAL/blob/master/apps/pdal.pc.in) file does *not* have this flag set at all.

I take the liberty to merge this now.

nilason:
> I take the liberty to merge this now.

Second thought, I hold on to this a bit more, removing the C++ standard flag and PDAL_CPPFLAGS implies using the C++ GRASS is configured to may lead to issues:
- if GRASS is explicitly configured to `std=c++11` (which is the present official standard for GRASS), while using PDAL 2.4+ (which needs std=c++17), building will fail.

If the above situation is unacceptable, this PR needs to be reworked.

It is, however, still possible to build GRASS in the above mentioned example, by either excluding the `std=c++11` flag or by adding GRASS configure flag `std=c++17` if the PDAL version used is 2.4+.

Any thoughts on this?

neteler:
@manisandro: do you have any suggestion here?

manisandro:
As I see it I would set the grass c++ language standard based on the highest required standard of any dependency. What one could do here is require c++11 for PDAL < 2.5.0 and c++17 for PDAL >= 2.5.0.

nilason:
Being C++11 standard requirement for GRASS, means --as I see it -- that there are no guarantees, nor even probable, it will build with pre-c++11 (e.g. c++98). There is no requirement compiling with explicitly stating 'std=-c++11'. In this sense, using the present state of this PR there will be two options, very much in line with @manisandro stated:

building GRASS with:
- PDAL < 2.4
  it is possible, but not required to use the flag `CXXFLAGS="std=-c++11"` for GRASS configure
- PDAL >=2.4
   it is possible, but not required to use the flag `CXXFLAGS="std=-c++17"` for GRASS configure

It is always possible to make a PDAL version check in configure, but unless absolutely necessary, I wouldn't choose that road. I'd vote for current state of the PR, but I wanted to bring this issue to the table.

neteler:
> I'd vote for current state of the PR, but I wanted to bring this issue to the table.

Given the option to use the flag `CXXFLAGS="std=-c++17"` for GRASS `configure` I suggest to merge and backport this PR.

nilason:
> > I'd vote for current state of the PR, but I wanted to bring this issue to the table.
> 
> Given the option to use the flag `CXXFLAGS="std=-c++17"` for GRASS `configure` I suggest to merge and backport this PR.

I agree, merged. On newer systems with PDAL 2.5, it is likely not even necessary to add the flag for GRASS. Will work anyway.

nilason:
Backport to 8.2 in separate PR #2757

PR #2758, opened 2023-01-18
Add r.in.pdal test for alpine image

anikaweinmann:
Add `r.in.pdal` test for alpine image so that it is executed in GHA.

Additional removement of `r.in.pdal` addon installation in Ubuntu and Debian docker images (replaced in C++ version in #1200).


neteler (approved):
Thanks, also for the cleanup.

PR #2759, opened 2023-01-18
Add pre-commit hook file

nilason:
This adds [pre-commit](https://pre-commit.com) hook file, which runs and checks black, flake8 and clang-format on staged files.

```bash
cd grass-source-dir
python -m pip install pre-commit
# following step once per repo
pre-commit install

# edit files ...

git add <modified_files>

# following step triggers the hook
git commit
```

This is by far the simplest way to keep the files in a good shape. Also it will save **a lot** of CI builds, with no need for correcting pushed and failed commits.

The black and flake8 part may be backported to 8.2.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,10 @@ repos:
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
+  - repo: https://github.com/igorshubovych/markdownlint-cli
+    rev: v0.33.0
+    hooks:
+      - id: markdownlint
   - repo: https://github.com/psf/black
     rev: 22.3.0
     hooks:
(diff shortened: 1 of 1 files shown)


nilason:
I accidentally hit the return key and posted this PR before I even started. Now I have updated the description.

nilason:
Just added `trailing-whitespace` and `end-of-file-fixer` hooks, which both check all types of files. I also added `markdownlint` hook.

nilason:
Any objections on merging this `.pre-commit-config.yaml` file?
I think it is a very convenient workflow, I find myself manually putting it in while working on other PRs.

Note: if `pre-commit` is **not** installed, this addition  will not in any way affect present workflow.

neteler:
> This is by far the simplest way to keep the files in a good shape. Also it will save **a lot** of CI builds, with no need for correcting pushed and failed commits.

This sounds very good! And it will speed up things.

Issue #2747, opened 2023-01-15
[Feat]  Fix extra compiler warnings


nilason:
Enabling `-Wextra` compiler flag in combination with `-Wall` causes a number of warnings of following types:

- [ ] -Wdeprecated-non-prototype #2727 #2798
- [x] -Wempty-body #2755
- [x] -Wextra #2754
- [x] -Wignored-qualifiers #2753
- [x] -Wimplicit-fallthrough= #2760
- [x] -Wliteral-range #2748
- [x] -Wmissing-field-initializers #2753
- [x] -Wsign-compare #2761
- [x] -Wsometimes-uninitialized #2751
- [x] -Wtype-limits #2756 #2799
- [x] -Wunused-but-set-variable #2752
- [x] -Wunused-parameter #2770

with GCC 9.4 and Clang 16.0.0

The issues are of various degree of seriousness, but addressing them would improve if only a little bit on code quality and keeping the code in good condition.
This is a follow-up to #2156.

nilason:
The GCC (gcc 11.3.0 + glibc 2.35) and Mac (clang 14.0.6 + MacOSX13.sdk) CI runners now compiles with
`-Wall -Wextra -Werror`.

There remain two code parts with `-Wdeprecated-non-prototype` warnings, which are issued with clang 16 (still not released). Those are non-trivial and I have prepared fixes (but not yet PRs). They are not urgent fixes and may be bumped to version 8.4.

Closing this as implemented.

wenzeslaus:
Cool! Having `-Wall -Wextra -Werror` is great!

PR #2760, opened 2023-01-18
Fix -Wimplicit-fallthrough warnings

nilason:
Fixes partly #2747.

### DIFF
--- a/lib/bitmap/bitmap.c
+++ b/lib/bitmap/bitmap.c
@@ -152,6 +152,7 @@ int BM_set_mode(int mode, int size)
     case BM_FLAT:
     case BM_SPARSE:
         Mode = mode;
+        break;
     default:
         fprintf(stderr, "BM_set_mode:  Unknown mode: %d\n", mode);
         ret--;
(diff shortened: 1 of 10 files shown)


nilason:
@metzm Could you please take a look at this?
There are some changes in logic, which I think is rather bugfixes, but I could be wrong.

lib/bitmap/bitmap.c
```diff
@@ -152,6 +152,7 @@ int BM_set_mode(int mode, int size)
     case BM_FLAT:
     case BM_SPARSE:
         Mode = mode;
+        break;
```
nilason, line 155: This changes the logic. However, the function `BM_set_mode()` is not used anywhere. No code to break (pun intended:).

lib/gmath/la.c
```diff
@@ -861,11 +861,13 @@ int G_matvect_extract_vector(mat_struct *mt, vtype vt, int indx)
     case RVEC: {
         mt->type = ROWVEC_;
         mt->v_indx = indx;
+        break;
     }
 
     case CVEC: {
         mt->type = COLVEC_;
         mt->v_indx = indx;
+        break;
```
nilason, line 870: These too changes the logic. Same situation here, `G_matvect_extract_vector()` is not used.

PR #2761, opened 2023-01-18
Fix -Wsign-compare warnings

nilason:
Fixes partly #2747.


nilason:
If there are no objections I will merge this in the next few days.

PR #2770, opened 2023-01-22
Fix '-Wunused-parameter' compiler warnings

nilason:
The absolute majority of the "-Wunused-parameter" warnings are addressed by adding `__attribute__((unused))` in the form of a macro, defined as following:

`#define UNUSED __attribute__((unused))`

In this sense, this PR is a trivial fix, in spite of the high number of affected files.

Fixes partly #2747


nilason:
If there are no objections, I'll merge this in a few days.


(CC @metzm )

PR #2785, opened 2023-01-24
Fix trailing whitespace and end-of-file errors

nilason:
This removes trailing whitespace and makes sure text files end with a newline and only a (one) newline.

This was done using pre-commits `trailing-whitespace` and `end-of-file-fixer` hooks.

Similar to https://github.com/OSGeo/grass-addons/pull/853

**Update:**

Now after #2787 I have expanded this PR a little and added another formatting with clang-format (for a few .json, .js and .m files) as well as one markdown file. I have also excluded some files from pre-commit formatting, which potentially may affect unit-tests.

GRASS source directory now passes:

```bash
pre-commit run trailing-whitespace --all-files
pre-commit run end-of-file-fixer --all-files
pre-commit run clang-format --all-files
pre-commit run markdownlint --all-files
```

After this PR there should be no "side effects" of using pre-commit on any part of GRASS source base. It is also advantageous to make this change in one commit, which can be added to `.git-blame-ignore-revs`.


nilason:
This now ready for review. I updated the PR description with latest modifications.

PR #2786, opened 2023-01-25
pre-commit: handle only c and c++ files

nilason:
The pre-commit hook 'clang-format' can handle a range of files (c++, c, c#, cuda, java, javascript, json, objective-c, proto). Presently there is only settings for C/C++ in `.clang-format`.

This PR makes pre-commit handle only C/C++ files.

PR #2787, opened 2023-01-25
pre-commit/ClangFormat: enable formatting all supported files

nilason:
This is an alternative, in my opinion better, to #2786 .

This adds formatting support for javascript, json and Objective-C files for both `clang-format` and `pre-commit`.

As there are only a few javascript, json and Objective-C files in source, ~~I included a formatted version of them here.~~

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,12 +1,15 @@
 ---
-Language:        Cpp
-BasedOnStyle:  LLVM
+BasedOnStyle: LLVM
+IndentWidth: 4
+
+---
+Language: Cpp
 AccessModifierOffset: -4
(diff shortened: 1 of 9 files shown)


nilason:
The formatted files are the result of

`pre-commit run clang-format --all-files`

We can always drop the formatting commit from this PR and they will be formatted next time they will be modified.
It was foremost added as illustration for the settings.

nilason:
@neteler Any thoughts on this one, or shall we go for #2786?

neteler:
I like this json support here :-)

neteler (approved):
Looks good (honestly, I cannot really judge the impact of this PR).

PR #2790, opened 2023-01-27
pre-commit: add yaml file lint

nilason:
This adds yaml linter https://github.com/adrienverge/yamllint ~~and formatter https://github.com/jumanjihouse/pre-commit-hook-yamlfmt~~ to `.pre-commit-config.yaml`.

All .yml files have also been fixed to pass without warning and formatted (for unified standard).

With CI's super-linter testing for yaml files, it is preferable to pre-empt any failures.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,7 +1,6 @@
 ---
 BasedOnStyle: LLVM
 IndentWidth: 4
-
 ---
 Language: Cpp
 AccessModifierOffset: -4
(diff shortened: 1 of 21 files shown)


nilason:
I dropped the `yamlfmt` formatter from .pre-commit-config.yaml. I wasn’t convinced of its result in particular with the line-width changes. The linter should be quite enough for us.

PR #2793, opened 2023-01-28
t.rast.univar: Add region_relation option for spatial filtering STDS by computational region

ninsbl:
This PR adds a `spatial_relation` filter to `t.rast.univar` that makes the module compute univariate statistis only for maps with the given spatial relation to the computational region.

It also fixes an issue in the temporal library.

This is intended to be a first of a series of PRs that add the spatial filter module to other relevant temporal modules. So, the PR is in a way a template and a careful review would be very much appreciated.

Tests are added for the new option.

Two questions in this regards:
1) If no maps are returned because of a given `where` clause or because the computational region is else where, should that really lead to a fatal error? Or is not a warning and no output more appropriate?
2) Is more documentation of the added option in the manual needed?


veroandreo:
I went through the changes in the test file, looks good to me :+1:

As per your questions: 
1. If the where returns nothing or the region is elsewhere do not seem like errors to me, so perhaps just a warning with a proper indication is enough - i.e., `the selection yielded nothing or the region is elsewhere, please check` or so.
2. Yes, definitely. However, as it will be something common to more temporal modules, perhaps we can include an explanation here: https://github.com/OSGeo/grass/blob/main/temporal/temporalintro.html

ninsbl:
> I went through the changes in the test file, looks good to me +1

Thanks, @veroandreo for looking into this and your feedback.

Will add something to the documentation and turn the error into a warning then.

ninsbl:
@veroandreo Now I added some documentation. Please have a look, if that is an OK basis for adding functionality to other relevant modules!?!

temporal/t.rast.univar/t.rast.univar.html
```diff
@@ -4,18 +4,32 @@ <h2>DESCRIPTION</h2>
...
 each zone (category) in that input raster map. The <em>zones</em> option
-does not support Spatio-Temporal-Raster-Datasets (STRDS) but only a single,
+does not support space time raster datasets (STRDS) but only a single,
 static raster map.
+<p>
+Space time raster datasets may contain raster maps with varying spatial
+extent like for example series of scenes of satellite images. With the
+<em>spatial_relation</em> option, computations can be limited to only
+maps of the space time raster dataset, that have a given spatial relation
+to the current computational region. Supported spatial relations are:
```
veroandreo, line 24:
```suggestion
<em>spatial_relation</em> option, computations can be limited to 
maps of the space time raster dataset that have a given spatial relation
to the current computational region. Supported spatial relations are:
```

temporal/temporalintro.html
```diff
@@ -160,6 +160,24 @@ <h4>Querying and map calculation</h4>
     <li><a href="t.rast.neighbors.html">t.rast.neighbors</a></li>
 </ul>
 
+<p>
+Similar to the <em>where</em> option, selected modules like
+<a href="t.rast.univar.html">t.rast.univar</a> support a
+<em>spatial_relation</em> option to limit computations to only maps of
+the space time raster dataset, that have a given spatial relation to the
```
veroandreo, line 167:
```suggestion
<em>spatial_relation</em> option to limit computations to maps of
the space time raster dataset that have a given spatial relation to the
```

veroandreo (approved):

wenzeslaus:
Is this meant for 8.3.0? Given that it is a first of a series of PRs, there might be some benefit in leaving this for 8.4.0.

wenzeslaus (changes requested):

wenzeslaus:
> 1. If no maps are returned...

I agree with @veroandreo that the it is simply an empty output, not an error. I would add that this usually the desired behavior when batch processing and automating. There is couple of these cases in the code. For db.univar, the new JSON output gives "nulls" while the original output formats fails ([PR code](https://github.com/OSGeo/grass/commit/120f198fca22a564334d7721fabbe8c6e299bf6b#diff-f6090a21aa8aa8cd6611287e6679168529a88d648b634d93f514d32a4bee5c22L176)). Notably, I decided not to include any warning at all. Text won't help in automation, but you can always test for it in the parsed JSON if needed. I think it can be argued that warning is not appropriate, it is an expected output given the inputs, but a verbose message might be useful when you did not expect the output and need to investigate.

wenzeslaus:
As for the concept, I think the functionality is useful, but the option name is unclear. Spatial relation to what? The underlying function uses _spatial relation_ but in combination with _spatial extent_. However, the new interface is always spatial relation to computational region extent. If your plan is to add spatial relation to other things besides the computational region extent, then the name makes sense.

I'm not sure why `-r` cannot be combined with `spatial_extent`. You can select the maps, but compute for the whole map, no? Perhaps the "ignore" in the description would have to be more nuanced (e.g., "Use the raster map regions...instead current region"). This would be more pronounced if the spatial relation is to some other things not only the region.

python/grass/temporal/univar_statistics.py
```diff
@@ -137,21 +144,35 @@ def print_gridded_dataset_univar_statistics(
...
+            warn += " or no maps with the requested spatial_relation to the computational region exist"
         if where:
-            err += " or where condition is wrong"
-        gs.fatal(
-            _(err) % {"sp": sp.get_new_map_instance(None).get_type(), "i": sp.get_id()}
+            warn += " or where condition does not return any maps"
+        gs.warning(
+            _(warn) % {"sp": sp.get_new_map_instance(None).get_type(), "i": sp.get_id()}
         )
+        return
```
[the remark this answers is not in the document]
ninsbl: Addressed in [cf49c4e](https://github.com/OSGeo/grass/pull/2793/commits/cf49c4ed47903c3d06bb2eab536ee4961f80da86)

python/grass/temporal/univar_statistics.py
```diff
@@ -137,21 +144,35 @@ def print_gridded_dataset_univar_statistics(
...
+        warn = "Space time %(sp)s dataset <%(i)s> is empty"
+        if spatial_relation:
+            warn += " or no maps with the requested spatial_relation to the computational region exist"
         if where:
-            err += " or where condition is wrong"
-        gs.fatal(
-            _(err) % {"sp": sp.get_new_map_instance(None).get_type(), "i": sp.get_id()}
+            warn += " or where condition does not return any maps"
+        gs.warning(
+            _(warn) % {"sp": sp.get_new_map_instance(None).get_type(), "i": sp.get_id()}
```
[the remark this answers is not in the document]
ninsbl: Addressed in [cf49c4e](https://github.com/OSGeo/grass/pull/2793/commits/cf49c4ed47903c3d06bb2eab536ee4961f80da86)

python/grass/temporal/abstract_space_time_dataset.py
```diff
@@ -1717,10 +1717,10 @@ def _update_where_statement_by_spatial_extent(
...
             if spatial_relation == "overlaps":
                 spatial_where_template += " AND top > {b}" " AND bottom < {t}"
-        elif spatial_relation == "is_contained":
-            spatial_where_template += " AND top <= {t}" " AND bottom >= {b}"
-        elif spatial_relation == "contains":
-            spatial_where_template += " AND top >= {t}" " AND bottom <= {b}"
+            elif spatial_relation == "is_contained":
+                spatial_where_template += " AND top <= {t}" " AND bottom >= {b}"
+            elif spatial_relation == "contains":
+                spatial_where_template += " AND top >= {t}" " AND bottom <= {b}"
```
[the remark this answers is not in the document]
ninsbl: Yes, the code would fail, but the library function is not yet used... Could indeed be separated out, maybe together with corrections / improvements of the library function doc....

ninsbl:
> > 1. If no maps are returned...
> 
> I agree with @veroandreo that the it is simply an empty output, not an error. (...). Notably, I decided not to include any warning at all. Text won't help in automation, but you can always test for it in the parsed JSON if needed. I think it can be argued that warning is not appropriate, it is an expected output given the inputs, but a verbose message might be useful when you did not expect the output and need to investigate.

Changed to verbose message...

ninsbl:
> As for the concept, I think the functionality is useful, but the option name is unclear. Spatial relation to what? The underlying function uses _spatial relation_ but in combination with _spatial extent_. However, the new interface is always spatial relation to computational region extent. If your plan is to add spatial relation to other things besides the computational region extent, then the name makes sense.

Renamed the module option to `region_relation`. Hope that is a bit more clear/precise. Any alternative suggestions are welcome...
 
> I'm not sure why `-r` cannot be combined with `spatial_extent`. You can select the maps, but compute for the whole map, no? Perhaps the "ignore" in the description would have to be more nuanced (e.g., "Use the raster map regions...instead current region"). This would be more pronounced if the spatial relation is to some other things not only the region.

Indeed, removed the rule....

ninsbl:
I am happy to separate out bug-fixes for 8.3 and keep the rest for 8.4, esp if a first 8.3 RC is around the corner. However, if 8.3 is on a more long term schedule, the alternative would be to speed up adding functionality to other relevant modules. As of now, there is no 8.3 branch yet, so the PR would practically be on hold until the 8.3 branch is created...

wenzeslaus:
@neteler can comment on the schedule for 8.3, but in last few days, number of open PRs against main was reduced from >80 to <40 (by merging or moving the milestone), so we are getting reading for branching out.

neteler:
> @neteler can comment on the schedule for 8.3, but in last few days, number of open PRs against main was reduced from >80 to <40 (by merging or moving the milestone), so we are getting reading for branching out.

Seems all blocker are gone. Will write to grass-dev asap as the discussion doesn't fit here in a PR.

python/grass/temporal/abstract_space_time_dataset.py
```diff
@@ -1369,15 +1369,22 @@ def get_registered_maps_as_objects_with_gaps(
         In case more map information are needed, use the select()
         method for each listed object.
 
+        The combination of the spatial_extent and spatial_relation parameters
+        can be used to return only map objects with the given spatial relation
+        to the provided spatial extent
```
wenzeslaus, line 1374:
```suggestion
        to the provided spatial extent.
```

And same below (3x).

temporal/temporalintro.html
```diff
@@ -160,6 +160,24 @@ <h4>Querying and map calculation</h4>
...
+ <li>"overlaps": process only maps that spatially overlap ("intersect")
+with the current computational region</li>
+ <li>"is_contained": process only maps that are fully within the current
+computational region</li>
+ <li>"contains": process only maps that contain (fully cover) the current
+computational region</li>
+</ul>
+Space time raster datasets may contain raster maps with varying spatial
+extent like for example series of scenes of satellite images. For such
+cases the <em>spatial_relation</em> option can be useful.
```
wenzeslaus, line 179: Now this should say region_relation, no?

wenzeslaus: Same for the module doc.

wenzeslaus (changes requested):

ninsbl:
Thanks for your code review effort, @wenzeslaus . Good catches. Should all be addressed now.

temporal/temporalintro.html
ninsbl: Addressed in [593da28](https://github.com/OSGeo/grass/pull/2793/commits/593da2826502023b56be96197b994ae394f7018a) and [403072a](https://github.com/OSGeo/grass/pull/2793/commits/403072a2adeb318c4ad1c3a0b288736bd73bb644)

python/grass/temporal/univar_statistics.py
```diff
@@ -159,19 +161,30 @@ def print_gridded_dataset_univar_statistics(
...
+                "Space time {type} dataset <{id}> is empty".format(
+                    type=sp.get_new_map_instance(None).get_type(), id=sp.get_id()
+                )
+            )
         )
+        if region_relation:
+            gs.verbose(
+                _(
+                    "or no maps with the requested spatial relation"
+                    "to the computational region exist"
```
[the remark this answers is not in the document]
ninsbl: Addressed, together with the following comment in: [52eab35](https://github.com/OSGeo/grass/pull/2793/commits/52eab354b63ea56ad90537d3e3db1e54425bb904)

wenzeslaus (approved):
Thank you for updating it!

ninsbl:
OK to merge? I so, I will merge and take it as a template for adding spatial filtering to other temporal modules...

PR #2807, opened 2023-02-06
Use 'PRId64' from C99, instead of macro PRI_OFF_T

### LINKED ISSUE #2806: Problem with i18n #2
The end of this line is not added to translation file.
In the end, the po-file contains only the larger chunk of this line:

"Coor file of vector <%s@%s> is shorter than it should be (%"

https://github.com/OSGeo/grass/blob/ce08116b890253ee5fc79ba6fda1e43bd8ee3355/locale/templates/grasslibs.pot#L5635

So, the end of this line ("%" PRI_OFF_T " bytes excess") will not be translated. Looks like this happens due to quotes around PRI_OFF_T.

https://github.com/OSGeo/grass/blob/ce08116b890253ee5fc79ba6fda1e43bd8ee3355/lib/vector/Vlib/open_nat.c#L159

nilason:
Replace the macro `PRI_OFF_T` with the use of `PRId64`, which is a macro defined in the C99 `<inttypes.h>` file.

Fixes #2806
which reports that gettext cannot handle locally defined macros, whereas

https://www.gnu.org/software/gettext/manual/gettext.html#Preparing-Strings :

> The gettext tools and library have special support for these <inttypes.h> macros. You can therefore simply write
> printf (gettext ("The amount is %0" PRId64 "\n"), number);
> The PO file will contain the string "The amount is %0<PRId64>\n". The translators will provide a translation
> containing "%0<PRId64>" as well, and at runtime the gettext function’s result will contain the appropriate
> constant string, "d" or "ld" or "lld".

This was already discussed in: https://github.com/OSGeo/grass/pull/1256#issuecomment-775499164

### DIFF
--- a/imagery/i.segment/iseg.h
+++ b/imagery/i.segment/iseg.h
@@ -21,10 +21,13 @@
 
 #ifdef HAVE_LONG_LONG_INT
 #define LARGEINT long long
+#define PRI_LONG "lld"
 #elif defined HAVE_LARGEFILES
 #define LARGEINT off_t
+#define PRI_LONG PRI_OFF_T
 #else
 #define LARGEINT long
+#define PRI_LONG "ld"
 #endif
 
 /* methods */
--- a/imagery/i.segment/iseg.h
+++ b/imagery/i.segment/iseg.h
@@ -11,6 +11,7 @@
  *
  *****************************************************************************/
 
+#include <inttypes.h>
 #include <grass/segment.h>
 #include <grass/imagery.h>
 #include "flag.h"
@@ -24,7 +25,7 @@
 #define PRI_LONG "lld"
 #elif defined HAVE_LARGEFILES
 #define LARGEINT off_t
-#define PRI_LONG PRI_OFF_T
+#define PRI_LONG PRId64
 #else
 #define LARGEINT long
 #define PRI_LONG "ld"
--- a/imagery/i.segment/iseg.h
+++ b/imagery/i.segment/iseg.h
@@ -21,13 +21,10 @@
 
 #ifdef HAVE_LONG_LONG_INT
 #define LARGEINT long long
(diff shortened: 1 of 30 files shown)


nilason:
`i.segment`, due to the definition of `LARGEINT long long`:

https://github.com/OSGeo/grass/blob/main/imagery/i.segment/iseg.h#L23

requires an alternative solution using an intermediate buffer
(circumventing direct use of PRId64 for gettext).

glibc, apparently only defines `PRId64` as `"lld"` on 32-bit systems, see e.g.:

https://codebrowser.dev/glibc/glibc/stdlib/inttypes.h.html#43
and
https://codebrowser.dev/glibc/glibc/sysdeps/x86/bits/wordsize.h.html#4

if I'm understand things correctly.

I have not tested what msg_id strings gettext actually make out of this, as I'm unfamiliar with that procedure.
I'm grateful if someone could test this!

nilason:
This is the present result on main:

```pot
#: ../lib/vector/Vlib/open_nat.c:156
msgid "Coor file of vector <%s@%s> is shorter than it should be (%"
msgstr ""
```

with this PR:

```pot
#: ../lib/vector/Vlib/open_nat.c:159
#, c-format
msgid ""
"Coor file of vector <%s@%s> is shorter than it should be (%<PRId64> bytes "
"missing)."
msgstr ""
```

nilason:
present (main):

```pot
#: ../imagery/i.segment/mean_shift.c:395
msgid "Changes > threshold: %"
msgstr ""
```

this PR:

```pot
#: ..//imagery/i.segment/mean_shift.c:399
#, c-format
msgid "Changes > threshold: %s, largest change: %g"
msgstr ""
```

nilason:
Merged to main, manual backport to 8.2 in #2807.

PR #2812, opened 2023-02-09
grass.pygrass: Convert all values to string, not just numbers

wenzeslaus:
pathlib.Path and decimal.Decimal should just work as is when calling a tool though pygrass. This converts the value to a string first, i.e., there is no type checking given that everything in Python coverts to a string.


echoix:
Is this still relevant?

PR #2817, opened 2023-02-09
grass.pygrass: Fix parameter range error reporting

wenzeslaus:
For a parameter value which was out of range from key_desc specified as `0-`, grass.pygrass failed to construct the error message and resulting error message for the exception was:

```
TypeError: must be real number, not NoneType
```

This handles the cases with only min or max specified separately, fixes the error message and uses f-strings. The new message is:

```
ValueError: The Parameter <size> must be higher than 1.0, 0 is out of range.
```


wenzeslaus:
This passes existing tests and adds new tests, merging.

PR #2771, opened 2023-01-22
docker: update of Ubuntu image

neteler:
- Debian docker base image update
- PDAL compilation: WITH_LASZIP=OFF, keep only WITH_LAZPERF=ON
- new versions:

| Base image   | GRASS GIS  | PROJ  | GDAL  | PDAL  | Python | Image size |
|--------------|------------|-------|-------|-------|--------|------------|
| Ubuntu 22.04 | 8.3.dev    | 8.2.1 | 3.4.1 | 2.4.3 | 3.10.6 | 2.92 GB    |

LAZ-perf is enabled in PDAL, hence both LAS and LAZ files can be imported into GRASS GIS.

TODO in a separate PR: reduce image size

PR #2820, opened 2023-02-11
docker: update of Ubuntu image

neteler:
(sync to #2771)

- Ubuntu docker base image update
- PDAL compilation: WITH_LASZIP=OFF, keep only WITH_LAZPERF=ON
- new versions:

| Base image   | GRASS GIS  | PROJ  | GDAL  | PDAL  | Python | Image size |
|--------------|------------|-------|-------|-------|--------|------------|
| Ubuntu 22.04 | 8.3.dev    | 8.2.1 | 3.4.1 | 2.4.3 | 3.10.6 | 2.92 GB    |

LAZ-perf is enabled in PDAL, hence both LAS and LAZ files can be imported into GRASS GIS.

TODO in a separate PR: reduce image size

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,11 +5,9 @@ LABEL maintainer="tawalika@mundialis.de,neteler@mundialis.de,weinmann@mundialis.
 
 ENV DEBIAN_FRONTEND noninteractive
 
-# define versions to be used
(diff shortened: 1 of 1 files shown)


wenzeslaus (approved):
That should do it. Thanks.

PR #2826, opened 2023-02-13
tests: Disable test of core.gcmd.recv_some

wenzeslaus:
Test of a known (unfixed) bug which is always failing the CI needs to be disabled in the config file.


nilason:
Mac test conf is here https://github.com/OSGeo/grass/blob/main/.github/workflows/macos_gunittest.cfg

wenzeslaus:
Thanks @nilason I forgot about that. Fixed.

nilason (approved):

PR #2721, opened 2022-12-29
A test case for bug #2720

marisn:
A test code exposing an error in core.gcmd.recv_some funcion.

PR #2827, opened 2023-02-14
Fix Python deprecations

nilason:
Fixes some Python deprecation warnings:

- `invalid escape sequence`
- `Menu.Append() is deprecated`

I came across these warnings and fixed them in November last year, I therefore can't remember versions involved, but I believe it was Python 3.10 and wxPython 4.2.


petrasovaa (approved):

nilason:
This apparently introduced some regression. I'm looking into it.

wenzeslaus:
To avoid cases like this and #2721 in the future, selected tests are now mandatory to pass before merging.

nilason:
> To avoid cases like this and #2721 in the future, selected tests are now mandatory to pass before merging.

If I only paid attention to the tests... Really sorry for this one, sloppy work from my side. #2830 will hopefully fix this.

PR #2828, opened 2023-02-14
g.extension: Remove six dependency

wenzeslaus:
* Removes imports from six, uses standard Python 3 imports.
* Removes Python <=2.4 ElementTree import.
* Removes future print_function imports.
* Removes six from g.extension and g.extension.all. Leaves the rest as is.

### DIFF
--- a/scripts/g.extension.all/g.extension.all.py
+++ b/scripts/g.extension.all/g.extension.all.py
@@ -35,19 +35,16 @@
 # % key: f
 # % label: Force operation (required for removal)
 # % end
-from __future__ import print_function
+
 import http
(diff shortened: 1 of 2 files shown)


wenzeslaus:
While python3-six package is in Travis CI, the build may fail as if there is no six package. Something like this happened before (possibly in Travis) and right now it happens for all runs.

This removes six from g.extension and g.extension.all, but not from anything else. This should allow the build to run without six.

petrasovaa (approved):

neteler:
Any objections to backport this PR to G8.2?

PR #2830, opened 2023-02-15
python/script: revert to byte string

nilason:
Fixes regression caused by #2827.


nilason:
> This makes sense since the input to convert_xml_to_utf8 is binary string from Popen.communicate, but I wonder what was the original reason for the change?

That was made some time ago, so I don't remember this one specifically. Probably I searched for regex patterns without raw literal strings and "happened" to change this one too.

petrasovaa (approved):

nilason:
Merged. Again, sorry for messing up!

PR #2832, opened 2023-02-16
wxGUI Profile: Don't repeat the first profile point (0,elev) in output

HuidaeCho:
This PR removes the first pair of repeating coordinates from output. It happens because the geometry type of transect line is "line" and the first "line" always starts with two identical start points.

### DIFF
--- a/gui/wxpython/wxplot/profile.py
+++ b/gui/wxpython/wxplot/profile.py
@@ -271,7 +271,9 @@ class ProfileFrame(BasePlotFrame):
             "r.profile",
             parent=self,
             input=raster,
-            coordinates=coords,
+            # since a transect "line" starts with two identical start points,
+            # don't repeat the first pair of coordinates in output
+            coordinates=",".join(coords.split(",")[2:]),
             resolution=transect_res,
             null="nan",
             quiet=True,


wenzeslaus:
Can you please add what is doing that just to be clear?

HuidaeCho:
> Can you please add what is doing that just to be clear?

Before this PR
image
r.profile command (not exact coordinates, just an example):
```bash
$ r.profile input=elevation coordinates=639014.08,222260.56,639014.08,222260.56,639542.25,223809.86   # 1st point given twice
Using resolution: 10 [meters]
Output columns:
Along track dist. [meters], Elevation
Approx. transect length: 0.000000 [meters]
 0.000000 109.053154     # <---- this output line should be removed because it's repeated below
Approx. transect length: 1636.854923 [meters]
 0.000000 109.053154     # you see this line again because the starting coordinates are given twice by the transect line
 10.000000 109.276176
 20.000000 108.157120
 30.000000 108.445526
...
```

After this PR, the first repeating row marked X is removed.
image
r.profile command:
```bash
$ r.profile input=elevation coordinates=639014.08,222260.56,639542.25,223809.86   # 1st point given only once
Using resolution: 10 [meters]
Output columns:
Along track dist. [meters], Elevation
Approx. transect length: 1636.854923 [meters]
 0.000000 109.053154     # no more repeating output line
 10.000000 109.276176
 20.000000 108.157120
 30.000000 108.445526
```

petrasovaa:
This solution is not ideal because it doesn't fix the actual source of the problem. It would give you wrong results when you click, hold mouse button down, move and release (as opposed to click and immediate release at the same pixel). In that case the point is not duplicate. I suggest this change:

```
--- a/gui/wxpython/mapwin/analysis.py
+++ b/gui/wxpython/mapwin/analysis.py
@@ -70,6 +70,11 @@ class AnalysisControllerBase:
 
         :param x,y: east north coordinates
         """
+        # avoid duplicating point in the beginning
+        begin = self._mapWindow.mouse["begin"]
+        end = self._mapWindow.mouse["end"]
+        if begin == end:
+            return
         # add new point and calculate distance
         item = self._registeredGraphics.GetItem(0)
         coords = item.GetCoords() + [[x, y]]
```

I tested it but it would be good to see if this works for others.

petrasovaa:
The problem described here was fixed in #2896.

PR #905, opened 2020-08-12
Add time function constants to temporal algebra

ninsbl:
This PR adds a parser rule for time functions that return a numerical constant for a map (like `start_doy`) in the temporal algebra. This allows to use e.g. the `start_doy` from the time stamp of a map as a constant in temporal algebra expressions in `t.rast.algebra`. 

Example:
`t.rast.algebra -d expression="R = if(A >= 3, start_doy(A, 0), 0)" basename="r" suffix="gran"`

Addresses #891

Issue #891, opened 2020-08-11
[Feat] t.rast.algebra: add time functions (start_doy, start_year, ...) that are currently unsupported in the spatial part of the expression


ninsbl:
**Describe the bug**
According to function documentation, [time functions](https://github.com/OSGeo/grass/blob/master/lib/python/temporal/temporal_algebra.py#L506) that return a float or integer (like e.g. start_doy, start_year) should be usable as constants in the spatial part of the expression. However, that does not seem to work, currently.
If e.g. one needs the day of the year as a numerical constant in t.rast.algebra (for being able to use it along with temporal selection and temporal windows) one has to run t.rast.algebra first and then t.rast.mapcalc. That causes unnecessary processing overhead.
The time delta (td) time function works partly, maybe because of this(?): https://github.com/OSGeo/grass/blob/master/lib/python/temporal/temporal_raster_base_algebra.py#L322
t.rast.algebra is a quite involved module and I do not fully get my head around how library functions are interrelated, so any help is highly appreciated....

**To Reproduce**

Using artificial maps from the testsuite, this
```
t.rast.algebra -d basename=r expression="R=start_doy(A)"
t.rast.algebra -d basename=r expression="R=A+start_doy(A)"
t.rast.algebra -d basename=r expression="R=if(A>12,start_doy(A),0)"
t.rast.algebra -d basename=r expression="R=td(A)"
t.rast.algebra -d basename=r expression="R=if(A>12,td(A),0)"
```
does not work (syntax error at e.g. token START_DOY), while this:

```
t.rast.algebra -d basename=r expression="R=A+td(A)"
t.rast.algebra -d basename=r expression="R=if(A>12,1,0)*td(A)"
t.rast.algebra -d basename=r expression="R=if(A>12,A+td(A),0)"
t.rast.algebra -d basename=r expression="R=if(A>12 && A[1]>12,A+td(A),0)"
```
does work and returns reasonable mapcalculator exressions.

**Expected behavior**
[Time functions](https://github.com/OSGeo/grass/blob/master/lib/python/temporal/temporal_algebra.py#L506) that return a float or integer (like e.g. start_doy, start_year) should be evaluated to numerical constants and included in the spatial part of the expression.

I would like to be able to do something like:
```
# Return day of the year when temperature (A) exceeds a certain value in three consecutive days
t.rast.algebra basename=r expression="R=if(start_month(A)<9, if(A>50 && A[1]>50 && A[2]>50, start_doy(A), 0))"
```
[...]

PR #2835, opened 2023-02-17
TGIS: Add time function constants to temporal algebra

ninsbl:
This PR adds a version of time functions returning numeric values for elements of a time series.
For this function type a neighborhood modifier is added that a) distinguishes it from the function used for temporal selection (for the lexical analyser) and b) adds more options for usage.

The function looks e.g. like this:
start_doy(A, -1)

Addresses #891

Finally, rebased #905, which caused a mess again. This PR is thus a clean replacement of #905.

Unsure about the milestone, though. Should be in mergable state and may go to 8.3, but given that it was on hold quite a bit, we can also wait for 8.4...


ninsbl:
OK, cleaned up #905 instead...

PR #2767, opened 2023-01-20
Replace sprintf() with snprintf() in C++ code

nilason:
Starting with Xcode 14.2 (and with the SDK following it), usage of sprintf() in C++ code issues a `-Wdeprecated-declarations` warning. This replaces it with the safer snprintf().

Fixes #2766

PR #2838, opened 2023-02-17
v.random: use unqualified names in db columns

### LINKED ISSUE #2473: [Bug] v.random doesn't strip mapset from column names
If a fully qualified vector map name is passed as a restrict parameter, it is used to create columns in target table in verbatim. As FQN is in form map@mapset and the value is not escaped or quoted, it causes an SQL error.
Instead of using plain param.input->answer for a column name, only the map part should be passed here.
https://github.com/OSGeo/grass/blob/da182541396a29288e8f23a66bd3a7150eb8fe4e/vector/v.random/main.c#L291

marisn:
GUI passes fully qualified map names as input parameters but SQL drivers do not properly escape passed values thus causing an SQL error on @ symbol.
Fixes #2473

### DIFF
--- a/vector/v.random/main.c
+++ b/vector/v.random/main.c
@@ -288,7 +288,7 @@ int main(int argc, char *argv[])
             ncols = db_get_table_number_of_columns(table);
             for (icol = 0; icol < ncols; icol++) {
                 col = db_get_table_column(table, icol);
-                sprintf(buf, ",%s_%s %s", xname,
+                snprintf(buf, DB_SQL_MAX, ",%s_%s %s", xname,
                         db_get_column_name(col),
                         db_sqltype_name(db_get_column_sqltype(col)));
                 db_append_string(&sql, buf);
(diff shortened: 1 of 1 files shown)


vector/v.random/main.c
```diff
@@ -286,7 +288,7 @@ int main(int argc, char *argv[])
             ncols = db_get_table_number_of_columns(table);
             for (icol = 0; icol < ncols; icol++) {
                 col = db_get_table_column(table, icol);
-                sprintf(buf, ",%s_%s %s", parm.input->answer,
+                sprintf(buf, ",%s_%s %s", xname,
```
nilason, line 291: Please change to snprintf() (see #2767) while you’re there. This will be an issue for C code too and a lot of work on GRASS source…
```suggestion
                snprintf(buf, DB_SQL_MAX, ",%s_%s %s", xname,
```

PR #2842, opened 2023-02-20
r.cost and r.walk: fix integer overflows

metzm:
`r.cost` and `r.walk` do not report correct percentages on the progress of the actual cost analysis for regions larger than INT_MAX (at most 2³¹ - 1) cells. Further on, memory allocation for the flag structure may fail.

Needs to be backported to all maintained GRASS 8 branches.


neteler:
(PR title renamed to make release note generator happy: https://github.com/OSGeo/grass/blob/main/utils/release.yml)

nilason (approved):

PR #2863, opened 2023-02-27
t.register: support mapset name in input file

ninsbl:
Before this PR, input files for registering maps with fully classified map names like here:
```
prec_1@$MAPSET|2001-01-01|2001-04-01|semantic_label
prec_2@$MAPSET|2001-04-01|2001-07-01|semantic_label
prec_3@$MAPSET|2001-07-01|2001-10-01|semantic_label
prec_4@$MAPSET|2001-10-01|2002-01-01|semantic_label
prec_5@$MAPSET|2002-01-01|2002-04-01|semantic_label
prec_6@$MAPSET|2002-04-01|2002-07-01|semantic_label
```
threw an error. Now, maps in the register file can be specified with a fully qualified map name (including the mapset name), which should be even faster than using unqualified map names, as existence check is skipped.


python/grass/temporal/register.py
```diff
@@ -226,7 +226,9 @@ def register_maps_in_space_time_dataset(
                 found = gscript.find_file(element=type, name=mapname)
                 if found["mapset"] is not None and len(found["mapset"]) > 0:
                     map_mapset = found["mapset"]
```
wenzeslaus, line 228: This practically says "if mapset was retrieved, use that else use the current one" is that correct? Perhaps add sentences to describe different states.

python/grass/temporal/register.py
```diff
@@ -226,7 +226,9 @@ def register_maps_in_space_time_dataset(
                 found = gscript.find_file(element=type, name=mapname)
                 if found["mapset"] is not None and len(found["mapset"]) > 0:
```
wenzeslaus, line 227: Since you are revising this part anyway, I suggest to change to be more readable:

1. `found -> result`
2. The test can be just `if found["mapset"]` in Python; that's what's written there.

python/grass/temporal/register.py
```diff
@@ -226,7 +226,9 @@ def register_maps_in_space_time_dataset(
...
                     map_mapset = found["mapset"]
-            row["id"] = AbstractMapDataset.build_id(mapname, map_mapset)
+                row["id"] = AbstractMapDataset.build_id(mapname, map_mapset)
+            else:
+                row["id"] = AbstractMapDataset.build_id(*mapname.split("@")[0:2])
```
wenzeslaus, line 231: We had a long discussion how to do this right. Perhaps a library function would be the best solution to cover the split or for the whole procedure in this block of code.

wenzeslaus (changes requested):
Thanks for updating the test. Can you please enable the test by moving the file to testsuite directory? (I did that recently in b59a04717bc6571117c7c71ed56325ae18a3b3e6 for r3.out.vtk.) There it would run in CI and fail it if the test fails. A different question is what the test actually tests automatically, but just testing that the modules don't fail is not bad for starters.

ninsbl:
> Thanks for updating the test. Can you please enable the test by moving the file to testsuite directory? (I did that recently in [b59a047](https://github.com/OSGeo/grass/commit/b59a04717bc6571117c7c71ed56325ae18a3b3e6) for r3.out.vtk.) There it would run in CI and fail it if the test fails. A different question is what the test actually tests automatically, but just testing that the modules don't fail is not bad for starters.

Sure. There are more test files though. And I am not sure if all tests are currently successful. I will have a look at them and evtl. port them to Python...

python/grass/temporal/register.py
ninsbl: Indeed, that makes me wonder does it make sense? If a map is not found on the current search path, should it then be registered as part of the current mapset? Isn`t a an error more appropriate then?

wenzeslaus: If it is not found by find_file, then it does not exist and that's it, no? ...but I don't know the full context here. A function with clearly defined purpose might be go good step if it is even needed here.

wenzeslaus:
Seems totally unrelated, but v.to.lines test failed for macOS, have we seen this before?

```
 Running ./scripts/v.to.lines/testsuite/test_v_to_lines.py...
.========================================================================
FAILED ./scripts/v.to.lines/testsuite/test_v_to_lines.py - Timeout >300.0s
```

nilason:
> Seems totally unrelated, but v.to.lines test failed for macOS, have we seen this before?
> 
> ```
>  Running ./scripts/v.to.lines/testsuite/test_v_to_lines.py...
> ========================================================================
> .========================================================================
> FAILED ./scripts/v.to.lines/testsuite/test_v_to_lines.py - Timeout >300.0s
> ```

Mac timeout caused fails do occur, but rarely enough for me to not see any pattern.

python/grass/temporal/register.py
```diff
@@ -81,6 +119,14 @@ def register_maps_in_space_time_dataset(
...
     end_time_in_file = False
     semantic_label_in_file = False
+    element = {
+        "rast": "cell",
+        "raster": "cell",
+        "rast3d": "grid3",
+        "raster3d": "grid3",
+        "raster_3d": "grid3",
+        "vector": "vector",
+    }[type]
```
wenzeslaus, line 129: How comes it worked before without this?

python/grass/temporal/register.py
```diff
@@ -31,6 +31,44 @@
 ###############################################################################
 
 
+def build_abstract_map_dataset(map_item, element):
+    """Build AbstractMapDataset from map item and element type.
```
wenzeslaus, line 35: It build an id. The whole AbstractMapDataset here and in return doc is misleading.

But perhaps more importantly, compare this to the build_id. That's already what the function is trying to do, its doc says: _Convenient method to build the unique identifier_.

python/grass/temporal/register.py
```diff
@@ -31,6 +31,44 @@
...
+    if not map_mapset:
+        result = gscript.find_file(element=element, name=map_name)
+        if result["mapset"]:
+            map_mapset = result["mapset"]
+        else:
+            gscript.fatal(
+                _("{type} map <{map_name}> not found on search path").format(
+                    type=type, map_name=map_name
+                )
+            )
```
wenzeslaus, line 68: This is not in the original code. The requirement for map existence needs to be stated in the doc.

python/grass/temporal/register.py
```diff
@@ -209,24 +245,18 @@ def register_maps_in_space_time_dataset(
...
 
-            if start_time_in_file and end_time_in_file:
+            if start_time_in_file:
                 row["start"] = line_list[1].strip()
-                row["end"] = line_list[2].strip()
 
-            if start_time_in_file and not end_time_in_file:
-                row["start"] = line_list[1].strip()
+            if end_time_in_file:
+                row["end"] = line_list[2].strip()
```
wenzeslaus, line 252: I lost a track why this change is here.

python/grass/temporal/register.py
```diff
@@ -31,6 +31,44 @@
...
+        >>> tgis.register.build_abstract_map_dataset("rastermap@PERMANENT", "cell")
+        'rastermap@PERMANENT'
+        >>> tgis.register.build_abstract_map_dataset("raster3dmap@PERMANENT", "grid3d")
+        'raster3dmap@PERMANENT'
+
+    """
+    map_name, map_mapset = (
+        map_item.split("@")[0:2] if "@" in map_item else (map_item, None)
+    )
+    map_name, map_layer = map_name.split(":") if ":" in map_name else (map_name, None)
```
wenzeslaus, line 58: In both cases, you can limit the number of splits which should be performed. The first call will be simpler and faster, the second one faster and safer.

wenzeslaus (changes requested):
I like that the similar code is replaced by a function, but I have doubts about when the function actually does and what is considered an error.

python/grass/temporal/register.py
wenzeslaus: Here, I meant a function to do the splitting of map@mapset (name@subproject), but the case here is more specific given the layer number is involved too. Also the error handling (fatal versus raise) and existence requirement might be different here than other places (existing vs new).

python/grass/temporal/register.py
ninsbl: I assume you mean in the doc string of the function?!

That said, the function does throw an error if the map is not found, only later in the code (see: https://github.com/OSGeo/grass/blob/b5c478d714077e31a5a481bb52fac78f10ec08d9/python/grass/temporal/register.py#L295-L299)

Here is the current behavior:
```
t.create output=strds type=strds title=test description=test
t.register input=strds maps=map_does_not_exist type=raster start="2001-01-01" end="2001-01-01"
ERROR: Unable to update raster map <map_does_not_exist@m2>. The map does
not exist.
```
So this addition, basically makes the register function fail earlier...

python/grass/temporal/register.py
ninsbl: It worked because

  a) the `find_file` function replaces "raster" with "cell" (and then gives a message if in verbose mode that the user may not be able to understand, see: https://github.com/OSGeo/grass/blob/ca3b0a8f2b1109867c2034dc41fe3b146ef7b447/python/grass/script/core.py#L1345-L1347) and
  b) if a 3d map was not found (which I guess never was the case) the function simply assumed it is in the current mapset, which may be true for the tests.

wenzeslaus: Makes sense, but perhaps best hidden in a library function which actually understands the user-facing types and does not complain about them, but that's for a different PR.

python/grass/temporal/register.py
wenzeslaus: So, perhaps the new function is existing_map_to_id() or something like that. In C, I sometime end up with x_parameter_to_y_or_fail just to make it clear that fatal/exit will be called.

Will the check later still make sense then? Maybe just check for failure (e.g. try-except) instead of checking ahead there. This is lower priority, though.

Yes, I mean the docstring.

python/grass/temporal/register.py
ninsbl: Right. The looking at the `build_id` function does almost exactly the same:
https://github.com/OSGeo/grass/blob/1a96f69ccf430caa03bec8dbfc990c07c81f3164/python/grass/temporal/abstract_map_dataset.py#L168-L194

So, `build_id` could be extended with an _elements_ key-word that defaults to _None_, and the method would check for map presence only if:

1. an element type to look for is provided,
3. the mapset is not contained in the name, and
4. 3. no mapset is given.

That way current behavior is kept, however, the method would get an additional if check with three conditions...

wenzeslaus: Well, the build_id function currently constructs an id from whatever it is given regardless of the existence or non-existence of the element. Adding a check for existence for a specific combination of parameters, but not the others would be hard to document and tricky to use.

get_mapset_for_map or get_mapset_for_existing_map makes more sense to me as that's more clear about what is required and what may happen even without reading the documentation (even get_mapset_for_map suggests that it must exists since only existing maps have mapsets).

It could also be something like `map, layer, mapset = ensure_mapset(name, element)`. Than the result will be passed to build_id by the caller. In any case, these would untangle the existence requirements from build_id (or its potential wrapper).

Unfortunately, the function cannot be more general than that because here we need the `:layer` piece which is not used elsewhere AFAIK.

Or maybe both your new function and build_id call a same new function to parse out the name, split_name_layer_mapset or something like that. While build_id then proceeds to create the id, your function does the existence check and ensures mapset is provided if missing.

Just to be clear, I don't see a one clear solution. Nevertheless, I think the primary goal should be to avoid creating code with unclear purpose and functions with hard-to-guess behavior.

ninsbl:
In order to disentangle the issues a bit here, I created two different feature branches in my fork from which I can create PRs if the direction to solve this is right.

1) Regarding `find_file`, I replaced the current translation of _raster_ and _rast_ to _cell_ with a more comprehensive replacement for such element types that are used in the code. Here, the logic does not change but message level is decreased from verbose to debug, as it is more relevant for development any way. See: https://github.com/OSGeo/grass/compare/main...ninsbl:grass:find_file_elements for proposed changes.
2) Regarding the `build_id` function in `abstract_map_dataset`: Here, I would suggest to add a keyword argument _check_mapset_element_, which searches for a given element with the given name on the current search path, but only if an element type and no mapset is given. So the behavior remains the same. The costs for that function increase by one if-check, though. See here for how I would propose to implement it: https://github.com/OSGeo/grass/compare/main...ninsbl:grass:bstract_map_dataset_build_id

With those two changes in place, the function added in this PR would be obsolet and we could do:
```
# Vector maps
# With checks executed
AbstractMapDataset.build_id("streams:1", None, None, check_mapset_element="vector")
# Without existence checks
AbstractMapDataset.build_id("streams:1", "PERMANENT", None, check_mapset_element="vector")
AbstractMapDataset.build_id("streams", "PERMANENT", 1, check_mapset_element="vector")
AbstractMapDataset.build_id("streams:1@PERMANENT", None, None, check_mapset_element="vector")
AbstractMapDataset.build_id("streams:1@PERMANENT")
# Note that the existence of a layer is not checked for vector maps

# Raster maps
# With checks executed
AbstractMapDataset.build_id("elevation", None, None, check_mapset_element="raster")
AbstractMapDataset.build_id("elevation", None, None, check_mapset_element="cell")
# Without existence checks
AbstractMapDataset.build_id("elevation", "PERMANENT", None, check_mapset_element="raster")
AbstractMapDataset.build_id("elevation", "PERMANENT")
AbstractMapDataset.build_id("elevation@PERMANENT")
AbstractMapDataset.build_id("elevation@PERMANENT", "mapset_that_does_not_exist", None)
```
All would basically return the same id.

That would appear to me as the clearest implementation with least duplication / repetition of functionality and without changing current behavior.

Any thoughts, should I open PRs for the suggested changes to the two functions in question?

ninsbl:
[a7c0030](https://github.com/OSGeo/grass/pull/2863/commits/a7c00300cfbbcea7bbbfa831113b008c7a80294f) illustrates how this would be solved if #2882 was merged... (because of missing kwarg that comes with #2882, the tests should currently fail. Would need a rebase in case)...

python/grass/temporal/register.py
ninsbl: Solved in the `build_id_from_search_path`function

python/grass/temporal/register.py
ninsbl: Solved in the `build_id_from_search_path` function

python/grass/temporal/register.py
ninsbl: Using the new `build_id_from_search_path` function now

python/grass/temporal/register.py
ninsbl: Addressed in `find_file`

python/grass/temporal/register.py
ninsbl: Solved in the `build_id_from_search_path` function and the documentation is updated accordingly.

python/grass/temporal/register.py
ninsbl: See below...

python/grass/temporal/register.py
ninsbl: I just stubled of those lines of could that could be simplified as I did. No logical change, just one line of code and two conditions less to check...

python/grass/temporal/register.py
ninsbl: Addressed in the `build_id_from_search_path` function

ninsbl:
Thanks @wenzeslaus 
There remain some shell script tests that have not been ported to Python, but that would be another PR.

wenzeslaus:
> ...There remain some shell script tests that have not been ported to Python...

Simple `git mv` will take care of enabling them. Shell scripts are supported (at least on unix-like systems) so that these legacy scripts can at least run with minimal or no changes (another reason was to have a shell-user friendly way of creating tests, but that really did not help with contributions).

PR #2876, opened 2023-03-06
Fix -Wpedantic compiler warnings

nilason:


lib/vector/Vlib/buffer2.c
```diff
@@ -1024,8 +1024,8 @@ void Vect_line_buffer2(const struct line_pnts *Points, double da, double db,
     Vect_line_prune((struct line_pnts *)Points);
 
     if (Points->n_points == 1)
-        return Vect_point_buffer2(Points->x[0], Points->y[0], da, db, dalpha,
-                                  round, tol, oPoints);
+        return; /* Vect_point_buffer2(Points->x[0], Points->y[0], da, db,
+                   dalpha, round, tol, oPoints)*/
```
wenzeslaus, line 1028: This changes what it does, no? The function was called before, or at least that seems like that was the intention.

nilason: Yes, but no. Even if it could return some value, any use of this function treat it as void. I commented it out as it may have been an intention to return something and I figured to keep it in this way in case…

nilason: Good you took this up, `Vect_point_buffer2()` is also void, and probably needed! Will update accordingly, thanks.

nilason: Updated with [785ab26](https://github.com/OSGeo/grass/pull/2876/commits/785ab26638ab8ef47f2831129f9e0667e42bc220)

PR #2877, opened 2023-03-06
g.proj: Fix -Wembedded-directive compiler warnings

nilason:
Addresses warnings with the description:
> warning: embedding a directive within macro arguments has undefined behavior

### DIFF
--- a/general/g.proj/main.c
+++ b/general/g.proj/main.c
@@ -107,7 +107,7 @@ int main(int argc, char *argv[])
 #ifdef HAVE_OGR
     dontprettify->description = _("Print 'flat' output with no linebreaks "
                                   "(applies to WKT and PROJ.4 output)");
-#elif
+#else
     dontprettify->description =
         _("Print 'flat' output with no linebreaks (applies to PROJ.4 output)");
 #endif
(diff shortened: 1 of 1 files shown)


wenzeslaus (approved):
Correctly me if I'm mistaken, but this not only addresses a compiler warning, but it also fixes the strings for translation which cannot be constructed using using ifdefs or ifs.

nilason:
> Correctly me if I'm mistaken, but this not only addresses a compiler warning, but it also fixes the strings for translation which cannot be constructed using using ifdefs or ifs.

It seems to me in this case it worked as it should, in spite of the warnings, example taken from main:

```po
#: ../general/g.proj/main.c:113
msgid ""
"Print 'flat' output with no linebreaks (applies to WKT and PROJ.4 output)"
msgstr "Ausgabe ohne Zeilenumbrüche (nur bei WKT- und PROJ.4-Ausgaben)"

#: ../general/g.proj/main.c:319
#, c-format
msgid "Only one of -%c, -%c, -%c, -%c, -%c or -%c flags may be specified"
msgstr ""
"Nur ein Schalter von  -%c, -%c, -%c, -%c, -%c oder -%c kann angegeben werden"

```

Either way, there is no translation for the case of `HAVE_OGR` not being defined (however unlikely). This depends on the settings on the machine generating the .po files. In this respect the macro condition doesn't make much sense. Alternative fix may be to reformulate the message to work for both ways. Suggestions?!
cc @neteler

nilason:
On the other hand, this is not an exceptional situation. There are lots of "missing" translatable messages for different states (compiled with this or that etc.).

Issue #2880, opened 2023-03-06
[Feat] Fix compiler warnings from -Wpedantic


nilason:
Compiling with -Wall -Wextra and `-Wpedantic` triggers a number of mostly (if not only) trivial warnings.
Addressing these warnings contribute to keep the code in good shape according to ISO standards, moreover potentially enables the use of -Wpedantic in combination with -Werror.

Following warnings are issued:

- -Wstrict-prototypes #2870
- -Winvalid-utf8 #2873
- -Wextra-semi #2871
- -Wformat-pedantic #2872
- -Wempty-translation-unit #2874
- -Wgnu-pointer-arith #2875
- -Wpedantic #2876
- -Wembedded-directive #2877
- -Wformat #2878
- -Woverflow #2879

Follow-up on #2747.

Background in parenthesis:
The initial reason for looking at also this level of warnings, was the four GCC CI runs, which all behaves as one and the same. I believe there is potential to reduce them to one runner. But that is a question for another issue.

wenzeslaus:
> ...the four GCC CI runs, which all behaves as one and the same.

Part of the original motivation was indeed the different levels of required "code correctness" and understanding that while version A works, it not okay for version B. However, the other reason was compatibility. It explicitly checks all supported versions. I also remember fixing some C++ issue where two different versions allowed two different constructs, but only one one of these was accepted by both. However, I don't remember if this was more about compiler versions (both clang and GCC) or about standard versions. It was also without the pedantic behavior. The tests for all supported versions still makes sense to me, though.

nilason:
GRASS code in itself now compiles without errors using -Wpedantic, only #2899 is needed to enable the flag on the CI runners.

PR #2873, opened 2023-03-06
Fix -Winvalid-utf8 compiler warning

nilason:

PR #2891, opened 2023-03-17
lib/raster3d: attempt to add prototype by removing const

nilason:
The `Rast3d_readWriteWindow()`:

https://github.com/OSGeo/grass/blob/d10ac4f5e17bc6bcdbe97a2bce115bc49fc8761d/lib/raster3d/windowio.c#L12

function both reads and writes files using function pointers without function prototypes, which is deprecated in all versions of C and is not supported in C2x.

Adding missing prototype for this function causes incompatibility with the set functions defined with `const` whereas the get functions are not. This PR removes the const qualifier for the set functions, which are however correct, but leads to the least modification overall.

An alternative to this approach, would be to rewrite this and separate the read and write code all together.

Putting this up as a draft, as the solution to this may be up to discussion.

### DIFF
--- a/include/grass/defs/raster3d.h
+++ b/include/grass/defs/raster3d.h
@@ -132,11 +132,11 @@ int Rast3d_key_get_double(struct Key_Value *, const char *, double *);
 int Rast3d_key_get_string(struct Key_Value *, const char *, char **);
 int Rast3d_key_get_value(struct Key_Value *, const char *, char *, char *, int,
                          int, int *);
-int Rast3d_key_set_int(struct Key_Value *, const char *, const int *);
-int Rast3d_key_set_double(struct Key_Value *, const char *, const double *);
-int Rast3d_key_set_string(struct Key_Value *, const char *, char *const *);
-int Rast3d_key_set_value(struct Key_Value *, const char *, const char *,
-                         const char *, int, int, const int *);
+int Rast3d_key_set_int(struct Key_Value *, const char *, int *);
+int Rast3d_key_set_double(struct Key_Value *, const char *, double *);
+int Rast3d_key_set_string(struct Key_Value *, const char *, char **);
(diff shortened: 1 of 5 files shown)


wenzeslaus:
> ...This PR removes the const qualifier for the set functions, which are however correct, but leads to the least modification overall.

Would some casting help? Like const_cast would do in C++? Too many changes.

> An alternative to this approach, would be to rewrite this and separate the read and write code all together.

That seems like the appropriate solution. Maybe a quick wrapper which does the const cast?

Just to be clear. You marked this as 8.4. The approach for 8.3 is to disable (or not enable) the warning/error. Just thinking how quick/non-invasive the fix needs to be.

nilason:
> > ...This PR removes the const qualifier for the set functions, which are however correct, but leads to the least modification overall.
> 
> Would some casting help? Like const_cast would do in C++? Too many changes.

No, "down casting" const is not an option in C, if we want to get rid of compiler warnings.

> > An alternative to this approach, would be to rewrite this and separate the read and write code all together.
> 
> That seems like the appropriate solution. Maybe a quick wrapper which does the const cast?

Present code is clever in reusing the code for get and set depending on a boolean flag. The code may be split with some perhaps duplicated code in two functions instead of using this boolean flag.

> Just to be clear. You marked this as 8.4. The approach for 8.3 is to disable (or not enable) the warning/error. Just thinking how quick/non-invasive the fix needs to be.

If I'm not mistaken using -Wpedantic (#2880) will be possible for the GCC builds after #2873, #2874, #2875 and #2876 are merged. They are all trivial and safe for 8.3. I personally will have little time the coming week for attempting an alternative for this PR. And #2894 is also needed for in combo with this, for the macOS clang build for -Wpedantic. That's why the 8.4 milestone.

wenzeslaus:
Thanks for clarifying. It makes sense.

To correct myself, I think I just wanted to say "...done without too many changes" not "Too many changes!" I'm happy with the change.

PR #2894, opened 2023-03-17
libvector : reading files with 'struct Map_info*' objects is technically not const

nilason:
Addressing -Wdeprecated-non-prototype compiler warnings and adding missing prototypes in [`lib/vector/Vlib/read.c`](https://github.com/OSGeo/grass/blob/main/lib/vector/Vlib/read.c), reveals API errors in lib functions passing `struct Map_info *` as const. Reading vector files **does** in fact modify the Map_info struct, therefore the `const` qualifier is not technically correct. Correcting this API in lib cascades through quite a few files, but functionally this doesn't change anything.

Alternative to this may be to separate the file reading variables to a separate "read object", but that will modify the API in a very drastic way and would be anything but trivial.


lib/vector/Vlib/read.c
```diff
@@ -19,21 +19,22 @@
 #include <grass/vector.h>
 #include <grass/glocale.h>
 
-static int read_dummy()
+static int read_dummy(struct Map_info *, struct line_pnts *, struct line_cats *)
```
marisn, line 22: You seem to have forgotten to provide a dummy parameter names too.

lib/vector/Vlib/read.c
```diff
@@ -19,21 +19,22 @@
...
 {
     G_warning("Vect_read_line() %s", _("for this format/level not supported"));
     return -1;
 }
 
 #if !defined HAVE_OGR || !defined HAVE_POSTGRES
-static int format()
+static int format(struct Map_info *, struct line_pnts *, struct line_cats *)
```
marisn, line 29: Ibid.

marisn (changes requested):
Seems fine. Fixing of those two functions in read.c should fix failing tests.

PR #2899, opened 2023-03-19
checks: suppress compiler warnings from GDAL

nilason:
Enabling compiler flag `-Wpedantic` unleashes a great number of warnings originating from non-GRASS sources. They all (on GCC CI runners) come from ogr_core.h with non-iso compliant enum values, eg.

```
/opt/local/include/ogr_core.h:460:5: warning: ISO C restricts enumerator values to range of 'int' (2147483649 is too large) [-Wpedantic]
    wkbPoint25D = 0x80000001, /**< 2.5D extension as per 99-402 */
    ^             ~~~~~~~~~~
```

This is not likely to be addressed upstreams in the near future: https://github.com/OSGeo/gdal/issues/2322.

I'm not too happy for the required solution of adding 12 line of macro for each include, but I reckon this is the only way to assure good code (`-Wpedantic -Werror` in the CI runners) in the GRASS source.

### DIFF
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -59,8 +59,6 @@ jobs:
           ldd --version
       - name: Build
         env:
-          # TODO: -pedantic-errors here won't go through ./configure (with GNU C)
-          CFLAGS: -std=${{ matrix.c }} -fPIC -Wall -Wextra
-          # TODO: -pedantic-errors here won't compile
-          CXXFLAGS: -std=${{ matrix.cpp }} -fPIC -Wall -Wextra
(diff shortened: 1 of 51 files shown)


lbartoletti:
> I'm not too happy for the required solution of adding 12 line of macro for each include, but I reckon this is the only way to assure good code (`-Wpedantic -Werror` in the CI runners) in the GRASS source.

not only in the GRASS source, but everywhere you want to disable these warnings via pragma.

Another solution - not tested - could be to include a "grass/gdal" header once, including all these pragmas and the gdal/ogr header.

nilason:
#3245 is intended to supersede this PR.

PR #2906, opened 2023-03-28
DRAFT: r.patch race condition

jameshu15869:
This is my current draft implementation for making an array of statf structures for each thread. 

Null count is being counted correctly in each thread and the values add up to what is expected, but I am unsure how to proceed for merging the trees. 

Before merging:
- [x] add test specifically testing CELL rasters
- [x] run benchmark, compare with and without -s flag
- [x] run valgrind


raster/r.patch/main.c
```diff
@@ -154,14 +169,18 @@ int main(int argc, char *argv[])
             out_type = DCELL_TYPE;
 
         Rast_init_cell_stats(&statf[i]);
```
petrasovaa, line 171: I would call `Rast_init_cell_stats` in a loop here instead of memcpy.

raster/r.patch/main.c
```diff
@@ -135,6 +146,10 @@ int main(int argc, char *argv[])
     for (t = 0; t < nprocs; t++)
         infd[t] = G_malloc(nfiles * sizeof(int));
     statf = G_malloc(nfiles * sizeof(struct Cell_stats));
```
petrasovaa, line 148: I would get rid of this one.

raster/r.patch/main.c
petrasovaa: Right. I think there are more ways to proceed, but first you need to make sure you understand how support() works and how it's "merging" the stats from multiple rasters. It's using the stats from the first raster as a base (see the function documentation).  So then one way it could work is that you first "merge" the stats for the first raster (with `Rast_find_cell_stat` and `Rast_update_cell_stat`) and then continue the loop in support() but looping through the rest of your stats (adding a loop over threads). Or you can merge them all before calling support() and then you don't need to modify anything there. I don't know what is better performance-wise.

HuidaeCho:
I think this race condition was introduced by parallelism, but I haven't tried it yet. Looking at the code, have you tried:
```c
#pragma omp parallel private(i, row) if (nprocs > 1)
```
in line 208 in main.c? `row` should be private to avoid data races.

petrasovaa:
> I think this race condition was introduced by parallelism, but I haven't tried it yet. Looking at the code, have you tried:
> 
> ```c
> #pragma omp parallel private(i, row) reduction(+:computed) if (nprocs > 1)
> ```
> 
> in line 208 in main.c? `row` should be private to avoid data races and `computed` should be reduced by addition. I think we can remove line 250 `#pragma omp atomic update` with this reduction.

Please see the original problem in #2410 and the issue #2411 this PR is addressing. What you suggest may be a good improvement, but AFAIU it's not really related to this PR. We had an initial conversation about this on gitter.

HuidaeCho:
> > I think this race condition was introduced by parallelism, but I haven't tried it yet. Looking at the code, have you tried:
> > ```c
> > #pragma omp parallel private(i, row) reduction(+:computed) if (nprocs > 1)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in line 208 in main.c? `row` should be private to avoid data races and `computed` should be reduced by addition. I think we can remove line 250 `#pragma omp atomic update` with this reduction.
> 
> Please see the original problem in #2410 and the issue #2411 this PR is addressing. What you suggest may be a good improvement, but AFAIU it's not really related to this PR. We had an initial conversation about this on gitter.

Oh, it was a different problem! Sorry for the noise.l

raster/r.patch/main.c
```diff
@@ -282,9 +317,50 @@ int main(int argc, char *argv[])
          */
         G_verbose_message(_("Creating support files for raster map <%s>..."),
                           new_name);
-        support(names, statf, nfiles, &cats, &cats_ok, &colr, &colr_ok,
-                out_type);
+
+        // try to merge all of thread_statf[] into a single statf at
+        // thread_statf[0]
+        int next_cell_stat;
```
petrasovaa, line 323: This should be type CELL

raster/r.patch/main.c
```diff
@@ -282,9 +317,50 @@ int main(int argc, char *argv[])
...
+                if (out_type == CELL_TYPE) {
+                    Rast_rewind_cell_stats(thread_statf[t] + i);
+                    while (Rast_next_cell_stat(&next_cell_stat, &merge_count,
+                                               thread_statf[t] + i)) {
+                        if (next_cell_stat &&
+                            !Rast_find_cell_stat(next_cell_stat, &merge_count,
+                                                 thread_statf[0] + i)) {
+                            Rast_update_cell_stats(&next_cell_stat, 1,
+                                                   thread_statf[0] + i);
+                        }
```
petrasovaa, line 338: I think this works fine for this case. Practically, the merged stats won't reflect the actual stats, because here you are  interested in whether the category is there and not really how many cells of that category is there, hence `Rast_update_cell_stats(&next_cell_stat, 1, ...)`  will work although you are adding just 1 cell of that category in there.

petrasovaa (changes requested):
Good job, I think this is close to finished! Please cleanup the code (address the issues, remove comments, old statf, print functions). The new code for merging should be ideally in a separate function (in support.c).

Before merging this, we should also write test and run benchmark to ensure this is correct and to see what impact this has on performance. But at this point I suggest you to finish the cleanup and then we brainstorm (on Gitter) the GSoC project and proposal if you are interested, the deadline is Apr 4th. This PR could be finalized as part of the project.

jameshu15869:
@petrasovaa 
I've gone back and cleaned up the code. I also created a `merge_threads` function in support that will actually perform the merge. Regarding testing, I found a test in `testsuite/test_rpatch_artificial.py` and my implementation passed. I tried to run the benchmark file, but got an error. Is it possible that this error came from the benchmark itself and not my implementation?
```
  File "benchmark/benchmark_r_patch.py", line 52, in generate_map
    Module(
  File "/usr/local/grass83/etc/python/grass/pygrass/modules/interface/module.py", line 610, in __init__
    self.__call__(*args, **kargs)
  File "/usr/local/grass83/etc/python/grass/pygrass/modules/interface/module.py", line 634, in __call__
    return self.run()
  File "/usr/local/grass83/etc/python/grass/pygrass/modules/interface/module.py", line 823, in run
    self.wait()
  File "/usr/local/grass83/etc/python/grass/pygrass/modules/interface/module.py", line 844, in wait
    raise CalledModuleError(
grass.exceptions.CalledModuleError: Module run `r.surf.fractal dimension=2.05 number=0 output=r_patch_reference_tmp --o` ended with an error.
The subprocess ended with a non-zero return code: -9. See errors above the traceback or in the error output.
````

Please let me know if there's anything else I should add to this PR. Thanks!

File: raster/r.patch/support.c [hunk 5]
```diff
@@ -1,6 +1,39 @@
...
+/**
+ * merges the statf[] arrays for each thread into
+ * Thread 0 for use in the support computation.
+ *
+ * the funciton takes in a 2D array thread_statf[][], where
+ * the rows are for represent each thread and each column is
+ * a file that is being used to patch. The end result is
+ * that thread_statf[0] is an array of size [nfiles] that holds
+ * the merged stats information.
+ *
```
petrasovaa, line 13: Please reread for typos. Start with uppercase as a sentence.

jameshu15869: Sorry, that was a careless mistake and has been corrected.

petrasovaa: No worries, thanks for fixing that.

jameshu15869:
> Looking at the code of the benchmark, there is a fallback if r.surf.fractal fails that runs r.random.surface instead, did the fallback work for you? I think the fallback is there because r.surf.fractal might require lot of memory, is it possible you are running out of memory? Try to lower the size or raster and see if that helps.

I found the error - when attempting to generate a map with size 4000, both functions require around 8.5 GB of memory. My computer only has 8 GB so I suspect this is why `r.surf.fractal` and the fallback `r.random.surface` both failed. However, I was able to run benchmarks for size `7071`, `14142`, and `10000`. The graph is attached below. 


benchmarking without 2000

My laptop only has 4 threads, so should I ignore the time increase as threads increase? My best guess is that the program does not know how to assign 12 threads to 4 threads, for example. However, other than that, this graph shows that the OpenMP does make a difference, right?

As a side note, running the benchmarks takes around 30-40 minutes for each run on my laptop. Is there a way to reduce the number of threads being tested so only nprocs=1,2,4 are run?

Thanks!

petrasovaa:
> FWIW, you can use a swap file (!) to "enlarge" the RAM if you have some spare GB of disk space:

Good to know, but I wonder if this affects the performance.

jameshu15869:
Sounds good. For now I will plan to ask others to help out with the testing. 

One other thing I noticed was that the GitHub check is currently failing for Ubuntu on the test for `t.rast.algebra`. The error message is 

```
Default TGIS driver / database set to:
driver: sqlite
database: $GISDBASE/$LOCATION_NAME/$MAPSET/tgis/sqlite.db
WARNING: Temporal database connection defined as:
/home/runner/nc_spm_full_v2alpha2/__temporal_t_rast_algebra_test_raster_algebra_fv_az574_655_32269/tgis/sqlite.db
But database file does not exist.
Creating temporal database: /home/runner/nc_spm_full_v2alpha2/__temporal_t_rast_algebra_test_raster_algebra_fv_az574_655_32269/tgis/sqlite.db
```

However, the `Ubutnu 22.04` test passed for one of my previous commits (https://github.com/OSGeo/grass/pull/2906/commits/2d2daa0451dde9e72b24663838a249b2b05a7e45). It seems that the only change between that commit and the most recent failing commits is running the additional tests. I quickly scanned through the `t.rast.algebra` test and it doesn't seem like any of the code I worked on should be tested in this file. Could there be an issue with the GitHub CI itself? Or is there something else going on?

Thanks!

jameshu15869 (commented):
All the changes have been resolved and I believe the PR is ready for a final review.

jameshu15869 (commented):
All the files have been reviewed, but I'm not sure how to get rid of the "merging block"

jameshu15869:
@petrasovaa 
Hi, I'm just following up with this PR. All the checks passed for my latest commit. GitHub still says "changes requested", even though I've made all the necessary changes and viewed them through GitHub's user interface. What should the next steps for this PR be?

petrasovaa (approved):
The code looks good, but I still need to run the benchmarks (I added tasks in the PR description), will get to it hopefully next week. You don't need to do anything right now.

jameshu15869:
> The code looks good, but I still need to run the benchmarks (I added tasks in the PR description), will get to it hopefully next week. You don't need to do anything right now.

Sounds good. Thank you so much for all your help!

petrasovaa:
I ran a benchmark with and without s flag:

with s flag:
rpatch_benchmark_s

without (default):
rpatch_benchmark

So running with s flag is slightly faster, but that's expected. Otherwise the benchmark seems fine.

jameshu15869:
@petrasovaa 
I'm glad that it works! Did testing reveal any issues with the implementation that need to be addressed?

jameshu15869:
> I think it's all fine, ready to be merged. The failing mac test is unrelated.
> 
> But could you please look at Huidae's suggestion [above](https://github.com/OSGeo/grass/pull/2906#issuecomment-1487197477) and see if that works and if it improves anything? If yes, that would be a different PR.

@HuidaeCho @petrasovaa 
I've implemented Huidae's suggestion and modified the old `#pragma omp atomic update` and this was the graph I got after running only the 50M r.patch benchmark (Since I am working at school on my laptop). I was also able to pass all 4 test cases in `testsuite/test_rpatch_artificial.py`. It seems that there was definitely an improvement, though I believe more testing on a more powerful computer is required. I haven't committed the changes yet since Anna mentioned that this could be a separate PR. Should I move these onto a new branch/open a new PR specifically for this issue?

rpatch_benchmark

petrasovaa:
> It seems that there was some minor improvement, though I'm not exactly sure how to quantify it. I haven't committed the changes yet since Anna mentioned that this could be a separate PR. Should I move these onto a new branch/open a new PR specifically for this issue?

Yes, please create a new PR and I will look at it. I will merge this one after the checks are done.

jameshu15869:
I've opened a new pull request for the reduction (https://github.com/OSGeo/grass/pull/2941). However, it looks like all of my changes from this PR are still on the new PR. Would this potentially become an issue later on?

petrasovaa:
> I've opened a new pull request for the reduction (#2941). However, it looks like all of my changes from this PR are still on the new PR. Would this potentially become an issue later on?

Right, this needs to be fixed. I am not sure what is the easiest way, you can update your main, create a new branch, put your changes there and the force push (something like `git push -f origin branch-with-new-changes:branch-you-are-pushing-to`)

PR #2914, opened 2023-04-07
pygrass: retrieve 3D geometries for 3D vector maps

landam:
## Current behaviour

PyGRASS currently handles geometries for 3D vector maps as 2D

### Steps to reproduce

```sh
v.random out=p3d n=1 -z zmax=1
```

Sample PyGRASS script:

```py
from grass.pygrass.vector import VectorTopo

with VectorTopo('p3d') as vmap:
    print('Map 3D?', vmap.is_3D())
    for f in vmap.viter('points'):
        print('Point 3D?', not f.is2D)
        print('Z:', f.z)
```

Output:

```
Map 3D? True
Point 3D? False
Z: None
```

## Expected behaviour

```
Map 3D? True
Point 3D? True
Z: 0.4480268010321531
```

### DIFF
--- a/python/grass/pygrass/vector/__init__.py
+++ b/python/grass/pygrass/vector/__init__.py
@@ -450,6 +450,7 @@ class VectorTopo(Vector):
 
             >>> test_vect.close()
         """
+        is2D = bool(libvect.Vect_is_3d(self.c_mapinfo) != 1)
         if vtype in _GEOOBJ.keys():
             if _GEOOBJ[vtype] is not None:
                 ids = (indx for indx in range(1, self.number_of(vtype) + 1))
@@ -461,6 +462,7 @@ class VectorTopo(Vector):
                         c_mapinfo=self.c_mapinfo,
                         table=self.table,
                         writeable=self.writeable,
+                        is2D=is2D
                     )
                     for indx in ids
                 )
--- a/python/grass/pygrass/vector/__init__.py
+++ b/python/grass/pygrass/vector/__init__.py
@@ -462,7 +462,7 @@ class VectorTopo(Vector):
                         c_mapinfo=self.c_mapinfo,
                         table=self.table,
                         writeable=self.writeable,
-                        is2D=is2D
+                        is2D=is2D,
                     )
                     for indx in ids
                 )
--- a/python/grass/pygrass/vector/__init__.py
+++ b/python/grass/pygrass/vector/__init__.py
@@ -450,7 +450,7 @@ class VectorTopo(Vector):
 
             >>> test_vect.close()
         """
-        is2D = bool(libvect.Vect_is_3d(self.c_mapinfo) != 1)
+        is2D = not self.is_3D()
         if vtype in _GEOOBJ.keys():
             if _GEOOBJ[vtype] is not None:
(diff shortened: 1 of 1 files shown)


python/grass/pygrass/vector/__init__.py
```diff
@@ -450,6 +450,7 @@ def viter(self, vtype, idonly=False):
 
             >>> test_vect.close()
         """
+        is2D = bool(libvect.Vect_is_3d(self.c_mapinfo) != 1)
```
wenzeslaus, line 453: The same code is in the Geo class. What put me on that trail was that slightly better code would be just `not bool(Vect_is_3d(...)`, but my conclusion now it that for this PR, this line is simply not needed.

python/grass/pygrass/vector/__init__.py
```diff
@@ -461,6 +462,7 @@ def viter(self, vtype, idonly=False):
                         c_mapinfo=self.c_mapinfo,
                         table=self.table,
                         writeable=self.writeable,
+                        is2D=is2D,
```
wenzeslaus, line 465: The Geo class can do it automatically by using `is2D=None`, however, it is probably better to fix it in Geo class itself. It has `is2D=True` in parameters, but `None` is a more reasonable default especially given that it determines the 2D/3D automatically if not given.

wenzeslaus (changes requested):
This seems quite helpful fix, but please, do the fix in the underlying [grass.pygrass.vector.geometry.Geo](https://grass.osgeo.org/grass82/manuals/libpython/grass.pygrass.vector.html#grass.pygrass.vector.geometry.Geo) class.

python/grass/pygrass/vector/__init__.py
landam: Yes, but passing `is2D=None` to the `Geo` class will force performing 3D check for every geometry in vector map. It's not expected that you have mixed 2D/3D geometries in single 3D vector map. That's reason why this PR sets `is2D` variable before looping over geometries.

landam: Changing `is2D` default value to `None` would make sense. But than it should be also changed in other parts of the code, eg. https://github.com/OSGeo/grass/blob/main/python/grass/pygrass/vector/geometry.py#L1928

python/grass/pygrass/vector/__init__.py
landam: See my comments below. I changed this line to call `is_3D()` similarly as it's done here: https://github.com/OSGeo/grass/blob/main/python/grass/pygrass/vector/__init__.py#L518

wenzeslaus: Much simpler!

python/grass/pygrass/vector/__init__.py
wenzeslaus: You are right, checking for every geometry wouldn't be good. `self.is_3D()` call is just better. No need to solve the Geo issue here, but the default definitively seems wrong.

wenzeslaus (approved):
`self.is_3D()` makes sense here and I see it used elsewhere.

PR #2923, opened 2023-04-18
grass.experimental: Add object to access modules as functions

wenzeslaus:
This adds a Tools class which allows to access GRASS tools (modules) to be accessed using methods. Once an instance is created, calling a tool is calling a function (method) similarly to grass.jupyter.Map. Unlike grass.script, this does not require generic function name and unlike grass.pygrass module shortcuts, this does not require special objects to mimic the module families.

Outputs are handled through a returned object which is result of automatic capture of outputs and can do conversions from known formats using properties.

Usage example is in the _test() function in the file.

The code is included under new grass.experimental package which allows merging the code even when further breaking changes are anticipated.


python/grass/experimental/tools.py
```diff
@@ -0,0 +1,104 @@
...
+
+        def wrapper(**kwargs):
+            # Run module
+            return self.run(grass_module, **kwargs)
+
+        return wrapper
+
+
+def _test():
+    gs.setup.init("~/grassdata/nc_spm_08_grass7/user1")
```
neteler, line 88: I got errors with shell expansion using `~/` - perhaps better use somthing like:

```
from pathlib import Path
home = str(Path.home())
```
?

wenzeslaus: Well, that should work because the underlying code calls _expanduser_: 

https://github.com/OSGeo/grass/blob/main/python/grass/script/setup.py#L317

Can you please test this more and report an issue against _grass.script.setup.init_ if needed?

BTW, the _str_ in `str(Path.home())` should not be needed because most things will accept a Path object. If not (and we have some places like that), that's something to fix.

landam:
It seems to be useful addition. On the other hand we have already two API to run GRASS modules: `grass.script.*_command()` and `grass.pygrass.modules` which is already confusing for the user. What is a benefit of the third one? It would be useful to merge existing APIs into single one instead introducing another one.

wenzeslaus:
> It seems to be an useful addition.

I still need to provide more context for this, but do you see some benefits already?

> On the other hand we have already two APIs to run GRASS modules: `grass.script.*_command()` and `grass.pygrass.modules` which is already confusing for the user.

The intro to this is obviously [xkcd Standards](https://xkcd.com/927/).

I'm not happy with the two competing interfaces. It's almost three, because we have Module and than also shortcuts.

As far as I understand, `grass.script.*_command()` was written to closely mimic the Bash experience with minimal involvement of Python. Python layer is mostly just avoiding need to pass all parameters as strings.

`grass.pygrass.modules` was written to mimic the `grass.script.*_command()` API and to manipulate the module calls themselves.

> What is a benefit of the third one?

The design idea is 1) to make the module (tool) calls as close to Python function calls as possible and 2) to access the results conveniently. To access the (text) results, it tries to mimic _subprocess.run_.

Additionally, it tries to 1) provide consistent access to all modules and 2) allow for extensibility, e.g., associating session parameters or computational region with a Tools object rather than passing it to every method.

The existing APIs are more general in some ways, especially because they make no assumptions about the output or its size. This API makes the assumption that you want the text output Python or that it is something small and you can just ignore that. If not, you need to use a more general API. After all, Tools itself, is using pipe_command to do the job.

> It would be useful to merge existing APIs into single one instead introducing another one.

Given the different goals of the two API, I was not able to figure out how these can be merged. For example, the Module class from _grass.pygrass_ was supposed to be a drop-in replacement for _run_command_, but it was not used that way much (maybe because it forces you to use class as an function). Any suggestions? What would be the features and aspects of each API worth keeping? For example, the Tools object might be able to create instances of the Module class.

I can also see that some parts of the new API could be part of the old ones like output-parsing related properties for the Module class, but there are some existing issues which the new API is trying to fix such as `r.slope_aspect` spelling in PyGRASS shortcuts and Python function name plus tools name as a string in grass.script.

Finally, the _subprocess_ changed too over the years, introducing new functions with _run_ being the latest addition, so reevaluation of our APIs seems prudent even if it involves adding functions as _subprocess_ did.

Anyway, I think some unification would be an ideal scenario.

wenzeslaus:
This is how exceptions look like now:  The error (whole stderr) is part of the exception, i.e., always printed with the traceback, not elsewhere, and it is under the traceback, not above.

```
  File "experimental/tools.py", line 252, in <module>
    _test()
  File "experimental/tools.py", line 241, in _test
    tools_pro.feed_input_to("13.45,29.96,200").v_in_ascii(
  File "experimental/tools.py", line 185, in wrapper
    return self.run(grass_module, **kwargs)
  File "experimental/tools.py", line 148, in run
    raise gs.CalledModuleError(
grass.exceptions.CalledModuleError: Module run `v.in.ascii input=- output=point format=xstandard` ended with an error.
The subprocess ended with a non-zero return code: 1. See the following errors:

ERROR: Value <xstandard> out of range for parameter <format>
	Legal range: point,standard
```

```
  File "experimental/tools.py", line 252, in <module>
    _test()
  File "experimental/tools.py", line 241, in _test
    tools_pro.feed_input_to("13.45,29.96,200").v_in_ascii(
  File "experimental/tools.py", line 185, in wrapper
    return self.run(grass_module, **kwargs)
  File "experimental/tools.py", line 148, in run
    raise gs.CalledModuleError(
grass.exceptions.CalledModuleError: Module run `v.in.ascii input=- output=point format=standard` ended with an error.
The subprocess ended with a non-zero return code: 1. See the following errors:
WARNING: Vector map <point> already exists and will be overwritten
WARNING: Unexpected data in vector header:
         [13.45,29.96,200]
ERROR: Import failed
```

echoix:
Solved conflicts

wenzeslaus:
While the code quality is down because my node version is too old and I can't run pre-commit, this can now be used to access JSON output directly (and naturally fails to parse if there is not JSON output):

```python
mean_value = tools.r_univar(map="surface", format="json")[0]["mean"]
```

It can also suggest (one or more) tools in case of misspelling:

```python
tools.r_sloppy_respect(elevation="surface", slope="slope")[0]["mean"]
```

```
AttributeError: Tool r.sloppy.respect not found. Did you mean: r.slope.aspect?
```

echoix:
> While the code quality is down because my node version is too old and I can't run pre-commit,

Since somewhere during the code sprint, in the PR that updated markdown lint in pre-commit, I had to fix this usability problem as the Ubuntu 22.04 runners defaulted to node 18 (others are installed but not active), but is already EOL or very close to, and the tool dropped support since. I configured it in a way that most normal users won't need to do anything or think about. Pre-commit uses the system node if available, otherwise sets up one in an environment similar to venvs for python.

If you would to try again today (with the pre-commit file from the main branch, or the branch updated), it should work as expected.

wenzeslaus:
## Experimental CLI

This is not really that important for this PR specifically, but similarly to what I did when adding mapset locking to the Python library, I added CLI for testing.

Set up needed without FHS:

```sh
export PYTHONPATH=$(./bin.x86_64-pc-linux-gnu/grass --config python-path)
```

There is only temporary XY project set up, so it is not really useful at this point, except for things like help or g.extension:

```sh
python -m grass.app run g.extension -l
python -m grass.app run g.region --help
python -m grass.app run r.slope.aspect --interface-description
```

I'm using the CLI as an unusual use case for the Tools API, so it helps me understand that standard input and output work as expected, and they do:

```sh
# stdin
echo "a = 6" | python -m grass.app run r.mapcalc file=-

# stdout
python -m grass.app run g.region -p | grep res
```

The CLI will become more interesting over time with changes such as #5877 and #5843.

wenzeslaus:
## Comparing to and migrating from run_command family of functions 

Here are examples of how the different use cases of run_command and friends look like with the Tools API, organized by the grass.script counterparts to Tools API calls.

You can just thumbs up this if you find that reasonable, but feel free to comment, too. The new API keeps the focus on the tools themselves rather than having user go through different  functions to call the tool with different inputs and outputs (run_command vs parse_command vs read_command vs write_command) or even through dedicated wrappers to get the output of the tool in a form reasonable in Python context (g.region as region, g.list as list_strings, etc.)

### Imports

```python
# original:
import grass.script as gs
```

```python
# replacement
from grass.experimental.tools import Tools
import io  # only needed when stdin is used
```

### run_command - just run the tool

```python
# original:
gs.run_command(
    "r.random.surface", output="surface", seed=42
)
```

```python
# replacement using the run function which is syntactically close to run_command:
tools = Tools()  # same for one or multiple calls
tools.run("r.random.surface", output="surface2", seed=42)  # name as a string
```

```python
# assuming we already have tools and using the function syntax:
tools.r_random_surface(output="surface3", seed=42)  # name as a function
```

### write_command - provide standard input (text)

```python
# original:
gs.write_command(
    "v.in.ascii",
    input="-",
    output="point1",
    separator=",",
    stdin="13.45,29.96,200\n",
)
```

```python
# replacement:
tools.run(
    "v.in.ascii",
    input=io.StringIO("13.45,29.96,200\n"),
    output="point2",
    separator=",",
)
```

```python
# or with function name syntax:
tools.v_in_ascii(
    input=io.StringIO("13.45,29.96,200\n"),
    output="point3",
    separator=",",
)
```

### read_command - get standard output (text)

```python
# original:
assert (
    gs.read_command("g.region", flags="c")
    == "center easting:  0.500000\ncenter northing: 0.500000\n"
)
```

```python
# replacement:
assert (
    tools.run("g.region", flags="c").stdout
    == "center easting:  0.500000\ncenter northing: 0.500000\n"
)
```

```python
# or with function name syntax:
assert (
    tools.g_region(flags="c").text
    == "center easting:  0.500000\ncenter northing: 0.500000"
)
```

### parse_command - get machine readable standard output

```python
# original (numbers are strings):
assert gs.parse_command(
    "g.region", flags="c", format="shell"
) == {
    "center_easting": "0.500000",
    "center_northing": "0.500000",
}
# numbers are always numbers with JSON:
assert gs.parse_command(
    "g.region", flags="c", format="json"
) == {
    "center_easting": 0.5,
    "center_northing": 0.5,
}
```

```python
# replacement with format=shell (numbers are not strings, but actual numbers as in JSON
# if they convert to Python int or float):
assert tools.run("g.region", flags="c", format="shell").keyval == {
    "center_easting": 0.5,
    "center_northing": 0.5,
}
```

```python
# parse_command with JSON and the function call syntax:
assert tools.g_region(flags="c", format="json").json == {
    "center_easting": 0.5,
    "center_northing": 0.5,
}
```

### parse_command storing JSON output in a variable and accessing individual values

```python
# original:
data = gs.parse_command(
    "g.region", flags="c", format="json"
)
assert data["center_easting"] == 0.5
assert data["center_northing"] == 0.5
```

```python
# replacement:
data = tools.g_region(flags="c", format="json")
assert data["center_easting"] == 0.5
assert data["center_northing"] == 0.5
```

### Dedicated wrappers: r.mapcalc

```python
# mapcalc wrapper of r.mapcalc
# original:
gs.mapcalc("a = 1")
```

```python
# replacement for short expressions:
tools.r_mapcalc(expression="b = 1")
# replacement for long expressions:
tools.r_mapcalc(file=io.StringIO("c = 1"))
```

### Dedicated wrappers: g.list

```python
# test data preparation (for comparison of the results):
names = ["a", "b", "c", "surface", "surface2", "surface3"]
# original:
assert gs.list_grouped("raster")["PERMANENT"] == names
```

```python
# replacement (using the JSON output of g.list):
assert [
    item["name"]
    for item in tools.g_list(type="raster", format="json")
    if item["mapset"] == "PERMANENT"
] == names
# original and replacement (directly comparing the results):
assert gs.list_strings("raster") == [
    item["fullname"] for item in tools.g_list(type="raster", format="json")
]
# original and replacement (directly comparing the results):
assert gs.list_pairs("raster") == [
    (item["name"], item["mapset"])
    for item in tools.g_list(type="raster", format="json")
]
```

### Dedicated wrappers: all other tools

```python
# Wrappers in grass.script usually parse shell-script style key-value pairs,
# and convert values from strings to numbers, e.g. g.region:
assert gs.region()["rows"] == 1
```

```python
# Conversion is done automatically in Tools and/or with JSON, and the basic tool
# call syntax is more lightweight, so the direct tool call is not that different
# from a wrapper. Direct tool calling also benefits from better defaults (e.g.,
# printing more in JSON) and more consistent tool behavior (e.g., tools accepting
# format="json"). So, direct call of g.region to obtain the number of rows:
assert tools.g_region(flags="p", format="json")["rows"] == 1
```

### run_command with returncode

```python
# original:
assert (
    gs.run_command(
        "r.mask.status", flags="t", errors="status"
    )
    == 1
)
```

```python
# replacement:
tools = Tools(errors="ignore")
assert tools.run("r.mask.status", flags="t").returncode == 1
assert tools.r_mask_status(flags="t").returncode == 1
```

### run_command with overwrite

```python
# original:
gs.run_command(
    "r.random.surface",
    output="surface",
    seed=42,
    overwrite=True,
)
```

```python
# replacement:
tools = Tools()
tools.r_random_surface(output="surface", seed=42, overwrite=True)
# or with global overwrite:
tools = Tools(overwrite=True)
tools.r_random_surface(output="surface", seed=42)
```

wenzeslaus:
I updated documentation of the Tools class and documentation of the test functions. I also updated the PR description to reflect the latest state.

Does anyone have any unanswered questions about this PR or the Tools API in general? I'm leaning towards moving it from _grass.experimental.tools_ to _grass.tools_.

wenzeslaus:
This is now ready for review. The original description is updated. A follow up PR #6015 will add this to the generated tool documentation.

python/grass/tools/session_tools.py
```diff
@@ -0,0 +1,308 @@
...
+
+"""The module provides an API to use GRASS tools (modules) as Python functions"""
+
+from __future__ import annotations
+
+import os
+
+import grass.script as gs
+
+from .support import ToolResult, ParameterConverter, ToolFunctionResolver
```
echoix: isort+formatting gives:

```suggestion
from .support import ParameterConverter, ToolFunctionResolver, ToolResult
```

python/grass/tools/tests/grass_tools_session_tools_test.py
```diff
@@ -0,0 +1,812 @@
...
+
+import io
+import os
+
+import pytest
+
+import grass.script as gs
+from grass.experimental.mapset import TemporaryMapsetSession
+from grass.tools import Tools
+from grass.exceptions import CalledModuleError
```
echoix: isort+formatting gives:

```suggestion
import io
import os

import pytest

import grass.script as gs
from grass.exceptions import CalledModuleError
from grass.experimental.mapset import TemporaryMapsetSession
from grass.tools import Tools
```

python/grass/tools/support.py
```diff
@@ -0,0 +1,252 @@
...
+"""
+
+from __future__ import annotations
+
+import json
+import shutil
+from io import StringIO
+
+
+import grass.script as gs
```
echoix: isort+formatting gives:

```suggestion
from __future__ import annotations

import json
import shutil
from io import StringIO

import grass.script as gs
```

python/grass/tools/tests/conftest.py
```diff
@@ -0,0 +1,15 @@
+"""Fixtures for grass.experimental.tools"""
```
echoix: Not in experimental anymore

python/grass/tools/support.py
```diff
@@ -0,0 +1,252 @@
...
+        return self.text_split(None)
+
+    def text_split(self, separator=None) -> list:
+        """Parse text output read as list separated by separators
+
+        Any leading or trailing newlines are removed prior to parsing.
+        """
+        # The use of strip is assuming that the output is one line which
+        # ends with a newline character which is for display only.
+        return self.text.split(separator)
```
echoix: Pylance and Mypy complains because above, `self.text` is annotated with `str | None`. If `self.text` can be none, it will cause an exception.

A test of this function with `self.text = None` could be useful, and would specify the wanted handling when this occurs

python/grass/tools/support.py
```diff
@@ -0,0 +1,252 @@
...
+class ToolFunctionResolver:
+    def __init__(self, *, run_function, env):
+        self._run_function = run_function
+        self._env = env
+        self._names = None
+
+    # def __getattr__(self, name):
+    #    self.get_function(name, exception_type=AttributeError)
+
+    def get_function(self, name, exception_type):
```
echoix: Pylint complains that `exception_type` is unused. There's two references of this function. Where it is defined, here, and in session_tools.py, in the `Tools.__getattr__`, that returns `return self._name_resolver.get_function(name, exception_type=AttributeError)`. The exception argument isn't used as far as I understand

python/grass/tools/session_tools.py
```diff
@@ -0,0 +1,308 @@
...
+        return list(static_attrs) + self._name_resolver.names()
+
+    def __enter__(self):
+        """Enter the context manager context.
+
+        :returns: reference to the object (self)
+        """
+        return self
+
+    def __exit__(self, type, value, traceback):
```
echoix: The signature in the python docs is usually:
```suggestion
    def __exit__(self, exc_type, exc_value, traceback):
```

https://docs.python.org/3/reference/datamodel.html#object.__exit__
In typeshed, it is most often like that, but not always, like in  https://github.com/python/typeshed/blob/7c7f8cd184ccd96980477e56c7496d1d87784d1d/stdlib/contextlib.pyi

Otherwise, Pylint is complaining with [W0622 / redefined-builtin](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/redefined-builtin.html) because of `type` that is a built-in function

echoix (commented):
Some comments

python/grass/tools/support.py
```diff
@@ -179,32 +188,44 @@ def conversion(value):
...
+        """
         return self.text_split(",")
 
     @property
     def space_items(self) -> list:
-        """Text output read as whitespace-separated list"""
-        return self.text_split(None)
+        """Text output read as whitespace-separated list
+
+        Empty or no output results in an empty dictionary.
```
echoix:
```suggestion
        Empty or no output results in an empty list.
```
Do I understand well that it is not a dictionary but really a list that is returned by text_split?

python/grass/tools/support.py
```diff
@@ -179,32 +188,44 @@ def conversion(value):
...
+        """
+        return self.text_split(separator=None)
 
     def text_split(self, separator=None) -> list:
         """Parse text output read as list separated by separators
 
         Any leading or trailing newlines are removed prior to parsing.
+        Empty or no output results in an empty dictionary.
```
echoix:
```suggestion
        Empty or no output results in an empty list.
```
Same question

python/grass/tools/support.py
wenzeslaus: Good catch. I polished handling of these cases and added test cases.

python/grass/tools/support.py
wenzeslaus: It was indeed unused. I added tests for that and some additional tests previously covered only indirectly through usage in Tools.

python/grass/tools/support.py
wenzeslaus: Ah, copy-paste all over the place. Yes, it is a list. All of these are empty lists or dicts, except JSON because `import json; json.loads('')` produces an error, so this is kept in a way that when accessing the attribute *json* or accessing the values through indexing, exception is produced.

python/grass/tools/support.py
```diff
@@ -0,0 +1,272 @@
...
+    def __getattr__(self, name):
+        return self.get_function(name, exception_type=AttributeError)
+
+    @staticmethod
+    def levenshtein_distance(text1: str, text2: str) -> int:
+        if len(text1) < len(text2):
+            return ToolFunctionResolver.levenshtein_distance(text2, text1)
+
+        if len(text2) == 0:
+            return len(text1)
```
echoix, line 100: According to code coverage, almost all 100% of coverage is there. But,?there's missing a test with text1 having some length, and text2 having a length of 0, to trigger the return.

python/grass/tools/support.py
```diff
@@ -0,0 +1,272 @@
...
+            # We are testing just std out and letting rest to the parse and the user.
+            # This makes no assumption about how JSON is produced by the tool.
+            try:
+                self._cached_json = json.loads(self._stdout)
+                return self._cached_json
+            except json.JSONDecodeError as error:
+                if self._kwargs and self._kwargs.get("format") == "json":
+                    raise
+                if self._command and "format=json" in self._command:
+                    raise
```
echoix, line 241: These two raises (not the third one), isn't triggered in any tests

python/grass/app/cli.py
```diff
@@ -20,10 +20,47 @@
...
+                elif any(item in command for item in SPECIAL_FLAGS):
+                    # This is here basically because of how --json behaves,
+                    # two JSON flags are accepted, but --json currently overridden by
+                    # other special flags, so later use of --json in tools will fail
+                    # with the other flags active.
+                    tools.no_nonsense_run_from_list(command)
+                else:
+                    tools.run_from_list(command)
+            except subprocess.CalledProcessError as error:
+                return error.returncode
```
echoix: According to code coverage, almost all new code of this file is touched by tests. But no test uses `subcommand_run_tool()`

python/grass/app/cli.py
```diff
@@ -120,5 +169,12 @@ def main(args=None, program=None):
...
 
-    parsed_args = parser.parse_args(args)
+    # Parsing
+    parsed_args, other_args = parser.parse_known_args(args)
+    # Standard help already exited, but we need to handle tools separately.
+    if parsed_args.subcommand == "run":
+        if parsed_args.tool is None and parsed_args.help:
+            run_subparser.print_help()
+            return 0
+        return parsed_args.func(parsed_args, other_args, help=parsed_args.help)
```
echoix: According to code coverage, no tests enter subcommand "run"

python/grass/tools/support.py
wenzeslaus: Fixed in 6b2b4fa.

python/grass/tools/support.py
wenzeslaus: Now, I'm testing this in and the other symmetrical case in a separate test.

python/grass/tools/support.py
wenzeslaus: Now, the four cases (also the already indirectly tested cases) are tested explicitly.

python/grass/app/cli.py
wenzeslaus: I added tests for the four cases here. This improved quality of the code because now the return code is passed regardless of the value. Before, the code assumed that 0 is the only possible non-error code (which _is_ a correct assumption at this point).

python/grass/app/cli.py
wenzeslaus: I'm testing the actual command with main, so this is now covered by the tests of subcommand_run_tool.

wenzeslaus:
Thanks for looking at the coverage @echoix. I still struggle with evaluating it. This should be pretty well covered and my goal indeed is to be at 100% coverage or very close to it.

echoix:
> Thanks for looking at the coverage @echoix. I still struggle with evaluating it. This should be pretty well covered and my goal indeed is to be at 100% coverage or very close to it.

Indeed, 100% is not always desirable. I just took a quick look, and indeed, almost 100% patch coverage, except two cases, one I think is more easily testable. I’ll write a little comment in these places

python/grass/app/cli.py
```diff
@@ -20,10 +20,48 @@
...
+        project_path = Path(tmp_dir_name) / project_name
+        gs.create_project(project_path)
+        with gs.setup.init(project_path) as session:
+            tools = Tools(session=session, capture_output=False)
+            try:
+                # From here, we return the subprocess return code regardless of its
+                # value. Error states are handled through exceptions.
+                if print_help:
+                    # We consumed the help flag, so we need to add it explicitly.
+                    return tools.call_cmd([*command, "--help"]).returncode
```
echoix, line 55: This return isn't covered in a test yet

python/grass/tools/support.py
```diff
@@ -0,0 +1,276 @@
...
+
+    @property
+    def text(self) -> str | None:
+        """Text output as decoded string"""
+        if self._text is not None:
+            return self._text
+        if self._stdout is None:
+            return None
+        if isinstance(self._stdout, bytes):
+            decoded_stdout = gs.decode(self._stdout)
```
echoix, line 160: The bytes case isn't tested (yet), maybe on Windows it will and I'm just too early.

Otherwise, maybe harder to test if we don't really use bytes anywhere.

echoix (dismissed):
I'll still approve if you think you don't need to add anything, as it's the cherry on top

python/grass/app/cli.py
wenzeslaus: Good catch. It was and I rewrote it to something else.

python/grass/tools/support.py
wenzeslaus: Testing the object directly make it possible to test for things like that, so added.

wenzeslaus:
Not only that this is ready, but #6111 (replacing grass.script by grass.tools in written documentation) and #6015 (adding grass.tools to generated tools documentation) are close to going in after this.

python/grass/app/tests/grass_app_cli_test.py
```diff
@@ -6,11 +6,50 @@
 
 
 def test_cli_help_runs():
+    """Check help the main command"""
```
echoix:
```suggestion
    """Check help of the main command"""
```

echoix:
> Not only that this is ready, but #6111 (replacing grass.script by grass.tools in written documentation) and #6015 (adding grass.tools to generated tools documentation) are close to going in after this. 

I'll make sure this gets merged for the weekend

echoix:
You made it! 100% patch coverage: 
IMG_4345

echoix:
@wenzeslaus Do you want to prepare the commit message?

petrasovaa (approved):
I look forward to start using the new API.

PR #2929, opened 2023-04-21
g.extension: fix extracting Git version string

tmszi:
**Describe the bug**
Installing extension via g.extension module fail.

**To Reproduce**

1. Try to install some extension `g.extension db.join`
2. See error

```
in _get_version
    return float(
           ^^^^^^
ValueError: could not convert string to float: 'Git-143)'
```

**Expected behavior**
Installing extension via g.extension module should work without error message.

**System description:**

- Operating System: macOS
- GRASS GIS version 8.3.dev

**Additional context**
To be backported with https://github.com/OSGeo/grass/pull/2895.

### DIFF
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -218,6 +218,7 @@ class GitAdapter:
         self,
         addons=[],
         url="https://github.com/osgeo/grass-addons",
+        git="git",
         working_directory=None,
         official_repository_structure=True,
         major_grass_version=None,
@@ -227,6 +228,8 @@ class GitAdapter:
     ):
         #: Attribute containing list of addons names
         self._addons = addons
+        #: Attribute containing Git command name
+        self._git = git
         #: Attribute containing the URL to the online repository
         self.url = url
         self.major_grass_version = major_grass_version
@@ -240,6 +243,8 @@ class GitAdapter:
 
         # Check if working directory is writable
         self.__check_permissions()
+        # Check if Git is installed
+        self._is_git_installed()
 
         #: Attribute containing available branches
         self.branches = self._get_branch_list()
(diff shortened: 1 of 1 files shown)


scripts/g.extension/g.extension.py
```diff
@@ -256,15 +261,21 @@ def __init__(
...
+                _("Failed to get Git version.\n{error}").format(
+                    gs.decode(stderr),
+                )
             )
-        )
+        git_version = re.search(r"\d+.(\d+.\d+|\d+)", gs.decode(git_version))
+        if not git_version:
+            gs.fatal(_("Failed to get Git version."))
+        git_version = git_version.group()
+        if git_version.count(".") == 2:
```
ninsbl, line 276: Unfortunately, I cannot test this weekend. Using re looks like a good choice in general. Is there any chance that count(".") != 2?

nilason:
Mac runner fails with `WARNING: The download of the commit from the GitHub API server wasn't successful, <rate limit exceeded>. ` Wasn't the use of git a way to circumvent this?

scripts/g.extension/g.extension.py
tmszi: Regular expression `git_version = re.search(r"\d+.(\d+.\d+|\d+)", gs.decode(git_version))` catch following Git version example:

1. major.minor.patch -> 2.39.2
2.  major.minor -> 2.39

Condition `if git_version.count(".") == 2:` handle first case (line 277 extract major.minor string from major.minor.patch string 2.39.2 -> 2.39). 

https://github.com/OSGeo/grass/blob/b5a119a90050f6f79a3d0fc4e332bc3db897c581/scripts/g.extension/g.extension.py#L276-L277

if `git_version` var string value is **major.minor** string e.g. 2.39, conversion to float is done on the line 278.

https://github.com/OSGeo/grass/blob/b5a119a90050f6f79a3d0fc4e332bc3db897c581/scripts/g.extension/g.extension.py#L278

tmszi:
> Mac runner fails with `WARNING: The download of the commit from the GitHub API server wasn't successful, <rate limit exceeded>. ` Wasn't the use of git a way to circumvent this?

Yes you are right, but it is another PR #2717.

scripts/g.extension/g.extension.py
tmszi:
```python
import re

In [7]: re.search(r"\d+.(\d+.\d+|\d+)", "git version 2.39.2 (Apple Git-143)").group()
Out[7]: '2.39.2'

In [8]: re.search(r"\d+.(\d+.\d+|\d+)", "git version 2.39 (Apple Git-143)").group()
Out[8]: '2.39'
```

nilason:
Have not tested, but the regex looks good to me!

PR #2937, opened 2023-04-26
WIP: pygrass: add get_json_dict function

ninsbl:
This PR adds a `get_json_dict` function to the pygrass Module interface.

Motivation is to make it easier to create JSON intput to the actinia REST API for GRASS GIS.

Here is a usage example:
```
import json
from grass.pygrass.modules.interface import Module

json.dumps(Module("r.info", map="elevation", run_=False).get_json_dict())
```

Suggestions for changes are most welcome...

### DIFF
--- a/python/grass/pygrass/modules/interface/module.py
+++ b/python/grass/pygrass/modules/interface/module.py
@@ -1,15 +1,7 @@
-from __future__ import (
-    nested_scopes,
-    generators,
-    division,
-    absolute_import,
-    with_statement,
-    print_function,
-    unicode_literals,
-)
-import sys
-from multiprocessing import cpu_count, Process, Queue
 import time
+
+from multiprocessing import cpu_count, Process, Queue
+from itertools import zip_longest
 from xml.etree.ElementTree import fromstring
 
 from grass.exceptions import CalledModuleError, GrassError, ParameterError
@@ -21,13 +13,6 @@ from .flag import Flag
 from .typedict import TypeDict
 from .read import GETFROMTAG, DOC
 from .env import G_debug
-
-if sys.version_info[0] == 2:
-    from itertools import izip_longest as zip_longest
-else:
-    from itertools import zip_longest
-
-    unicode = str
 
 
 def _get_bash(self, *args, **kargs):
@@ -528,9 +513,7 @@ class Module(object):
     """
 
     def __init__(self, cmd, *args, **kargs):
-        if isinstance(cmd, unicode):
-            self.name = str(cmd)
-        elif isinstance(cmd, str):
+        if isinstance(cmd, str):
             self.name = cmd
         else:
             raise GrassError("Problem initializing the module {s}".format(s=cmd))
@@ -539,8 +522,9 @@ class Module(object):
             get_cmd_xml = Popen([cmd, "--interface-description"], stdout=PIPE)
         except OSError as e:
             print("OSError error({0}): {1}".format(e.errno, e.strerror))
-            str_err = "Error running: `%s --interface-description`."
-            raise GrassError(str_err % self.name)
+            raise GrassError(
+                "Error running: `{} --interface-description`.".format(self.name)
+            )
         # get the xml of the module
         self.xml = get_cmd_xml.communicate()[0]
         # transform and parse the xml into an Element class:
@@ -675,7 +659,7 @@ class Module(object):
                 # verbose and quiet) work like parameters
                 self.flags[key].value = val
             else:
-                raise ParameterError("%s is not a valid parameter." % key)
+                raise ParameterError("{} is not a valid parameter.".format(key))
 
     def get_bash(self):
         """Return a BASH representation of the Module."""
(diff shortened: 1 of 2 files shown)


python/grass/pygrass/modules/interface/module.py
```diff
@@ -755,6 +739,82 @@ def check(self):
...
+            "SQLite": "vector",
+            "CSV": "file",
+            "TXT": "file",
+        }
+        special_flags = ["overwrite", "verbose", "quiet"]
+        skip = ["stdin", "stdout", "stderr"]
+
+        # Check export formats
+        if export and export not in export_dict:
+            raise GrassError("Invalid Export format.")
```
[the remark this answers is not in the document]
ninsbl: This was addressed in [e4af988](https://github.com/OSGeo/grass/pull/2937/commits/e4af98859c1dfff2f45a5fcaba14708405fd9c21)

python/grass/pygrass/modules/interface/module.py
```diff
@@ -539,8 +522,9 @@ def __init__(self, cmd, *args, **kargs):
             get_cmd_xml = Popen([cmd, "--interface-description"], stdout=PIPE)
         except OSError as e:
             print("OSError error({0}): {1}".format(e.errno, e.strerror))
-            str_err = "Error running: `%s --interface-description`."
-            raise GrassError(str_err % self.name)
+            raise GrassError(
+                "Error running: `{} --interface-description`.".format(self.name)
+            )
```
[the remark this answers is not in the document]
ninsbl: This was addressed in [e4af988](https://github.com/OSGeo/grass/pull/2937/commits/e4af98859c1dfff2f45a5fcaba14708405fd9c21)

python/grass/pygrass/modules/interface/testsuite/test_modules.py
```diff
@@ -53,6 +53,44 @@ def test_rsun(self):
...
+
+        Module(
+            "r.slope.aspect",
+            elevation="elevation",
+            slope="slope",
+            aspect="aspect",
+            overwrite=True,
+            verbose=True,
+            run_=False,
+        ).get_json_dict(export="GTiff")
```
wenzeslaus, line 91: Nothing against smoke tests in general, but these tests are unnecessary simple. Please add assertions to move it beyond smoke tests. It does not really tell what is the expected output.

python/grass/pygrass/modules/interface/testsuite/test_modules.py
```diff
@@ -53,6 +53,44 @@ def test_rsun(self):
         out.close()
 
 
+class TestModulesJsonDictExport(TestCase):
```
wenzeslaus, line 56: Consider using pytest for these test. No data needed. Assert functionality suited for for plain values.

wenzeslaus (changes requested):

python/grass/pygrass/modules/interface/testsuite/test_modules.py
echoix: And we can run them faster/in parallel! (I was exploring making pytest run for windows and macOS the other day, but still had problems making it find the pytest I installed on the windows CI)

ninsbl: This test case is now removed and replaced with pytest in new file: `pygrass_modules_interface_json_test.py` which has assertations. Hope that is roughly along your lines of thought...

wenzeslaus (changes requested):
Looking at the actual concept here, why this is actually needed? We already have the same JSON emitted from the tools themselves with `--json`:

```bash
grass --tmp-project XY --exec python
```

```
>>> import subprocess
>>> import json
>>> json.loads(subprocess.run(["r.info", "map=elevation", "--json"], capture_output=True).stdout)
{'module': 'r.info', 'id': 'r.info_1804289383', 'inputs': [{'param': 'map', 'value': 'elevation'}]}
```

This PR's code:

```
>>> from grass.pygrass.modules.interface import Module
>>> Module("r.info", map="elevation", run_=False).get_json_dict()
{'module': 'r.info', 'id': 'r_info_bb8b711437c149fdb00e1a4ccdb77839', 'inputs': [{'param': 'map', 'value': 'elevation'}]}
```

Difference:

```bash
grass --tmp-project XY --exec r.info map=elevation --json | jq > old.json
grass --tmp-project XY --exec python -c \
    'from grass.pygrass.modules.interface import Module; import json; print(json.dumps(Module("r.info", map="elevation", run_=False).get_json_dict()))' \
    | jq > new.json
diff -u old.json new.json
```

```diff
 {
   "module": "r.info",
-  "id": "r.info_1804289383",
+  "id": "r_info_4722a5e5637f47cb8bf561591f8a4435",
   "inputs": [
     {
       "param": "map",
```

ninsbl:
> Looking at the actual concept here, why this is actually needed? We already have the same JSON emitted from the tools themselves with --json

Valid question. My use case for that function is prototyping actinia workflows in jupyter.
Here, multiple modules can be combined in a dictionary, that then can be send as JSON to actinia. Note, that the functon also can add the export section for maps or stdout.
```
from grass.pygrass.modules.interface import Module
Module("r.info", map="elevation", run_=False).get_json_dict(stdout_export="table", stdout_delimiter="|", stdout_id="raster_info")
```
Overall, I think the _get_json()_ function is more user-friendly, at least  in the context of my usecase than the `--json` approach (though I considered that as well). It also alignes withh other similar functions in the modules interface, like _get_bash()_...

wenzeslaus:
> Overall, I think the _get_json()_ function is more user-friendly, at least in the context of my usecase than the `--json` approach (though I considered that as well).

Why you just don't call the tools with `--json` from _get_json_ and add what you need there or in the library? 

As far as the current state, you talk about interface and functionality, but I'm concerned with maintenance. If this would be merged, we would be left with maintenance of two JSON generators which are actinia-specific (so need to at least somewhat follow changes in actinia) and the JSON format is not well specified, at least not within GRASS GIS, further hindering the maintenance.

> It also aligns with other similar functions in the modules interface, like _get_bash()_...

The Bash and Python functions don't duplicate any substantial code in the library. _get_json_ does.

More generally, I was wondering if you even know about `--json` because using it or wrapping it would be an obvious solution, but it is undocumented and you did not mention it in the description.

ninsbl:
I would not be disappointed if the decision is not to merge this. I can add that function to the Module object in my downstream code... No problem.

So please feel free to close this as wontfix, if that is preferable...

wenzeslaus:
Why don't you just call the tool with `--json` in the `get_json_dict` function?

Check other recent grass.pygrass PRs or issues. The overall issue there is that it tries to mimic what the actual command line parser is doing, but it fails for edge cases. Using the tool to figure out these cases would be a better solution, but it requires more substantial rewrite, so I was just doing smaller fixes here and there. Using `--json` would fit with this approach.

echoix:
Resolved conflict

python/grass/pygrass/modules/interface/module.py
```diff
@@ -777,6 +777,109 @@ def check(self):
...
+
+        # Handle outputs
+        outputs = []
+        for key, val in self.outputs.items():
+            if val.value:
+                param = {
+                    "param": key,
+                    "value": (
+                        ",".join(val.value)
+                        if type(val.value) == list
```
echoix:
```suggestion
                        if isinstance(val.value, list)
```

or

```suggestion
                        if type(val.value) is list
```

python/grass/pygrass/modules/interface/module.py
```diff
@@ -777,6 +777,109 @@ def check(self):
...
+                    for flg in self.flags
+                    if self.flags[flg].value and flg not in special_flags + ["help"]
+                ]
+            ),
+            "inputs": [
+                {
+                    "param": key,
+                    "value": (
+                        ",".join(val.value)
+                        if type(val.value) == list
```
echoix:
```suggestion
                        if isinstance(val.value, list)
```

or

```suggestion
                        if type(val.value) is list
```

ninsbl:
Revisiting this old PR...
I fixed the formating issues. One reason for not using `--json` is that handling for overwrite and verbose / quiet would have to be added anyway (in addition to export), as ` --json` does not seem to support them.
That said, I understand that actinia.specific functionality may be better served in another package!?

wenzeslaus:
1. The functionality should definitively go to --json in C. For better or worse, this was designed and added specifically for actinia. For the future, --json should be general enough to accommodate various use cases, including your use case.

2. Try:

```python
>>> from grass.tools import Tools
>>> import grass.script as gs
>>> gs.setup.init("~/grassdata/nc_spm_08_grass7")
>>> tools = Tools()
>>> tools.call_cmd(["r.blend", "first=abc", "second=def", "output=ghi", "--json"]).json
{'module': 'r.blend', 'id': 'r.blend_1804289383', 'inputs': [{'param': 'first', 'value': 'abc'}, {'param': 'second', 'value': 'def'}, {'param': 'percent', 'value': '50'}], 'outputs': [{'param': 'output', 'value': 'ghi'}]}
```

ninsbl:
Replaced by #6697

PR #2939, opened 2023-04-28
contributing: Link Open Collective contribute page

wenzeslaus:
Given how Open Collective page is organzied, the page with clear contributing options is contribute which has all the options and nothing else, not the home (index) page for a collective. GitHub support suggests to change the configuration by adding /contribute to make the change.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,2 +1,2 @@
 ---
-open_collective: grass
+open_collective: grass/contribute


wenzeslaus:
This works as expected, but it makes the displayed link longer because not it includes the `/contribute` part: [opencollective.com/grass/contribute](https://opencollective.com/grass/contribute). It is an improvement, so I will leave it as is.

PR #2942, opened 2023-04-29
i18n: Merge Transifex translations

HuidaeCho:
This PR merges additional translations from Transifex. The merging process was approximately as follows:
```bash
# install new tx CLI
curl -o- https://raw.githubusercontent.com/transifex/cli/master/install.sh | bash

# move to the grass repo
cd .../grass

# create a new branch
git checkout -B transifex_final_merge

cd locale
# new tx seems to require file_filter and just overwrites
# PO files in place; make a backup copy of PO files
cp -a po po.bak
# pull Transifex translations into locale/po/*
tx pull -a -f
# move Transifex translations to transifex/.tx
mv po/* transifex/.tx
# restore the PO backup files
rmdir po
mv po.bak po

# recreate the old transifex/.tx folder structure
(
cd transifex/.tx

mkdir grass7.grasslibspot
for i in grasslibs_*; do
  j=`echo $i | sed 's/grasslibs_//; s/\.po/_translation/'`
  mv $i grass7.grasslibspot/$j
done

mkdir grass7.grassmodspot
for i in grassmods_*; do
  j=`echo $i | sed 's/grassmods_//; s/\.po/_translation/'`
  mv $i grass7.grassmodspot/$j
done

mkdir grass7.grasswxpypot
for i in grasswxpy_*; do
  j=`echo $i | sed 's/grasswxpy_//; s/\.po/_translation/'`
  mv $i grass7.grasswxpypot/$j
done
)

# use this modified version of transifex_merge.sh;
# only does msgcat
transifex/transifex_merge_2.sh

# check for new translations; here, we ignore any files
# that have only line-wrapping changes without new
# translations
#(
#cd po
#for i in *.po; do
#  # if there are any changes in msgstr, keep them
#  if git diff $i | grep -q '^[-+]msgstr\(\|\[[0-9]*\]\) "[^"]'; then
#    echo "$i: New translations found"
#  else # otherwise, copy back old PO file
#    echo "$i: No new translations"
#    cp ../po.bak/$i .
#  fi
#done
#)

# clean up
rm -rf po.bak
```


HuidaeCho:
I might have missed
```
msgstr ""
"new translation"
```
Maybe, we should skip the last "check for new translations"?

neteler:
> I might have missed
> 
> ```
> msgstr ""
> "new translation"
> ```
> 
> Maybe, we should skip the last "check for new translations"?

The problem you mention isn't clear to me.

But the po files look good, thanks for your work.

After merge, weblate will simply update from GitHub and then we continue from there and close our transifex project?

HuidaeCho:
> > I might have missed
> > ```
> > msgstr ""
> > "new translation"
> > ```
> > 
> >   
> > Maybe, we should skip the last "check for new translations"?
> 
> The problem you mention isn't clear to me.

My script above overwrote any merged PO files with the current PO if they don't have any new translation from Transifex to avoid unnecessary commits. This check is simply see if there are any `+msgstr "..."` or `-msgstr "..."`. If there were
```
+msgstr ""
+"new translation"
```
we missed it. A simple solution would be to skip this check and commit everything even including those same msgstr with simply different line wrapping. Then, we would see a lot of false *changes*. I think this latter method should be preferred *if* Weblate can handle massive line-wrapping-only changes.

> 
> But the po files look good, thanks for your work.
> 
> After merge, weblate will simply update from GitHub and then we continue from there and close our transifex project?

At least, that's my plan if everything goes smoothly.

neteler:
> A simple solution would be to skip this check and commit everything even including those same msgstr with simply different line wrapping. Then, we would see a lot of false _changes_. I think this latter method should be preferred _if_ Weblate can handle massive line-wrapping-only changes.

Let's go for this. If Weblate rejects it we may go for the other way.

PR #2947, opened 2023-05-03
README.md: Thanks to all contributors

neteler:
Add an auto-generated section listing all contributors with their icons at the bottom of `README.md`.

Make this README.md file less boring :-)

image

Uses:
https://contrib.rocks/preview?repo=OSGeo%2Fgrass

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,3 +107,11 @@ this issue, clean all the compiled files from the source code:
 ```bash
 make distclean
 ```
+
+
+## Thanks to all contributors ❤
+
+ <a href = "https://github.com/OSGeo/grass/graphs/contributors">
+   <img src = "https://contrib.rocks/image?repo=OSGeo/grass"/>
+ </a>
+


veroandreo (approved):
beautiful :heart:

marisn:
Oh no! How is possible that this got merged?!? As I have enabled pre-commit hook (thanks @nilason ) now I can not commit anything unless I change this code too – a no-no in the middle of a merge.
Either this code needs to be changed or pre-commit linters need an exception.

```
$ git commit
[INFO] Checking merge-conflict files only.
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
markdownlint.............................................................Failed
- hook id: markdownlint
- exit code: 1

README.md:111 MD012/no-multiple-blanks Multiple consecutive blank lines [Expected: 1; Actual: 2]
README.md:114:2 MD033/no-inline-html Inline HTML [Element: a]
README.md:115:4 MD033/no-inline-html Inline HTML [Element: img]

black....................................................................Passed
flake8...................................................................Passed
clang-format.............................................................Passed
yamllint.................................................................Passed
```

neteler:
> Either this code needs to be changed or pre-commit linters need an exception.

Done both, see #2971

wenzeslaus:
I think `--no-verify` is meant for these cases like merge of non-complaint files ([SE](https://stackoverflow.com/questions/7230820/skip-git-commit-hooks)).

wenzeslaus:
> ... How is possible that this got merged?!? ...

> ...`--no-verify` is meant for these cases...

With that said, this should not happen. I understood that a check in CI is not possible yet because the not all files are compliant yet and doing only the "PR diff" is not readily available. Please, correct me if if I'm wrong, @nilason.

nilason:
Temporarily disabling hooks: https://pre-commit.com/#temporarily-disabling-hooks

> With that said, this should not happen.

To that I absolutely agree!

> I understood that a check in CI is not possible yet because the not all files are compliant yet and doing only the "PR diff" is not readily available. Please, correct me if if I'm wrong, @nilason.

You're essentially right, but let me elaborate:

The main problem is that the `superlinter`, which checks (among others) markdown and yaml files are not in complete sync with the settings in `.pre-commit-config.yaml`. Furthermore, the `trailing-whitespace` check does not even have an equivalent, neither does the `end-of-file-fixer` check (the latter is however visible in GH code viewer as a stop sign).
The clang-format, black and flake8 checks are reliable however.

At the time I introduced pre-commit all files in repo were passing the checks (with the rules and exceptions in `.pre-commit-config.yaml`) in ordinary committing usage. It is not, however, possible to do a `pre-commit run --all-files` because the way pre-commit works in that mode: the `.flake8` file is then not used and perhaps the black setting are also ignored.

I figured using superlinter is close enough (and considering its ease of use and with the main advantage to avoid unnecessary CI runs, *everyone* would use pre-commit anyway...), but close enough is obviously not good enough!

Back then I flagged for the the possible solution for this might be https://pre-commit.ci, but I haven't got to it to explore how that might be implemented.

Issue #2956, opened 2023-05-12
[Bug] r.univar: zone option introduces empty rows without t-flag


ninsbl:
**Describe the bug**
r.univar introduces empty rows if the zone option is used without t-flag

**To Reproduce**
```
# Prepare input data
$ g.region rast=landclass96 -p
$ r.mapcalc -s expression="test=if(row()>100,2,9)" --o
$ r.category test

# Compare
$ r.univar map=elevation zone=test -t
2||196350|374|62.1162147521973|156.269027709961|94.1528129577637|108.129578388876|108.129578388876|18.7151490791832|350.256805056052|17.308075512767|21231242.7166557|21231242.7166557
9||51975|625|56.0590705871582|155.669967651367|99.610897064209|119.101608795489|119.101608795489|23.4318847141117|549.05322125542|19.6738607908705|6190306.11714554|6190306.11714554

# to
$ r.univar map=elevation zone=test -g | grep zone
zone 2
zone 3
zone 4
zone 5
zone 6
zone 7
zone 8
zone 9

# or
$ r.univar map=elevation zone=test -g | grep zone
zone=2;
zone=3;
zone=4;
zone=5;
zone=6;
zone=7;
zone=8;
zone=9;
```

**Expected behavior**
r.univar should not introduce empty rows / data points if zones are used

**System description (please complete the following information):**

- Operating System: Ubuntu Linux 20.04
- GRASS GIS version 8.3

wenzeslaus:
The fix is merged, closing as fixed. The backport label is enough to say that it needs a backport.

PR #2957, opened 2023-05-12
r.univar: print only statistics for existing zones

ninsbl:
Fix for #2956

### DIFF
--- a/raster/r.univar/stats.c
+++ b/raster/r.univar/stats.c
@@ -116,6 +116,10 @@ int print_stats(univar_stat *stats)
         unsigned int i;
         size_t qpos_25, qpos_75, *qpos_perc;
 
+        /* stats collected for this zone? */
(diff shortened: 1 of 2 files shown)


ninsbl:
If there are no objections I merge this relatively simple fix tonight... The test failure seems unrelated...

wenzeslaus (commented):
This looks okay. I think the current output is wrong because it prints non-existent categories. I'm not sure about how non-existent versus not present in the computational region, but I guess that's the same for `-t`. Adding a test would clarify the desired behavior. The test will become part of code and the commit rather than just commit linking the PR and PR linking the issue which has an example.

ninsbl:
>  Adding a test would clarify the desired behavior. The test will become part of code and the commit rather than just commit linking the PR and PR linking the issue which has an example.

Thanks for looking into this. Right. Now I added a test for this case...

PR #2965, opened 2023-05-17
contributing: Add suggest message to version script

wenzeslaus:
The helper script for udpates of versions suggests a commit message right after the update, but when additional changes need to be done, the message from the script can be lost. This adds a subcommand which can generate the message anytime. The message only makes sense when doing the version update, but the script does not check that. It also does not check (does not know) if the RC was just released. We use different commit message after that, but in that case, the commit happens right after the change from RC to dev, so there is no need to use this subcommand.

### DIFF
--- a/utils/update_version.py
+++ b/utils/update_version.py
@@ -236,6 +236,28 @@ def status(args):
         status_as_yaml(version_info=version_info, today=today, version=version, tag=tag)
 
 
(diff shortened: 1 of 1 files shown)


neteler:
@wenzeslaus Can this please be backported to G8.3.1?

PR #2967, opened 2023-05-17
version: Update Sphinx config under man

wenzeslaus:
Version is updated to 8.4 and references to version are reduced. Word documentation is removed from names. Related duplicated values are reused using variables.


wenzeslaus:
I did run `make html` in `man/sphinx`, but besides that I don't know how to actually test this. Suggestions welcome.

(To actually run it, I created an empty file in the current directory. I got an empty documentation as a result.)

PR #2968, opened 2023-05-17
doc: Sync grass package and wxGUI Sphinx conf

wenzeslaus:
Synchronize Sphix configuration between the grass package and wxGUI code. Update wxGUI to make it work.


wenzeslaus:
Documentation for Python API is generated in CI, so libpython should still work. Merging.

PR #2969, opened 2023-05-17
version: Replace references to 8.2

wenzeslaus:
- [x] Links to latest (current) stable in notebooks use the grass-stable URL.
- [x] singularity: Use tag instead of branch. Use an old tag for the example. Avoid using version in the image name.

8.2 used in examples (no change needed?):

- [x] lib/init/grass.html

Fix based on updates to 8.3 branch (same text or same style):

- [x] general/g.version/g.version.html (example of an output; change needed?)
- [x] rpm/grass.spec (copy the updated file? what is the procedure for syncs or updates?) #2977 
- [x] doc/howto_release.md (code for copy-pasting; prepare for 8.4?) #2980
- [x] doc/development/submitting/general.md (instructions for contributors; generalize to avoid need for updates?) #2978

### DIFF
--- a/singularity/debian/README_debian.md
+++ b/singularity/debian/README_debian.md
@@ -16,7 +16,7 @@ cd grass
 __Build the singularity with__:
 
 ```bash
-sudo singularity build grass_development.simg singularity/debian/singularity_debian
+sudo singularity build grass_gis.simg singularity/debian/singularity_debian
 ```
 
 __To build a stable version__:
@@ -24,17 +24,17 @@ __To build a stable version__:
 change to the releasebranch or tag you want to build:
 
 ```bash
-git checkout remotes/origin/releasebranch_8_2
+git checkout 8.2.0
 ```
 
 and build and enter with:
 
 ```bash
-sudo singularity build grass_8_2.simg singularity/debian/singularity_debian
+sudo singularity build grass_gis.simg singularity/debian/singularity_debian
 ```
 
 The image can be used as:
 
 ```bash
-singularity exec containers/grass_8.2.simg grass --version
+singularity exec containers/grass_gis.simg grass --version
 ```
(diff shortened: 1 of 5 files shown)


landam:
@wenzeslaus Still relevant?

wenzeslaus:
Given the lack of ideas and suggestions here, I just updated the numbers from 8.2 to 8.4. I was trying to avoid these changes elsewhere because these need to be updated with each release. I fixed some dates, years and dropped the dev suffix.

This is now ready for review and inclusion into 8.4.

singularity/debian/README_debian.md
```diff
@@ -16,25 +16,25 @@ cd grass
...
+sudo singularity build grass_gis.simg singularity/debian/singularity_debian
 ```
 
 __To build a stable version__:
 
 change to the releasebranch or tag you want to build:
 
 ```bash
-git checkout remotes/origin/releasebranch_8_2
+git checkout 8.2.0
```
neteler, line 27: Since 8.4 will be the new stable very soon, better use `8.4.0` here?

wenzeslaus: Maybe I was not looking for underscores here.

Sure, between 8.2 and 8.4, this PR is trying to update to 8.4.

However, better than updating is to remove these references which needed to be updated every time the number changes. Including micro version (8.2.0) creates even more need for updates than just the branch number (8_2): every release (multiple times a year) as opposed to every major or minor (once a year or so).

Any suggestions for removing the need to update? My suggestions:

```bash
git checkout 3.4.0
git checkout {version}
git checkout {version number or branch name}
git checkout {tag or branch}
```

As for the desired frequency of updates, note that this PR is on main updating from 8.2.

wenzeslaus, line 27:
```suggestion
git checkout {tag or branch}
```

neteler (approved):
Looks good, thanks.

PR #2972, opened 2023-05-19
init: Do not report creating new mapset for tmp mapset

wenzeslaus:
When running grass with --tmp-mapset, printing a message about mapset creation is not needed. Mapset is created and then deleted just as the user asked.


wenzeslaus:
Given that this PR got zero traction, I don't think it is a priority for backporting. It is not even a bug fix, so it should not be backported anyway. I'm more worried about finishing things like #2969.

PR #2978, opened 2023-05-23
contributing: Align backporting example with text

wenzeslaus:
The first part of the paragraph talks about backporting to all branches in between main and furthest backport. However, the example pushed the idea of backport to the furthest instead. The new example is now only about backporting to the in-between branch while the paragraph addresses the backport in general and packport to the closest branch first.

The new example uses version numbers with the major version 3 because it is only an example, so it does not need to be updated.

### DIFF
--- a/doc/development/submitting/general.md
+++ b/doc/development/submitting/general.md
@@ -147,11 +147,11 @@ available), and compilation of the whole source tree (ideally after the
 the original commit to "main". Also note that not all these steps have to be
 done manually, you can take an advantage of our [​CI](https://github.com/OSGeo/grass/actions).
 
-Often there is more than one active release branch, if you are backporting, make
-sure you always backport to all the branches between "main" and the furthest
-branch you are backporting to. For example, let's say we have "main" and release
-branches 3.5 and 3.6, if you backport to 3.5, you should backport to 3.6, too.
-You can also choose to backport only to the closest branch, in our example 3.6.
+Often there is more than one active release branch. You can also choose to backport
+only to the closest branch. If you are backporting to other release branches than
+just the closets one, make sure you always backport to all the branches between "main"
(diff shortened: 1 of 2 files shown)


neteler:
> grass.spec: update to latest grass.spec from Fedora git

I guess this should be in a separate PR.

wenzeslaus:
> I guess this should be in a separate PR.

An in fact it is: #2977. I must have used that branch instead of main to create this one branch. I'll just revert that commit.

doc/development/submitting/general.md
```diff
@@ -149,9 +149,9 @@ done manually, you can take an advantage of our [​CI](https://github.com/OSGeo
 
 Often there is more than one active release branch, if you are backporting, make
 sure you always backport to all the branches between "main" and the furthest
-branch you are backporting to. For example, let's say we have "main" and branches
-8.2 and 7.8, if you backport to 8.2, you should also consider to backport to 7.8.
-You can also choose to backport only to the closest branch, in our example 8.2.
+branch you are backporting to. For example, let's say we have "main" and release
+branches 3.5 and 3.6, if you backport to 3.5, you should backport to 3.6, too.
```
neteler, line 153: Honestly, I find it confusing to first mention 3.5 and then 3.6.
Normally we *back*port, i.e. we go from higher to lower numbers.

wenzeslaus: I did that because that's what the text before the example is describing, but I think I can modify it a bit to still keep it relevant to the text. Or maybe, the whole paragraph needs to be changed like putting the last sentence first. (The previous example was example of something else than what the text was describing.)

wenzeslaus: The paragraph is now different and the example lists branches in the reverse order.

wenzeslaus:
This is ready for review. (The conversation can be resolved, too.)

PR #2997, opened 2023-06-03
Show ModelRelation in white on dark mode

landam:
## Before PR

Screenshot from 2023-06-03 12-06-47

## After PR

Screenshot from 2023-06-03 12-06-05

Note: Black brush is used in dark mode only when exporting model into image:

dark_fixed


wenzeslaus (approved):
Works for me.

## Screenshot
Screenshot from 2023-06-03 14-07-44

## Exported
test_model

PR #3005, opened 2023-06-04
wxGUI/gmodeler: Support actinia export

pesekon2:
Added support for https://github.com/actinia-org/actinia-core


neteler:
I have tested this PR by

- adding "data": lakes (North Carolina sampling dataset)
- adding `r.buffer` with 100m buffer
- the output is then self-added once clicking on the `r.buffer` icon

Then I didn't really know where to find the actinia export (not visible yet under File > Export which would be nice).

Using the `Script` tab I then hoped to find it but got this error:

```
(Thu Oct 12 13:10:51 2023) Command finished (0 sec)                             
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-
linux-gnu/gui/wxpython/gmodeler/frame.py", line 245, in
OnPageChanged

self.pythonPanel.RefreshScript()
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-
linux-gnu/gui/wxpython/gmodeler/frame.py", line 2104, in
RefreshScript

self.write_object(fd, self.parent.GetModel())
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-
linux-gnu/gui/wxpython/gmodeler/model.py", line 3230, in
__init__

self._writePython()
  File "/home/mneteler/software/grass_main/dist.x86_64-pc-
linux-gnu/gui/wxpython/gmodeler/model.py", line 3290, in
_writePython

for flag in item.GetParameterizedParams()["flags"]:

^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
^
AttributeError
:
'ModelData' object has no attribute 'GetParameterizedParams'

```

pesekon2:
Hi @neteler:

Thanks for the test. You are right, it is indeed missing in `File > Export...` Currently, the export is only in the `Script` tab, as `Save as...`/`Save`. I'll add it to `File > Export...`.

The second problem you find, I am also able to reproduce. I didn't test adding data to the model previously, so the issue is connected to that. But testing it, adding data to the model apparently destroys even a normal Python export, not only the actinia one. A separate issue should be opened for the data inclusion -- feel free to do so (and even tag me, I should be able to take a look at that within a few days).

pesekon2:
@neteler: Both issues should fixed now. [36536c9](https://github.com/OSGeo/grass/pull/3005/commits/36536c96a244a65c9586867a12774af098410e0e) adds it to the `File` menu and [f332fb0](https://github.com/OSGeo/grass/pull/3005/commits/f332fb091ad82e15f47a24c53131f082066eefba) fixes the issue with models containing `ModelData`-type items.

The issue with models including `ModelData`-type items was a general bug in `g.gui.gmodeler`. There is a parallel PR fixing this issue for the already-existing exports: #3209.

landam (approved):

PR #3008, opened 2023-06-04
i.signatures: Add imagery classifier signature management module

marisn:
For the first time of GRASS history this PR tries to introduce a tool capable of managing signature files.


marisn:
One thing to discuss: how should "type" values be named? At the moment they follow internal naming of signature files (i.gensig -> "sig"; i.gensigset -> "sigset") but it might be hard to get it right for users. An alternative could be naming after consumer ("maxlik" and "smap").

Other issue is that the GUI at the moment can not handle signature file name selection as it requires linking "type" with "copy", "rename" and "remove" to list signatures of selected type.

veroandreo:
> One thing to discuss: how should "type" values be named? At the moment they follow internal naming of signature files (i.gensig -> "sig"; i.gensigset -> "sigset") but it might be hard to get it right for users. An alternative could be naming after consumer ("maxlik" and "smap").

I'd prefer to name them as the module that generates them, i.e., i.gensig -> "sig"; i.gensigset -> "sigset", seems more consistent to me.

wenzeslaus:
The sig and sigset are indeed confusing, but the producer modules are named that way as well as the internal files, so using sig and sigset is hopefully less confusion overall. Perhaps the consumer modules need to be really good at documenting what they need.

imagery/i.signatures/i.signatures.html
```diff
@@ -0,0 +1,43 @@
...
+<a href="i.svm.train.html">i.svm.train</a>.
+The module can perform multiple actions per run. The order of execution
+is "copy", "remove", "rename". If no actions are specified, then the module
+lists (prints) all signature files of specified type.
+
+<h2>EXAMPLES</h2>
+
+Print names of all signature files:
+<div class="code"><pre>
+i.signatures
```
wenzeslaus, line 15: Historically and in terms of consistency, we have issues with calls to tools without parameters. Should it start GUI or do something else? I suggest `i.signatures -p` for this.

marisn:
> Perhaps the consumer modules need to be really good at documenting what they need.

Consumer module documentation lists the correct workflow. Also the parser code will not let to pass wrong signature file. This is only an issue for this module as it allows to operate on all of them.

imagery/i.signatures/i.signatures.html
```diff
@@ -0,0 +1,43 @@
...
+</pre></div>
+
+Delete signature file called "foo" of type "sig" (i.gensig / i.maxlik).
+<div class="code"><pre>
+i.signatures remove=foo type=sig
+</pre></div>
+
+Copy signature file "bar" from mapset "baz" to current mapset
+<div class="code"><pre>
+i.signatures copy=bar@baz,best_version type=sigset
```
wenzeslaus, line 30: The copy=from,to really makes more sense for g.copy which copies many different things, but it is not necessarily the best example of an interface. In Bash it does not matter much. In Python, ends up as awkward:

```python
gs.run_command("i.signatures", copy=(old_name, new_name), type="sigset")  
```

As opposed to:

```python
gs.run_command("i.signatures", from=old_name, to=new_name, type="sigset")  
```

Maybe we can discuss in-person.

imagery/i.signatures/main.c
```diff
@@ -0,0 +1,225 @@
...
+    G_option_requires(parms.copy, parms.type, NULL);
+    G_option_requires(parms.rename, parms.type, NULL);
+    G_option_requires(parms.remove, parms.type, NULL);
+
+    if (G_parser(argc, argv))
+        exit(EXIT_FAILURE);
+
+    if (!print->answer && !parms.copy->answer && !parms.rename->answer &&
+        !parms.remove->answer)
+        G_fatal_error(_("At least one action should be specified"));
```
wenzeslaus, line 170: G_option_required() https://grass.osgeo.org/programming8/defs_2gis_8h.html#a5cabe9d5db03c21390a59fe346e6a5c4

wenzeslaus (changes requested):

imagery/i.signatures/i.signatures.html
marisn: I changed to use a flag for printing

imagery/i.signatures/i.signatures.html
marisn: As this tool is not expected to be used too often, we'll keep it as is as each of alternatives has its downsides.

marisn:
If there are no objections, I'll merge.

imagery/i.signatures/main.c
```diff
@@ -0,0 +1,225 @@
...
+ * COPYRIGHT:    (C) 2023 by Maris Nartiss and the GRASS Development Team
+ *
+ *               This program is free software under the GNU General Public
+ *               License (>=v2). Read the file COPYING that comes with GRASS
+ *               for details.
+ *
+ *****************************************************************************/
+#include <stdlib.h>
+#include <string.h>
+#include <stdbool.h>
```
nilason, line 16: Is including stdbool.h necessary in this file?

marisn: Not any more. Removed.

Issue #3013, opened 2023-06-04
[Bug] Default single-window size is too small to see anything


pesekon2:
**Describe the bug**

When I open the GUI using `g.gui`, the default size is extremely small and covers more-or-less only the tabs. See the screenshot below.

**To Reproduce**

`g.gui`

**Screenshots**

![image](https://github.com/OSGeo/grass/assets/17025405/17ed06b1-fc20-4ae4-a16b-73121ef90396)

**System description:**

```
System Info                                          
Operating System: Debian testing                           
GRASS version: 8.4.0dev      
Code revision: e887353f6c
Python: 3.11.2
wxPython: 4.2.0 gtk3 (phoenix)
wxWidgets: 3.2.2
libgtk-3-0: 3.24.37
```

veroandreo:
I get the same after a clean recompilation and switch to 8.4dev. At the end of the compilation I get: 
```
--------------------------------------------------
Started compilation: Sun Jun  4 01:30:28 PM -03 2023
--
Errors in:
/home/veroandreo/software/grass-dev/python/libgrass_interface_generator
--
In case of errors please change into the directory with error and run 'make'.
If you get multiple errors, you need to deal with them in the order they
appear in the error log. If you get an error building a library, you will
also get errors from anything which uses the library.
--
Finished compilation: Sun Jun  4 01:32:15 PM -03 2023
make: *** [Makefile:71: default] Error 1
```
and when starting grassdev I get tons of GTK errors in the terminal.

```
System Info                                                                     
GRASS version: 8.4.0dev                                                         
Code revision: e887353f6c                                                       
Build date: 2023-06-04                                                          
Build platform: x86_64-pc-linux-gnu                                             
GDAL: 3.4.3                                                                     
PROJ: 8.2.1                                                                     
GEOS: 3.10.2                                                                    
SQLite: 3.36.0                                                                  
Python: 3.10.11                                                                 
wxPython: 4.0.7                                                                 
Platform: Linux-6.2.15-100.fc36.x86_64-x86_64-with-glibc2.35                    
```

tmszi:
I have not been able to reproduce this behavior on Gentoo GNU/Linux distribution with Python 3.10.11 and wxPython
4.2.0 gtk3 (phoenix) wxWidgets 3.2.2.1.

tmszi:
> and when starting grassdev I get tons of GTK errors in the terminal.
> 
> ```
> System Info                                                                     
[...]

PR #3024, opened 2023-06-05
wxGUI: Default Single-Window size is too small

lindakladivova:
Fixes https://github.com/OSGeo/grass/issues/3013

### DIFF
--- a/gui/wxpython/gui_core/preferences.py
+++ b/gui/wxpython/gui_core/preferences.py
@@ -1830,10 +1830,18 @@ class PreferencesDialog(PreferencesBaseDialog):
         #
         # update default window dimension
         #
-        if (
-            self.settings.Get(group="general", key="defWindowPos", subkey="enabled")
-            is True
-        ):
+        single_window = self.settings.Get(
+            group="general",
+            key="defWindowPos",
+            subkey="enabled",
+        )
+        windows_pos = self.settings.Get(
+            group="appearance",
+            key="singleWindow",
+            subkey="enabled",
+        )
+        # Multiple windows mode
+        if windows_pos and not single_window:
             dim = ""
             # layer manager
             pos = self.parent.GetPosition()
@@ -1855,6 +1863,17 @@ class PreferencesDialog(PreferencesBaseDialog):
 
             self.settings.Set(
                 group="general", key="defWindowPos", subkey="dim", value=dim
+            )
+        # Single window mode
+        elif windows_pos and single_window:
+            pos = self.parent.GetPosition()
+            size = self.parent.GetSize()
+            dim = f"{pos[0]},{pos[1]},{size[0]},{size[1]}"
+            self.settings.Set(
+                group="general",
+                key="defWindowPos",
+                subkey="dimSingleWindow",
+                value=dim,
             )
 
         return True
--- a/gui/wxpython/gui_core/preferences.py
+++ b/gui/wxpython/gui_core/preferences.py
@@ -1830,51 +1830,48 @@ class PreferencesDialog(PreferencesBaseDialog):
         #
         # update default window dimension
         #
-        single_window = self.settings.Get(
+        windows_pos = self.settings.Get(
             group="general",
             key="defWindowPos",
             subkey="enabled",
         )
-        windows_pos = self.settings.Get(
+        single_window = self.settings.Get(
             group="appearance",
             key="singleWindow",
             subkey="enabled",
         )
-        # Multiple windows mode
-        if windows_pos and not single_window:
-            dim = ""
-            # layer manager
+        if windows_pos:
+            # get dimension of main window
             pos = self.parent.GetPosition()
             size = self.parent.GetSize()
-            dim = "%d,%d,%d,%d" % (pos[0], pos[1], size[0], size[1])
-            # opened displays
+            dim = f"{pos[0]},{pos[1]},{size[0]},{size[1]}"
+
+            # extend dimension by dimensions of opened displays
             for mapdisp in self._giface.GetAllMapDisplays():
                 pos = mapdisp.GetPosition()
                 size = mapdisp.GetSize()
 
                 # window size must be larger than zero, not minimized
-                # when mapdisp is inside single window (panel has no IsIconized), don't save dim
+                # do not save dim when mapdisp is docked within single window
                 if (
-                    hasattr(mapdisp, "IsIconized")
-                    and not mapdisp.IsIconized()
+                    not mapdisp.IsDocked()
                     and (size[0] > 0 and size[1] > 0)
                 ):
-                    dim += ",%d,%d,%d,%d" % (pos[0], pos[1], size[0], size[1])
+                    dim += f",{pos[0]},{pos[1]},{size[0]},{size[1]}"
 
-            self.settings.Set(
-                group="general", key="defWindowPos", subkey="dim", value=dim
-            )
-        # Single window mode
-        elif windows_pos and single_window:
-            pos = self.parent.GetPosition()
-            size = self.parent.GetSize()
-            dim = f"{pos[0]},{pos[1]},{size[0]},{size[1]}"
-            self.settings.Set(
-                group="general",
-                key="defWindowPos",
-                subkey="dimSingleWindow",
-                value=dim,
-            )
+            if single_window:
+                # single window mode
+                self.settings.Set(
+                    group="general",
+                    key="defWindowPos",
+                    subkey="dimSingleWindow",
+                    value=dim,
+                )
+            else:
+                # multi window mode
+                self.settings.Set(
+                    group="general", key="defWindowPos", subkey="dim", value=dim
+                )
 
         return True
 
--- a/gui/wxpython/gui_core/preferences.py
+++ b/gui/wxpython/gui_core/preferences.py
@@ -1853,7 +1853,9 @@ class PreferencesDialog(PreferencesBaseDialog):
 
                 # window size must be larger than zero, not minimized
                 # do not save dim when mapdisp is docked within single window
-                if not mapdisp.IsDocked() and (size[0] > 0 and size[1] > 0):
+                if (not mapdisp.IsDockable() or not mapdisp.IsDocked()) and (
+                    size[0] > 0 and size[1] > 0
+                ):
                     dim += f",{pos[0]},{pos[1]},{size[0]},{size[1]}"
 
             if single_window:
--- a/gui/wxpython/gui_core/preferences.py
+++ b/gui/wxpython/gui_core/preferences.py
@@ -1853,10 +1853,7 @@ class PreferencesDialog(PreferencesBaseDialog):
 
                 # window size must be larger than zero, not minimized
                 # do not save dim when mapdisp is docked within single window
-                if (
-                    not mapdisp.IsDocked()
-                    and (size[0] > 0 and size[1] > 0)
-                ):
+                if not mapdisp.IsDocked() and (size[0] > 0 and size[1] > 0):
                     dim += f",{pos[0]},{pos[1]},{size[0]},{size[1]}"
 
             if single_window:
--- a/gui/wxpython/main_window/frame.py
+++ b/gui/wxpython/main_window/frame.py
@@ -194,10 +194,8 @@ class GMFrame(wx.Frame):
                 self.SetSize((w, h))
             except Exception:
                 pass
-            self.Layout()
-        else:
-            self.Layout()
-
+        
+        self.Layout()
         self.Show()
 
         # load workspace file if requested
--- a/gui/wxpython/main_window/frame.py
+++ b/gui/wxpython/main_window/frame.py
@@ -194,7 +194,7 @@ class GMFrame(wx.Frame):
                 self.SetSize((w, h))
             except Exception:
                 pass
-        
+
         self.Layout()
         self.Show()
 
--- a/gui/wxpython/main_window/frame.py
+++ b/gui/wxpython/main_window/frame.py
@@ -195,13 +195,8 @@ class GMFrame(wx.Frame):
             except Exception:
                 pass
             self.Layout()
-            if w <= globalvar.GM_WINDOW_SIZE[0] or h <= globalvar.GM_WINDOW_SIZE[1]:
-                self.Fit()
         else:
             self.Layout()
-            self.Fit()
-            # does center (of screen) make sense for lmgr?
-            self.Centre()
 
         self.Show()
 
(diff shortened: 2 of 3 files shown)


pesekon2:
Fixes the problem for me. @tmszi does it fix also your `TypeError` from #3013?

lindakladivova:
> Fixes the problem for me. @tmszi does it fix also your `TypeError` from #3013?

@pesekon2 Most likely, it is not going to help. The MenuItem error has nothing to do with the size of Single Window. I would suggest creating a different PR for it.

petrasovaa:
Should this go to 8.3 release? It looks like a an important fix. Why is it draft? The TypeError is unrelated.

lindakladivova:
> Should this go to 8.3 release? It looks like a an important fix. Why is it draft? The TypeError is unrelated.

Yes, it should go to the release 8.3.0, I initially set it up for version 8.3.0.

This is a draft because I would like to test it more (it was very fast written).

landam:
Single Window should probably start maximized by default. @lindakladivova @petrasovaa Any opinion?

tmszi:
> Well @landa, it actually is maximized by default but the problem is that we take the parameter defWindowPos (set by globalvar.GM_WINDOW_SIZE variable) from UserSettings if defined (rows 181 - 196). You can go to Settings -> General -> and check Suppress positioning Layer Manager window, close GRASS and start again and it will launch as maximized.

By default without using this PR, single window mode size depend on the DE, respectively the type of window layout (floating, tiled, monocle) and OS wxGTK, wxMSW.

wxMSW (floating windows layout) default single window size and position:

wxmsw_swm_size_pos_floating

wxGTK (floating windows layout with dwm window manager) default single window size and position (issue #3013):

wxgtk_smw_size_pos_dwm_floating

wxGTK (floating windows layout with Gnome DE) default single window size and position:

wxgtk_smw_size_pos_gnome_floating

Important user settings is **Save current window layout as default** which works as expected. You can change position and size of window (single window mode too).

> IMHO This User setting does not make sense for Single-Window, I would suggest removing rows 181 - 196 in frame/main_window.py. @petrasovaa what do you think?

It is not a good idea, if you remove these lines, the wxGUI single window mode will not respect the client area and will be overlapped by the DE panel(s).

pesekon2:
> Further, I have one more thing here, I would like to try to suppress the GTK warnings as @vero mentioned... it is also painful but I will at least try it and we will see..

This would be the best thing to happen in GRASS for many years, in my eyes (I have the experience that when the users see those warnings and criticals, they consider GRASS to be broken and stop using it). Have had a discussion about that with @wenzeslaus and he can give a verbose, loquacious monologue on that. But we shouldn't make this issue a discussion forum keep sticking to the topic.

tmszi:
Seems it is regression between revisions (respectively GRASS GIS versions)

1. Single window mode launch in the full screen mode by default with GRASS GIS version [8.3.0dev](https://grass.osgeo.org/grass83/binary/linux/snapshot/)

wxGTK (floating windows layout with Gnome DE) default single window size and position:

wxgtk_swm_def_pos_size_floating

```
GRASS world_latlong_wgs84/PERMANENT:~ > g.version -g
version=8.3.0dev
date=2023
revision=567ea066bd
build_date=2023-06-07
```

2. Single window mode doesn't  launch in the full screen mode by default with GRASS GIS version [8.4.0dev](https://grass.osgeo.org/grass84/binary/linux/snapshot/)

wxGTK (floating windows layout with Gnome DE) default single window size and position:

wxgtk_swm_def_pos_size_floating


```
GRASS world_latlong_wgs84/PERMANENT:~ > g.version -g
version=8.4.0dev
date=2023
revision=24fd2a38f4
build_date=2023-06-08
```

lindakladivova:
> 

@tmszi  I will try to explain how I understand the problem, I might be wrong, please correct me.

The size of wx.Frame is in the constructor given by wx.Display().GetGeometry().GetSize() by default (in my case e.g. (1920, 1080)).

But the problem is that after calling self.Fit(), for some reason, the size of the main window is being changed and this is why I removed all self.Fit() callings. Not sure what happened between 8.3.0dev and 8.4.0dev versions but suddenly self.Fit() did not work.

But! When I tested this Pull request I was quite annoyed by the fact that even after removing the Fit() calling, I cannot reach the maximized extent (1920, 1080). The problem was that I was actually still overwriting the desired extent by the extent I had saved in the UserSettings (rows 181-197).

Maybe I am not right, but I think that using just wx.Display().GetGeometry().GetSize() could be enough for setting the size of the main window. (I went from the fact that also other software I use always starts over the whole screen at the beginning), I think that making Single-Window smaller and saving this into the User Settings does not make much sense...

But we can let it as it is, I have no problem with that. I was just critically thinking about the code and if these 17 rows have some meaning..

lindakladivova:
> > Further, I have one more thing here, I would like to try to suppress the GTK warnings as @vero mentioned... it is also painful but I will at least try it and we will see..
> 
> This would be the best thing to happen in GRASS for many years, in my eyes (I have the experience that when the users see those warnings and criticals, they consider GRASS to be broken and stop using it). Have had a discussion about that with @wenzeslaus and he can give a verbose, loquacious monologue on that. But we shouldn't make this issue a discussion forum keep sticking to the topic.

I have spent two hours thinking about how to suppress these GTK warnings when launching GRASS and no results. I just succeed in making more of them :)...

wenzeslaus:
> > > ...I would like to try to suppress the GTK warnings as @vero mentioned... it is also painful but I will at least try it and we will see..
> > 
> > This would be the best thing to happen in GRASS for many years, in my eyes (I have the experience that when the users see those warnings and criticals, they consider GRASS to be broken and stop using it). Have had a discussion about that with @wenzeslaus and he can give a verbose, loquacious monologue on that...
> 
> ...I have spent two hours thinking about how to suppress these GTK warnings when launching GRASS and no results...

#3049 removes the stderr when GUI is started together with GRASS GIS (not for g.gui).

tmszi:
> > 
> 
> @tmszi I will try to explain how I understand the problem, I might be wrong, please correct me.
> 
> The size of wx.Frame is in the constructor given by wx.Display().GetGeometry().GetSize() by default (in my case e.g. (1920, 1080)).
> 
> But the problem is that after calling self.Fit(), for some reason, the size of the main window is being changed and this is why I removed all self.Fit() callings. Not sure what happened between 8.3.0dev and 8.4.0dev versions but suddenly self.Fit() did not work.
> 
> But! When I tested this Pull request I was quite annoyed by the fact that even after removing the Fit() calling, I could not reach the maximized extent (1920, 1080). The problem was that I was actually still overwriting the desired extent by the extent I had saved in the UserSettings (rows 181-197).
> 
> Maybe I am not right, but I think that using just wx.Display().GetGeometry().GetSize() could be enough for setting the size of the main window. (I went from the fact that also other software I use always starts over the whole screen at the beginning), I think that making Single-Window smaller and saving the adapted extent into the User Settings does not make much sense...
> 
> But we can let it as it is, I have no problem with that. I was just critically thinking about the code and if these 17 rows have some meaning..

@lindakladivova First thanks for your explanation.

I attached my patch which solve two issues:

1. Maximize single window mode by default
2. Respect user custom window position and size (custom win pos and size for single win mode is saved as standalone settings separate from multi win mode settings) changed via wxGUI Preference dialog **Save current window layout as default** checkbox widget

`git am 0001-wxGUI-main_window-maximize-window-by-default.patch`

[0001-wxGUI-main_window-maximize-window-by-default.patch.zip](https://github.com/OSGeo/grass/files/11711847/0001-wxGUI-main_window-maximize-window-by-default.patch.zip)

lindakladivova:
> > > 
> > 
> > 
> > @tmszi I will try to explain how I understand the problem, I might be wrong, please correct me.
> > The size of wx.Frame is in the constructor given by wx.Display().GetGeometry().GetSize() by default (in my case e.g. (1920, 1080)).
> > But the problem is that after calling self.Fit(), for some reason, the size of the main window is being changed and this is why I removed all self.Fit() callings. Not sure what happened between 8.3.0dev and 8.4.0dev versions but suddenly self.Fit() did not work.
> > But! When I tested this Pull request I was quite annoyed by the fact that even after removing the Fit() calling, I could not reach the maximized extent (1920, 1080). The problem was that I was actually still overwriting the desired extent by the extent I had saved in the UserSettings (rows 181-197).
> > Maybe I am not right, but I think that using just wx.Display().GetGeometry().GetSize() could be enough for setting the size of the main window. (I went from the fact that also other software I use always starts over the whole screen at the beginning), I think that making Single-Window smaller and saving the adapted extent into the User Settings does not make much sense...
> > But we can let it as it is, I have no problem with that. I was just critically thinking about the code and if these 17 rows have some meaning..
> 
> @lindakladivova First thanks for your explanation.
> 
> I attached my patch which solve two issues:
> 
>     1. Maximize single window mode by default
> 
>     2. Respect user custom window position and size (custom win pos and size for single win mode is saved as standalone settings separate from multi win mode settings) changed via wxGUI Preference dialog **Save current window layout as default** checkbox widget
> 
> 
> `git am 0001-wxGUI-main_window-maximize-window-by-default.patch`
> 
> [0001-wxGUI-main_window-maximize-window-by-default.patch.zip](https://github.com/OSGeo/grass/files/11711847/0001-wxGUI-main_window-maximize-window-by-default.patch.zip)

@tmszi  Thanks for sharing it. I think now, it is more clear what is going on. I have used your changes and made little fixes [73fb223](https://github.com/OSGeo/grass/pull/3024/commits/73fb223d05ff5250d2bc29d9eda98bc12fa57180). Please look at them and if alright from your side, we could merge this.

landam:
I tested PR on Debian testing (Gnome). Before starting GRASS I have deleted wx settings. Single layout started maximized as expected. Can we merge PR and backport to 8.3?

landam (approved):

pesekon2:
Starting the GUI maximized solves the issue for me too.

pesekon2 (approved):

tmszi:
Tested on Gentoo with [dwm](https://dwm.suckless.org/) with activated float window mode and wxGUI has been launched maximized by default and custom user settings (win position and size) works as expected too.

petrasovaa (changes requested):
I tested this with multi-window and currently it ignores the map display size and position, because `IsDocked` is True for mutli-window. This should fix it:

```
--- a/gui/wxpython/gui_core/preferences.py
+++ b/gui/wxpython/gui_core/preferences.py
@@ -1853,7 +1853,9 @@ class PreferencesDialog(PreferencesBaseDialog):
 
                 # window size must be larger than zero, not minimized
                 # do not save dim when mapdisp is docked within single window
-                if not mapdisp.IsDocked() and (size[0] > 0 and size[1] > 0):
+                if (not mapdisp.IsDockable() or not mapdisp.IsDocked()) and (
+                    size[0] > 0 and size[1] > 0
+                ):
                     dim += f",{pos[0]},{pos[1]},{size[0]},{size[1]}"
 
             if single_window:
```

landam:
I tested multi-window with the patch above applied.

1. started GUI in multi-window layout
2. changed windows position
3. go to preferences, save
4. restart GUI

Windows appeared in original position, new position was ignored.

It seems that there is still an issue.

landam (changes requested):
multi-window position is ignored

landam:
I have 

```
      "dim": "26,23,634,622",
```

in settings regardless how I organize windows. It seems to be different issue not related to this PR.

landam:
Strangely, when I change size of windows it works. But position is always the same:

```
26,23,866,802,26,23,951,762
```

landam:
`GetPosition()` always returns `(26,23)` in my case (wxpython 4.2.0). But it seems to be not related to this PR.

gui/wxpython/gui_core/preferences.py
```diff
@@ -1830,32 +1830,45 @@ def _updateSettings(self):
...
                 # window size must be larger than zero, not minimized
-                # when mapdisp is inside single window (panel has no IsIconized), don't save dim
-                if (
-                    hasattr(mapdisp, "IsIconized")
-                    and not mapdisp.IsIconized()
-                    and (size[0] > 0 and size[1] > 0)
-                ):
-                    dim += ",%d,%d,%d,%d" % (pos[0], pos[1], size[0], size[1])
+                # do not save dim when mapdisp is docked within single window
+                if not mapdisp.IsDocked() and (size[0] > 0 and size[1] > 0):
```
landam, line 1856:
```suggestion
                if (not mapdisp.IsDockable() or not mapdisp.IsDocked()) and (
                    size[0] > 0 and size[1] > 0
                ):
                     dim += f",{pos[0]},{pos[1]},{size[0]},{size[1]}"
```

gui/wxpython/main_window/frame.py
```diff
@@ -179,30 +179,30 @@ def show_menu_errors(messages):
...
+                        x = client_disp[0]
+                    if y == 1:
+                        # Get client display y offset (OS panel)
+                        y = client_disp[1]
+                    self.SetPosition((x, y))
+                    self.SetSize((w, h))
+                except Exception:
+                    pass
+            else:
+                self.Maximize(True)
```
nilason, line 205: `self.Maximize()`is never reached in a "first run" simulation (with no existing`~/.grass8` dir), placing the window partially under the OS menu on Mac:

Screenshot_default


Outdenting the `else: self.Maximize()´ seems to work for me :

```py
        if UserSettings.Get(group="general", key="defWindowPos", subkey="enabled"):
            single_window_dim = {
                "group": "general",
                "key": "defWindowPos",
                "subkey": "dimSingleWindow",
            }
            dim = UserSettings.Get(**single_window_dim)
            default_dim = UserSettings.Get(**single_window_dim, settings_type="default")
            if dim != default_dim:
                try:
                    x, y = map(int, dim.split(",")[0:2])
                    w, h = map(int, dim.split(",")[2:4])
                    client_disp = wx.ClientDisplayRect()
                    if x == 1:
                        # Get client display x offset (OS panel)
                        x = client_disp[0]
                    if y == 1:
                        # Get client display y offset (OS panel)
                        y = client_disp[1]
                    self.SetPosition((x, y))
                    self.SetSize((w, h))
                except Exception:
                    pass
        else:
            self.Maximize(True)
```

lindakladivova:
> `GetPosition()` always returns `(26,23)` in my case (wxpython 4.2.0). But it seems to be not related to this PR.

Yep, that's unrelated (probably a wxPython 4.2.0 error).

petrasovaa:
> > `GetPosition()` always returns `(26,23)` in my case (wxpython 4.2.0). But it seems to be not related to this PR.
> 
> Yep, that's unrelated (probably a wxPython 4.2.0 error).

For me the position is saved as expected, I also use wxPython 4.2.0, so I am unsure why it doesn't work.

petrasovaa (approved):

nilason (approved):

petrasovaa:
In the interest of timely release I took liberty to merge and backport to 8_3 branch.

PR #3035, opened 2023-06-06
contributing: Add more expressions to match release notes groups

wenzeslaus:
- Two modules can now be separated with comma (2 modules are quite common, more is probably something else).
- Weblate commit messages now match.
- Packaging actually can use the word packaging.
- Docker can have OS after slash.
- Changes of contributors can now use the plain word contributors.

### DIFF
--- a/utils/release.yml
+++ b/utils/release.yml
@@ -26,7 +26,7 @@ notes:
       regexp: '(winGRASS|win|[Ww]indows): '
 
     - title: Packaging, Configuration, Portability, and Compilation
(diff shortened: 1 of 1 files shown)


neteler:
Nice. Locally patched and tested with 8.3.0RC1 : works.

landam:
> Nice. Locally patched and tested with 8.3.0RC1 : works.

@neteler Please consider approving this PR.

neteler (approved):

echoix (approved):
Seems like a justified improvement in our processes, I don't see why we couldn't try it out right away

PR #3039, opened 2023-06-06
t.rast.univar: Add support for percentiles

ninsbl:
This PR adds support for percentiles to `t.rast.univar`.


ninsbl:
If there are no objections I would like to merge this relatively soon...

PR #3051, opened 2023-06-13
g.extension: sparse checkout with history for fast commit log access

### LINKED ISSUE #2979: [Bug] g.extension test often times out in CI
**Describe the bug**

The `scripts/g.extension/testsuite/test_addons_download.py` test times out quite often.

Duration of a successful execution in the CI is 1 min 42 sec. The time out is currently set to 4 mins.

**To Reproduce**

Hard to say if this specific issue can be reproduced locally, but happens quite often on main and in PRs at least for the macOS job.

Locally, the test fails for me and runs around 3 min 20 sec, but given that `test_github_download_official_module_src_code_only` test (function) fails, I don't know if the longer run time is relevant to the CI issue.

**Expected behavior**

Test always finishes successfully in the given time limit (whatever that is).

**Screenshots**

```
Running ./scripts/g.extension/testsuite/test_addons_download.py...
WARNING: Bug in UI description. Missing module description
.......========================================================================
FAILED ./scripts/g.extension/testsuite/test_addons_download.py - Timeout >300.0s
```

**System description**

- Operating System: macOS in CI virtual machine
- GRASS GIS version: main

ninsbl:
` git -log -1` is slow with sparse checkout if `--filter=tree:0` is used because history is not checked out. Cloning is similarly fast with `--filter=blob:none` but then the history is checked out too and `git log -1` should be fast again.

Please try e.g.:
```
dirs="src/raster/r.centroids src/imagery/i.pysptools.unmix"
cd /tmp
time (git clone -q --no-checkout --branch=grass8 --filter=blob:none https://github.com/osgeo/grass-addons
cd grass-addons
git sparse-checkout set $dirs
git checkout grass8)

cd grass-addons
for dir in $dirs
do
  time git log --first-parent -1 $dir
  
done
cd ..
rm -rf grass-addons
```

Fixes #2979

I marked this as blocker as 30 seconds wait time per addon is a pretty bad user experience. Feel free to remove the label if you disagree...


tmszi (approved):
Tested and works as expected.

This code line required changes too.

https://github.com/OSGeo/grass/blob/2ca49e6cf0f14cda158ec15c020bd1280a7074d1/utils/mkhtml.py#L712

Small note:

With `blob:none` change size of downloaded data from 2.6M to 6.5M.

ninsbl:
> Tested and works as expected.
> 
> This code line required changes too.
> 
> https://github.com/OSGeo/grass/blob/2ca49e6cf0f14cda158ec15c020bd1280a7074d1/utils/mkhtml.py#L712


Thanks for testing and pointing out the missing change in `mkhtml.py`... Addressed in 48a86df.

ninsbl:
Cannot be backported to 8.2 directly, should be added to a bunch of g.extension-fixes that should be backported...

PR #3067, opened 2023-06-29
Update v.reclass.html: either 'rules' or 'column' option must be specified

agiudiceandrea:
Add a note about the fact that either 'rules' or 'column' option must be specified.

See https://github.com/OSGeo/grass/blob/5cb4e430b5e405bce4ab30e6d20c15222b10815f/vector/v.reclass/main.c#L103-L107.


neteler:
Thanks!

PR #3070, opened 2023-06-30
parallelized raster modules: fix buffer sizes

metzm:
With the parallelization of various raster modules, a memory option for a new output buffer was introduced. The calculation of the number of buffered rows did not take into account that a potentially large amount of RAM is already used for input data. This PR adjusts the calculation of the number of rows to be buffered for output maps for the modules `r.neighbors`, `r.patch`, `r.resamp.filter`, and `r.series`, avoiding accidental OOM errors. These modules should now use not (much) more than the user-specified amount of RAM.

### DIFF
--- a/raster/r.patch/main.c
+++ b/raster/r.patch/main.c
@@ -41,7 +41,7 @@ int main(int argc, char *argv[])
     int colr_ok;
     int outfd;
     RASTER_MAP_TYPE out_type, map_type;
-    size_t out_cell_size;
+    size_t out_cell_size, in_buf_size, out_buf_size;
     struct History history;
     void **presult, **patch;
     void *outbuf;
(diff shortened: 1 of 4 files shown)


petrasovaa (approved):

nilason (approved):
Looks good to me!

raster/r.patch/main.c
```diff
@@ -180,7 +180,17 @@ int main(int argc, char *argv[])
     nrows = Rast_window_rows();
     ncols = Rast_window_cols();
 
-    bufrows = atoi(memory->answer) * (((1 << 20) / out_cell_size) / ncols);
+    /* memory reserved for presult and patch */
+    in_buf_size = out_cell_size * ncols * nprocs * 2;
```
wenzeslaus, line 184: "reserved for result? and patch" or just "reserved for patch". I don't get the result part.

metzm: `presult` is the preliminary result (initially the first raster map), see https://github.com/metzm/grass/blob/openmp_buffered_rows/raster/r.patch/main.c#L233
`patch` are the remaining input maps, see https://github.com/metzm/grass/blob/openmp_buffered_rows/raster/r.patch/main.c#L250
`presult` and `patch` are allocated here: https://github.com/metzm/grass/blob/e0b4aa9db08a8a067fd633273ee2098bedf1739a/raster/r.patch/main.c#L170-L177

PR #3075, opened 2023-07-07
docker: CI update and simplification

mmacata:
This PR simplifies the github workflow for docker build, tag and push.
Due to some tag restructuring and new features of [docker/metadata-action](https://github.com/docker/metadata-action) it now only uses one job and, depending on the trigger action, creates different tags. It will create four different types of tags:

- `latest` for main branch and with ubuntu os
- `{{ branch }}-{{ os }}` for all configured branches, e.g. `releasebranch_8_3-alpine`
- `{{ tag }}-{{ os }}` for all published releases. e.g. `8.3.0-alpine`
- `current-{{ os }}` for a still hard-coded branch, here update to releasebranch_8_3, e.g `current-alpine`

This new naming conforms more with the GRASS GIS naming style of versions. While naming of branches and tags and content of `latest` tag is still the same, it adds `current-{{ os }}` and abandons former naming style with `latest-{{ os }}` and `stable-{{ os }}` like `latest-alpine` and `stable-alpine` to avoid confusion.

The plain `latest` tag is still created because it is default for `docker pull osgeo/grass-gis` and would be nice to have, although usage of it is not recommended. No strong opinion if it should point to `main` branch or `releasebranch_8_3`. Open for discussion. See also https://github.com/OSGeo/grass-website/pull/371

As the action was triggered twice for a release, `on: [tags ](tags: ['*.*.*'])` was outcommented.

Whole thing was tested as can be seen in [first commit of this PR](https://github.com/OSGeo/grass/commit/18e7811f3aff7b40133efaff3cba8c8af9cada30) on a branch named `docker-workflow-update` with mmacata remote and dockerhub account. Created and pushed image tags can be seen [here on dockerhub](https://hub.docker.com/r/mmacata/grass-gis/tags).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -61,7 +61,7 @@ jobs:
             type=ref,event=tag
             type=ref,event=branch
             type=raw,value=current,enable=${{ github.ref == format('refs/heads/{0}', 'releasebranch_8_3') }}
-            type=raw,value=latest,event=tag,enable=${{ matrix.os == 'ubuntu' }},suffix=
+            type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/') && matrix.os == 'ubuntu' }},suffix=
           flavor: |
             latest=false
             suffix=-${{ matrix.os }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,7 +17,7 @@ on:
       - main
       - releasebranch_*
       - '!releasebranch_7_*'
-      - docker-workflow-update
+      - docker-gh-workflow-update
     # tags: ['*.*.*']
     paths-ignore: [doc/**]
   release:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,7 +17,6 @@ on:
       - main
       - releasebranch_*
       - '!releasebranch_7_*'
-      - docker-gh-workflow-update
     # tags: ['*.*.*']
     paths-ignore: [doc/**]
   release:
@@ -36,7 +35,7 @@ jobs:
   #     8.3.0-alpine, 8.3.0-debian, 8.3.0-ubuntu
   docker-os-matrix:
     name: build and push ${{ matrix.os }} for ${{ github.ref }}
-    if: github.repository_owner == 'mmacata'
+    if: github.repository_owner == 'OSGeo'
     runs-on: ubuntu-latest
 
     strategy:
@@ -57,7 +56,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v4
         with:
-          images: mmacata/grass-gis
+          images: osgeo/grass-gis
           tags: |
             type=ref,event=tag
             type=ref,event=branch
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,6 +17,7 @@ on:
       - main
       - releasebranch_*
       - '!releasebranch_7_*'
+      - docker-workflow-update
     # tags: ['*.*.*']
     paths-ignore: [doc/**]
   release:
@@ -35,7 +36,7 @@ jobs:
   #     8.3.0-alpine, 8.3.0-debian, 8.3.0-ubuntu
   docker-os-matrix:
     name: build and push ${{ matrix.os }} for ${{ github.ref }}
-    if: github.repository_owner == 'OSGeo'
+    if: github.repository_owner == 'mmacata'
     runs-on: ubuntu-latest
 
     strategy:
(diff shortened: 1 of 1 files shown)


wenzeslaus:
While the main branch of GRASS GIS is fairly stable, I think it is more expected that it is the latest release rather than the latest state of the code.

mmacata:
Updated workflow to build and push `latest` tag for latest stable release only. Tested again with mmacata remote.

mmacata:
@wenzeslaus I updated the file according to your suggestion. Do you agree now?

neteler (approved):
Thanks, looks good to me (but another reviewer is welcome)

.github/workflows/docker.yml
```diff
@@ -58,125 +52,19 @@ jobs:
...
-      - name: Checkout
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-      - name: Create image and tag name
+      - name: Docker meta
         id: meta
         uses: docker/metadata-action@v4
         with:
-          # images: ${DOCKERHUB_REPOSITORY}
```
wenzeslaus, line 166: No commented out code. Discuss in a comment if you think some code alternatives need to be preserved right in the code.

.github/workflows/docker.yml
```diff
@@ -5,43 +5,37 @@ name: Docker
...
+# * creates tag latest for last stable release with ubuntu os
 
 on:
   push:
     branches:
       - main
       - releasebranch_*
       - '!releasebranch_7_*'
-    tags: ['*.*.*']
+    # tags: ['*.*.*']
```
wenzeslaus, line 20: I'm not sure about removing tags. The tags are created later on after a commit was made and pushed, so when the branch is updated with a commit, there is no tag. The tag is added only later. Will this be triggered? It is not clear to me if your tests included tags or not.

wenzeslaus (changes requested):
I appreciate that this is much shorter. The need to understand the many options of docker/metadata-action is unfortunate, but seem like a better alternative to the long custom code.

I'm still unclear about how some of the cases will work out here. I left individual comments for that.

.github/workflows/docker.yml
mmacata: The outcommented line of code is removed here, not added. Before, the variable was used elsewhere in the file but couldn't be used here. As not used anymore, I removed the outcommented code and have no further alternatives to this.

.github/workflows/docker.yml
mmacata: My intention was to simplify things and only have docker builds as described above. I also tested that case.
In these changes a tag itself is not triggered, so e.g. `8.3.0RC1-ubuntu` would not be build and tagged anymore, only the published releases like `8.3.0`. This also prevents the build from running twice on a release. Before for each published release the docker images were build two times - once for the tag and once for the release itself. This might not happen now with the simplification into a single workflow, but I have tested only the case without the tags.
So do you think also pre-releases are important to be build and tagged on dockerhub? Then I can readd this line.

wenzeslaus: I missed that release part. Thanks for the explanation.

.github/workflows/docker.yml
wenzeslaus: Thanks. I missed that. Makes sense.

.github/workflows/docker.yml
wenzeslaus: And no, I don't think not having RCs is a problem. However, I think these are still published releases, so they fulfill the condition or are they excluded because they are pre-releases? It seems that pre-releases are to excluded from released but not from published ([doc](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release)):

> Note: The prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.

.github/workflows/docker.yml
```diff
@@ -58,125 +52,19 @@ jobs:
...
         id: meta
         uses: docker/metadata-action@v4
         with:
-          # images: ${DOCKERHUB_REPOSITORY}
           images: osgeo/grass-gis
           tags: |
             type=ref,event=tag
+            type=ref,event=branch
+            type=raw,value=current,enable=${{ github.ref == format('refs/heads/{0}', 'releasebranch_8_3') }}
+            type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/') && matrix.os == 'ubuntu' }},suffix=
```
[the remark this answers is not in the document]
mmacata: I would proceed with this as in the past - this one file for all active release branches. So now it contains 8.3 and will be backported. Once 8.4 comes out, this needs to be changed (as before) and backported as well. So for older branches the conditions would then not fit anymore.
If a better solution comes up, it can be approached in a different PR - as this PR is not introducing this poblem but only not solving it.

.github/workflows/docker.yml
mmacata: Ah ok then, so `published` simply sorts out the `draft` state of the releases. Thanks for diving into the docs :)
I don't have a strong opinion on having the RCs be build and pushed or not. As that doesn't happen too often, I am also fine with keeping them. Would you prefer only type `released` or we leave it as is?

.github/workflows/docker.yml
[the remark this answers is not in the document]
wenzeslaus: Right, if we backport to all active branches, then the code is fine, that is with the `refs/tags/8.3` suggestion.

.github/workflows/docker.yml
wenzeslaus: Leave it as is. Having the RCs makes sense, you can easily test the release candidate and testing is certainly what RCs are meant for.

wenzeslaus (approved):
With the clarifications and plan for backporting, this makes a lot of sense and seems to fit with the GitHub and docker/metadata-action doc.

neteler:
Great work, I take liberty to merge it and backport now.

PR #3079, opened 2023-07-12
Python: Remove code for Python 2, remove six from requirements

a-detiste:
Here is the remainder of the Python2 removal.

There remain usage of `sys.version_info.major` to verify;
and most `from __future__ import`s can be removed.


a-detiste:
> icons/icon.py:48:33: R1733: Unnecessary dictionary index lookup, use 'img' instead (unnecessary-dict-index-lookup)

It looks like these two lines of code do absolutely nothing, since forever.
Whats was their meaning ?

petrasovaa:
> unnecessary-dict-index-lookup

I think initially, the code made sense and later it was rewritten, but not carefully enough. Feel free to remove that part.

a-detiste:
I did à "git diff HEAD~10000" and saw it was there since forever. It looks
like it does nothing but I m curious to find What let to this situation.

Anyway I m on Holliday now.

I would do this in a separate PR.

Greetings

Le mer. 12 juil. 2023, 15:36, Anna Petrasova ***@***.***> a
écrit :

a-detiste:
Thank you for finishig up this.
I wouldn't have touch the CI code itself,
I don't have the infrastructure to test it.

----

The next big thing in Python3 CI is mypy & annotations.

MyPy will find some problems without even without any annotations/guiding.

Example: it  complains here about more forgotten Python2 code:
```
mypy man/parser_standard_options.py
man/parser_standard_options.py:13: error: Name "urlopen" already defined (possibly by an import)  [no-redef
```

It will also find a lot of `strings` vs `bytes` mistakes.

Some people find annotations annoying or "anti-Pythonic"
so it's better to discuss it first elsewhere.

Greetings

a-detiste:
One last note: with `from something import *`,
Pyflakes cannot identifiy missing/extraneous symbols very well;
these could be changed to explicit symbol importing. (later...)

wenzeslaus:
> The next big thing in Python3 CI is mypy & annotations.
> ...
> Some people find annotations annoying or "anti-Pythonic"
> so it's better to discuss it first elsewhere.

We can talk about that on grass-dev mailing list or GitHub Discussions. I think doing a PR for a small piece of code might be a good idea. As for myself, I would like to see PEP 484 type hints introduced in the new code such as grass.jupyter and grass.grassdb.

petrasovaa (approved):
I would merge this, it's getting too big. Let's keep additional changes for another PR. Thanks!

a-detiste:
I m done 😁

petrasovaa:
Thanks again for the effort!

PR #3085, opened 2023-07-14
wxGUI/psmap: fix close vector map properties dialog after hit OK button

### LINKED ISSUE #3084: [Bug] Cartographic composer locks up GUI when setting vector properties
**Describe the bug**
Trying to set vector properties in the cartographic composer locks the GUI so that clicking anywhere just generates an error beep. Nothing works. I can't even open the console to see if there is an error message. The only recourse is to type "exit" in the terminal and kill the entire GRASS session. 

**To Reproduce**
1. use nc_spm demo data set
2. start cartographic composer
3. set frame to match the elevation raster
4. add streams vector map and try to set the properties
5. after setting line color, the dialog window won't close and everything else in the GUI is locked. 

- Operating System: MacOS 13.4.1
- GRASS GIS version 8.3.0

tmszi:
Fixes #3084.

### DIFF
--- a/gui/wxpython/psmap/dialogs.py
+++ b/gui/wxpython/psmap/dialogs.py
@@ -3290,6 +3290,7 @@ class VPropertiesDialog(PsmapDialog):
     def OnOK(self, event):
         self.update()
         event.Skip()
+        self.Destroy()
 
 
 class LegendDialog(PsmapDialog):
--- a/gui/wxpython/psmap/dialogs.py
+++ b/gui/wxpython/psmap/dialogs.py
@@ -1963,7 +1963,7 @@ class VectorPanel(Panel):
             id = self.vectorList[pos][2]
 
             dlg = VPropertiesDialog(
-                self,
+                self.parent,
                 id=id,
                 vectors=self.vectorList,
                 tmpSettings=self.tmpDialogDict[id],
--- a/gui/wxpython/psmap/dialogs.py
+++ b/gui/wxpython/psmap/dialogs.py
@@ -1965,14 +1965,12 @@ class VectorPanel(Panel):
             dlg = VPropertiesDialog(
                 self,
                 id=id,
-                settings=self.instruction,
-                env=self.env,
                 vectors=self.vectorList,
                 tmpSettings=self.tmpDialogDict[id],
             )
-            dlg.ShowModal()
-
-            self.parent.FindWindowById(wx.ID_OK).SetFocus()
+            if dlg.ShowModal() == wx.ID_OK:
+                dlg.update()
+            dlg.Destroy()
 
     def enableButtons(self, enable=True):
         """Enable/disable up, down, properties, delete buttons"""
@@ -2124,20 +2122,16 @@ class MainVectorDialog(PsmapDialog):
         pass
 
 
-class VPropertiesDialog(PsmapDialog):
-    def __init__(self, parent, id, settings, vectors, tmpSettings, env):
-        PsmapDialog.__init__(
+class VPropertiesDialog(Dialog):
+    def __init__(self, parent, id, vectors, tmpSettings):
+        Dialog.__init__(
             self,
             parent=parent,
-            id=id,
-            title="",
-            settings=settings,
-            env=env,
-            apply=False,
         )
 
         vectorList = vectors
         self.vPropertiesDict = tmpSettings
+        self.spinCtrlSize = (65, -1)
 
         # determine map and its type
         for item in vectorList:
@@ -2194,6 +2188,26 @@ class VPropertiesDialog(PsmapDialog):
             self.OnPattern(None)
 
         self._layout(notebook)
+
+    def _layout(self, panel):
+        # buttons
+        btnCancel = Button(self, wx.ID_CANCEL)
+        btnOK = Button(self, wx.ID_OK)
+        btnOK.SetDefault()
+
+        # sizers
+        btnSizer = wx.StdDialogButtonSizer()
+        btnSizer.AddButton(btnCancel)
+        btnSizer.AddButton(btnOK)
+        btnSizer.Realize()
+
+        mainSizer = wx.BoxSizer(wx.VERTICAL)
+        mainSizer.Add(panel, proportion=1, flag=wx.EXPAND | wx.ALL, border=5)
+        mainSizer.Add(btnSizer, proportion=0, flag=wx.EXPAND | wx.ALL, border=5)
+
+        self.SetSizer(mainSizer)
+        mainSizer.Layout()
+        mainSizer.Fit(self)
 
     def _DataSelectionPanel(self, notebook):
         panel = Panel(
@@ -2924,9 +2938,7 @@ class VPropertiesDialog(PsmapDialog):
 
         styleText = StaticText(panel, id=wx.ID_ANY, label=_("Choose line style:"))
         penStyles = ["solid", "dashed", "dotted", "dashdotted"]
-        self.styleCombo = PenStyleComboBox(
-            panel, choices=penStyles, validator=TCValidator(flag="ZERO_AND_ONE_ONLY")
-        )
+        self.styleCombo = PenStyleComboBox(panel, choices=penStyles)
         # self.styleCombo = wx.ComboBox(panel, id = wx.ID_ANY,
         ##                            choices = ["solid", "dashed", "dotted", "dashdotted"],
         # validator = TCValidator(flag = 'ZERO_AND_ONE_ONLY'))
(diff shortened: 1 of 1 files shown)


cmbarton:
Has this been merged yet? Doesn't look like it.

petrasovaa:
> Has this been merged yet? Doesn't look like it.

Not yet. I looked at it briefly some time ago and I had some issues reproducing it. Can you confirm this PR actually fixes it?

tmszi:
> > Has this been merged yet? Doesn't look like it.
> 
> Not yet. I looked at it briefly some time ago and I had some issues reproducing it. Can you confirm this PR actually fixes it?

The bug is still reproducible following the steps in the reported issue #3084 and this PR fixes it. I carefully tested each of my PRs before creating them.

petrasovaa:
Sorry it took so long, but finally I got to this. @tmszi yes, I can reproduce it on Ubuntu and I agree with the fix. The problem is ShowModal is not called the standard way with this dialog (testing outcome of ShowModal and then destroying the dialog would be better, but would require larger changes).

@cmbarton since this seems to fix it for Ubuntu (and I assume @tmszi tested it on linux as well, no?), could you make sure the patch is correctly applied? But it sounds like maybe the behavior with this fix is different? It could be something Mac specific.

tmszi:
> Sorry it took so long, but finally I got to this. @tmszi yes, I can reproduce it on Ubuntu and I agree with the fix. The problem is ShowModal is not called the standard way with this dialog (testing outcome of ShowModal and then destroying the dialog would be better, but would require larger changes).

My complex debug info:

Problematic is **line vector map props dialog** only, especially `PenStyleComboBox()` widget

https://github.com/OSGeo/grass/blob/69dd32429a8c53e9e72011332cbf30e09e53ccf7/gui/wxpython/psmap/dialogs.py#L2927-L2929

If you try comment code lines 2927-2960, confirming **line** vector map props dialog Ok button widget works as expected.

Others vector map props dialog (**point** and **area type)**, confirming Ok button widget works as expected.


Another bug (new PR) is in `PenStyleComboBox` class `OnDrawItem` method (`DrawText` and `DrawLine` method). under wxGTK.

```
  File "/usr/lib64/grass84/gui/wxpython/psmap/dialogs.py", line 182, in OnDrawItem
    dc.DrawText(
TypeError: DC.DrawText(): arguments did not match any overloaded call:
  overload 1: argument 3 has unexpected type 'float'
  overload 2: argument 2 has unexpected type 'int'
```

https://github.com/OSGeo/grass/blob/69dd32429a8c53e9e72011332cbf30e09e53ccf7/gui/wxpython/psmap/dialogs.py#L182-L192

> @cmbarton since this seems to fix it for Ubuntu (and I assume @tmszi tested it on linux as well, no?),

I tested this PR under wxGTK and wxMSW.

nilason:
I can confirm that the problem persists on Mac even with this present solution, using
Python 3.11.5 (main, Aug 25 2023, 13:55:04) [Clang 13.1.6 (clang-1316.0.21.2.5)]
wxPython 4.2.1 osx-cocoa (phoenix) wxWidgets 3.2.2.1

nilason:
The vector property window, once opened cannot be closed. By simply open it, doing nothing, then close (via cancel button), only puts the window behind all other windows, but the window doesn't close nor the modal state cannot be terminated. Only solution is to force quit gui (wxPython).

nilason:
I tried commenting out the lines 2927-2960, but it didn't do any difference.

petrasovaa:
> Maybe trying `self.EndModal(wx.ID_OK)` instead of `self.Destroy()` may make a difference?

Could someone on Mac try using this in the OnOK method?

nilason:
> > Maybe trying `self.EndModal(wx.ID_OK)` instead of `self.Destroy()` may make a difference?
> 
> Could someone on Mac try using this in the OnOK method?

I missed that suggestion, yes it seems to work! Now, only cancel remain the same (problematic).

nilason:
Following works for me:

```py
    def OnCancel(self, event):
        self.EndModal(wx.ID_CANCEL)

    def OnOK(self, event):
        self.update()
        event.Skip()
        self.EndModal(wx.ID_OK) 
```

One (rather trivial, but still) problem remains, after closing "Vector Properties" window, the "Map settings" window looses focus and is placed behind "Cartographic Composer" window.

cmbarton:
> One (rather trivial, but still) problem remains, after closing "Vector Properties" window, the "Map settings" window looses focus and is placed behind "Cartographic Composer" window.

This problem seems to be pervasive in the custom-designed interfaces that wrap GRASS commands (e.g., the raster and vector import tools). Trivial but annoying. 

In class today, I needed to show a student why they could not import a file. After selecting the file in the raster import tool, the import window disappears behind the main GRASS GUI window. They did not know this and wondered what happened.

nilason:
> > One (rather trivial, but still) problem remains, after closing "Vector Properties" window, the "Map settings" window looses focus and is placed behind "Cartographic Composer" window.
> 
> This problem seems to be pervasive in the custom-designed interfaces that wrap GRASS commands (e.g., the raster and vector import tools). Trivial but annoying.
> 
> In class today, I needed to show a student why they could not import a file. After selecting the file in the raster import tool, the import window disappears behind the main GRASS GUI window. They did not know this and wondered what happened.

Yes, now with the single window mode it is particularly annoying because it is virtually impossible to see (if the main window fills the screen). I was fooled by this several times testing this PR.

petrasovaa:
> Following works for me:
> 
> ```python
>     def OnCancel(self, event):
>         self.EndModal(wx.ID_CANCEL)
> 
>     def OnOK(self, event):
>         self.update()
>         event.Skip()
>         self.EndModal(wx.ID_OK) 
> ```
> 

I decided to slightly refactor this, the dialog is now derived from standard wx dialog, so the hack above is not needed. Testing welcome. I did have some issues with a validator, so I removed it, it's not very important anyway,

petrasovaa:
> One (rather trivial, but still) problem remains, after closing "Vector Properties" window, the "Map settings" window looses focus and is placed behind "Cartographic Composer" window.

That looks Mac-specific, you could try calling `self.Raise()` after vector properties dialog is closed.

But some of these problems with single window are general, we could explore some window settings like 'wx.FRAME_FLOAT_ON_PARENT'.

tmszi:
Successfully tested on wxGTK, wxMSW.

nilason:
> I decided to slightly refactor this, the dialog is now derived from standard wx dialog, so the hack above is not needed. Testing welcome. I did have some issues with a validator, so I removed it, it's not very important anyway,

That works fine!

nilason:
> > One (rather trivial, but still) problem remains, after closing "Vector Properties" window, the "Map settings" window looses focus and is placed behind "Cartographic Composer" window.
> 
> That looks Mac-specific, you could try calling `self.Raise()` after vector properties dialog is closed.
> 
> But some of these problems with single window are general, we could explore some window settings like 'wx.FRAME_FLOAT_ON_PARENT'.

I found what I believe a better solution. In `VectorPanel::OnProperties()` initialise the `VPropertiesDialog` with `self.parent` instead of the present `self`.  "self" is only the Panel, but we need to have the MapDialog as parent.

gui/wxpython/psmap/dialogs.py
```diff
@@ -1965,14 +1965,12 @@ def OnProperties(self, event):
             dlg = VPropertiesDialog(
                 self,
```
nilason, line 1966:
```suggestion
                self.parent,
```

cmbarton:
If this is committed I can test too.

petrasovaa:
> I found what I believe a better solution. In `VectorPanel::OnProperties()` initialise the `VPropertiesDialog` with `self.parent` instead of the present `self`. "self" is only the Panel, but we need to have the MapDialog as parent.

That makes sense. I assume you tested it, does this solve the problem with hiding the dialog on Mac?

nilason:
> That makes sense. I assume you tested it, does this solve the problem with hiding the dialog on Mac?

Yes

cmbarton:
This solves all reported problem with the cartographic composer. Thanks to all!

Can we add the fix to the hidden window to the raster and vector import tools too? Probably some other places need it, but I don't know which right now.

nilason:
> This solves all reported problem with the cartographic composer. Thanks to all!

👍 

> Can we add the fix to the hidden window to the raster and vector import tools too? Probably some other places need it, but I don't know which right now.

I looked for similar problems in the import tools, but I couldn't find any. Please, take a closer look and report in an another ticket.

cmbarton:
OK. Thanks again to you, Thomas, and Anna for fixing this.

Michael
____________________
Michael Barton

PR #3090, opened 2023-07-18
v.db.select: Add column names and types to JSON

wenzeslaus:
The JSON output now has column names and types under info/columns.

The example in documentation does not include the unclear note about booleans (perhaps a discussion for another place).
A command to reproduce the example output is now included.

Documentation and test of the new functionality are included.

### DIFF
--- a/vector/v.db.select/main.c
+++ b/vector/v.db.select/main.c
@@ -337,8 +337,32 @@ int main(int argc, char **argv)
     if (format == JSON) {
         if (flags.region->answer)
             fprintf(stdout, "{\"extent\":\n");
-        else
-            fprintf(stdout, "{\"records\":[\n");
+        else {
+            fprintf(stdout, "{\"info\":\n{\"columns\":[\n");
+            for (col = 0; col < ncols; col++) {
+                column = db_get_table_column(table, col);
+                if (col)
+                    fprintf(stdout, "},\n");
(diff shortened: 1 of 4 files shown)


wenzeslaus:
List of column names in Python are obtained by:

```python
column_names = [item["name"] for item in data["info"]["columns"]]
```

Is this good? Convenient enough?

cwhite911:
> List of column names in Python are obtained by:
> 
> ```python
> column_names = [item["name"] for item in data["info"]["columns"]]
> ```
> 
> Is this good? Convenient enough?

Do you plan on including any other information in `info`, such as, record count, layer name, query, group by, and null value?

wenzeslaus:
> Do you plan on including any other information in `info`, such as, record count, layer name, query, group by, and null value?

Not now, but in the future, yes. I was thinking about including output of the bounding box flag (-r Print minimal region extent of selected vector features instead of attributes). That would go on the top level next to info and records.

Record count would make sense. I would avoid repeating random inputs unless we want to start including provenance.

I guess this suggests that info should be skipped.

I was also thinking about including `column_names` as an additional info, but that can be done on the top level as well.

cwhite911:
If the plan is to include more to ```info``` in the future I would include it now to avoid any backwards compatibility issues.

PR #3097, opened 2023-07-24
r.sun: fix -Wimplicit-int warnings

nilason:
This is a partial, manual backport of https://github.com/OSGeo/grass/pull/1405.

The warning is treated as an error by Clang 16+ and the compilation fails.

Considering next 7.8 release likely will be the last, this should be a blocker.

### DIFF
--- a/raster/r.sun/local_proto.h
+++ b/raster/r.sun/local_proto.h
@@ -44,7 +44,7 @@ void setHorizonInterval(double val);
 void setAngularLossDenominator();
 
 
-void cube(int, int);
+/* void cube(int, int); */
 
 double com_sol_const(int no_of_day);
 
--- a/raster/r.sun/main.c
+++ b/raster/r.sun/main.c
@@ -118,8 +118,8 @@ int OUTGR(void);
 int min(int, int);
 int max(int, int);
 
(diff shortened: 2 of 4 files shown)


wenzeslaus:
Please, link a PR which is merged. #1405 is closed without merging and this change which comments out code requires at least some explanation.

nilason:
> Please, link a PR which is merged. #1405 is closed without merging and this change which comments out code requires at least some explanation.

That PR was created just before the change to "Squash and merge"-only policy, so I cherry picked and pushed each commit in the PR manually (to keep each fix separate). But you're right about the lack of explanation, I incorporated the original commit message with updates:

> r.sun: remove unused function 'cube'
> 
> For now only comments out relevant parts.
> Addresses -Wimplicit-int compiler warnings.
> The warning is treated as an error by Clang 16+ and the compilation fails.
> 
> This is a partial, manual backport of https://github.com/OSGeo/grass/pull/1405,
> commit 594c947ce8984f07faf28d7d71e4c902f4532069.

neteler (approved):
Compilation tested fine with gcc version 12.3.1 on Fedora 37.

PR #3108, opened 2023-07-31
r.texture manual: fix output parameter in example

neteler:
Fixes example and registers old `prefix` in `lib/gis/renamed_options`.

(was missing in e212c96107110eb9d29f69e03c3ae226a928eefd)

### DIFF
--- a/lib/gis/renamed_options
+++ b/lib/gis/renamed_options
@@ -386,6 +386,7 @@ r.surf.area|input:map
 r.terraflow|stream_dir:directory
 # r.texture
 r.texture|measure:method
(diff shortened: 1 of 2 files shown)


wenzeslaus:
I don't know if anyone needs this, but I would just emphasize that this - if I understand correctly - makes 8.4 backwards compatible with 6.4 and which is sort of cool.

neteler:
> makes 8.4 backwards compatible with 6.4 and which is sort of cool.

Yeah - interestingly the example was broken since then... now fixed.

PR #3110, opened 2023-08-03
v.to.rast: Change val to value

wenzeslaus:
The use parameter (option) had value 'val' while the parameter which is using that is called value. This makes it consistent, so use=value and value=1, not use=val and value=1. Both can be shortened, so even use=val and val=1 are still valid parameters.

This changes code which is using v.to.rast.

This does not change the other possible values (options), namely attr, cat, and dir.


marisn (approved):

wenzeslaus:
I'm getting error with using through grass.pygrass:

```
...
   File "/home/runner/install/grass84/etc/python/grass/pygrass/modules/interface/module.py", line 653, in update
    self.inputs[key].value = val
  File "/home/runner/install/grass84/etc/python/grass/pygrass/modules/interface/parameter.py", line 212, in _set_value
    self._value, self._rawvalue = _check_value(self, value)
  File "/home/runner/install/grass84/etc/python/grass/pygrass/modules/interface/parameter.py", line 110, in _check_value
    raise ValueError(must_val % (param.name, param.values))
ValueError: The Parameter <use>, must be one of the following values: ['attr', 'cat', 'value', 'z', 'dir']
```

Possibly, the check there needs to be less strict.

PR #3115, opened 2023-08-08
CITATION.cff: make license entry SPDX conform

neteler:
Use proper short code from https://spdx.org:

https://spdx.org/licenses/GPL-2.0-or-later.html

Hopefully fixes automated Zenodo.org upload, see discussion at
https://lists.osgeo.org/pipermail/grass-dev/2023-August/095989.html

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -106,7 +106,7 @@ authors:
     email: hamish_b@yahoo.com
 
 repository-code: https://github.com/OSGeo/grass
-license: GNU General Public License v2 or later
+license: GPL-2.0-or-later
 doi: 10.5281/zenodo.4621728
 keywords:
   - GIS


wenzeslaus (approved):
Thank you. The new code is correct according to the [Citation File Format documentation](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#license):

> license...description: The [SPDX license identifier(s)](https://spdx.dev/ids/) for the license...

And the value checks even against [CFF's license-enum](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionslicense-enum).

Given that the other branches just work, I hope that this fixes the Zenodo issue, too.

wenzeslaus:
> ...other branches just work...

Correction: That would be 7.8 only because that's the only one with recent release. No backport is needed there. We are using the automated Zenodo metadata extraction there.

neteler:
In fact, we do not have the `CITATION.cff` file in GRASS GIS 7.

PR #3133, opened 2023-08-25
v.info: Rename location to project in outputs

wenzeslaus:
- Translatable plain text output now generates "Project" instead of "Location".
- Extended shell-style output now generates "project=" in addtion to "location=".
- Examples in documentation now show project only.
- History continues using LOCATION because history file is printed as is and the parsing rules applied to it are unclear, so consequences of adding a new key there are not clear to me.


cmbarton (approved):
Yes! I agree with this standardized usage and recommend it for all other places where these terms are used.

petrasovaa (approved):

PR #3141, opened 2023-09-06
CI: drop outdated CentOS 7 compilation

neteler:
Remove of CentOS 7 CI compilation as no longer compatible with new GHA checkout v4 (see #3138 - Background: default runtime now node20. Node16 has end of life on 11 Sep 2023).


wenzeslaus (approved):

wenzeslaus:
Thanks!

I checked the PR which added that and there was also yum.txt file which is not used elsewhere, so I removed that.

PR #3163, opened 2023-09-18
diglib: Calculate the total size of topo file to get the correct off_t_size

HuidaeCho:
`off_t` sizes for coor and topo files can be different. For example, from an EPSG:4326 location,
```bash
> v.mkgrid map=grids_00625deg position=coor coor=-180,-90 box=0.0625,0.0625 grid=2880,5760
> d.mon wx0
> d.vect grids_00625deg
ERROR: Unable to seek: Invalid argument

> ls -al .../vector/grids_00625deg
-rw-r--r-- 1 hcho users 1708966098 Sep 17 22:35 coor
-rw-r--r-- 1 hcho users         95 Sep 17 22:33 dbln
-rw-r--r-- 1 hcho users        173 Sep 17 22:33 head
-rw-r--r-- 1 hcho users        260 Sep 17 22:33 hist
-rw-r--r-- 1 hcho users 6706995200 Sep 17 22:45 sidx
-rw-r--r-- 1 hcho users 2173694570 Sep 17 22:45 topo
```
`coor_size` is less than `PORT_LONG_MAX` (2^31 - 1 = 2147483647; max `int`), but the topo size is greater than that. Setting `ptr->off_t_size = 4` in https://github.com/OSGeo/grass/blob/fd59e94985870d8558e236cf8015a99b8d83e5ff/lib/vector/diglib/plus_struct.c#L677-L682 causes an integer overflow and `ptr->Isle_offset` is read as -2121341854 in https://github.com/OSGeo/grass/blob/fd59e94985870d8558e236cf8015a99b8d83e5ff/lib/vector/diglib/plus_struct.c#L641

This PR calculates the total size of topo file to get the correct `off_t_size`. Before this PR, the size of coor file was used to determine this offset size for topo file.


lib/vector/diglib/plus_struct.c
```diff
@@ -673,13 +673,68 @@ int dig_Wr_Plus_head(struct gvfile *fp, struct Plus_head *ptr)
     if (0 >= dig__fwrite_port_C((char *)buf, 5, fp))
         return (-1);
 
-    /* determine required offset size from coor file size */
-    if (ptr->coor_size > (off_t)PORT_LONG_MAX) {
-        /* can only happen when sizeof(off_t) == 8 */
-        ptr->off_t_size = 8;
```
metzm, line 679: Please don't remove this test. `ptr->off_t_size` must still be set to 8 if the coor file is larger than `PORT_LONG_MAX`, no matter how large the topo file would become.

Suggestion: First check the coor file size, if `ptr->off_t_size` is not set to 8, use your new check if the topo file would become larger than `PORT_LONG_MAX`.

HuidaeCho: Yep! I missed that. Obviously, `ptr->off_t_size` is shared between coor and topo files as of now.
https://github.com/OSGeo/grass/blob/6ffd18f8083c888fddfdae86ceb771e0e33cbb25/lib/vector/diglib/plus_struct.c#L762

I cannot see anywhere else `ptr->off_t_size` is used for coor files other than the above line and the reading part where its size is written and read. If we use a different-size variable for the coor size in these two lines, theoretically, we can only save 4 bytes only if coor file is less than `PORT_LONG_MAX`, but topo file is greater than that.

wenzeslaus:
> ...ptr->off_t_size is shared between coor and topo files as of now...

Can you please add documentation for these things? Preferably also something in the Doxygen API-level doc.

HuidaeCho:
Oh! I messed it up... Closing it for now.

HuidaeCho:
> > ...ptr->off_t_size is shared between coor and topo files as of now...
> 
> Can you please add documentation for these things? Preferably also something in the Doxygen API-level doc.

Added it in #3351

PR #3167, opened 2023-09-20
grass.script: Add function for colored message

ecodiv:
Functions to print colored messages to the screen. They can be used to print error messages in red, warning messages in light purple and tips in green.  More examples with different colors on https://www.geeksforgeeks.org/print-colors-python-terminal/, from where I got these functions. (p.s. a better option, perhaps, is to add a color option to g.message?)

### DIFF
--- a/python/grass/script/core.py
+++ b/python/grass/script/core.py
@@ -661,6 +661,30 @@ def message(msg, flag=None):
     run_command("g.message", flags=flag, message=msg, errors="ignore")
 
 
+def prRed(msg):
+    """Display a message in red to highlight error message
+
+    :param str msg: message to be displayed
+    """    
+    print("\033[91m {}\033[00m".format(msg))
+
+
+def prLightPurple(msg):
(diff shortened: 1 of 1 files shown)


ecodiv:
I noticed that colored message/warnings are already used, at least in r.patch; see attached screenshot.

rwXPu2NHTo

veroandreo:
I've seen such colored messages in the GUI only, i.e., in module's GUI or in the Console tab... Would your functions work for the terminal too?

ecodiv:
> I've seen such colored messages in the GUI only, i.e., in module's GUI or in the Console tab... Would your functions work for the terminal too?

It works in the terminal. It does, in fact, not work in the console or module's GUI when running grass gis on Windows. I am not sure about Linux, and I can't test it now because of problems with my Linux computer. However, you can easily test yourself as I implemented the function in r.vif. Set the region outside the bounding box of the intput maps. Then, run the r.vif function. This should give you the error message shown in the screenshot below. 

image

PR #3173, opened 2023-09-22
db.columns: Add separator= and exclude= options

HuidaeCho:
This PR adds separator= and exclude= options to allow a CSV-like output using comma and exclusion of selected columns. The default separator is still newline for compatibility.


wenzeslaus (changes requested):
Can you please add exclusion and custom separator to the test? Existing functionality is already tested. It would be good to have also the new one covered.

db/db.columns/testsuite/test_dbcolumns.py
scripts/db.dropcolumn/testsuite/test_db_dropcolumn.py

PR #3174, opened 2023-09-22
v.db.join: Add exclude_columns= option

HuidaeCho:
This PR adds exclude_columns= option to exclude certain columns from other table (e.g., cat).


HuidaeCho:
BTW, overwriting `cat` is dangerous in general. If other table is the attribute table of another vector, we should exclude `cat` automatically?

wenzeslaus (approved):
Looks good. I'm for merging this as is and fixing the cat column issue separately.

landam:
@HuidaeCho Are you going to merge this PR?

HuidaeCho:
> @HuidaeCho Are you going to merge this PR?

Yes. Doing it now. Thanks for the reminder!

Issue #3183, opened 2023-09-23
[Bug] utils/grass_clang_format.sh fails to detect version


marisn:
On Debian version number is the fourth item:
`
clang-format --version
Debian clang-format version 16.0.6 (15)
`
But the script expects it to be the third item:
`clang_version=$(echo "${clang_version_full}" | cut -f3 -d" " | cut -f1 -d".")`
https://github.com/OSGeo/grass/blob/main/utils/grass_clang_format.sh#L56

The version check should be rewritten to deal with customized version strings.

nilason:
Could you please test:

```bash
clang-format --version | sed -En 's/.*version ([0-9]+)\.[0-9]+\.[0-9]+.*/\1/p'
```

It should print the major version.

PR #3186, opened 2023-09-24
utils: fix clang-format version check

nilason:
- Adopt to the fact the result of `clang-format --version` may look slightly different on various platforms.
- Make sure 'clang-format' version 15 is used (v 16+ results in unwanted changes)

Closes https://github.com/OSGeo/grass/issues/3183

### DIFF
--- a/utils/grass_clang_format.sh
+++ b/utils/grass_clang_format.sh
@@ -53,10 +53,11 @@ if ! (${fmt} --version >/dev/null); then
 fi
 
 clang_version_full=$(${fmt} --version)
-clang_version=$(echo "${clang_version_full}" | cut -f3 -d" " | cut -f1 -d".")
-if [ "${clang_version}" -lt "${req_cf_v}" ]; then
+clang_version=$(echo "${clang_version_full}" | \
+    sed -En 's/.*version ([0-9]+)\.[0-9]+\.[0-9]+.*/\1/p')
+if [ "${clang_version}" -ne "${req_cf_v}" ]; then
     echo "Error: ${clang_version_full}"
-    echo "  is used, but version ${req_cf_v} or newer is required."
+    echo "  is used, but version ${req_cf_v} is required."
     echo "  Consider setting the global variable GRASS_CLANG_FORMAT to"
     echo "  the clang-format version needed."
     exit 1


marisn (approved):
Version extraction works. Can not comment on requirement of exact version match.

nilason:
> Version extraction works. Can not comment on requirement of exact version match.

Thanks for testing!

I'm afraid we have to stick to clang-format version 15 for the time being. I tested with version 16 and following files were changed:

```
lib/gis/lz4.c
lib/gis/lz4.h
lib/gmath/solvers_direct.c
lib/gpde/test/test_arrays.c
lib/rst/interp_float/segmen2d_parallel.c
raster/r.mfilter/execute.c
raster/r.proj/main.c
raster/r.sim/simlib/hydro.c
raster/r.slope.aspect/main.c
raster/r.sun/main.c
```

Just for testing, I ran clang-format again on the above changed files, now with version 15. The following files changed again:
```
lib/gis/lz4.h
lib/rst/interp_float/segmen2d_parallel.c
raster/r.mfilter/execute.c
raster/r.proj/main.c
raster/r.sim/simlib/hydro.c
raster/r.slope.aspect/main.c
raster/r.sun/main.c
```

So, mixing versions will not work, the same as with Black.

wenzeslaus:
> So, mixing versions will not work, the same as with Black.

Any idea how and when to update the version? For Black, the major (breaking) releases are yearly and having the latest seems to me like the only reasonable policy because everything else seems arbitrary. At the same time, using pip to install Black was necessary at least up until recently, so having the latest version was not an issue.

nilason:
> > So, mixing versions will not work, the same as with Black.
> 
> Any idea how and when to update the version? For Black, the major (breaking) releases are yearly and having the latest seems to me like the only reasonable policy because everything else seems arbitrary. At the same time, using pip to install Black was necessary at least up until recently, so having the latest version was not an issue.

I don't think there will be reason to change this in at least a year. Most platforms provide packages with versioned clang-format and as I understand it version 15 is broadly supported. For reference: GDAL just updated to clang-format 15.

The changes after running version 16 was mainly (if not exclusively) re-formatting of pragma statements. Nothing urgent.

PR #3199, opened 2023-10-05
configure: remove or replace obsolete macros

nilason:
This eliminates warnings running autoconf with version 2.71. Note: the file 'configure' is still created by and fully compatible with version 2.69. This does **not** represent a change to generate the configure file by 2.71, but that change will eventually be merely formal. 

One issue involving <time.h> vs. <sys/time.h> remains and now issues a separate warning. Solving that includes changes in code and must be addressed separately.

The changes of this PR is in practice a subset of #2281, without the official move to 2.71.

### DIFF
--- a/configure
+++ b/configure
@@ -619,6 +619,7 @@ ac_includes_default="\
 # include <unistd.h>
 #endif"
 
+ac_header_list=
 ac_subst_vars='LTLIBOBJS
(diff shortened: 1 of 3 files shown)


nilason:
I reckon this will not be needed as #2281 is on the line.

PR #3214, opened 2023-10-22
v.db.connect: Replace -o flag with the standard --overwrite

### LINKED ISSUE #3149: [Bug] v.db.connect does not recognize --overwrite flag
**Describe the bug**
v.db.connect does not recognize --overwrite and equivalent to -o. Hence, it throws an error if you try to overwrite from the GUI, which uses the long form of --overwrite instead of -o. 

When modules were converted to also use --overwrite, somehow v.db.connect was missed. 

- Operating System: Mac OS 13.5.1
- GRASS GIS version 8.3.0

HuidaeCho:
This PR fixes #3149 by using the standard `--overwrite` flag.

### DIFF
--- a/lib/gis/renamed_options
+++ b/lib/gis/renamed_options
@@ -468,6 +468,8 @@ v.buffer|bufcolumn:column
 v.clean|thresh:threshold
 # v.colors
 v.colors|volume:raster_3d
+# v.db.connect
+v.db.connect|-o:--overwrite
 # v.db.join
 v.db.join|otable:other_table
 v.db.join|ocolumn:other_column
--- a/vector/v.db.connect/main.c
+++ b/vector/v.db.connect/main.c
@@ -34,7 +34,7 @@ int main(int argc, char **argv)
     struct GModule *module;
     struct Option *inopt, *dbdriver, *dbdatabase, *dbtable, *field_opt, *dbkey,
         *sep_opt;
-    struct Flag *overwrite, *print, *columns, *delete, *shell_print;
+    struct Flag *print, *columns, *delete, *shell_print;
     dbDriver *driver;
     dbString table_name;
     dbTable *table;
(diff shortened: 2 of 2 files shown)


marisn:
I'm sorry, but this shall not pass. You should not remove a flag in non-major release.
Please add a printf with warning whenever `-o` is used: "The -o flag is deprecated and will be removed in future releases. Use --o instead"
To future proof the code, wrap the old `-o` in a macro `#if GRASS_VERSION_MAJOR < 9`

landam (changes requested):
@HuidaeCho It make sense to me to add `--overwrite` flag. On the other hand we cannot break backward compatibility in GRASS 8 releases. The `-o` flag could be removed in GRASS 9. I suggest to add depreciation warning that `-o` will be removed in GRASS 9.

HuidaeCho:
This PR requires #3256.

HuidaeCho:
@marisn, @landam Will this work with #3256?

marisn (approved):

marisn:
> @marisn, @landam Will this work with #3256?

Yes. I tested together with #3256 and backwards compatibility is in place.

wenzeslaus (commented):
Maybe a test to prove that both `-o` and `--overwrite` work? Probably using `gs.run_command(...)` (with `flags="o"` and `overwrite=True`) rather than the `assert*()` functions which rely on grass.pygrass Module interface which performs checks on its own and does not support `lib/gis/renamed_options`.

landam (approved):
I successfully tested:

```
v.db.connect map=voda_aoi table=VodniTok
ERROR: Use --overwrite to overwrite existing link for layer <1>

v.db.connect map=voda_aoi table=VodniTok --o
The table <VodniTok> is now part of vector map <voda_aoi> and may be
deleted or overwritten by GRASS modules

v.db.connect map=voda_aoi table=VodniTok --o
The table <VodniTok> is now part of vector map <voda_aoi> and may be
deleted or overwritten by GRASS modules
```

landam:
@HuidaeCho Please feel free to merge (you have more than zero approvals)...

PR #3230, opened 2023-11-09
parser: add markdown support

landam:
This PR:

- adds a new option `--md-description` to produce module usage in Markdown
- updates RestructuredText to be synced with HTML interface description
- uses common codebase for Markdown and RestructuredText

Example:

```
v.buffer --md-description
```

Comparision of HTML (right) and Markdown (left):

Screenshot from 2023-11-09 11-51-31


lib/gis/parser_rest_md.c
```diff
@@ -0,0 +1,415 @@
+/*!
+   \file lib/gis/parser_rest.c
```
neteler, line 2:
```suggestion
   \file lib/gis/parser_rest_md.c
```

neteler:
A small example here would be nice:

https://github.com/OSGeo/grass/blob/4744671171117de550d8880430104c0fba7e2400/general/g.parser/g.parser.html#L297

Like this:

```
<h2>Markdown</h2>

The flag <b>--md-description</b> added to a GRASS command
generates module interface description in Markdown, a lightweight
markup language. Example:

<div class="code"><pre>
v.in.db --md-description
</pre></div>

```

landam:
Python script to validate generated Markdown files:

```py

import subprocess
import grass.script as gs

# https://github.com/markdownlint/markdownlint
# sudo apt install ruby-mdl

for cmd in gs.get_commands()[0]:
    out_fn = "/tmp/" + cmd + ".md"
    fd = open(out_fn, "w")
    print([cmd, "--md-description"])
    p = subprocess.Popen([cmd, "--md-description"], stdout=fd)
    p.wait()
    fd.close()
    p = subprocess.Popen(["mdl", out_fn])
    p.wait()
```

wenzeslaus:
Please, add the validation script to utils and run that in CI for couple tool in the additional checks workflow. This will ensure that both this PR is okay and any subsequent tweaks don't break it.

neteler:
There is currently a compile error:

```
parser_rest_md.c: In function ‘usage_rest_md’:
parser_rest_md.c:267:29: error: format not a string literal and no format arguments [-Werror=format-security]
  267 |             fprintf(stdout, image_spec_rest);
      |                             ^~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

landam:
> Please, add the validation script to utils and run that in CI for couple tool in the additional checks workflow. This will ensure that both this PR is okay and any subsequent tweaks don't break it.

In this stage I would say that there is no need for to run validation script in CI. CI validation will be probably performed when we will modify system for generating manuals to use markdown. This PR is just about adding `--md-description` option to the parser.

utils/md_isvalid.py
```diff
@@ -0,0 +1,68 @@
+# Check if Markdown produced by --md-description is valid
+#
+# Requirements:
+# - https://github.com/markdownlint/markdownlint
+#   (on Debian/Ubuntu - apt install markdownlint)
```
neteler, line 5:
```suggestion
#   (on Debian/Ubuntu - apt install markdownlint)
#   (using Ruby installer - gem install mdl)
```

neteler (approved):
Great work, tested with EN,DE, JA locales. Thanks!

PR #3253, opened 2023-11-20
contributing: Optimize commands, text in release procedure (G83)

neteler:
Partial sync to `main`, following #2414:

- Add cron job update
- Restructure updates to website and wiki
- Remove Launchpad notes (apparently not used, at least not by the release manager)
- Change of procedure for the mailing list announcements
- Improves precision in the text

Note that there are still differences in the initial `git` sync part: also to be synced?

### DIFF
--- a/doc/howto_release.md
+++ b/doc/howto_release.md
@@ -33,12 +33,13 @@ git diff
 git diff --staged
 # Should give no output:
 git log upstream/releasebranch_8_3..HEAD
-# Should give the same as last commits visible on GitHub:
+# There should be no commits which are not visible on GitHub:
 git log --max-count=5
 ```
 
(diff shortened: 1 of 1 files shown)


wenzeslaus:
> [...] differences in the initial git sync part: also to be synced?

I would say so. The files should be almost identical except for version numbers and where we are changing the procedure (I don't recall that we do). I tried to create a diff online, but it didn't work, so local Meld is probably the best ([this is what I got](https://github.com/neteler/grass/compare/doc_howto_release_83...OSGeo:grass:main#diff-c64a1d3d50aca7dac2970359a33d3e63d2364a03cac7f20c9d461fe612225e40) - I don't know why it shows 8.2 stuff in red. I expected 8.3).

wenzeslaus (approved):

PR #3258, opened 2023-11-22
added support for Darwin 64bit arch and for ASTER L1T data products

afrigeri:
this adds to r.in.aster support for:

 1.  Darwin 64bit architecture
 2.  ATER Level 1 Precision Terrain Corrected Registered At-Sensor Radiance ([L1T](https://doi.org/10.5067/ASTER/AST_L1T.031)) containing calibrated at-sensor radiance, which corresponds with the ASTER Level 1B [(AST_L1B)](https://doi.org/10.5067/ASTER/AST_L1B.003), that has been geometrically corrected, and rotated to a north-up UTM projection.

updated code and manual.  

The code expects 15 bands, while The L1T dataset does not include the aft-looking VNIR band 3 (3b), so 3b is the copy of 3n.  Future improvements shall consider this aspect specific to L1T.

best,

Alessandro

### DIFF
--- a/scripts/r.in.aster/r.in.aster.py
+++ b/scripts/r.in.aster/r.in.aster.py
@@ -36,7 +36,7 @@
 # % key: proctype
 # % type: string
 # % description: ASTER imagery processing type (Level 1A, Level 1B, or relative DEM)
-# % options: L1A,L1B,DEM
+# % options: L1A,L1B,L1T,DEM
 # % answer: L1B
 # % required: yes
 # %end
@@ -82,6 +82,25 @@ bands = {
         "5": "SWIR_Swath:ImageData5",
         "6": "SWIR_Swath:ImageData6",
         "7": "SWIR_Swath:ImageData7",
+        "8": "SWIR_Swath:ImageData8",
+        "9": "SWIR_Swath:ImageData9",
+        "10": "TIR_Swath:ImageData10",
+        "11": "TIR_Swath:ImageData11",
+        "12": "TIR_Swath:ImageData12",
+        "13": "TIR_Swath:ImageData13",
+        "14": "TIR_Swath:ImageData14",
+    },
+    "L1T": {
+        "4": "SWIR_Swath:ImageData4",
+        "5": "SWIR_Swath:ImageData5",
+        "6": "SWIR_Swath:ImageData6",
+        "7": "SWIR_Swath:ImageData7",
+        "8": "SWIR_Swath:ImageData8",
+        "9": "SWIR_Swath:ImageData9",
+        "1": "VNIR_Swath:ImageData1",
+        "2": "VNIR_Swath:ImageData2",
+        "3n": "VNIR_Swath:ImageData3N",
+        "3b": "VNIR_Swath:ImageData3N", # A placeholder for consistency with other bands
         "8": "SWIR_Swath:ImageData8",
         "9": "SWIR_Swath:ImageData9",
         "10": "TIR_Swath:ImageData10",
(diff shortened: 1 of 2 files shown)


scripts/r.in.aster/r.in.aster.py
```diff
@@ -166,7 +183,10 @@ def import_aster(proj, srcfile, tempfile, output, band):
     grass.debug("gdalwarp -t_srs %s %s %s" % (proj, srcfile, tempfile))
 
     if platform.system() == "Darwin":
```
nilason, line 185: I suspect this architecture detection for Mac dates back to FAT binary era. But even then I don't understand why this had to be explicitly added to the command.

May I suggest to try removing lines 185-190 and out-indent the line with `cmd = ["gdalwarp", "-t_srs", proj, srcfile, tempfile]` (which will be the only remaining one).

afrigeri: hei @nilason! that's a good point.  I have been conservative, as I expected it to be something related also with M1/arm which I can not test (intel machine here).   

from the arch manual:

```     
     ...
     The other use of the arch command is to run a selected architecture of a
     universal binary.  A universal binary contains code that can run on
     different architectures.  By default, the operating system will select
     the architecture that most closely matches the processor type.  A 64-bit
     architecture is preferred over a 32-bit architecture on a 64-bit
     processor, while only 32-bit architectures can run on a 32-bit processor.

     When the most natural architecture is unavailable, the operating system
     will try to pick another architecture.  On 64-bit processors, a 32-bit
     architecture is tried.  Otherwise, no architecture is run, and an error
     results.
     ...
```

I suppose gdal for OSX is _universal binary_ , and don't know about the FAT binary era.  So it seem safe to simplify with the solution you suggest!

nilason: An ARM machine would need "arm64", eventually the list will be long...  :)

afrigeri: Got it!  I took some time to study the historical context of the [fat binary](https://en.wikipedia.org/wiki/Fat_binary).  

As far as I understand, nowadays we have gdal compiled natively for the platform we are running GRASS in.   So I'd go with your suggested code, adding a comment about this #3258 to keep track of this discussion for future reference.

nilason: It _could_ be a universal build (eg. x86_64+arm64), but it should work either way. You can safely remove the comment, the history will be in git!

scripts/r.in.aster/r.in.aster.py
```diff
@@ -133,20 +149,25 @@ def main():
         "13",
         "14",
     ]
+    
+    # Band 3b is not included ASTER L1T
+    if proctype == 'L1T':
+        allbands.pop(3)
```
nilason, line 155:
```suggestion
        allbands.remove("3b")
```

Selfdocumenting and no need to keep in sync with construction of the variable `allbands` (if there ever will be changes).

scripts/r.in.aster/r.in.aster.py
```diff
@@ -165,10 +186,9 @@ def import_aster(proj, srcfile, tempfile, output, band):
     grass.message(_("Georeferencing aster image ..."))
     grass.debug("gdalwarp -t_srs %s %s %s" % (proj, srcfile, tempfile))
 
-    if platform.system() == "Darwin":
-        cmd = ["arch", "-i386", "gdalwarp", "-t_srs", proj, srcfile, tempfile]
-    else:
-        cmd = ["gdalwarp", "-t_srs", proj, srcfile, tempfile]
+    # We assume gdal is build natively for the platform GRASS is running on.
+	# See #3258
```
nilason, line 190:
```suggestion
```

nilason (changes requested):
I added some more suggestions.
You can also delete line 55 ´import platform`, as it is no longer needed.

You will need to run the Python module `Black` (preferable version 23.1.0) on the py file as there are some formatting issues that need to be corrected (see https://github.com/OSGeo/grass/blob/main/doc/development/submitting/python.md#automation).

nilason (approved):
Looks good to me and simple local test (on Mac) succeeded.

nilason:
@afrigeri Thank you for your contribution!

afrigeri:
> @afrigeri Thank you for your contribution!

Yay!

PR #3282, opened 2023-12-02
v.in.ogr: add fix for unclosed rings

metzm:
Sometimes there are unclosed polygon rings in input vectors which is a violation of the OGR simple feature specification. `ogrinfo` reports in these cases:
```
Warning 1: Non closed ring detected. To avoid accepting it, set the OGR_GEOMETRY_ACCEPT_UNCLOSED_RING configuration option to NO
```
`ogr2ogr -makevalid` fails with
```
ERROR 1: IllegalArgumentException: Points of LinearRing do not form a closed linestring
```
This PR adds a fix by appending the first vertex to the end of an unclosed ring, thus closing it. It seems that this is usually the reason for unclosed rings: someone or some software forgot to append the first vertex as the last vertex.

This fix in `v.in.ogr` is only applied if polygon cleaning is not disabled with the `-c` flag, otherwise `v.in.ogr` will fail with a new fatal error: `Found unclosed [outer|inner] polygon ring, can be closed when cleaning polygons is enabled`

In the current behaviour, such unclosed rings will end up as invalid boundaries in the output or are removed altogether during cleaning, in the extreme case resulting in an empty output vector without any geometries.

### DIFF
--- a/vector/v.in.ogr/geom.c
+++ b/vector/v.in.ogr/geom.c
@@ -368,8 +368,9 @@ int geom(OGRGeometryH hGeomAny, struct Map_info *Map, int field, int cat,
             Points->z[0] != Points->z[lastidx]) {
             if (mk_centr) {
                 /* do not clean polygons */
-                G_fatal_error(_("Found unclosed outer polygon ring, can be "
-                                "closed when cleaning polygons is enabled"));
+                G_fatal_error(
+                    _("Found unclosed outer polygon ring, can be "
+                      "closed when cleaning polygons is not disabled"));
             }
             else {
                 G_warning(_("Closing unclosed outer polygon ring"));
@@ -435,7 +436,7 @@ int geom(OGRGeometryH hGeomAny, struct Map_info *Map, int field, int cat,
                         /* do not clean polygons */
                         G_fatal_error(
                             _("Found unclosed inner polygon ring, can be "
-                              "closed when cleaning polygons is enabled"));
+                              "closed when cleaning polygons is not disabled"));
                     }
                     else {
                         G_warning(_("Closing unclosed inner polygon ring"));
--- a/vector/v.in.ogr/geom.c
+++ b/vector/v.in.ogr/geom.c
@@ -274,6 +274,7 @@ int geom(OGRGeometryH hGeomAny, struct Map_info *Map, int field, int cat,
     OGRGeometryH hRing;
     double x, y;
     double size;
+    int lastidx;
 
     G_debug(3, "geom() cat = %d", cat);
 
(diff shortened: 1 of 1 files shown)


metzm:
Here is a test polygon as geojson, any UTM zone will do for testing:
```
{
    "features": [
        {
            "geometry": {
                "coordinates": [
                    [
                        [
                            549672.61533,
                            5937238.005984
                        ],
                        [
                            549672.778399,
                            5937238.059685
                        ],
                        [
                            549672.814345,
                            5937238.064736
                        ],
                        [
                            549673.058112,
                            5937238.119269
                        ],
                        [
                            549673.218296,
                            5937238.121337
                        ],
                        [
                            549673.414677,
                            5937238.169171
                        ],
                        [
                            549673.623124,
                            5937238.231385
                        ],
                        [
                            549673.675263,
                            5937238.246337
                        ],
                        [
                            549673.720277,
                            5937238.252535
                        ],
                        [
                            549673.921698,
                            5937238.311333
                        ],
                        [
                            549674.117898,
                            5937238.461934
                        ],
                        [
                            549674.277766,
                            5937238.484133
                        ],
                        [
                            549674.366072,
                            5937238.496335
                        ],
                        [
                            549674.555007,
                            5937238.512771
                        ],
                        [
                            549674.384837,
                            5937238.923591
                        ],
                        [
                            549673.985323,
                            5937238.923591
                        ],
                        [
                            549673.921666,
                            5937238.913261
                        ],
                        [
                            549673.652655,
                            5937238.82359
                        ],
                        [
                            549672.751883,
                            5937238.82359
                        ],
                        [
                            549672.517568,
                            5937238.589275
                        ],
                        [
                            549672.517568,
                            5937237.976638
                        ]
                    ]
                ],
                "type": "Polygon"
            },
            "properties": {
                "height": 0.046967506408691406,
                "label_id": "1",
                "label_name": "unknown",
                "object_id": 0,
            },
            "type": "Feature"
        }
    ],
    "type": "FeatureCollection"
}
```

And yes, milestone should be `8.4.0`, thanks for the hint!

vector/v.in.ogr/geom.c
```diff
@@ -361,6 +362,22 @@ int geom(OGRGeometryH hGeomAny, struct Map_info *Map, int field, int cat,
...
         Vect_line_prune(Points);
 
+        lastidx = Points->n_points - 1;
+        if (Points->x[0] != Points->x[lastidx] ||
+            Points->y[0] != Points->y[lastidx] ||
+            Points->z[0] != Points->z[lastidx]) {
+            if (mk_centr) {
+                /* do not clean polygons */
+                G_fatal_error(_("Found unclosed outer polygon ring, can be "
+                                "closed when cleaning polygons is enabled"));
```
neteler, line 372: Small remark: from the message it is not clear how to enable "cleaning polygons". Is snapping (`snap` parameter) meant?

vector/v.in.ogr/geom.c
```diff
@@ -407,6 +424,28 @@ int geom(OGRGeometryH hGeomAny, struct Map_info *Map, int field, int cat,
...
+                        IPoints[valid_isles]->x[lastidx] ||
+                    IPoints[valid_isles]->y[0] !=
+                        IPoints[valid_isles]->y[lastidx] ||
+                    IPoints[valid_isles]->z[0] !=
+                        IPoints[valid_isles]->z[lastidx]) {
+                    if (mk_centr) {
+                        /* do not clean polygons */
+                        G_fatal_error(
+                            _("Found unclosed inner polygon ring, can be "
+                              "closed when cleaning polygons is enabled"));
```
neteler, line 438: (see comment above)

vector/v.in.ogr/geom.c
metzm: Cleaning polygons is enabled by default and can be disabled with the `-c` flag: "Do not clean polygons (not recommended)". Thus maybe a more correct message would be "Found unclosed outer polygon ring, can be closed when cleaning polygons is not disabled"?

PR #3286, opened 2023-12-04
v.db.join: speed up processing by using fewer db.execute commands

griembauer:
This PR speeds up `v.db.join` for large tables. Previously, individual `v.db.addcolumn` and `db.execute` commands were run for **each** column to join.
This PR alters `v.db.addcolumn` and `v.db.join` such that only one `db.execute` command is executed for each, using a temporary sql_file instead of stdin as input for the SQL statement. 

Previous approach (deprecated): 
~~In this PR, one single `v.db.addcolumn` and `db.execute` command is executed for a chunk of max 100 columns at once. 
Here, all individual `v.db.addcolumn` commands are grouped into one single command. Additionally, all `db.execute UPDATE table...` commands to add data to the new columns are grouped into a single command. As the SQL command may become very long this way (and `incomplete input in sqlite3_prepare()` errors may be encountered, see also https://github.com/OSGeo/grass/pull/3273), commands with a total string length of > ~10.000 are again split into individual `SQL UPDATE` commands.~~


scripts/v.db.join/v.db.join.py
```diff
@@ -170,32 +168,57 @@ def main():
...
-        grass.verbose(_("Updating column <%s> of vector map <%s>...") % (colname, map))
+        set_str += cur_set_str
+        cols_in_str.append(colname)
+        if len(set_str) < max_length and i < last_idx:
+            set_str += ", "
+        else:
+            set_str_list.append(set_str)
+            set_str = ""
+            cols_list.append(cols_in_str)
+            cols_in_str = []
```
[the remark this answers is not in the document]
griembauer: Thanks, I now added an even cleaner version(from my point of view), where everything is executed for chunks of 500 columns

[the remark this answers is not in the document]
metzm: Yes, it looks better and cleaner now!

tmszi:
The problem with `db.execute input="-"` is the statically defined character buffer size if the SQL string is read from stdin `input="-"`.

https://github.com/OSGeo/grass/blob/ea8f3eae3299d70be3e46f2d1beeb9a72635e525/db/db.execute/main.c#L189-L191


https://github.com/OSGeo/grass/blob/ea8f3eae3299d70be3e46f2d1beeb9a72635e525/include/grass/dbmi.h#L142

This line truncates the SQL string to 8206 chars.

https://github.com/OSGeo/grass/blob/ea8f3eae3299d70be3e46f2d1beeb9a72635e525/db/db.execute/main.c#L196

Example of long SQL string:

1. Generate long SQL string with 30145 chars (create table) with Python

```python
sql = "CREATE TABLE soils_test (cat integer, soiltype varchar(10)"
for i in range(1200):
    sql += f", soiltype{i} varchar(10)"
sql += ")"
```
2. Try to execute this SQL string with `db.execute input="-"` command
3. `db.execute` command fails because SQL string is truncated and not valid

For this long SQL string, it is better to use **read SQL string from the file** `db.execute sql=/tmp/create_table.sql`

griembauer:
> The problem with `db.execute input="-"` is the statically defined character buffer size if the SQL string is read from stdin `input="-"`.
> 
> https://github.com/OSGeo/grass/blob/ea8f3eae3299d70be3e46f2d1beeb9a72635e525/db/db.execute/main.c#L189-L191
> 
> https://github.com/OSGeo/grass/blob/ea8f3eae3299d70be3e46f2d1beeb9a72635e525/include/grass/dbmi.h#L142
> 
> This line truncates the SQL string to 8206 chars.
> 
> https://github.com/OSGeo/grass/blob/ea8f3eae3299d70be3e46f2d1beeb9a72635e525/db/db.execute/main.c#L196
> 
> Example of long SQL string:
> 
>     1. Generate long SQL string with 30145 chars (create table) with Python
> 
> 
> ```python
> sql = "CREATE TABLE soils_test (cat integer, soiltype varchar(10)"
> for i in range(1200):
>     sql += f", soiltype{i} varchar(10)"
> sql += ")"
> ```
> 
>     2. Try to execute this SQL string with `db.execute input="-"` command
> 
>     3. `db.execute` command fails because SQL string is truncated and not valid
> 
> 
> For this long SQL string, it is better to use **read SQL string from the file** `db.execute input=/tmp/create_table.sql`

Thanks! I now updated both `v.db.join` and `v.db.update` so that they both use `db.execute`with a temporary sql file as input instead of stdin. For joining 500 INT columns to a test vector with ~750 objects this brings calculation time down from 110s to 9s on my local machine. However, in the sql_files there are still separate sql statements:
```
[...]
UPDATE builtup_vectorized_base SET test_column_491 = (SELECT test_column_491 FROM builtup_vectorized_attributes WHERE builtup_vectorized_attributes.cat=builtup_vectorized_base.cat);
UPDATE builtup_vectorized_base SET test_column_492 = (SELECT test_column_492 FROM builtup_vectorized_attributes WHERE builtup_vectorized_attributes.cat=builtup_vectorized_base.cat);
UPDATE builtup_vectorized_base SET test_column_493 = (SELECT test_column_493 FROM builtup_vectorized_attributes WHERE builtup_vectorized_attributes.cat=builtup_vectorized_base.cat);
UPDATE builtup_vectorized_base SET test_column_494 = (SELECT test_column_494 FROM builtup_vectorized_attributes WHERE builtup_vectorized_attributes.cat=builtup_vectorized_base.cat);
[...]
```
instead of having just one `UPDATE` statement. I am not sure if this would significantly improve the processing time again, but I couldn't manage to put together a single working `UPDATE` statement - input is welcome!

neteler:
Probably a set of SQL commands should be wrapped into a [TRANSACTION](https://www.sqlite.org/lang_transaction.html)?

```
    sql = ["BEGIN TRANSACTION"]
   ...
    sql.append("END TRANSACTION")
```

Random example:

https://github.com/OSGeo/grass/blob/24365e1d47a5684be65648e3b99e3021270c80e3/scripts/v.dissolve/v.dissolve.py#L200

scripts/v.db.join/v.db.join.py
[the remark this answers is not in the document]
griembauer: This is an outdated version, so I will resolve this conversation. Please feel free to review again @metzm

scripts/v.db.addcolumn/v.db.addcolumn.py
```diff
@@ -40,11 +40,25 @@
...
+rm_files = []
+
+
+def cleanup():
+    for file in rm_files:
+        if os.path.isfile(file):
+            try:
+                os.remove(file)
+            except Exception as e:
+                grass.warning(_("Unable to remove file %s: %s" % (file, e)))
```
tmszi, line 57: You can use `.format()` func, but don't use f-string for trans string, please.

```python
grass.warning(
    _("Unable to remove file {file}: {message}").format(
        file=file, message=e
    )
)
```

scripts/v.db.addcolumn/v.db.addcolumn.py
```diff
@@ -87,20 +102,27 @@ def main():
             grass.error(_("Column <%s> is already in the table. Skipping.") % col_name)
             continue
         grass.verbose(_("Adding column <%s> to the table") % col_name)
-        p = grass.feed_command(
-            "db.execute", input="-", database=database, driver=driver
+        add_str += "ALTER TABLE {} ADD COLUMN {};\n".format(table, col)
```
tmszi, line 105: You can use f-string here.

```python
add_str += f"ALTER TABLE {table} ADD COLUMN {col};\n"
```

scripts/v.db.addcolumn/v.db.addcolumn.py
```diff
@@ -87,20 +102,27 @@ def main():
...
         )
-        res = "ALTER TABLE {} ADD COLUMN {}".format(table, col)
-        p.stdin.write(encode(res))
-        grass.debug(res)
-        p.stdin.close()
-        if p.wait() != 0:
-            grass.fatal(_("Unable to add column <%s>.") % col)
-
+    except CalledModuleError:
+        grass.fatal(_(f"Error adding columns {cols_add_str}"))
```
tmszi, line 120: For trans string use `.format()` method instead of f-string.

```python
grass.fatal(_("Error adding columns {}").format(cols_add_str))
```

scripts/v.db.join/v.db.join.py
```diff
@@ -173,28 +183,41 @@ def main():
...
-        try:
-            grass.write_command(
-                "db.execute", stdin=stmt, input="-", database=database, driver=driver
-            )
-        except CalledModuleError:
-            grass.fatal(_("Error filling column <%s>") % colname)
+        update_str += cur_up_str
+    update_str += "END TRANSACTION"
+    grass.debug(update_str, 1)
+    grass.verbose(_(f"Updating columns {cols_added_str} of vector map {map}..."))
```
tmszi, line 207: For trans string use `.format()` method instead of f-string, please.

```python
grass.verbose(
    _("Updating columns {columns} of vector map {map_name}...").format(
        columns=cols, map_name=map
    )
)
```

scripts/v.db.join/v.db.join.py
```diff
@@ -173,28 +183,41 @@ def main():
...
+        write_file.write(update_str)
+    try:
+        grass.run_command(
+            "db.execute",
+            input=sql_file,
+            database=database,
+            driver=driver,
+        )
+    except CalledModuleError:
+        grass.fatal(_(f"Error filling columns {cols_added_str}"))
```
tmszi, line 220: For trans string use `.format()` method instead of f-string, please.

```python
grass.fatal(_("Error filling columns {}").format(cols_added_str))
```

tmszi (approved):
First I appologize for long delay. 

Looks good to me. I tested this fix with adding 500 column and time was reduced from approximately 1 minute to  0.721s.

metzm (approved):
Looks good to me, impressive speed-up!

PR #3315, opened 2023-12-25
wxGUI/mapdisp: save/load Map Display used output img size for session

### LINKED ISSUE #3215: [Feat] g.gui: Remember last image size for Save display to file

When I save the display to file multiple times, I'd like to use the same image size to be consistent.

It would be very useful to remember last image size for Save display to file.

Remember the last used image size and manually type it every time.

**Additional context**

Creating multiple map images in the same consistent size would be great!

tmszi:
Closes #3215.


petrasovaa (approved):
I didn't test, but looks good and indeed useful!

PR #3325, opened 2024-01-03
Use integer port argument with mysql_real_connect

fweimer-rh:
The port argument is not a string, but a plain integer.  In the past, compilers ignore such type errors and produced a warning only, but this is changing, so this change also addresses a potential build failure.

Note that I can't really test this, but at least the compiler error is gone. Found as part of:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/db/drivers/mysql/db.c
+++ b/db/drivers/mysql/db.c
@@ -53,10 +53,9 @@ int db__driver_open_database(dbHandle *handle)
         db_get_login2("mysql", name, &user, &password, &host, &port);
 
         connection = mysql_init(NULL);
-        res = mysql_real_connect(connection, host, user, password,
-                                 connpar.dbname,
-                                 port != NULL ? atoi(port) : 0,
-                                 NULL, 0);
+        res =
+            mysql_real_connect(connection, host, user, password, connpar.dbname,
+                               port != NULL ? atoi(port) : 0, NULL, 0);
 
         if (res == NULL) {
             db_d_append_error("%s\n%s", _("Connection failed."),
--- a/db/drivers/mysql/db.c
+++ b/db/drivers/mysql/db.c
@@ -54,7 +54,9 @@ int db__driver_open_database(dbHandle *handle)
 
         connection = mysql_init(NULL);
         res = mysql_real_connect(connection, host, user, password,
-                                 connpar.dbname, port, NULL, 0);
+                                 connpar.dbname,
+                                 port != NULL ? atoi(port) : 0,
+                                 NULL, 0);
 
         if (res == NULL) {
             db_d_append_error("%s\n%s", _("Connection failed."),


wenzeslaus (approved):
Thank you for the fix!

This is consistent with _mysql_real_connect_ and _db_get_login2_ which creates the string.

nilason (approved):
Thanks @fweimer-rh !
[`mysql_real_connect()`](https://dev.mysql.com/doc/c-api/8.2/en/mysql-real-connect.html) does indeed use `unsigned int` for the port (in comparison to [PQsetdbLogin()](https://www.postgresql.org/docs/16/libpq-connect.html#LIBPQ-PQSETDBLOGIN) which uses `char *`).

wenzeslaus:
Anyone has a trick for quickly fixing clang-format issues without pulling the code locally? Unlike with Black, I'm not good at guessing what the right formatting is from the message.

db/drivers/mysql/db.c
```diff
@@ -54,7 +54,9 @@ int db__driver_open_database(dbHandle *handle)
 
         connection = mysql_init(NULL);
         res = mysql_real_connect(connection, host, user, password,
```
echoix, line 56:
```suggestion
        res =
            mysql_real_connect(connection, host, user, password, connpar.dbname,
```
image

```
        res =
            mysql_real_connect(connection, host, user, password, connpar.dbname,
                               port != NULL ? atoi(port) : 0, NULL, 0);
```

echoix:
Its ugly.

neteler:
(For the record for those not familiar with `gh`: it is fairly straighforward)

```
gh pr checkout 3325
utils/grass_clang_format.sh db/drivers/mysql/db.c
git diff
git commit -m"utils/grass_clang_format.sh db/drivers/mysql/db.c" db/drivers/mysql/db.c
git push 
```

Done in fc9803f

nilason:
> Anyone has a trick for quickly fixing clang-format issues without pulling the code locally? Unlike with Black, I'm not good at guessing what the right formatting is from the message.

#3255 ;-)

nilason:
I believe we have no CI test with MySQL driver, I’ll look into this. I suppose this should have been caught.

wenzeslaus:
> [The formatting is] ugly. I wasn't able to create a good suggestion.

I agree. We didn't necessarily choose the format. We used whatever was already there and what clang-format supported at the time hoping to get back to creating a nicer style with a more powerful version of clang-format.

nilason:
> > [The formatting is] ugly. I wasn't able to create a good suggestion.
> 
> I agree. We didn't necessarily choose the format. We used whatever was already there and what clang-format supported at the time hoping to get back to creating a nicer style with a more powerful version of clang-format.

Attractive or not, the point of using formatters like ClangFormat and Black is to avoid non-productive discussions on formatting with potentially no end ...... :)

wenzeslaus:
> ...the point of using formatters like ClangFormat and Black is to avoid non-productive discussions on formatting with potentially no end ...... :)

Again, I [deny](https://wenzeslaus.github.io/code-quality-measures-foss4g-2022/#/14) ever taking a part in such discussion.

Issue #3343, opened 2024-01-08
Build errors with missing lex


marisn:
If lex/flex is missing, it is not a fatal error in configuration phase, but results in errors during compilation phase. Thus either:

- code must be conditioned to skip lex dependent parts if it is missing or
- missing lex should be a fatal error of configure call

As one of failing modules is r.mapcalc, I vote for second option (fatal error of configure script)

PR #3345, opened 2024-01-08
configure: fail configuration with missing lex/flex

nilason:
Closes https://github.com/OSGeo/grass/issues/3343

### DIFF
--- a/configure
+++ b/configure
@@ -6444,7 +6444,9 @@ fi
 rm -f conftest.l $LEX_OUTPUT_ROOT.c
 
 fi
-if test "$LEX" = "lex"; then
+if test "x$LEX" = "x:"; then
+    as_fn_error $? "*** Unable to locate lex." "$LINENO" 5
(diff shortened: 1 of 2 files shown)


marisn (approved):

nilason:
I'm not aware of any package distributing system building GRASS without [f]lex.
 @neteler ,  @lbartoletti Maybe you have some idea if this can be safe?

nilason:
> The real question is: Should `r[3].mapcalc` and `lib/db/sqlp/sqlp.y` be mandatory for compilation or not? (I guess yes)
> 
> `REQUIREMENTS.md` lists `flex` under "General requirements".

Thanks! Well "General requirements" means in my understanding simply "required". Unless we want to change that this should be ok.
Now Debian and MacPorts already lists directly `flex` as dependency; FreeBSD, Fedora and Ubuntu delivers with r[3].mapcalc (at least implicitly using flex).

PR #3348, opened 2024-01-08
lib/vector/Vlib: fix count of removed dangles

metzm:
The count of removed dangles in [dangles()](https://github.com/OSGeo/grass/blob/main/lib/vector/Vlib/dangles.c#L257) was wrong, counting all detected dangles. This PR fixes the count of removed dangles.


nilason (approved):
Looks good to me!

PR #3353, opened 2024-01-11
g.extension: fix PYTHONPATH

anikaweinmann:
Fixes https://github.com/OSGeo/grass-addons/issues/1013 (relevant for addons which consist of several Python files).

### DIFF
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -2055,6 +2055,9 @@ def install_extension_std_platforms(name, source, url, branch):
 
     os.chdir(srcdir)
 
+    # add TMPDIR to PYTHONPATH
+    os.environ["PYTHONPATH"] = f"{TMPDIR}:{os.environ['PYTHONPATH']}"
+
     gs.message(_("Compiling..."))
     if not os.path.exists(os.path.join(gisbase, "include", "Make", "Module.make")):
         gs.fatal(_("Please install GRASS development package"))
--- a/scripts/g.extension/g.extension.py
+++ b/scripts/g.extension/g.extension.py
@@ -2055,8 +2055,8 @@ def install_extension_std_platforms(name, source, url, branch):
 
     os.chdir(srcdir)
 
-    # add TMPDIR to PYTHONPATH
-    os.environ["PYTHONPATH"] = f"{TMPDIR}:{os.environ['PYTHONPATH']}"
+    # add srcdir to PYTHONPATH
+    os.environ["PYTHONPATH"] = f"{srcdir}:{os.environ['PYTHONPATH']}"
 
     gs.message(_("Compiling..."))
     if not os.path.exists(os.path.join(gisbase, "include", "Make", "Module.make")):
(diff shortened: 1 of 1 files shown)


scripts/g.extension/g.extension.py
```diff
@@ -2055,6 +2055,9 @@ def install_extension_std_platforms(name, source, url, branch):
 
     os.chdir(srcdir)
 
+    # add TMPDIR to PYTHONPATH
+    os.environ["PYTHONPATH"] = f"{TMPDIR}:{os.environ['PYTHONPATH']}"
```
neteler, line 2059:
```suggestion
    # add srcdir to PYTHONPATH
    os.environ["PYTHONPATH"] = f"{srcdir}:{os.environ['PYTHONPATH']}"
```

neteler: With this change it also works on Fedora 38.

neteler (approved):
Thanks a lot!

scripts/g.extension/g.extension.py
```diff
@@ -2055,6 +2055,9 @@ def install_extension_std_platforms(name, source, url, branch):
 
     os.chdir(srcdir)
 
+    # add srcdir to PYTHONPATH
+    os.environ["PYTHONPATH"] = f"{srcdir}:{os.environ['PYTHONPATH']}"
```
ninsbl, line 2059: The following would be more crossplatform, but since it is not relevant on Windows anyway, you can go with :, I guess...
```suggestion
    os.environ["PYTHONPATH"] = f"{srcdir}{os.pathsep}{os.environ['PYTHONPATH']}"
```

ninsbl (approved):
Good catch! Thanks for the contribution.

scripts/g.extension/g.extension.py
ninsbl: Sorry, I was not aware this was blocking a merge...

neteler: Is it?

ninsbl: It was blocking. Any unresolved conversation blocks merging. See "unresolve" button below...

ninsbl: And now, after clicking unresolved, it should be possible...

neteler: Yes, indeed! Done.

tmszi:
My suggestion is to use a lazy import in the ./src/imagery/i.fusion.hpf/i.fusion.hpf.py script file.


```python
def main():
    # import modules from "etc"
    from high_pass_filter import (
        get_high_pass_filter,
        get_modulator_factor,
        get_modulator_factor2,
    )
```

```
GRASS nc_spm_08_grass7/test:~ > g.extension i.fusion.hpf url=/home/tomas/src/grass-addons/src/imagery/i.fusion.hpf/
Compiling...
Installing...
Updating extensions metadata file...
Updating extension modules metadata file...
Installation of <i.fusion.hpf> successfully finished
```

tmszi:
Better fix at addon level https://github.com/OSGeo/grass-addons/pull/1014.

PR #3365, opened 2024-01-16
wxGUI/history: Create panel displaying info about command

lindakladivova:
This PR implements new functionality to the History Browser panel, particularly showing info about a command from history through the new Command Info dialog.

- The history log (.wxgui_history) is newly stored as a JSON file. It is backward compatible with the plain .wxgui_history. But in the case that a user creates a new mapset, a new history file is automatically created as a JSON file.

- This PR further implements a new option in the context menu called "Show Info." "Show Info" displays a small dialog called **Command info** about the respective process stored in the JSON file.

- JSON file stores the following info for each executed command:

1. Execution time
2. Runtime duration
3. Status (In process / Successfully finished/ Failed)
4. If a 2D mask was set
5. Executor login
6. Region settings (separate notebook page in Command info dialog)

Screenshot from 2024-01-16 21-45-28

Example of the JSON structure:

```
[
  {
    "command": "v.what map=boundaries@PERMANENT coordinates=1.0658362989323844,0.7413997627520759",
    "command_info": {
      "Execution time": "Mon Jan 22 10:08:16",
      "Runtime duration": "0 sec",
      "Status code": "Successfully finished",
      "2D mask set": true,
      "Executor login": "linduska",
      "Region settings": {
        "projection": "99 (ETRS89-extended / LAEA Europe)",
        "zone": "0",
        "datum": "etrs89",
        "ellipsoid": "grs80",
        "north": "1",
        "south": "0",
        "west": "0",
        "east": "1",
        "nsres": "1",
        "ewres": "1",
        "rows": "1",
        "cols": "1",
        "cells": "1"
      }
    }
  },
  {
    "command": "v.what map=boundaries@PERMANENT coordinates=0.8190984578884934,0.6500593119810201",
    "command_info": {
      "Execution time": "Mon Jan 22 10:09:09",
      "Runtime duration": "0 sec",
      "Status code": "Successfully finished",
      "2D mask set": true,
      "Executor login": "linduska",
      "Region settings": {
        "projection": "99 (ETRS89-extended / LAEA Europe)",
        "zone": "0",
        "datum": "etrs89",
        "ellipsoid": "grs80",
        "north": "3017225",
        "south": "3000000",
        "west": "4623050",
        "east": "4657625",
        "nsres": "25",
        "ewres": "25",
        "rows": "689",
        "cols": "1383",
        "cells": "952887"
      }
    }
  }
]
```

### DIFF
--- a/gui/wxpython/history/browser.py
+++ b/gui/wxpython/history/browser.py
@@ -27,6 +27,8 @@ import wx.lib.scrolledpanel as SP
 from gui_core.wrap import SearchCtrl, StaticText, StaticBox, Button
 from history.tree import HistoryBrowserTree
 
+import grass.grassdb.history as history
+
 from grass.pydispatch.signal import Signal
 
 from core.gcmd import GError
@@ -353,7 +355,7 @@ class HistoryBrowser(wx.SplitterWindow):
 
     def OnCmdExportHistory(self, event):
         """Export the history of executed commands to a selected file."""
-        if self.tree.history_manager.filetype == "json":
+        if history.get_extension() == ".json":
             defaultFile = "grass_cmd_log.json"
             wildcard = _("{json} (*.json)|*.txt|{files} (*)|*").format(
                 json=_("JSON files"), files=_("Files")
@@ -373,7 +375,7 @@ class HistoryBrowser(wx.SplitterWindow):
         if dlg.ShowModal() == wx.ID_OK:
             path = dlg.GetPath()
             try:
-                self.tree.history_manager.copy(path)
+                history.copy(path)
                 self.showNotification.emit(
                     message=_("Command history saved to '{}'".format(path))
                 )
--- a/gui/wxpython/history/browser.py
+++ b/gui/wxpython/history/browser.py
@@ -234,9 +234,11 @@ class HistoryInfo(SP.ScrolledPanel):
 
     def showCommandInfo(self, command_info):
         """Show command info input."""
-        self._updateGeneralInfoBox(command_info)
-        self._updateRegionSettingsBox(command_info)
-
+        if command_info:
+            self._updateGeneralInfoBox(command_info)
+            self._updateRegionSettingsBox(command_info)
+        else:
+            self.clearCommandInfo()
         self.SetupScrolling(scroll_x=False, scroll_y=True)
         self.Layout()
 
--- a/gui/wxpython/history/browser.py
+++ b/gui/wxpython/history/browser.py
@@ -19,23 +19,15 @@ for details.
 @author Tomas Zigo
 """
 
+from datetime import datetime
+
 import wx
 import wx.lib.scrolledpanel as SP
 
-<<<<<<< HEAD
-from gui_core.wrap import SearchCtrl, Button
-=======
-from datetime import datetime
-
-from gui_core.wrap import SearchCtrl, StaticText, StaticBox
->>>>>>> 193e7de2f0 (first proposal for splitter window, WIP)
+from gui_core.wrap import SearchCtrl, StaticText, StaticBox, Button
 from history.tree import HistoryBrowserTree
 
 from grass.pydispatch.signal import Signal
-from grass.grassdb.history import (
-    copy_history,
-    get_current_mapset_gui_history_path,
-)
 
 from core.gcmd import GError
 
@@ -293,23 +285,14 @@ class HistoryBrowser(wx.SplitterWindow):
         self.search.Bind(wx.EVT_TEXT, lambda evt: self.tree.Filter(evt.GetString()))
         self.search.Bind(wx.EVT_SEARCHCTRL_CANCEL_BTN, lambda evt: self.tree.Filter(""))
 
-<<<<<<< HEAD
         # buttons
-        self.btnCmdExportHistory = Button(parent=self, id=wx.ID_ANY)
+        self.btnCmdExportHistory = Button(parent=self.panelInfo, id=wx.ID_ANY)
         self.btnCmdExportHistory.SetLabel(_("&Export history"))
         self.btnCmdExportHistory.SetToolTip(
             _("Export history of executed commands to a file")
         )
         self.btnCmdExportHistory.Bind(wx.EVT_BUTTON, self.OnCmdExportHistory)
 
-        # tree with layers
-        self.tree = HistoryBrowserTree(self, giface=giface)
-        self.tree.SetToolTip(_("Double-click to run selected tool"))
-        self.tree.showNotification.connect(self.showNotification)
-        self.tree.runIgnoredCmdPattern.connect(self.runIgnoredCmdPattern)
-
-=======
->>>>>>> 193e7de2f0 (first proposal for splitter window, WIP)
         self._layout()
 
     def _layout(self):
@@ -323,7 +306,10 @@ class HistoryBrowser(wx.SplitterWindow):
             flag=wx.ALL | wx.EXPAND,
             border=5,
         )
-<<<<<<< HEAD
+        self.browserSizer.Add(
+            self.tree, proportion=1, flag=wx.EXPAND | wx.LEFT | wx.RIGHT, border=5
+        )
+
         btnSizer = wx.BoxSizer(wx.HORIZONTAL)
         btnSizer.AddStretchSpacer()
         btnSizer.Add(
@@ -332,18 +318,16 @@ class HistoryBrowser(wx.SplitterWindow):
             flag=wx.EXPAND | wx.LEFT | wx.RIGHT,
             border=5,
         )
-        sizer.Add(
-=======
-        self.browserSizer.Add(
->>>>>>> 193e7de2f0 (first proposal for splitter window, WIP)
-            self.tree, proportion=1, flag=wx.EXPAND | wx.LEFT | wx.RIGHT, border=5
-        )
-        sizer.Add(btnSizer, proportion=0, flag=wx.EXPAND | wx.ALL, border=5)
-
         self.infoSizer.Add(
             self.infoPanel,
             proportion=1,
             flag=wx.EXPAND | wx.LEFT | wx.RIGHT | wx.TOP,
+            border=5,
+        )
+        self.infoSizer.Add(
+            btnSizer,
+            proportion=0,
+            flag=wx.ALL | wx.EXPAND,
             border=5,
         )
 
@@ -359,8 +343,8 @@ class HistoryBrowser(wx.SplitterWindow):
 
         # split window
         self.SplitHorizontally(self.panelBrowser, self.panelInfo, 0)
-        self.SetMinimumPaneSize(100)
-        self.SetSashPosition(self.GetSize().y - 400)
+        self.SetMinimumPaneSize(self.btnCmdExportHistory.GetSize()[1] + 100)
+        self.SetSashPosition(self.GetSize().y - 500)
         self.SetSashGravity(1.0)
 
         # layout
@@ -368,23 +352,28 @@ class HistoryBrowser(wx.SplitterWindow):
         self.Layout()
 
     def OnCmdExportHistory(self, event):
-        """Export the history of executed commands stored
-        in a .wxgui_history file to a selected file."""
+        """Export the history of executed commands to a selected file."""
+        if self.tree.history_manager.filetype == "json":
+            defaultFile = "grass_cmd_log.json"
+            wildcard = _("{json} (*.json)|*.txt|{files} (*)|*").format(
+                json=_("JSON files"), files=_("Files")
+            )
+        else:
+            defaultFile = "grass_cmd_log.txt"
+            wildcard = _("{txt} (*.txt)|*.txt|{files} (*)|*").format(
+                txt=_("Text files"), files=_("Files")
+            )
         dlg = wx.FileDialog(
             self,
             message=_("Save file as..."),
-            defaultFile="grass_cmd_log.txt",
-            wildcard=_("{txt} (*.txt)|*.txt|{files} (*)|*").format(
-                txt=_("Text files"), files=_("Files")
-            ),
+            defaultFile=defaultFile,
+            wildcard=wildcard,
             style=wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT,
         )
-
         if dlg.ShowModal() == wx.ID_OK:
             path = dlg.GetPath()
-            history_path = get_current_mapset_gui_history_path()
             try:
-                copy_history(path, history_path)
+                self.tree.history_manager.copy(path)
                 self.showNotification.emit(
                     message=_("Command history saved to '{}'".format(path))
                 )
--- a/gui/wxpython/history/browser.py
+++ b/gui/wxpython/history/browser.py
@@ -43,49 +43,40 @@ from core.gcmd import GError
 class CommandInfoMapper:
     """Class for mapping command info values to the structure used in GUI."""
 
+    TRANSLATION_KEYS = {
+        "timestamp": _("Timestamp: "),
+        "runtime": _("Runtime duration: "),
+        "status": _("Status: "),
+        "mask2d": _("Mask 2D: "),
+        "mask3d": _("Mask 3D: "),
+        "n": _("North: "),
+        "s": _("South: "),
+        "w": _("West: "),
+        "e": _("East: "),
+        "nsres": _("North-south resolution: "),
+        "ewres": _("East-west resolution: "),
+        "rows": _("Number of rows: "),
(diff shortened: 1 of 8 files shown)


petrasovaa:
I would revise the JSON format, generally I would use lower case and leave formatting of time to GUI.

* "Execution time" -> "timestamp" and use something like ISO 8601 format (right now there is not even a year there)
* "Runtime duration" -> "runtime" maybe use integer representing number of seconds
* "Status code" -> "status" could be one of failed/success/aborted
* "2D mask set" -> "mask"
* "Executor login": not sure how useful this is, get_mapset_owner would always give the same result for all commands (None on Windows), maybe use something different?
* "Region settings" -> "region", values should be floats/integer, not strings, and I would drop the lines that refer to projection. I don't know why g.region even prints that.

Then in the GUI the labels should be translatable. So for example JSON will have status "success" and GUI will use `_("Success")`.

It would be better to have the command itself in a parsed format.

Then regarding the dialog, wouldn't it be better to display the information within the history browser, e.g. at the bottom? What would be even nicer I think is to expand the command node and show the info below the command, but maybe we don't need to do this in this PR.

lindakladivova:
> I would revise the JSON format, generally I would use lower case and leave formatting of time to GUI.
> 
>     * "Execution time" -> "timestamp" and use something like ISO 8601 format (right now there is not even a year there)
> 
>     * "Runtime duration" -> "runtime" maybe use integer representing number of seconds
> 
>     * "Status code" -> "status" could be one of failed/success/aborted
> 
>     * "2D mask set" -> "mask"
> 
>     * "Executor login": not sure how useful this is, get_mapset_owner would always give the same result for all commands (None on Windows), maybe use something different?
> 
>     * "Region settings" -> "region", values should be floats/integer, not strings, and I would drop the lines that refer to projection. I don't know why g.region even prints that.
> 
> 
> Then in the GUI the labels should be translatable. So for example JSON will have status "success" and GUI will use `_("Success")`.
> 
> It would be better to have the command itself in a parsed format.
> 
Thanks Anna for review. Ok, I will focus more on a JSON file and incorporate suggested stuff.

> Then regarding the dialog, wouldn't it be better to display the information within the history browser, e.g. at the bottom? What would be even nicer I think is to expand the command node and show the info below the command, but maybe we don't need to do this in this PR.

Yes @petrasovaa , you are right, The Command Info dialog is quite hidden for users, to have the info directly accessible at the bottom of the History browser would be better. What do you think about the structure of the dialog content? Are two notebooks "General info" and "Region settings" okay, or would you devise something different?

As I was thinking about your suggestions, I also tried to focus more on the better user experience and I think it would be also nice to have other functionalities as part of this PR:
1) Info about the command is shown as part of OnItemSelected event (no need for context menu option)
2) When the command is launched by "Run", the new entry in the history browser is automatically selected (info is automatically displayed) - then we really do not need to have the "In process" status, since it is quite clear that if the status is not filled, the command is probably still running
3) After GRASS startup, the last entry in the history browser is automatically selected (and its info is displayed)

I agree that expanding the command node is an interesting functionality for the next PR.

petrasovaa:
> > Then regarding the dialog, wouldn't it be better to display the information within the history browser, e.g. at the bottom? What would be even nicer I think is to expand the command node and show the info below the command, but maybe we don't need to do this in this PR.
> 
> Yes @petrasovaa , you are right, The Command Info dialog is quite hidden for users, to have the info directly accessible at the bottom of the History browser would be better. What do you think about the structure of the dialog content? Are two notebooks "General info" and "Region settings" okay, or would you devise something different?

I am not completely sure, I would perhaps merge it into one panel. Then place it at the bottom, similarly to how command console has it, and include a vertical scrollbar, so the region info would be hidden most of the time, which is fine. Eventually, it may be good to have an option (a button?) to set the region based on the command. It brings up a question when you rerun the command, whether it should rerun it with the same region or the current one (probably current one).




> As I was thinking about your suggestions, I also tried to focus more on the better user experience and I think it would be also nice to have other functionalities as part of this PR:
> 
> 1. Info about the command is shown as part of OnItemSelected event (no need for context menu option)
yes

> 2. When the command is launched by "Run", the new entry in the history browser is automatically selected (info is automatically displayed) - then we really do not need to have the "In process" status, since it is quite clear that if the status is not filled, the command is still running

AFAIU yes

> 3. After GRASS startup, the last entry in the history browser is automatically selected (and its info is displayed)

I don't think you have to have that, the info will be empty, which is fine.

> 
> I agree that expanding the command node is an interesting functionality for the next PR.

lindakladivova:
> I would revise the JSON format, generally I would use lower case and leave formatting of time to GUI.
> 
>     * "Execution time" -> "timestamp" and use something like ISO 8601 format (right now there is not even a year there)
> 
>     * "Runtime duration" -> "runtime" maybe use integer representing number of seconds
> 
>     * "Status code" -> "status" could be one of failed/success/aborted
> 
>     * "2D mask set" -> "mask"
> 
>     * "Executor login": not sure how useful this is, get_mapset_owner would always give the same result for all commands (None on Windows), maybe use something different?
> 
>     * "Region settings" -> "region", values should be floats/integer, not strings, and I would drop the lines that refer to projection. I don't know why g.region even prints that.
> 
> 
> Then in the GUI the labels should be translatable. So for example JSON will have status "success" and GUI will use `_("Success")`.
> 
> It would be better to have the command itself in a parsed format.
> 
> Then regarding the dialog, wouldn't it be better to display the information within the history browser, e.g. at the bottom? What would be even nicer I think is to expand the command node and show the info below the command, but maybe we don't need to do this in this PR.

What about 3D mask? Make it sense to have it here as well as the attribute?

lindakladivova:
> I would revise the JSON format, generally I would use lower case and leave formatting of time to GUI.
> 
>     * "Execution time" -> "timestamp" and use something like ISO 8601 format (right now there is not even a year there)
> 
>     * "Runtime duration" -> "runtime" maybe use integer representing number of seconds
> 
>     * "Status code" -> "status" could be one of failed/success/aborted
> 
>     * "2D mask set" -> "mask"
> 
>     * "Executor login": not sure how useful this is, get_mapset_owner would always give the same result for all commands (None on Windows), maybe use something different?
> 
>     * "Region settings" -> "region", values should be floats/integer, not strings, and I would drop the lines that refer to projection. I don't know why g.region even prints that.

 I used g.region with the -g flag instead of the previously used -p flag. There's still some info about the projection but only "projection" and "zone" and both as integers ("datum" and "ellipsoid" are missing). I can drop it but I think that it is more consistent just to preserve the output from g.region as it is when it adds only these two extra info.. 

The JSON structure looks as follows now (the command itself will be parsed in the next step).

```
[
  {
    "command": "v.what map=boundaries@PERMANENT coordinates=4632304.615724382,3013759.0437279153",
    "command_info": {
      "timestamp": "2024-01-25T15:56:19.743376",
      "mask": false,
      "region": {
        "projection": 99,
        "zone": 0,
        "n": 3017225,
        "s": 3000000,
        "w": 4623050,
        "e": 4657625,
        "nsres": 25,
        "ewres": 25,
        "rows": 689,
        "cols": 1383,
        "cells": 952887
      },
      "runtime": 0,
      "status": "success"
    }
  }
]

```

> 
> Then in the GUI the labels should be translatable. So for example JSON will have status "success" and GUI will use `_("Success")`.
> 
> It would be better to have the command itself in a parsed format.
> 
> Then regarding the dialog, wouldn't it be better to display the information within the history browser, e.g. at the bottom? What would be even nicer I think is to expand the command node and show the info below the command, but maybe we don't need to do this in this PR.

petrasovaa:
> What about 3D mask? Make it sense to have it here as well as the attribute?

Yes, although it's low priority. I wonder if the GUI should show the info about mask only when it was active to keep the command info short and relevant.

lindakladivova:
I am sharing the picture of the current version of GUI (implementation itself works but the code needs some cleaning and refactoring):
Screenshot from 2024-01-30 11-19-18

After clicking on the command, the info appears at the bottom part of the History pane.
What do you think about the current proposal @petrasovaa/ @tmszi  ?

petrasovaa:
I didn't test, but it looks fine. I wouldn't spend too much time on the look of the GUI yet, it would be best to get this PR in and then see if we want to change the appearance. Thanks @tmszi for testing!

lindakladivova:
> > After clicking on the command, the info appears at the bottom part of the History pane. What do you think about the current proposal @petrasovaa/ @tmszi ?
> 
> Looks good to me.
> 
> Some info from the my quickly testing:
Thank @tmszi for testing!
> 
>     1. During my testing sometimes `d.vect` command has General info with Timestamp info only, and sometimes has extended info as Runtime Duration and Status.

This is caused by the fact that the self._giface.entryInHistoryUpdated is called on one specific place - on OnCmdDone. If the commands d.vect, d.rast are executed from the console pane, they go through the RunCmd method in gconsole.py (so a history entry is correctly added) but then they are evaluated as ignoredCmdPattern and they never get to the call of OnCmdDone. But I think that it is not a big issue and we can let it as it is for now (not many people are using d.vect directly from the command line)..

> 
>     2. When I remove the selected command item from the history tree, the info panel should also be empty
Fixed.
> 
>     3. Is there some command/internal method/func which provide transition from existed old simple .wxgui_history file to the new JSON format? Because with old simple history file, History panel command history tree items is not correctly rendered and it is not working as expected.

To point 3 - I fixed it, but generally I was not much convinced about the solution I made (the new JSON file is created only when a user creates a new mapset) so yesterday I discussed it with @landam and we decided to automatically convert the plain text file to JSON file when launching a new command. This is however a bit problematic regarding the fact that we wanted to have the parsed command in a JSON file (as list - I think @petrasovaa suggested it), so whenever this conversion would happen it would require parsing all commands from old plain-text history log into lists. So, from this point of view, it would be better to preserve the logic of commands stored as strings in the JSON file similarly as in the plain text. So we will basically have two "read" methods (from JSON and from plain-text) but only one write method which will always write into JSON. I will change the current solution to this variant in a separate commit.

gui/wxpython/history/browser.py
```diff
@@ -17,22 +19,229 @@
...
-    copy_history,
-    get_current_mapset_gui_history_path,
-)
 
 from core.gcmd import GError
 
 
-class HistoryBrowser(wx.Panel):
-    """History browser panel for executing the commands from history log.
+class CommandInfoMapper:
```
petrasovaa, line 37: Perhaps 2 separate functions may be better here instead of the class.

petrasovaa (changes requested):

python/grass/grassdb/history.py
```diff
@@ -18,104 +19,343 @@
...
+    history_filename = ".wxgui_history"
+
+    txt_path = os.path.join(base_path, history_filename)
+    json_path = os.path.join(base_path, f"{history_filename}.json")
+
+    # Return path to plain-text file if exists otherwise return JSON
+    return txt_path if os.path.exists(txt_path) else json_path
+
+
+def get_extension():
```
petrasovaa, line 32: The name could be more descriptive, like `get_history_file_extension`.

python/grass/grassdb/history.py
```diff
@@ -18,104 +19,343 @@
...
+                pass
+        except OSError as e:
+            raise OSError(
+                _("Unable to create history file {}").format(history_path)
+            ) from e
 
 
-def create_history_file(history_path):
-    """Set up a new GUI history file."""
+def _read_from_PT(history_path):
```
petrasovaa, line 56: PT is unclear, use e.g. `read_from_plain_text`

python/grass/grassdb/history.py
```diff
@@ -18,104 +19,343 @@
...
+
+def get_extension():
+    """Return extension of the current mapset history file.
+    The standard long-term format is plain text.
+    Newly it is stored as JSON-formatted.
+    """
+    history_path = get_current_mapset_gui_history_path()
+    file_path, file_name = os.path.split(history_path)
+    file_name_no_ext, extension = os.path.splitext(file_name)
+    return extension
```
petrasovaa, line 40: Let's use pathlib for this.

lindakladivova: I used pathlib in the whole file, it is clearer.

python/grass/grassdb/history.py
```diff
@@ -18,104 +19,343 @@
...
+    """
+    from datetime import datetime
+    import grass.script as grass
+    from grass.script.utils import parse_key_val
+
+    # Execution timestamp in ISO 8601 format
+    exec_time = datetime.now().isoformat()
+
+    # 2D raster MASK presence
+    env = gisenv()
```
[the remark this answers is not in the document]
lindakladivova: Originally, I thought the same, but this variable is for some reason always empty..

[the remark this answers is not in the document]
petrasovaa: So in that case, default value of parameter env should be None and then call gisenv only when env was not provided.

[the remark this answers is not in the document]
lindakladivova: I am afraid that it does not work like that.. This env is something different than gisenv. I looked into other occurrences of env in GRASS (e.g. psmap) and it is used as os.environ.copy().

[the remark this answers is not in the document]
petrasovaa: Oh, right!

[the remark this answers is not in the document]
petrasovaa: So it should use `gisenv(env_run)` then.

python/grass/grassdb/history.py
```diff
@@ -9,113 +9,359 @@
...
+
+            # JSON file path
+            json_path = file_path / (file_name_no_ext + ".json")
+
+            # Write JSON data to a new file
+            with open(json_path, "w") as json_outfile:
+                json.dump(lines, json_outfile, indent=2)
+
+            # Remove the old plain text file
+            history_path.unlink()
```
petrasovaa, line 221: How about keeping the old file there? Is there any reason not to? When running in a mapset with an older version, it could still be using the plain text one. In that case you would probably have to change `get_current_mapset_gui_history_path` to primarily return json, not plain text file.

petrasovaa (approved):

python/grass/grassdb/history.py
```diff
@@ -9,113 +9,359 @@
...
+
+            # Extract file path and name without extension
+            file_path = history_path.parent
+            file_name_no_ext = history_path.stem
+
+            # JSON file path
+            json_path = file_path / (file_name_no_ext + ".json")
+
+            # Write JSON data to a new file
+            with open(json_path, "w") as json_outfile:
```
petrasovaa, line 217:
```suggestion
            with open(json_path, encoding="utf-8", mode="w") as json_outfile:
```

petrasovaa: Pylint W1514

python/grass/grassdb/history.py
```diff
@@ -9,113 +9,359 @@
...
+    """Read the content of the history file.
+
+    :param str history_path: path to the history log file
+    :return content_list: list of dictionaries
+    with 'command' and 'command_info' keys
+    """
+    if get_history_file_extension(history_path) == ".json":
+        return _read_from_JSON(history_path)
+    else:
+        return _read_from_plain_text(history_path)
```
petrasovaa, line 133:
```suggestion
    if get_history_file_extension(history_path) == ".json":
        return _read_from_JSON(history_path)
    return _read_from_plain_text(history_path)
```
Pylint W1514

python/grass/grassdb/history.py
```diff
@@ -9,113 +9,359 @@
...
+
+def _add_entry_to_JSON(history_path, entry):
+    """Add entry to JSON history file.
+
+    :param str history_path: path to the history log file
+    :param dict entry: entry consisting of 'command' and 'command_info' keys
+    """
+    try:
+        with open(history_path, encoding="utf-8", mode="r") as fileHistory:
+            existing_data = json.load(fileHistory)
```
petrasovaa, line 277:
```suggestion
        with open(history_path, encoding="utf-8", mode="r") as file_history:
            existing_data = json.load(file_history)
```

python/grass/grassdb/history.py
```diff
@@ -9,113 +9,359 @@
...
     try:
-        file_history = open(history_path, encoding="utf-8", mode="r+")
-        lines = file_history.readlines()
-        file_history.seek(0)
-        file_history.truncate()
-        for number, line in enumerate(lines):
-            if number not in [del_line_number]:
-                file_history.write(line)
+        with open(history_path, encoding="utf-8", mode="w") as fileHistory:
+            json.dump(existing_data, fileHistory, indent=2)
```
petrasovaa, line 284:
```suggestion
        with open(history_path, encoding="utf-8", mode="w") as file_history:
            json.dump(existing_data, file_history, indent=2)
```

PR #3373, opened 2024-01-24
Graphical Modeler: fix UI produced by Python export

landam:
Python scripts generated by Graphical Modeler (`Python editor -> Save As`) contain broken UI definition. Sample model attached: 
[buffer.gxm.zip](https://github.com/OSGeo/grass/files/14039187/buffer.gxm.zip)

```py
# %module
# % description: Script generated by wxGUI Graphical Modeler.
# %end
...
```

Extra space between `#` and `%` causes that `GConsole` skips interface completely: https://github.com/OSGeo/grass/blob/main/gui/wxpython/core/gconsole.py#L641

[Screencast from 2024-01-24 14-37-48.webm](https://github.com/OSGeo/grass/assets/5683186/592a45d2-9da4-4cfa-a118-448a6e7e9a84)

The PR fixes this bug:

[Screencast from 2024-01-24 14-45-45.webm](https://github.com/OSGeo/grass/assets/5683186/70512ee8-2036-4c56-93f2-5e943988fbf4)

### DIFF
--- a/gui/wxpython/gmodeler/model.py
+++ b/gui/wxpython/gmodeler/model.py
@@ -3120,9 +3120,9 @@ class WritePythonFile(WriteScriptFile):
         # UI
         self.fd.write(
             r"""
-# %module
-# % description: {description}
-# %end
+#% module
+#% description: {description}
+#% end
 """.format(
                 description=" ".join(properties["description"].splitlines())
             )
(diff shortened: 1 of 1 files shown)


landam:
> I was about to approve the changes, but I double checked the current python scripts in the repo. Why does the parser not fail with scripts in the repo, but would fail with the exported ones? See PR here when it was added #1287.

you are right, I was not aware of #1287.

> If its an urgent fix, this PR could go through, as it is fixing the consequence of the bug, but maybe fixing the source of the problem would be a better solution.

This PR is replaced by #3375

PR #3402, opened 2024-02-04
wxGUI/history: Move Export History button to history pane

lindakladivova:
This PR relocates the Export History button placed at the bottom of the Console pane to the History pane at the top.

Before:
Screenshot from 2024-02-04 20-05-23

After:
Screenshot from 2024-02-04 20-02-08

### DIFF
--- a/gui/wxpython/gui_core/goutput.py
+++ b/gui/wxpython/gui_core/goutput.py
@@ -194,13 +194,13 @@ class GConsoleWindow(wx.SplitterWindow):
             promptSizer.Add(helpText, proportion=0, flag=wx.EXPAND | wx.LEFT, border=5)
 
             btnSizer = wx.BoxSizer(wx.HORIZONTAL)
+            btnSizer.AddStretchSpacer()
             btnSizer.Add(
                 self.btnOutputSave,
                 proportion=0,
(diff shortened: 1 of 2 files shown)


petrasovaa:
Moving it to history makes sense. What I am unsure about is the placement of the button, right now it looks like it's related to the search (see advanced search in Tools tab), which is problematic, also because it doesn't export the filtered history if I understand it correctly. I would expect it to be at the bottom, below the history, no?

lindakladivova:
> Moving it to history makes sense. What I am unsure about is the placement of the button, right now it looks like it's related to the search (see advanced search in Tools tab), which is problematic, also because it doesn't export the filtered history if I understand it correctly. I would expect it to be at the bottom, below the history, no?

Yes, you are right, that is better.

lindakladivova:
Screenshot from 2024-02-06 13-07-45
I've put it to the bottom on the left, we can change the position also later once we add more buttons.
Could I merge it?

petrasovaa (approved):
Could you move the button to the right (like in Tools tab)? Otherwise this is good to go!

petrasovaa:
Also, it's minor, but wouldn't it be better to have Save and Clear buttons besides each other in the console tab (on the right)?

PR #3421, opened 2024-02-13
wxGUI/history: New button for setting current computational region based on executed commands

lindakladivova:
This PR enhances the history command info panel by adding the option for setting the current computational region settings to the one active upon the history command execution.

The original command info has only information about the computational region settings:
Screenshot from 2024-03-19 14-16-47

The new proposal adds information on whether the current region is equal to the region from the history or not. In the case that the two settings are not equal, the new button "Set as current region" is added:
Screenshot from 2024-03-19 14-05-38

If settings are equal the panel shows just the info message:
Screenshot from 2024-03-19 14-06-00

### DIFF
--- a/gui/wxpython/history/browser.py
+++ b/gui/wxpython/history/browser.py
@@ -135,6 +135,11 @@ class HistoryInfoPanel(SP.ScrolledPanel):
         self.sizer_region_settings_grid.SetCols(2)
         self.sizer_region_settings_grid.SetRows(9)
 
+        self.sizer_region_settings_text = wx.BoxSizer(wx.VERTICAL)
+        self.region_settings_box_sizer.Add(
+            self.sizer_region_settings_text, proportion=0, flag=wx.EXPAND, border=5
+        )
+
         self.region_settings_box_sizer.Add(
             self.sizer_region_settings_grid,
             proportion=1,
@@ -143,12 +148,6 @@ class HistoryInfoPanel(SP.ScrolledPanel):
         )
 
         self.sizer_region_settings_grid.AddGrowableCol(1)
-
-        self.sizer_region_settings_text = wx.BoxSizer(wx.VERTICAL)
-        self.region_settings_box_sizer.Add(
-            self.sizer_region_settings_text, proportion=0, flag=wx.EXPAND, border=5
-        )
-
         self.region_settings_box.Hide()
 
     def _general_info_filter(self, key, value):
@@ -244,7 +243,7 @@ class HistoryInfoPanel(SP.ScrolledPanel):
             )
 
             btnSetRegion = Button(parent=self.region_settings_box, id=wx.ID_ANY)
-            btnSetRegion.SetLabel(_("&Set as current region"))
+            btnSetRegion.SetLabel(_("&Set current region to following values"))
             btnSetRegion.SetToolTip(
                 _("Set current computational region to the region of executed command")
             )
--- a/gui/wxpython/history/browser.py
+++ b/gui/wxpython/history/browser.py
@@ -107,8 +107,8 @@ class HistoryInfoPanel(SP.ScrolledPanel):
     def _initImages(self):
         bmpsize = (16, 16)
         self.icons = {
-            "check": MetaIcon(img="grassdb").GetBitmap(bmpsize),
-            "cross": MetaIcon(img="location").GetBitmap(bmpsize),
+            "check": MetaIcon(img="check").GetBitmap(bmpsize),
+            "cross": MetaIcon(img="check").GetBitmap(bmpsize),
         }
 
     def _createGeneralInfoBox(self):
@@ -145,7 +145,7 @@ class HistoryInfoPanel(SP.ScrolledPanel):
         self.sizer_region_settings_grid.SetCols(2)
         self.sizer_region_settings_grid.SetRows(9)
 
-        self.sizer_region_settings_match = wx.BoxSizer(wx.VERTICAL)
+        self.sizer_region_settings_match = wx.BoxSizer(wx.HORIZONTAL)
         self.region_settings_box_sizer.Add(
             self.sizer_region_settings_match, proportion=0, flag=wx.EXPAND, border=5
         )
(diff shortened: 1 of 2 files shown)


gui/wxpython/history/browser.py
```diff
@@ -226,11 +274,35 @@ def showCommandInfo(self, command_info):
...
+        self.sizer_region_settings_text.Clear(True)
         self._createGeneralInfoBox()
         self._createRegionSettingsBox()
         self._layout()
 
+    def _get_current_region(self):
+        """Get current computational region settings."""
+        current_region = dict(
+            parse_key_val(gs.read_command("g.region", flags="g"), val_type=float)
+        )
```
petrasovaa, line 287: Why not `gs.region()`?

gui/wxpython/history/browser.py
```diff
@@ -226,11 +274,35 @@ def showCommandInfo(self, command_info):
...
+        for key, value in current_region.items():
+            current_region[key] = int(value) if value.is_integer() else value
```
petrasovaa, line 289: Not sure I understand the need for this?

gui/wxpython/history/browser.py
```diff
@@ -119,20 +124,32 @@ def _createGeneralInfoBox(self):
     def _createRegionSettingsBox(self):
         """Create a static box for displaying region settings of the command"""
         self.region_settings_box = StaticBox(
-            parent=self, id=wx.ID_ANY, label=_("Region settings")
+            parent=self,
+            id=wx.ID_ANY,
+            label=_("Computational region upon command execution"),
```
petrasovaa, line 129:
```suggestion
            label=_("Computational region during command execution"),
```

gui/wxpython/history/browser.py
lindakladivova: I was wondering whether there is a better way.. Thanks Anna. I also simplified the similar code in history.py.

petrasovaa (commented):
Can we move the button towards the top of the region box, otherwise it may be hidden...

lindakladivova:
> Can we move the button towards the top of the region box, otherwise it may be hidden...

Then, the description of the button should be probably changed.. I have changed it to the following:
Screenshot from 2024-04-19 14-03-18

petrasovaa:
There is a lot of text, but I am not sure how to improve it, @veroandreo any idea?

lindakladivova:
I am wondering if the arrangement in the bottom panel is sufficiently flexible. I also plan enhance the panel by the error info text (now we have only "failed" written and there is no clue what happened). So maybe we could divide the bottom panel to notebook with tabs: General info, Error info, Region settings and place particular boxes into these tabs?..

veroandreo:
> There is a lot of text, but I am not sure how to improve it, @veroandreo any idea?

Well, users need to click over a command to get the region info, so *Region settings* as it was seems enough to me. What about using also green checks when region setting on command matches the current and red cross when it doesn't? Then with a text box appearing when hovering inform what it means. The amount of text is reduced in that way and the button could say *update current region to* or simply *update current region*.

lindakladivova:
> > There is a lot of text, but I am not sure how to improve it, @veroandreo any idea?
> 
> Well, users need to click over a command to get the region info, so _Region settings_ as it was seems enough to me. What about using also green checks when region setting on command matches the current and red cross when it doesn't? Then with a text box appearing when hovering inform what it means. The amount of text is reduced in that way and the button could say _update current region to_ or simply _update current region_.


I like the idea of "Update current region" rather than "Set to .." but I do not understand much of how you think checks should be arranged. Where exactly would you put them, Vero? At the top? I would probably place the check at the top to the left and the button on the right-top side. Do you mean it like this?

veroandreo:
> > > There is a lot of text, but I am not sure how to improve it, @veroandreo any idea?
> > 
> > 
> > Well, users need to click over a command to get the region info, so _Region settings_ as it was seems enough to me. What about using also green checks when region setting on command matches the current and red cross when it doesn't? Then with a text box appearing when hovering inform what it means. The amount of text is reduced in that way and the button could say _update current region to_ or simply _update current region_.
> 
> I like the idea of "Update current region" rather than "Set to .." but I do not understand much of how you think a check should be arranged. Where exactly would you put it, Vero? At the top? I would probably place the check at the top to the left and the button on the right-top side. Do you mean it like this?

I imagined it could go either left or right from the box title, i.e., [check|cross] Region settings or Region settings [check|cross]... Does that make sense?

Though, just below the box title, with something like `Region match [check|cross]` and the button next to it if it does not match could also do, what do you think?

lindakladivova:
Waits for merging https://github.com/OSGeo/grass/pull/3679 to use proper icons

lindakladivova:
> > > > There is a lot of text, but I am not sure how to improve it, @veroandreo any idea?
> > > 
> > > 
> > > Well, users need to click over a command to get the region info, so _Region settings_ as it was seems enough to me. What about using also green checks when region setting on command matches the current and red cross when it doesn't? Then with a text box appearing when hovering inform what it means. The amount of text is reduced in that way and the button could say _update current region to_ or simply _update current region_.
> > 
> > 
> > I like the idea of "Update current region" rather than "Set to .." but I do not understand much of how you think a check should be arranged. Where exactly would you put it, Vero? At the top? I would probably place the check at the top to the left and the button on the right-top side. Do you mean it like this?
> 
> I imagined it could go either left or right from the box title, i.e., [check|cross] Region settings or Region settings [check|cross]... Does that make sense?
> 
> Though, just below the box title, with something like `Region match [check|cross]` and the button next to it if it does not match could also do, what do you think?

I implemented the variant with Region match [check|cross] and the button "Update current region" next to it. It is concise, fitting into one row. Thank @veroandreo for the suggestion with the icons, I think it looks really good. :-)

Screenshot from 2024-05-09 14-14-04

I plan to use icons for Check and Cross from #3679 (now there is only the Check icon among current icons)

lindakladivova:
> > > > > There is a lot of text, but I am not sure how to improve it, @veroandreo any idea?
> > > > 
> > > > 
> > > > Well, users need to click over a command to get the region info, so _Region settings_ as it was seems enough to me. What about using also green checks when region setting on command matches the current and red cross when it doesn't? Then with a text box appearing when hovering inform what it means. The amount of text is reduced in that way and the button could say _update current region to_ or simply _update current region_.
> > > 
> > > 
> > > I like the idea of "Update current region" rather than "Set to .." but I do not understand much of how you think a check should be arranged. Where exactly would you put it, Vero? At the top? I would probably place the check at the top to the left and the button on the right-top side. Do you mean it like this?
> > 
> > 
> > I imagined it could go either left or right from the box title, i.e., [check|cross] Region settings or Region settings [check|cross]... Does that make sense?
> > Though, just below the box title, with something like `Region match [check|cross]` and the button next to it if it does not match could also do, what do you think?
> 
> I implemented the variant with Region match [check|cross] and the button "Update current region" next to it. It is concise, fitting into one row. Thank @veroandreo for the suggestion with the icons, I think it looks really good. :-)
> 
> Screenshot from 2024-05-09 14-14-04
> 
> I plan to use icons for Check and Cross from #3679 (now there is only the Check icon among current icons)


After using proper icons and minor repositioning..
When regions do not match:
Screenshot from 2024-05-10 10-42-12
When regions match:
Screenshot from 2024-05-10 10-42-32

petrasovaa (approved):

PR #3416, opened 2024-02-11
v.clip: do not fail when clip map has no table connected

landam:
In GRASS 8.4 `v.clip` fails when `clip` map has no table connected. Steps to reproduce in NC location:

```
v.extract input=boundary_county cat=1 output=aoi
v.db.connect aoi -d
v.clip input=geology clip=aoi output=geology_clip --o
```

`v.clip` currently fails with:

```
ERROR: Database connection for map <aoi> is not defined in DB file
Traceback (most recent call last):
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/scripts/v.dissolve", line 686, in <module>
    main()
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/scripts/v.dissolve", line 566, in main
    aggregate_columns_exist_or_fatal(input_vector, layer, columns_to_aggregate)
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/scripts/v.dissolve", line 265, in aggregate_columns_exist_or_fatal
    column_names = gs.vector_columns(vector, layer).keys()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/etc/python/grass/script/vector.py", line 128, in vector_columns
    s = read_command(
        ^^^^^^^^^^^^^
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/etc/python/grass/script/core.py", line 532, in read_command
    return handle_errors(returncode, stdout, args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/etc/python/grass/script/core.py", line 344, in handle_errors
    raise CalledModuleError(module=module, code=code, returncode=returncode)
grass.exceptions.CalledModuleError: Module run `v.info --q -c map=aoi layer=1` ended with an error.
The subprocess ended with a non-zero return code: 1. See errors above the traceback or in the error output.
Traceback (most recent call last):
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/scripts/v.clip", line 226, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/scripts/v.clip", line 150, in main
[...]

PR #3423, opened 2024-02-14
v.clip: add test for no table

landam:
This PR extends `v.clip` module by testing inputs with no table attached. It runs on given combinations:

* input (no table) + clip
* input + clip (no table)
* input (no table) + clip (no table)

See related #3416


wenzeslaus (approved):
I'm happy to see this. Maybe a separate class would lead to better messages in case of failure, but this will do the trick.

neteler:
(@landam pls always set a milestone when merging. To re-add them in future is a pain and having them set is great to understand in which version a PR became part of a release. Thx)

PR #3432, opened 2024-02-16
grass.script: Create new location without a session

wenzeslaus:
The function create_location now works without a full session. Internally, it sets up a runtime environment to execute tools (modules) without the connection to a location (project) which is sufficient for executing g.proj.

The new test covers the new functionality, but to actually test it, the test needs to run by itself because grass.script.setup.init (still) creates a global session.

### DIFF
--- a/python/grass/script/core.py
+++ b/python/grass/script/core.py
@@ -1733,9 +1733,20 @@ def create_location(
     if datum_trans:
         kwargs["datum_trans"] = datum_trans
 
+    # Lazy-importing to avoid circular dependencies.
+    # pylint: disable=import-outside-toplevel
+    if os.environ.get("GISBASE"):
+        env = os.environ
+    else:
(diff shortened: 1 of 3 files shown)


wenzeslaus:
@pesekon2 and others, I originally thought this will have to wait for 8.5, but I did some prep for this when doing prep work for grass.jupyter, so the overall change is not so large and I think it may go to 8.4 just fine.

petrasovaa (dismissed):
This would be useful for 8.4

marisn (approved):

pesekon2:
Works perfect for me. Thanks.

pesekon2 (approved):

wenzeslaus:
Great! I have some updates for grass.script.setup.init taking advantage of the more flexible runtime setup, so we will see that part of this PR (not the create part) in action in this repo soon.

Issue #3420, opened 2024-02-13
[Bug] v.dissolve looses connection with database tables when running


geografin:
**Describe the bug**
When processing the vector map one of backend commands (seems to be v.reclass) change the database table name to map_somenumbersandchars (example map_ef0691c223f2_631@PERMANENT), after that following command (v.extract) can't find tables for the map.

**To Reproduce**

1. Take a vector with varchar column with duplicated values and several columns with double precision or integer values.
2. Start dissolving this vector on varchar column  v.dissolve (version grass 8.4) (aggregate_backend=sql)
3. See error ERROR: There is no table connected to <map>

**Expected behavior**
A dissolved map should be written.

**System description:**

- Operating System: Docker with GRASS GIS (debian)
- GRASS GIS version 8.4.0

ecodiv:
@geografin I can't reproduce the problem on GRASS 8.4.0 on Ubuntu 22.04, using aggregate_backend=sql.

wenzeslaus:
Thanks for reporting this. Can you please provide the code to reproduce ideally using the NC SPM sample dataset or other public data? The code can be just copied commands from the GUI if you are using it (there is _Copy_ button in the tool dialog).

geografin:
`v.dissolve input=over_geo column=subid aggregate_columns=a_length,a_source_elev,a_outlet_elev aggregate_methods=sum,max,min result_columns=LENGTH,SOURCE,OUTLET output=over_geo_diss --overwrite`
This is the command that ends with error. The map taken is a simple rivers layers generated with r.stream.order and overlayed with subwatersheds vector with `v.overlay`.

geografin:
I don't use GUI. I use bash linux with GRASS docker image (from main branch, the last one)

wenzeslaus:
Sorry @geografin what I meant was a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example). @ecodiv was not able to reproduce it, so I'm looking for a way to get the same data and situation you have. Bash is definitively better starting point than GUI, so that's great. Can you try to run relevant commands with the [NC SPM sample dataset](https://grass.osgeo.org/download/data/) and share these commands?

geografin:
Well. So take NorthCarolina dataset (full) and reproduce (I copy some maps form the dataset, that are identical to mine by type):
```bash
g.copy vect=roadsmajor,streams_init --overwrite
g.copy vect=zipcodes_wake,geodata_subbasins --overwrite
[...]

PR #3433, opened 2024-02-17
v.dissolve: Allow all types of features to be dissolved

ecodiv:
This seems to solve issue #3420


wenzeslaus:
Would you be able to provide a test for this? The current tests pass so that's great. A test for lines or points would be nice addition. v.dissolve has grass.gunittest-based tests in `testsuite` directory and pytest-based tests in `tests`. You can pick whatever makes more sense to you.

ecodiv:
> Would you be able to provide a test for this? The current tests pass so that's great. A test for lines or points would be nice addition. v.dissolve has grass.gunittest-based tests in `testsuite` directory and pytest-based tests in `tests`. You can pick whatever makes more sense to you.

I haven't figured out how those tests work. Something on my todo list.

wenzeslaus (dismissed):
`v.extract -d` operates on boundaries, but the dissolve for the other types happens in v.reclass if I understand that correctly, so this looks good. (For the record, the areas limitation was there at least since the translation from Bash to Python for v7.)

There is a lot of tests for the current functionality including some new ones through v.clip, so the tests would be nice, but perhaps not crucial.

There is, however, several places in the code and doc which mention area. Can you please update those?

ecodiv:
> There is, however, several places in the code and doc which mention area. Can you please update those?

I can give it a try. Most descriptions you read are about dissolving boundaries between adjacent polygons. That is the more common and also easily visualized user case. But in this case, adjacent or overlapping features can be dissolved irrespective of their type (at least, I tried with lines and polygons, I haven't tried the other types). That makes it a bit harder to clearly describe.

But perhaps first the observation that in the manual, it reads that:

> Multiple areas with the same category or the same attribute value which are not adjacent are merged together into one entity which consists of multiple areas, i.e., a multipolygon.

This is different behavior from the general description, which reads:

> Dissolves boundaries between adjacent areas sharing a common category number or attribute. 

I would have thought that the latter is the more common approach to dissolving features? 

I actually like how QGIS has implemented this. Its dissolve function offers the option to "keep disjoint features as separate". Selecting that option means that any features in the layer that do not share a common boundary with another feature will be kept as separate features.  I am not sure how ArcGIS handles things.

wenzeslaus:
> > There is, however, several places in the code and doc which mention area. Can you please update those?
> 
> I can give it a try. Most descriptions you read are about dissolving boundaries between adjacent polygons. That is the more common and also easily visualized user case.

For merging this PR, it would be enough just to fix the places where is says it does areas (only). If it says "areas and other types" that should be enough to clarify it is not areas only. 

> But perhaps first the observation that in the manual, it reads that:
> 
> > Multiple areas with the same category or the same attribute value which are not adjacent are merged together into one entity which consists of multiple areas, i.e., a multipolygon.
> 
> This is the actual behavior, which differs from what the general description suggests, which reads:
> 
> > Dissolves boundaries between adjacent areas sharing a common category number or attribute.
> 
> I would have thought that the latter is the more common approach to dissolving features?

Based on what I found, just a conjecture based on what are the defaults in different SW, the more common behavior is the same what the current v.dissolve does. It may be different from the expectations created by the more geometrically focused examples in GIS 101 classes, but that's the reality perhaps based on what is easier to implement. The v.dissolve implementation relies on two-step (three with aggregation) process with the geometry and attributes being resolved separately and at this point does not have an easy switch for changing to the other behavior. The update would have to happen in v.extract and practically it is likely just a question of how much time/money is someone willing to spend adding that feature.

ecodiv:
> > > There is, however, several places in the code and doc which mention area. Can you please update those?
> > 
> > 
> > I can give it a try. Most descriptions you read are about dissolving boundaries between adjacent polygons. That is the more common and also easily visualized user case.
> 
> For merging this PR, it would be enough just to fix the places where is says it does areas (only). If it says "areas and other types" that should be enough to clarify it is not areas only.
> 
> > But perhaps first the observation that in the manual, it reads that:
> > > Multiple areas with the same category or the same attribute value which are not adjacent are merged together into one entity which consists of multiple areas, i.e., a multipolygon.
> > 
> > 
> > This is the actual behavior, which differs from what the general description suggests, which reads:
> > > Dissolves boundaries between adjacent areas sharing a common category number or attribute.
> > 
> > 
> > I would have thought that the latter is the more common approach to dissolving features?
> 
> Based on what I found, just a conjecture based on what are the defaults in different SW, the more common behavior is the same what the current v.dissolve does. It may be different from the expectations created by the more geometrically focused examples in GIS 101 classes, but that's the reality perhaps based on what is easier to implement. The v.dissolve implementation relies on two-step (three with aggregation) process with the geometry and attributes being resolved separately and at this point does not have an easy switch for changing to the other behavior. The update would have to happen in v.extract and practically it is likely just a question of how much time/money is someone willing to spend adding that feature.

Fair enough. And it is possible to separate geographically separated features in a separate step using the code https://github.com/OSGeo/grass-addons/pull/1021, of that is required.

As for the text, I'll have a look.

wenzeslaus (changes requested):
Are you sure about the image the result has A and B values in key column, there should probably be cat column again with 1 and 2.

ecodiv:
> Are you sure about the image the result has A and B values in key column, there should probably be cat column again with 1 and 2.

You are right, I misunderstood the earlier text. Removed that part of the text and the figure.

scripts/v.dissolve/v.dissolve.html
```diff
@@ -1,126 +1,176 @@
...
 </center>
 
-<h3>Merging behavior</h3>
+<p>
+
+Instead of dissolving features based on the category values, the user 
+can define an integer or string column using the <b>column</b> 
+parameter. In that case, features that share the same value in that 
+column are dissolved. Note, the newly created layer does not retain the 
+cat values from the input layer, only the values of the new key column.
```
wenzeslaus, line 22: It "does retain values of new key column"? That does not make sense to me. Is this related to your comment about the figure?

scripts/v.dissolve/v.dissolve.html
```diff
@@ -1,126 +1,176 @@
...
 
-Multiple areas with the same category or the same attribute value
-which are not adjacent are merged together into one entity
-which consists of multiple areas, i.e., a multipolygon.
+<p>
+Note that multiple areas with the same category or the same attribute 
+value that are not adjacent are merged into one entity, which consists 
+of multiple features. I.e., this results in multipart features.
```
wenzeslaus, line 27:
```suggestion
Note that multiple areas with the same category or the same attribute 
value that are not adjacent are merged into one entity, which consists 
of multiple features, i.e., a multipart feature.
```

Or simply start a new sentence and restate the same in other words (Or say "In other words, ...").

scripts/v.dissolve/v.dissolve.html
```diff
@@ -1,126 +1,176 @@
...
+future.
 
 <p>
-If only <b>aggregate_columns</b> is provided, methods default to
-<em>n</em>, <em>min</em>, <em>max</em>, <em>mean</em>, and <em>sum</em>.
-If the <em>univar</em> backend is specified, all the available methods
-for the <em>univar</em> backend are used.
+If the <b>result_columns</b> is not provided, each method is applied to 
+each column specified by <b>aggregate_columns</b>. This results in 
+columns for all combinations. These result columns have auto-generated
```
wenzeslaus, line 107:
```suggestion
each column specified by <b>aggregate_columns</b>. This results in a column for each of the combinations. These result columns have auto-generated 
```

scripts/v.dissolve/v.dissolve.html
```diff
@@ -1,126 +1,176 @@
...
+names based on the aggregate column and method. For example, setting 
+the following parameters:
+
+<div class="code"><pre>
+aggregate_columns=A,B
+aggregate_methods=sum,n
+</pre></div>
+
+<p>
+Results in the following columns: A_sum, A_n, B_sum, B_n. Also see
```
wenzeslaus, line 117: Is this common to start here with a capital letter when it is continuation of a sentence from above? You could also split the "For example..." part and the result part into two sentences to avoid this completely.

Same issue couple lines above.

scripts/v.dissolve/v.dissolve.html
```diff
@@ -1,126 +1,176 @@
...
+the following parameters:
+
+<div class="code"><pre>
+aggregate_columns=A,B
+aggregate_methods=sum,n
+</pre></div>
+
+<p>
+Results in the following columns: A_sum, A_n, B_sum, B_n. Also see 
+the Example section.
```
wenzeslaus, line 118: Or Examples section? Same below.

wenzeslaus (changes requested):
I don't know if it was the formatting changes or just GitHub diff rendering, but I had to read everything to figure out what was changed. That took some time to get to it. It looks great overall. Hopefully next round can be quick.

ecodiv:
> I don't know if it was the formatting changes or just GitHub diff rendering, but I had to read everything to figure out what was changed. That took some time to get to it. It looks great overall. Hopefully next round can be quick.

Sorry about that. I have never figured out when GitHub shows the diff for the line in which the change took place, or whole blocks of text, or what I should do to make sure it only shows the line with changes.

scripts/v.dissolve/v.dissolve.html
```diff
@@ -1,126 +1,176 @@
...
-which <em>v.db.univar</em> uses by default,
-i.e., <em>n</em>, <em>min</em>, <em>max</em>, <em>range</em>,
-<em>mean</em>, <em>mean_abs</em>, <em>variance</em>, <em>stddev</em>,
-<em>coef_var</em>, and <em>sum</em>.
-When the <em>sql</em> backend is used, the methods in turn depends on the SQL
-database backend used for the attribute table of the input vector.
-For SQLite, it is at least the following
-<a href="https://www.sqlite.org/lang_aggfunc.html">build-in aggregate functions</a>:
-<em>count</em>, <em>min</em>, <em>max</em>,
+The default behavior is typically intended for interactive use and
```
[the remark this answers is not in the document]
ecodiv: I am not sure why I used the word typically here.

scripts/v.dissolve/v.dissolve.html
```diff
@@ -1,126 +1,176 @@
...
+
+<h4>aggregation using univar</h4>
+
+When <em>univar</em> is used, the methods available are the ones which 
+<em>v.db.univar</em> uses by default, i.e., <em>n</em>, <em>min</em>, 
+<em>max</em>, <em>range</em>, <em>mean</em>, <em>mean_abs</em>, 
+<em>variance</em>, <em>stddev</em>, <em>coef_var</em>, and 
+<em>sum</em>.
+
+<h4>Aggregation using univar</h4>
```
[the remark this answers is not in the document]
ecodiv: Corrected

scripts/v.dissolve/v.dissolve.html
```diff
@@ -1,126 +1,176 @@
...
+functions applied, e.g.,
 
 <p>
-Aggregate methods are specified by name in <b>aggregate_methods</b>
-or using SQL syntax in <b>aggregate_columns</b>.
-If <b>result_columns</b> is provided including type information
-and the <em>sql</em> backend is used,
-<b>aggregate_columns</b> can contain SQL syntax specifying both columns
-and the functions applied, e.g.,
 <code>aggregate_columns="sum(cows) / sum(animals)"</code>.
```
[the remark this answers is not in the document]
ecodiv: Should there be a <p> above  a code block?

scripts/v.dissolve/v.dissolve.html
ecodiv: I think. I was a bit confused about that part in the original text.  The important part, I guess, from the original text was that the original cat values are not retained. Deleting the second part of the sentence.

scripts/v.dissolve/v.dissolve.html
```diff
@@ -1,126 +1,177 @@
...
-<a href="https://www.sqlite.org/lang_aggfunc.html">build-in aggregate functions</a>:
-<em>count</em>, <em>min</em>, <em>max</em>,
+The default behavior is intended for interactive use and 
+testing. For scripting and other automated usage, explicitly specifying 
+the backend with the <b>aggregate_backend</b> parameter is strongly 
+recommended. When choosing, note that the <em>sql</em> aggregate 
+backend, regardless of the underlying database, will typically perform 
+significantly better than the <em>univar</em> backend.
+
+<h4>aggregation using univar</h4>
```
wenzeslaus, line 48:
```suggestion
<h4>Aggregation using univar backend</h4>
```

scripts/v.dissolve/v.dissolve.html
```diff
@@ -1,126 +1,177 @@
...
+
+When <em>univar</em> is used, the methods available are the ones which 
+<em>v.db.univar</em> uses by default, i.e., <em>n</em>, <em>min</em>, 
+<em>max</em>, <em>range</em>, <em>mean</em>, <em>mean_abs</em>, 
+<em>variance</em>, <em>stddev</em>, <em>coef_var</em>, and 
+<em>sum</em>.
+
+<h4>Aggregation using sql</h4>
```
wenzeslaus, line 56:
```suggestion
<h4>Aggregation using sql backend</h4>
```

wenzeslaus (approved):

wenzeslaus (approved):

wenzeslaus (changes requested):
Actually, the size of the images changed slightly, but nothing is changed otherwise. That needs to be reverted. (I will do it.)

wenzeslaus (approved):

wenzeslaus:
> ...I have never figured out when GitHub shows the diff for the line in which the change took place, or whole blocks of text, or what I should do to make sure it only shows the line with changes.

If there are changes in wrapping of paragraphs, than it shows the whole blocks as edited. Both Git and GitHub can show just changed words and they try to show changes around unchanged text, but if there is a change on every line, than everything looks new and the word diff is no longer possible. The versioning is line-oriented and if there is a change in a line it will show it (and it will not show it otherwise).

You can limit some of these issue by really editing just the necessary stuff, e.g., by picking only what you intended to change using `git add -p` (which is a really great tool in general). This may help with things like slightly modified images or automatically changed lines by your editor.

If you do changes everywhere, then, of course, you create a huge diff and that's fine if making changes everywhere is what you are really doing.

PR #3452, opened 2024-02-25
fixed scanf error to recognize EOF as a possible return value

naidneelttil:


raster/r.to.rast3elev/main.c
```diff
@@ -408,7 +408,7 @@ int main(int argc, char *argv[])
 
     /*Set the upper value */
     if (param.upper->answer) {
-        if (sscanf(param.upper->answer, "%lf", &db.upper))
+        if (sscanf(param.upper->answer, "%lf", &db.upper) == 1)
```
HuidaeCho, line 411: `G_parser()` does all sanity checks to ensure it returns a non-NULL only if any value is provided to an option. An empty value will return a NULL, not `"\0"`, so it shoud be enough to check for `NULL`.

https://github.com/OSGeo/grass/blob/87d2d42d3b26c01d6de08ae8845618c9918a20cc/lib/gis/parser.c#L1180-L1182

raster/r.to.rast3elev/main.c
```diff
@@ -419,7 +419,7 @@ int main(int argc, char *argv[])
 
     /*Set the lower value */
     if (param.lower->answer) {
-        if (sscanf(param.lower->answer, "%lf", &db.lower))
+        if (sscanf(param.lower->answer, "%lf", &db.lower) == 1)
```
HuidaeCho, line 422: See above.

raster/r.watershed/ram/init_vars.c
```diff
@@ -83,7 +83,7 @@ int init_vars(int argc, char *argv[])
         else if (sscanf(argv[r], "blocking=%s", ob_name) == 1)
             ob_flag++;
         else if (sscanf(argv[r], "disturbed_land=%s", ril_name) == 1) {
-            if (sscanf(ril_name, "%lf", &ril_value) == 0) {
+            if (sscanf(ril_name, "%lf", &ril_value) == 0 || sscanf(ril_name, "%lf", &ril_value) == EOF ){
```
HuidaeCho, line 86: `ril_name` was successfully read from the previous `sscanf()`, so it shouldn't be `"\0"` and there is no need for this `EOF` check.

raster/r.watershed/seg/init_vars.c
```diff
@@ -92,7 +92,7 @@ int init_vars(int argc, char *argv[])
         else if (sscanf(argv[r], "memory=%lf", &segs_mb) == 1)
             ;
         else if (sscanf(argv[r], "disturbed_land=%s", ril_name) == 1) {
-            if (sscanf(ril_name, "%lf", &ril_value) == 0) {
+            if (sscanf(ril_name, "%lf", &ril_value) == 0 || sscanf(ril_name, "%lf", &ril_value) == EOF ) {
```
HuidaeCho, line 95: See above.

raster3d/r3.out.vtk/writeVTKData.c
```diff
@@ -339,7 +339,7 @@ void write_vtk_data(FILE *fp, RASTER3D_Map *map, RASTER3D_Region region,
     depths = region.depths;
 
     /*the nullvalue */
-    if (!sscanf(param.null_val->answer, "%lf", &nullvalue)) {
+    if (!sscanf(param.null_val->answer, "%lf", &nullvalue) || sscanf(param.null_val->answer, "%lf", &nullvalue) == EOF ) {
```
HuidaeCho, line 342: `param.null_val` has a default value, so it cannot be `"\0"`.

HuidaeCho (commented):
Thanks for your PR. Most (all?) of these checks are for `G_parser()` options. `G_parser()` does all necessary sanity checks and returns non-NULL values only if there are any valid values. Even an empty string with an option name (e.g., `map=""`) returns a NULL in `answer`, so `EOF` from `sscanf()` is not really a concern when we already check for the nullity of `answer`.

general/g.region/main.c
```diff
@@ -834,7 +834,7 @@ int main(int argc, char *argv[])
     /* grow by number of cells */
     if ((value = parm.grow->answer)) {
         update_file = true;
-        if (sscanf(value, "%i", &pix)) {
+        if (sscanf(value, "%i", &pix) == 1 ) {
```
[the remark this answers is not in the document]
wenzeslaus: EOF "expands into a negative integral constant expression (generally, -1)" ([source](https://cplusplus.com/reference/cstdio/EOF/)), so `>=1` or, here, `==1` is required to get true for the "good" cases.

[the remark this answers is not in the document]
HuidaeCho: What I mean is, when `parm.grow->answer` is not `NULL` (line 835), `*parm.grow->answer` is not '\0'. In this case, `sscanf()` cannot return `EOF`. [A quick search](https://www.ibm.com/docs/en/i/7.1?topic=functions-sscanf-read-data):
> The return value is EOF when **the end of the string** is encountered before anything is converted.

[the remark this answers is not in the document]
echoix: [Here, it stays more general](https://en.cppreference.com/w/c/io/fscanf), noting that 

> 1-3) Number of receiving arguments successfully assigned (which may be zero in case a matching failure occurred before the first receiving argument was assigned), _or [EOF](https://en.cppreference.com/w/c/io) if input failure occurs before the first receiving argument was assigned._

What I understand from this is that there is some chance that another failure gets us an EOF, and it isn't only a question of strings or not. So handling correctly the return values _just because we can_ should be the correct thing to do.

raster/r.out.vtk/writeascii.c
```diff
@@ -165,7 +165,7 @@ void write_vtk_structured_coordinates(int fd, FILE *fp, char *varname UNUSED,
     G_debug(3, _("write_vtk_structured_coordinates: Writing Coordinates"));
 
     /*the nullvalue */
-    if (!sscanf(null_value, "%lf", &nullvalue)) {
+    if (!sscanf(null_value, "%lf", &nullvalue) || sscanf(null_value, "%lf", &nullvalue) == EOF ) {
```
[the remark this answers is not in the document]
echoix: In any of the CodeQL articles for these detections, like https://github.com/OSGeo/grass/security/code-scanning/529, they show that it is possible to assign to a variable and then use it more than once. Is that variable allocation a problem?

```c
 if (scanf("%d %d", &i, &j) == 2) { 
      // GOOD: the result is checked against 2
  }

  // GOOD: the result is compared directly
  int r = scanf("%d %d", &i, &j);
  if (r < 2) {
    return;
  }
  if (r == 1) { 
    j = 0;
  }
```

naidneelttil (commented):
in this PR, I removed the explicit EOF checks but kept the checks for a scanf return value, since this is generally good practice and one less thing to worry about  if ever refactoring.

raster/r.out.vtk/writeascii.c
```diff
@@ -219,7 +219,7 @@ void write_vtk_polygonal_coordinates(int fd, FILE *fp, char *varname UNUSED,
             _("write_vtk_polygonal_coordinates: Writing VTK Polygonal data"));
 
     /*the nullvalue */
-    if (!sscanf(null_value, "%lf", &nullvalue)) {
+    if (!sscanf(null_value, "%lf", &nullvalue) || sscanf(null_value, "%lf", &nullvalue) = EOF ) {
```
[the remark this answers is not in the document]
naidneelttil: thanks, removed

general/g.region/main.c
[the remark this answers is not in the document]
naidneelttil: since this seems to be good practice, Ill keep it and changes similar to it in the PR.

raster/r.out.vtk/writeascii.c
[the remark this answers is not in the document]
naidneelttil: Ill remove the explicit EOF check

raster/r.to.rast3elev/main.c
naidneelttil: kept in as it seems good practice as stated in an earlier comment

raster/r.watershed/ram/init_vars.c
naidneelttil: i removed this change.

raster/r.watershed/seg/init_vars.c
naidneelttil: i removed this change.

raster3d/r3.out.vtk/writeVTKData.c
naidneelttil: i removed this change.

raster/r.out.vtk/writeascii.c
```diff
@@ -329,7 +329,7 @@ void write_vtk_data(int fd, FILE *fp, char *varname, struct Cell_head region,
     G_debug(3, _("write_vtk_data: Writing VTK-Data"));
 
     /*the nullvalue */
-    if (!sscanf(null_value, "%lf", &nullvalue)) {
+    if (!sscanf(null_value, "%lf", &nullvalue) || sscanf(null_value, "%lf", &nullvalue) == EOF ) {
```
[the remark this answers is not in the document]
naidneelttil: I removed this change, thanks.

echoix:
@HuidaeCho with this PR now limited to only three changes with checks like ` == 1`, is this PR ready to go?

nilason:
Not only does the parser either return `NULL` or a non-empty string, the string is also guaranteed to contain a valid integer as it already been tested with https://github.com/OSGeo/grass/blob/8a3c6b7c7627036bf52681cdcb9c9833c398d6ac/lib/gis/parser.c#L1399 
There is no way a compiler/static analyser could know this.

I see two possible "solutions" to this:
1. Change to early exit strategy: eg. `if (sscanf(value, "%i", &pix) != 1)  die(parm.grow);`
2. Silence the warning with void: eg. `if ((void)sscanf(value, "%i", &pix)) {`

`if (sscanf(value, "%i", &pix) == 1) {` without response if the `if` condition is false, will only lead to new problems.

PR #3457, opened 2024-02-26
grass Python package: hide windows triggered by subprocess.Popen on MS Windows

landam:
When running Python script using GRASS Python API in third party environments (like QGIS) on MS Windows for every command triggered by `script.core.Popen()` a new window pops up. This can be very frustrating when running multiple GRASS modules in sequence. See short example below:

[Screencast_02_26_2024_06:28:46_AM.webm](https://github.com/OSGeo/grass/assets/5683186/a45d095e-74db-40cf-a31a-6264a2c3f4b4)

# How to tests

Run script below in QGIS:

```py
import os

import grass.script.setup as gsetup
from grass.pygrass.modules import Module

gisdb = f"{os.environ['HOMEPATH']}\\Documents\\grassdata"
gsetup.init(gisdb, "world_latlong_wgs84", 'PERMANENT')

Module("g.region", n=90, s=0, w=0, e=100, res=0.02, flags="p")
Module("r.random.surface", output="x", overwrite=True)
print("done")
```

# New behaviour

The PR changes this behaviour. No pop-up windows are show.

[Screencast_02_26_2024_06:31:13_AM.webm](https://github.com/OSGeo/grass/assets/5683186/3406f3c7-b046-42ad-901b-14c03b8c136c)

Windows can be show by setting a new environmental variable globally:

```py
from grass.script import env
env.subprocess.show_window = True
```

Or just for selected commands (see below):

```py
from grass.script import Env
with Env() as env:
    env.subprocess.show_window = True
    Module("r.random.surface", output="x", overwrite=True)
```

[Screencast_02_26_2024_06:39:18_AM.webm](https://github.com/OSGeo/grass/assets/5683186/b53ff1dd-3cac-4a24-932d-5fb1ab0f812c)

# Side effect

New concept of `Env` class allows to add new variables in the future, eg.

```py
env.output.overwrite = True
```

or similar.

### DIFF
--- a/lib/init/variables.html
+++ b/lib/init/variables.html
@@ -262,16 +262,6 @@ PERMANENT
   of G_fatal_error() will end in a segmentation violation. GDB can be used
   to trace the source of the error.</dd>
 
-  <dt>GRASS_SUBPROCESS_CREATE_NEW_CONSOLE</dt>
-  <dd>[Python grass package]<br>
-  This variable affects GRASS commands invoked in Python using
-  grass.script.core.Popen class. If this environment variable is defined
-  (and set to any value) a new console window is opened for each
-  command. This behavior was defaulted to GRASS GIS 8.3. Since version
-  8.4, new console windows are not automatically opened. Relevant only
-  for MS Windows operating system.
-  </dd>
-
   <dt>GRASS_PYTHON</dt>
   <dd>[wxGUI, Python Ctypes]<br>
     set to override Python executable.<br>
--- a/lib/init/variables.html
+++ b/lib/init/variables.html
@@ -262,6 +262,16 @@ PERMANENT
   of G_fatal_error() will end in a segmentation violation. GDB can be used
   to trace the source of the error.</dd>
 
+  <dt>GRASS_SUBPROCESS_CREATE_NEW_CONSOLE</dt>
+  <dd>[Python grass package]<br>
+  This variable affects GRASS commands invoked in Python using
+  grass.script.core.Popen class. If this environment variable is defined
+  (and set to any value) a new console window is opened for each
+  command. This behavior was defaulted to GRASS GIS 8.3. Since version
+  8.4, new console windows are not automatically opened. Relevant only
+  for MS Windows operating system.
+  </dd>
+
   <dt>GRASS_PYTHON</dt>
   <dd>[wxGUI, Python Ctypes]<br>
     set to override Python executable.<br>
--- a/python/grass/script/core.py
+++ b/python/grass/script/core.py
@@ -128,7 +128,8 @@ class Popen(subprocess.Popen):
                 kwargs["shell"] = True
                 args = [self._escape_for_shell(arg) for arg in args]
 
-            if env.subprocess.show_window:
+            if env.subprocess.show_window is False:
+                # do not show new window on MS Windows
                 si = subprocess.STARTUPINFO()
                 si.dwFlags = (
                     subprocess.CREATE_NEW_CONSOLE | subprocess.STARTF_USESHOWWINDOW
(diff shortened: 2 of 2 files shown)


wenzeslaus (changes requested):
Interesting that this was not bothering anyone yet. Does this show in GRASS GIS itself?

Anyway, I don't know whether the whole Env system is needed for this. Why not make the hidden window the only option? Will anyone ever want a terminal window popping up on Windows in the middle of some procedure for each (or selected) subprocess(es) called? The answer to me seems to be no, I don't want a separate window pop up for any of the subprocesses I call. Or is this something expected on Windows? The motivation for this PR is that it is unexpected and undesired, no?

While I enjoyed reading the new code and the idea of a new mechanism for handling general behavior, this comes in addition of the existing global settings for grass.script:

```
gs.set_raise_on_error(True)
gs.set_capture_stderr(True)
```

There is also the now broadly used (at least internally) passing of an _env_ dictionary to functions from the _run_command_ family. And one can pass other subprocess.Popen arguments such as _shell_ and, I assume, _startupinfo_.

Then there is of course the not so nice interface for global override you suggested to replace in the future:

```
os.environ["GRASS_OVERWRITE"] = "1"
```

Don't take me wrong, I do think we need a nice way how to handle all these, ideally in a consistent manner which would feels native to Python. The overwrite is not nice whether it is _os.environ_ or its copy. The globals for the package are disconnected from that etc.  

However, I'm reluctant to merge _Env_ at this point because it is focused on the global case while all the previous _env_ updates were going in the opposite direction allowing subprocesses to have their own unique set of variables. There is also the possible confusion in naming between _grass.script.env_ as an instance of _Env_ and _env_ parameters which are dictionaries, copies of _os.environ_.  

Finally, someone needs to analyze how this will interact with the session concept which is now present in grass.script.setup (even more with pending #3438), grass.jupyter, and experimental mapset sessions now merged #2367. I need some time (days-weeks) to consider that and I'm happy for input.

If that still addresses your main issue, I suggest removing the _Env_ part from this PR and focusing this PR on changing the behavior. Or in other words, I suggest splitting this PR into two and discussing the more complicated _Env_ part separately.

You could also use the existing global+function mechanism to do the settings if really needed. If I understand correctly, this should be compatible with the _Env_ mechanism if we merge that in the future.

landam:
> Interesting that this was not bothering anyone yet. Does this show in GRASS GIS itself?

No, it doesn't show in GRASS GIS itself. But it bothers at least QGIS. One of the reasons why don't use GRASS Python API to launch GRASS commands on background, see https://github.com/OSGeo/grass/pull/3460#issuecomment-1966299817

Background story: one colleague of mine simply rewrote one of his QGIS plugins to avoid GRASS commands just because of popping up windows. He replaced GRASS commands by equivalent QGIS tools.

Other real-life (SMODERP2D) example shown in https://github.com/OSGeo/grass/pull/3460

> Anyway, I don't know whether the whole Env system is needed for this. Why not make the hidden window the only 

Closing this PR by https://github.com/OSGeo/grass/pull/3460

neteler:
> Background story: one colleague of mine simply rewrote one of his QGIS plugins to avoid GRASS commands just because of popping up windows. He replaced GRASS commands by equivalent QGIS tools.

So, a good motivation to get the problem fixed in GRASS GIS.

wenzeslaus:
> No, it doesn't show in GRASS GIS itself. But it bothers at least QGIS. One of the reasons why don't use GRASS Python API to launch GRASS commands on background, see [#3460 (comment)](https://github.com/OSGeo/grass/pull/3460#issuecomment-1966299817)

After I added `--exec`, I know QGIS migrated to it. I think they were setting up their own session before. Is there a record of the discussion you are referring to here which is listing the reasons why QGIS does not use the Python API?

PR #3459, opened 2024-02-26
lib/vector/Vlib: always write out topo files in update mode

metzm:
When a vector with topology is opened in update mode, but nothing is changed, topology info (topo, cidx, sidx) is not written out. This PR fixes this.

I am not sure if this is a bug or an enhancement.

### DIFF
--- a/lib/vector/Vlib/open.c
+++ b/lib/vector/Vlib/open.c
@@ -572,6 +572,7 @@ int Vect__open_old(struct Map_info *Map, const char *name, const char *mapset,
             if (access(file_path, F_OK) == 0) /* fidx file exists? */
                 unlink(file_path);
         }
+        Map->support_updated = TRUE;
     }
 
     return level;
--- a/lib/vector/Vlib/open.c
+++ b/lib/vector/Vlib/open.c
@@ -550,9 +550,30 @@ int Vect__open_old(struct Map_info *Map, const char *name, const char *mapset,
         Vect_rewind(Map);
     }
 
-    /* DO NOT delete support files topo, sidx, cidx, fidx if native format
-     * was opened for update (not head_only)
-     * they will be deleted by Vect_close() if the map was actually updated */
+    /* delete support files if native format was opened for update (not
+     * head_only) */
+    if (update && !head_only) {
+        char file_path[GPATH_MAX];
+
+        Vect__get_element_path(file_path, Map, GV_TOPO_ELEMENT);
+        if (access(file_path, F_OK) == 0) /* topo file exists? */
+            unlink(file_path);
+
+        Vect__get_element_path(file_path, Map, GV_SIDX_ELEMENT);
+        if (access(file_path, F_OK) == 0) /* sidx file exists? */
+            unlink(file_path);
+
+        Vect__get_element_path(file_path, Map, GV_CIDX_ELEMENT);
+        if (access(file_path, F_OK) == 0) /* cidx file exists? */
+            unlink(file_path);
+
+        if (format == GV_FORMAT_OGR || format == GV_FORMAT_POSTGIS) {
+            Vect__get_element_path(file_path, Map, GV_FIDX_ELEMENT);
+            if (access(file_path, F_OK) == 0) /* fidx file exists? */
+                unlink(file_path);
+        }
+        Map->support_updated = TRUE;
+    }
 
     return level;
 }
--- a/lib/vector/Vlib/open.c
+++ b/lib/vector/Vlib/open.c
@@ -550,30 +550,9 @@ int Vect__open_old(struct Map_info *Map, const char *name, const char *mapset,
         Vect_rewind(Map);
     }
 
-    /* delete support files if native format was opened for update (not
-     * head_only) */
-    if (update && !head_only) {
-        char file_path[GPATH_MAX];
-
-        Vect__get_element_path(file_path, Map, GV_TOPO_ELEMENT);
-        if (access(file_path, F_OK) == 0) /* topo file exists? */
-            unlink(file_path);
-
-        Vect__get_element_path(file_path, Map, GV_SIDX_ELEMENT);
-        if (access(file_path, F_OK) == 0) /* sidx file exists? */
-            unlink(file_path);
-
-        Vect__get_element_path(file_path, Map, GV_CIDX_ELEMENT);
-        if (access(file_path, F_OK) == 0) /* cidx file exists? */
-            unlink(file_path);
-
-        if (format == GV_FORMAT_OGR || format == GV_FORMAT_POSTGIS) {
-            Vect__get_element_path(file_path, Map, GV_FIDX_ELEMENT);
-            if (access(file_path, F_OK) == 0) /* fidx file exists? */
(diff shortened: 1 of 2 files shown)


HuidaeCho (commented):
It works without `Vect_build_partial()` or `Vect_build()` as long as the map is `Vect_close()`. Still, `G_fatal_error()` needs to add an error handler to call `Vect_close()`. Is it possible just to crash without closing the map?

metzm:
When a vector is opened with `Vect_open_update2()`, the topo, cidx, and sidx files are deleted and need to be created anew which happens with `Vect_close()`. That means with a crash before `Vect_close(), these files and with them all of the topology info are lost.

HuidaeCho (dismissed):

metzm:
With [8ef73e8](https://github.com/OSGeo/grass/pull/3459/commits/8ef73e8a9b83bb659d6fe732a45456c02bc44e67) there is a better solution where a map opened in update mode survives with level 2, any changes would be discarded, i.e. if `Vect_close()` can not be called because of a crash earlier on, the old topo, sidx, cidx, fidx files would still exist.

HuidaeCho (approved):

HuidaeCho:
> With [8ef73e8](https://github.com/OSGeo/grass/pull/3459/commits/8ef73e8a9b83bb659d6fe732a45456c02bc44e67) there is a better solution where a map opened in update mode survives with level 2, any changes would be discarded, i.e. if `Vect_close()` can not be called because of a crash earlier on, the old topo, sidx, cidx, fidx files would still exist.

Works well! Thanks!

metzm:
I am not convinced with the current proposed solution because in case of a crash this can lead to a modified coor file (modified geometries) and old topo, sidx, cidx files no longer matching the new coor file which can cause strange errors.

Therefore I prefer my first solution were even with no changes, the topo, sidx, cidx files are created anew. In case of a crash, there will be no topo, sidx, cidx files and `v.build` will be needed to recreate them from the coor file.

HuidaeCho:
> I am not convinced with the current proposed solution because in case of a crash this can lead to a modified coor file (modified geometries) and old topo, sidx, cidx files no longer matching the new coor file which can cause strange errors.
> 
> Therefore I prefer my first solution were even with no changes, the topo, sidx, cidx files are created anew. In case of a crash, there will be no topo, sidx, cidx files and `v.build` will be needed to recreate them from the coor file.

Did a quick test.
## With your first solution,
```c
Vect_open_update2();
/* changes or no changes both with and without Vect_build() */
G_fatal_error();
```
Changes (or no changes) are preserved, but level is downgraded to 1 (v.info works, v.build upgrades it to 2).

```c
Vect_open_update2();
/* changes without Vect_build() */
Vect_close();
```
Changes are preserved, but level is downgraded to 1 (v.info works, v.build upgrades it to 2).


```c
Vect_open_update2();
/* changes with Vect_build() */
Vect_close()
```
Everything is good!

## With your second solution,
```c
Vect_open_update2();
/* no changes without Vect_build() */
G_fatal_error();
```
Level is still 2.

```c
Vect_open_update2();
/* changes without Vect_build() */
G_fatal_error(); /* or Vect_close(); */
```
v.info doesn't work. After v.build, v.info shows level 2 and changes are preserved.

```c
Vect_open_update2();
/* changes with Vect_build() */
G_fatal_error();
```
v.info doesn't work. After v.build, v.info shows level 2 and changes are preserved.

```c
Vect_open_update2();
/* changes with Vect_build() */
Vect_close()
```

Everything is good!

wenzeslaus:
This may not be obvious since we have mostly tests of individual tools, but these examples can be rewritten as tests in Python without a need to create a separate binary like we have for [3D raster library](https://grass.osgeo.org/grass83/manuals/test.raster3d.lib.html). We have examples of using ctypes to test the C library functions for both [grass.gunittest tests](https://github.com/OSGeo/grass/tree/main/lib/imagery/testsuite) using ctypes (also [without grass.pygrass](https://github.com/OSGeo/grass/tree/main/lib/gis/testsuite)) and [pytest-based tests](https://github.com/OSGeo/grass/tree/main/lib/gis/tests).

metzm:
@marisn This does not solve the problem of a crash.

I prefer solution 1 where the support files are deleted when opening and always written out anew when closing. Writing out these files is fast and does not require rebuilding topology. With solution 2, there is the danger of ending up with non matching/wrong support files which can lead to difficult to debug problems.

HuidaeCho:
> @marisn This does not solve the problem of a crash.
> 
> I prefer solution 1 where the support files are deleted when opening and always written out anew when closing. Writing out these files is fast and does not require rebuilding topology. With solution 2, there is the danger of ending up with non matching/wrong support files which can lead to difficult to debug problems.

@metzm Based on my test, I think solution 1 should be preferred because with either solution, we need an error handler to properly close the map and it'll be the developer's mistake if that doesn't happen and solution 1 will be less harmful in that case.

HuidaeCho (approved):

metzm:
There is a strange error in the tests with
`Running ./python/grass/pygrass/vector/testsuite/test_pygrass_vector_doctests.py...`
resulting in
`ERROR: Category index is not up to date`

I guess that before the test vector is closed, topology must be built before
https://github.com/OSGeo/grass/blob/main/python/grass/pygrass/vector/__init__.py#L621

The error in the test seems to be an error in the test: features are written to a new vector, but topology is not built with `GV_BUILD_ALL` before closing.

HuidaeCho:
I saw that same error between edits without topology building. I was wondering if it's always required to rebuild topology or if there are any cases where we don't need it when editing features. Will it be slow to rebuild it if there are a lot of features?

metzm:
> [...] Will it be slow to rebuild it if there are a lot of features?

If topology is already built and you just want that the cidx is marked up to date, it should be enough to call `Vect_build()` **without** calling `Vect_build_partial(GV_BUILD_NONE)` first, and this should be very fast.

metzm:
In pygrass, here https://github.com/OSGeo/grass/blob/main/python/grass/pygrass/vector/abstract.py#L454-L470
and here https://github.com/OSGeo/grass/blob/main/python/grass/pygrass/vector/abstract.py#L454
the vector is first closed, then topology is built, then the vector is closed again. This seems wrong to me, topology should be built when the vector is still open, or the other way around, the vector should only be closed after topology has been built, not before.

marisn:
> This seems wrong to me, topology should be built when the vector is still open, or the other way around, the vector should only be closed after topology has been built, not before.

Please create a PR for the documentation as neither Vect_build or Vect_close mention this important bit.

metzm:
> > This seems wrong to me, topology should be built when the vector is still open, or the other way around, the vector should only be closed after topology has been built, not before.
> 
> Please create a PR for the documentation as neither Vect_build or Vect_close mention this important bit.

`Vect_build()` needs to be called as `int Vect_build(struct Map_info *Map)`, thus a valid `struct Map_info` pointer must be passed which implies an opened vector map.

metzm:
With the fix in pygrass, all checks have now passed. Ready to merge?

neteler (approved):

PR #3462, opened 2024-02-28
vector: use full db connection settings for db.describe

metzm:
Some scripts did not specify database and driver when calling `grass.db_describe()`, causing errors if vector maps use non-standard db connections. This PR queries actual db connection settings for affected modules and passes these to `grass.db_describe()`


metzm:
Backported to G83

PR #3480, opened 2024-03-09
r.surf.fractal: Added seed option and -s flag to module

Shashankss1205:
Here is my proposed solution to the issue raised #1044.
I have written the code for adding `seed` and `-s flag` for reproducibility, along with a unit test for verifying it's functionality


.vscode/settings.json
marisn, line 1: This file has to be removed.

raster/r.surf.fractal/testsuite/test_r_surf_fractal.py
```diff
@@ -0,0 +1,110 @@
...
+
+from grass.gunittest.case import TestCase
+from grass.gunittest.main import test
+
+
+class FractalTestCase(TestCase):
+    """Test case for r.surf.fractal"""
+
+    # Raster map name to be used as output
+    output = "fractal_result"
```
marisn, line 29: Better use tempname: `output = gs.core.tempname(10)`

raster/r.surf.fractal/testsuite/test_r_surf_fractal.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        """Set up necessary environment"""
+        # Set up temporary computational region
+        cls.use_temp_region()
+        # Only 100,000,000 seem to reasonably (not 100%) ensure that all values
+        # are generated, so exceeding of ranges actually shows up.
```
marisn, line 37: What kind of ranges you are referring to? I didn't see any range check. Usually a tiny computational region suffices to save on CI execution time.

marisn (changes requested):

.vscode/settings.json
Shashankss1205: I am sorry for adding this file without double checking. I will remove it in the next commit.
Thank You @marisn.

raster/r.surf.fractal/testsuite/test_r_surf_fractal.py
Shashankss1205: Would you mind helping me with the test suite please, because I haven't previously worked with them and therefore I was struggling with how to write them.

Thanks @marisn

raster/r.surf.fractal/testsuite/test_r_surf_fractal.py
petrasovaa: For tests, we actually don't use the temporary names, the tests are usually ran in throw-away mapsets. So you can keep this.

raster/r.surf.fractal/main.c
```diff
@@ -54,9 +59,44 @@ int main(int argc, char *argv[])
...
+        _("Automatically generates random seed for random number"
+          " generator (use when you don't want to provide the seed option)");
+
+
     if (G_parser(argc, argv)) /* Performs the prompting for      */
         exit(EXIT_FAILURE);   /* keyboard input.                 */
 
+    /****** INITIALISE RANDOM NUMBER GENERATOR ******/
+    if (s_flag->answer) {
+        seed_value = G_math_srand_auto();
```
petrasovaa, line 83: Use instead functions [G_srand48_auto](https://grass.osgeo.org/programming8/defs_2gis_8h.html#a43a0a497f87c6ff53f2b4827bf05ea75) and G_srand48. You need to include `grass/gis.h`.

raster/r.surf.fractal/main.c
```diff
@@ -28,6 +28,11 @@ int main(int argc, char *argv[])
     struct Option *rast_out;   /* Structure for output raster     */
     struct Option *frac_dim;   /* Fractal dimension of surface.   */
     struct Option *num_images; /* Number of images to produce.    */
+    struct Option *seed;
+    struct Flag *s_flag;
```
ninsbl, line 32: Just wondering, is the s-flag needed given that seed is optional?

Or is it added for consistency with current practice in other modules? Possibly redundand? If `seed` option is given, seed should be used, no?

And looking at the logic in lines 82 to 95 it seems that setting the s-flag has the same effect as not setting the s-flag...

So, maybe the s-flag can be dropped?

petrasovaa: I looked back at when this was [introduced in r.mapcalc](https://lists.osgeo.org/pipermail/grass-dev/2014-July/070150.html) by Glynn:

> My main objection to automatic seeding is that people will inevitably produce non-repeatable results without even realising it.

That's why user needs to choose -s flag or seed option explicitly. So the focus there was reproducibility. We could argue though that convenience is priority an in that case we could decide to drop the -s flag and not introduce it here.

echoix: I wasn't following the development when the seed option was discussed for r.mapcalc. From an exterior point of view, the general trend of all software I have used yet that uses random numbers is that when you can set the seed, it is usually optional and for users who want know they want it reproducible. I almost always say it as an addition, not a requirement for just basic uses.

So, @Shashankss1205's interpretation is just like mine, and his implementation where not setting the `-s` flag results in a seed being generated, but with a warning with explanations, is my expected behaviour. Plus it doesn't break compatibility as well, and shows what was done. Maybe an even clearer message that makes the user aware and learn why setting a seed is wanted could be even better. If it's a pattern that we have for handling of -s flags, shouldn't these messages be common across GRASS and not specific to each module?

petrasovaa: Looks like the consensus so far is that -s flag could be dropped. @Shashankss1205 could you please rewrite this without the -s flag?

wenzeslaus: +1 for dropping the _s_ flag. Outside of this PR, but this will require making it optional wherever it is used.

Shashankss1205: Ok, I'll rewrite and send the code, my main aim was to keep the consistency of the previous code changes and abiding by the same rules. Will be happy to make the necessary changes.

Thank you for all of your's valuable insights.

Shashankss1205: Are there any suggestions related to the updated code, I think I have removed all occurences of -s_flag from my code, and also I wanted to **seek some guidance** on how to work further for getting into **GSOC'24 with this organization.**

raster/r.surf.fractal/main.c
```diff
@@ -54,9 +58,32 @@ int main(int argc, char *argv[])
...
+    if (seed->answer) {
+        seed_value = atol(seed->answer);
+        G_srand48(seed_value);
+        G_verbose_message(_("Read random seed from %s option: %ld"), seed->key,
+                          seed_value);
+    }
+    else {
+        /* default as it used to be */
+        seed_value = G_srand48_auto();
+        G_verbose_message(_("Warning set option seed. Generated "
```
petrasovaa, line 82: Remove the first sentence and the "-s"

raster/r.surf.fractal/testsuite/test_r_surf_fractal.py
petrasovaa: See for example test for r.random.cells and [test documentation](https://grass.osgeo.org/grass83/manuals/libpython/gunittest.html). Run r.surf.fractal first with certain fixed seed and save the raster statistics (computed with r.univar) for reference in the test. You can then use assertRasterFitsUnivar to ensure the test-generated raster with the same fixed seed will always give the same values. This type of test is not necessarily testing the output is correct, but it helps detect any regressions in the tool in the future and also that with a fixed seed you always get the same results.

petrasovaa (changes requested):
Check the failed CI logs, r.surf.fractal is currently not compiling.

raster/r.surf.fractal/main.c
petrasovaa: Great! Please introduce yourself on [grass-dev mailing list](https://lists.osgeo.org/mailman/listinfo/grass-dev) and let us know which [idea](https://grasswiki.osgeo.org/wiki/GRASS_GSoC_Ideas_2024) you would like to work on so that we can further discuss it. See the tips on the [idea](https://grasswiki.osgeo.org/wiki/GRASS_GSoC_Ideas_2024) page as well. You can also start drafting the application based on the template specified [here](https://wiki.osgeo.org/wiki/Google_Summer_of_Code_Recommendations_for_Students).

petrasovaa:
@Shashankss1205 do you plan to keep working on this? There is not that much missing (a simple test would suffice) and fixing the compilation error should be easy.

raster/r.surf.fractal/testsuite/test_r_surf_fractal.py
```diff
@@ -0,0 +1,110 @@
...
+COPYRIGHT: (C) 2024 by Shashank Shekhar Singh and the GRASS Development Team
+
+This program is free software under the GNU General Public
+License (>=v2). Read the file COPYING that comes with GRASS
+for details.
+"""
+
+import os
+
+import grass.script as gs
```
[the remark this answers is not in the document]
Shashankss1205: Is there something I need to do here?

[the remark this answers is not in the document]
echoix: You can apply the suggestion, as you don't use the os import or the grass.script import

petrasovaa:
Closing, solved in #5233

Issue #3481, opened 2024-03-10
[Bug] Display commands for `d.mon` can take a long time *unexpectedly* to update the `cmd` file when first run.


HuidaeCho:
**Describe the bug**
When we run a display command (`d.*`) from the terminal to add a layer to a `wx*` monitor, it runs twice:
1. Original process executed by the user without `GRASS_REGION` (display extent): This process adds a new command line to the `cmd` file. If the command takes a long time to finish for the entire raster region, we have to wait even if the display extent is very small.
2. Another process executed by the monitor with `GRASS_REGION`: The monitor runs the above command line saved in the `cmd` file, but this time with the current display extent. The same command can run quickly if the display extent is very small. Ultimately, we will only see this result.

**To Reproduce**

```bash
wget https://data.isnew.info/grass/nidp.pack https://data.isnew.info/grass/drain_tx.pack
grass -c epsg:5070 /tmp/grass_test
r.unpack -o nidp.pack
r.unpack -o drain_tx.pack
g.region rast=nidp -ap
d.mon wx0
d.rast nidp
# zoom to a very small region in the wx0 monitor (e.g., 50x50 grid)
d.rast.arrow -a drain_tx type=drainage # takes very long for a small region
```
See it takes long even though the display extent is very small

**Expected behavior**
Any display commands should *always* honor the current display extent.

**System description (please complete the following information):**

- Operating System: Linux
- GRASS GIS version: main branch

**Additional context**
`GRASS_REGION` is internal to the monitor and the terminal doesn't have access to it. Even if `GRASS_REGION` was exposed to the terminal, display commands would still run twice.

I changed

https://github.com/OSGeo/grass/blob/b45319c8a35c68d8d80c49f6e5e12da42a01961e/display/d.rast.arrow/main.c#L222-L223

to

```c
    /* Setup driver and check important information */
    D_open_driver();

    if (!getenv("GRASS_REGION")) {
        D_save_command(G_recreate_command());
        D_close_driver();
        exit(EXIT_SUCCESS);
    }
```
and it worked. Basically, `d.rast.arrow` just adds a new command line to the `cmd` file and exits. Then, it'll be executed by the monitor for actual rendering. This solution is not very elegant and takes the module developer's effort.

Maybe, we add to `D_open_driver()`:
```c
    if (!getenv("GRASS_REGION")) {
        D_save_command(G_recreate_command());
        D_close_driver();
        return 1;
    }
```
and in each display module:
```c
[...]

PR #3482, opened 2024-03-10
libdisplay: D_open_driver exits if GRASS_REGION is not defined from the terminal

HuidaeCho:
This PR attempts to address https://github.com/OSGeo/grass/issues/3481. Display commands can be very slow for a small display extent in the monitor because the process executed by the user from the terminal doesn't have access to that display extent through `GRASS_REGION`.


HuidaeCho:
Hmm... It doesn't work with `d.redraw` because `d.redraw` has no access to `GRASS_REGION` in the monitor.

HuidaeCho:
Testing other display modules...

echoix (dismissed):

wenzeslaus (changes requested):
More explanation or different implementation needed. Please see my comment in https://github.com/OSGeo/grass/issues/3481#issuecomment-1996236968

lib/display/r_raster.c
```diff
@@ -152,7 +155,28 @@ int D_open_driver(void)
 
     init();
 
-    return 0;
+    /* if GRASS_REGION is not defined, a display module can choose to render
+     * for the entire computational region or wait for the monitor to call it
+     * with GRASS_REGION set to the current display extent (what is shown) in
+     * the monitor; the monitor defines GRASS_REGION to the display extent
```
wenzeslaus, line 161: Usage of GRASS_REGION just happens to be the case in d.mon wx0. I don't know much about the display library, but I think the current API allows implementation of a monitor which does not use GRASS_REGION. This makes GRASS_REGION integral part of the monitor system. That's perhaps fine, but it is a change we should be sure about.

lib/display/r_raster.c
```diff
@@ -152,7 +155,28 @@ int D_open_driver(void)
...
+     * dynamically, but it's the display module who is responsible to contrain
+     * its rendering within this region to save computational time; if the
+     * display module does not check the return value of D_open_driver(), it
+     * will have to render the same content twice (once by the command line and
```
wenzeslaus, line 165: "render twice" - It seems to me that should never happen and the display modules should be required to work properly rather than opting in it.

lib/display/r_raster.c
```diff
@@ -152,7 +155,28 @@ int D_open_driver(void)
...
+     * once more by the monitor) and it is potentially more intensive
+     * computationally to render for the entire computational region, so it is
+     * recommended to exit from the module if D_open_driver() returns -1; upon
```
wenzeslaus, line 168: Rendering according to computation region seems like plain wrong when the expectation is _only_ to pass the command line deeper to the rendering backend. Is it possible that this GRASS_REGION issue is a symptom of a deeper issue in the monitor system?

lib/display/r_raster.c
```diff
@@ -152,7 +155,28 @@ int D_open_driver(void)
...
+     * exit, the module will be executed again automatically by the monitor
+     * with GRASS_REGION and D_open_driver() will return 0 then; now the module
+     * can proceed to render its content only for the display extent; here is
+     * the suggested pattern:
+     *
+       if (D_open_driver() == -1) {
+           D_close_driver();
+           exit(EXIT_SUCCESS);
+       }
```
wenzeslaus, line 177: An example usage should be in the Doxygen comment.

wenzeslaus (changes requested):

lib/display/r_raster.c
```diff
@@ -78,6 +78,9 @@ static void init(void)
   Default display driver is Cairo, if not available PNG is used.
 
   \return 0 on success
+  \return -1 if GRASS_REGION is not defined; use this return value to decide
+  what to do when rendering for the entire computational region can be slow
+  without constraining to the display extent in GRASS_REGION from the monitor
```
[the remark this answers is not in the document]
HuidaeCho: It's actually both. It can be slow and renders twice. It's rendered the first time in the same file, which will be overwritten the second time. You can see that first render using this patch to the monitor `render.py`:
```patch
--- MONITORS/wx0/render.py.orig 2024-03-14 01:46:02.172981610 -0600
+++ MONITORS/wx0/render.py      2024-03-14 01:45:37.172088784 -0600
@@ -2,6 +2,7 @@
 import os
 import sys
 import tempfile
+import shutil
 
 from grass.script import core as grass
 from grass.script import task as gtask
@@ -163,6 +164,7 @@
     read_stdin(cmd)
 
     render(cmd, mapfile)
+    shutil.copy(mapfile, "/tmp/first_render.ppm");
     update_cmd_file(os.path.join(path, "cmd"), cmd, mapfile)
     if cmd[0] == "d.erase" and os.path.exists(legfile):
         os.remove(legfile)
```
and run `d.rast nidp` for a small region. Check `/tmp/first_render.ppm` to see the entire Texas, not the small region.

lib/display/r_raster.c
HuidaeCho: `GRASS_REGION` is not specific to the `wx0` monitor. `d.mon`'s `render_cmd.py` already uses it to adjust the region:
https://github.com/OSGeo/grass/blob/9c3b67ec82ff1d5db64f6a2baba98390bfe9406c/display/d.mon/render_cmd.py#L115

and I think `GRASS_REGION` was mainly designed for this purpose. Otherwise, the monitor system would have to change the computational region every time it wants to change the display extent. How does `g.gui` do it?

From [variables.html](https://grass.osgeo.org/grass83/manuals/variables.html): _GRASS_REGION ... use is the same as WIND_OVERRIDE. ... This allows programs such as the GUI to run external commands on an alternate region without having to modify the WIND file then change it back afterwards._

lib/display/r_raster.c
HuidaeCho: I completely agree with you. However, at the moment, not all display modules **display** (e.g., `d.info`). Obviously, with my first `exit()` method, this module and any other modules relying on it didn't work.

Maybe, it would be better to add a new argument to `D_open_driver(exit_on_no_region)` to not exit and use `D_open_driver(1)` in displaying modules and `D_open_driver(0)` in non-displaying ones.

lib/display/r_raster.c
HuidaeCho: Like I said in https://github.com/OSGeo/grass/issues/3481#issuecomment-1996725006, we used to have an IPC-based (and socket-based) monitor system for UN*X where we didn't have to call display modules **three times**, IIRC, like we do now. Since we moved away from UN*X only systems and to support non-interactive file-based monitors, we don't have a server-client model anymore. Is it a deeper issue in the monitor system? Yes and no... Yes because it's very annoying to deal with this kind of multiple calls to one command and maybe no because it's a necessary evil within the current architecture when you deal with the terminal and GUI using only environment variables.

HuidaeCho:
I think I found a cleaner and better solution. Testing...

Issue #3497, opened 2024-03-13
[Bug] r.univar returns nan for rasters with only negative values


wenzeslaus:
**Describe the bug**

For rasters with only negative values, r.univar returns nan for maximum and range.

For positive values and a mix of positive and negative values, the module returns expected max and range values.

**To Reproduce**

```
grass -c ~/grassdata/nc_spm_08_grass7/test1
```

```
GRASS nc_spm_08_grass7/user1:grass > r.mapcalc "negative_const = -1"
 100%
GRASS nc_spm_08_grass7/user1:grass > r.mapcalc "negative_random = rand(-1, -0.01)" -s
 100%
GRASS nc_spm_08_grass7/user1:grass > r.univar negative_const
 100%
 100%
total null and non-null cells: 277200
total null cells: 0

Of the non-null cells:
----------------------
n: 277200
minimum: -1
maximum: nan
range: nan
mean: -1
mean of absolute values: 1
standard deviation: 0
variance: 0
variation coefficient: -0 %
sum: -277200
GRASS nc_spm_08_grass7/user1:grass > r.univar negative_random
 100%
 100%
total null and non-null cells: 277200
total null cells: 0

Of the non-null cells:
----------------------
n: 277200
minimum: -0.999997
maximum: nan
range: nan
mean: -0.505264
mean of absolute values: 0.505264
standard deviation: 0.285897
variance: 0.0817371
variation coefficient: -56.5837 %
sum: -140059.082419107
```

Same with the `g` flag.

**Expected behavior**

Maximum and range computed.

**System description**

- Operating System: Linux
- GRASS GIS version: main branch

I didn't do a full `git bisect`, but 8.2.0 is good and 8.3.0 is bad.

neteler:
Here's a candidate, in preparation for CMake support:

- _cmake support preparation: use of INFINITY and NAN_ (#2681), merged Jan 16, 2023
  - which was extracted from WIP PR #289

Then 8.3.0 was released on  Jun 24, 2023.

wenzeslaus:
I did _git bisect_ and it is indeed you guess: a946ca6bb7da6721617a43cf94ae4f67edc5d416 (#2681)

## _git bisect_

Test script `test_for_bisect.sh` (assuming the data from above):

```bash
#!/bin/bash
cd raster/r.univar
make && /.../bin.x86_64-pc-linux-gnu/grass /.../grassdata/nc_spm_08_grass7/test1/ \
   --exec r.univar negative_random \
[...]

PR #3512, opened 2024-03-18
r.univar: init max value with smallest possible value

nilason:
Init max value with smallest possible value, `DBL_MIN` is still a positive value. 

Closes: https://github.com/OSGeo/grass/issues/3497


nilason:
Just to clarify, the following was never true for a negative value (`val`) if `zd->max = DBL_MIN`:
https://github.com/OSGeo/grass/blob/b656141cbc9d86edc64e8947c5aac32e37a39b6c/raster/r.univar/r.univar_main.c#L465

echoix (approved):

PR #3515, opened 2024-03-19
bandit: add bandit to GRASS GIS CI

jadenabrams100:
This pull request will add a GitHub action that will run Bandit on pull requests and inform the maintainers of detected security vulnerabilities in the repositories Python code. Additional configurations to Bandit can be done in the .bandit file in the root of the repository.

This pull request includes work from Nolan Peters (@Ntp9413), Maryam Abkar (@naidneelttil), Jaden Abrams (@jadenabrams100), and Kira Polchow (@kpolchow).

Some additional configurations could be useful for Bandit in the future, but we did not include them because we did not know if the community wanted them excluded. One rule we chose not to ignore for now is B405, and B313-B320, which concerns XML parsing.

Additional tweaks may be made as we find false positives and find additional ways to make Bandit more accurate.


wenzeslaus (changes requested):
The CI reports deprecations for couple of the actions you are using. See Checks > Bandit. Maybe you used an outdated example, make sure to check the official doc for the actions. Maybe you just need to update the versions in the workflow.

echoix:
Even if the action has its commit hash pinned, and that I was able to review that newly introduced action complete code since it was simple, it is still an unmaintained action (2 years) and used by only one repo: an action playground.

So it doesn't really pass my trust test. Reputation of dependencies in the CI is a part of keeping software secure.

Since it is a simple action, what about just plainly running it inside our Python quality checks, either as steps (to use the same checkout, we are often time-and-runner constrained here), or as a separate job inside that file. If it is a separate job, add it to the required jobs of the verify success to add is as a required check.

jadenabrams100:
> Even if the action has its commit hash pinned, and that I was able to review that newly introduced action complete code since it was simple, it is still an unmaintained action (2 years) and used by only one repo: an action playground.
> 
> So it doesn't really pass my trust test. Reputation of dependencies in the CI is a part of keeping software secure.
> 
> Since it is a simple action, what about just plainly running it inside our Python quality checks, either as steps (to use the same checkout, we are often time-and-runner constrained here), or as a separate job inside that file. If it is a separate job, add it to the required jobs of the verify success to add is as a required check.

Thank you for your feedback! I see that there are deprecated portions of the action and I hear your concerns about them and the action in general. I will look into moving the Bandit scanning functionality into the Python quality checks workflow.

.github/workflows/python-code-quality.yml
```diff
@@ -68,6 +68,7 @@ jobs:
           pip install black==${{ matrix.black-version }}
           pip install flake8==${{ matrix.flake8-version }}
           pip install pylint==${{ matrix.pylint-version }} pytest-github-actions-annotate-failures
+          pip install bandit
```
echoix, line 71: If you could, it would be even better if the version was pinned. Otherwise, if everything uploads correctly, I think this PR is ok to go!

.github/workflows/python-code-quality.yml
```diff
@@ -149,6 +150,22 @@ jobs:
           name: sphinx-grass
           path: sphinx-grass
           retention-days: 3
+
+      - name: Bandit Vulnerability Scan
+        run: |
+          bandit -c pyproject.toml -iii -r . -f sarif -o bandit.sarif --exit-zero
+
+      - name: Upload Bandit Scan Results
+        uses: actions/upload-artifact@v4
```
echoix, line 159: Would you be able to pin like the clang-format workflow, including the comment at the end? Renovate keeps the has and the comment updated with the correct granularity.

.github/workflows/python-code-quality.yml
```diff
@@ -149,6 +150,22 @@ jobs:
...
+        with:
+          name: bandit.sarif
+          path: bandit.sarif
+
+      - name: Upload SARIF File into Security Tab
+        uses: github/codeql-action/upload-sarif@v3
```
echoix, line 165: Same thing here for the pinned version.

echoix:
@jadenabrams100 I think everything is ready! I assumed you checked that the upload of the SARIF works correctly on your side, but since this PR doesn't have any problems, I can't see a issue reported for that PR branch, which is normal.

echoix (approved):
Good!

echoix:
If ever the upload doesn't really work later on, we could handle it from there ourselves

wenzeslaus (approved):
This looks great. Thank you!

wenzeslaus:
...well, it looks great in terms of the action working, not in terms of issues to solve :-)

Screenshot from 2024-03-22 13-16-23

echoix:
We'll get there, but at least I assume it will do like CodeQL and comment preventing us from adding new ones

PR #3582, opened 2024-04-12
grassTask.get_param(): full match also for strings

landam:
`grassTask.get_param()` does not perform full match on string data type, see https://github.com/OSGeo/grass/blob/main/python/grass/script/task.py#L151

This may lead to buggy behaviour as explained below.

Sample code:

```py
from grass.script.task import grassTask

print(grassTask("r.mapcalc.simple").get_param('e')['name'])
```

returns `expression` instead of `e`.

### DIFF
--- a/python/grass/script/tests/test_script_task.py
+++ b/python/grass/script/tests/test_script_task.py
@@ -1,0 +1,10 @@
+import pytest
+from grass.script.task import grassTask as gtask
+
+def test_mapcalc_simple_e_name():
+    gt = gtask("r.mapcalc.simple")
+    assert gt.get_param('e')['name'] == 'e'
+
+def test_mapcalc_simple_expession_name():
+    gt = gtask("r.mapcalc.simple")
+    assert gt.get_param('expression')['name'] == 'expression'
--- a/python/grass/script/tests/test_script_task.py
+++ b/python/grass/script/tests/test_script_task.py
@@ -4,7 +4,8 @@ from grass.script.task import grassTask as gtask
 
 def test_mapcalc_simple_e_name():
     gt = gtask("r.mapcalc.simple")
-    assert gt.get_param('e')['name'] == 'e'
+    assert gt.get_param("e")["name"] == "e"
+
 
 def test_mapcalc_simple_expession_name():
     gt = gtask("r.mapcalc.simple")
--- a/python/grass/script/tests/test_script_task.py
+++ b/python/grass/script/tests/test_script_task.py
@@ -9,4 +9,4 @@ def test_mapcalc_simple_e_name():
 
 def test_mapcalc_simple_expession_name():
     gt = gtask("r.mapcalc.simple")
-    assert gt.get_param('expression')['name'] == 'expression'
+    assert gt.get_param("expression")["name"] == "expression"
--- a/python/grass/script/tests/test_script_task.py
+++ b/python/grass/script/tests/test_script_task.py
@@ -1,4 +1,3 @@
-import pytest
 from grass.script.task import grassTask as gtask
 
 
--- a/python/grass/script/tests/test_script_task.py
+++ b/python/grass/script/tests/test_script_task.py
@@ -1,6 +1,7 @@
 import pytest
 from grass.script.task import grassTask as gtask
 
+
 def test_mapcalc_simple_e_name():
     gt = gtask("r.mapcalc.simple")
     assert gt.get_param('e')['name'] == 'e'
--- a/python/grass/script/task.py
+++ b/python/grass/script/task.py
@@ -148,9 +148,6 @@ class grassTask:
             if isinstance(val, (list, tuple)):
                 if value in val:
                     return p
-            elif isinstance(val, (bytes, str)):
-                if p[element][: len(value)] == value:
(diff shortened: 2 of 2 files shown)


echoix:
I don't understand enough this part of the code to know what's going on.

Could the broken behaviour be written in a simple unit test so we would see it fail and then fixed? Either for grassTask or mapcalc.

Something like:
```python
import pytest
from grass.script.task import grassTask as gtask

def test_mapcalc_simple_e_name():
    gt = gtask("r.mapcalc.simple")
    assert gt.get_param('e')['name'] == 'e'

def test_mapcalc_simple_expession_name():
    gt = gtask("r.mapcalc.simple")
    assert gt.get_param('expression')['name'] == 'expression'

```

I just wrote it on the fly, there might be formatting errors

echoix:
Ok, I had the chance to run the equivalent test file in my branch. I purposely made it based on my set of branches that are behind the main repo, but advanced on code coverage and better pytest reporting.

So before the fix of this PR:
https://github.com/echoix/grass/pull/72

And after applying the fix of this PR:
https://github.com/echoix/grass/pull/73

The test fails before and after.
I think I might not understood your example correctly then, as we see from these traces

https://github.com/echoix/grass/actions/runs/8670103706/job/23777706024?pr=72#step:13:182

https://github.com/echoix/grass/actions/runs/8670117007/job/23777729558?pr=73#step:13:179

image
image

wenzeslaus (changes requested):
This looks like another case where our Python code is trying to mimic what is happening in the parser, but it fails to account for all the cases.

However, given the confusion around it, it would be better to have a test before moving forward here.

wenzeslaus (dismissed):
After the discussion during PSC meeting, it seems to me that there is was not strong reason in the past to have these lines there. It can always fallback to the full match now and before. So, we can behave like these lines were never there.

I don't have an opinion on how far this should be backported except that the only supported release series is 8.3 at this point (by us a the project; us as individuals contributors can pursue fixing and release other versions).

Let's proceed.

wenzeslaus:
> The test fails before and after.

@landam's example works for me:

```python
grass-dev --tmp-project XY --exec python -c 'from grass.script.task import grassTask; print(grassTask("r.mapcalc.simple").get_param("e")["name"])'
```

Having it in the code, it would be nice to have.

landam:
> ```python
> grass-dev --tmp-project XY --exec python -c 'from grass.script.task import grassTask; print(grassTask("r.mapcalc.simple").get_param("e")["name"])'
> ```

Before:

```
./bin.x86_64-pc-linux-gnu/grass --tmp-project XY --exec python3 -c 'from grass.script.task import grassTask; print(grassTask("r.mapcalc.simple").get_param("e")["name"])'
expression
```

-> wrong match

After:

```
./bin.x86_64-pc-linux-gnu/grass --tmp-project XY --exec python3 -c 'from grass.script.task import grassTask; print(grassTask("r.mapcalc.simple").get_param("e")["name"])'
e
```

-> correct match

> Having it in the code would be nice to have.

Truly to say I am not sure what do you mean by this statement. Can you clarify it?

echoix:
> > ```python
> 
> > grass-dev --tmp-project XY --exec python -c 'from grass.script.task import grassTask; print(grassTask("r.mapcalc.simple").get_param("e")["name"])'
> 
> > ```
> 
> 
> 
> Before:
> 
> 
> 
> ```
> 
> ./bin.x86_64-pc-linux-gnu/grass --tmp-project XY --exec python3 -c 'from grass.script.task import grassTask; print(grassTask("r.mapcalc.simple").get_param("e")["name"])'
> 
> expression
> 
> ```
> 
> 
> 
> -> wrong match
> 
> 
> 
> After:
> 
> 
> 
> ```
> 
> ./bin.x86_64-pc-linux-gnu/grass --tmp-project XY --exec python3 -c 'from grass.script.task import grassTask; print(grassTask("r.mapcalc.simple").get_param("e")["name"])'
> 
> e
> 
> ```
> 
> 
> 
> -> correct match
> 
> 
> 
> > Having it in the code would be nice to have.
> 
> 
> 
> Truly to say I am not sure what do you mean by this statement. Can you clarify it?

I assume this means written as a test?

python/grass/script/tests/test_script_task.py
```diff
@@ -0,0 +1,12 @@
+import pytest
```
echoix, line 1:
```suggestion
```

Seems like flake8 doesn't like that I didn't use any fixtures at all finally, so pytest is unused. Would this work?

wenzeslaus:
Yes, no need to import pytest for a basic pytest test. They don't run as
standalone scripts.

wenzeslaus:
Yes, having it written as a test is what I meant to say. Nice to have ==
not required but desirable.

echoix (approved):

landam:
> Yes, having it written as a test is what I meant to say. Nice to have == not required but desirable.

Done by @echoix (thanks).

echoix:
You were able to merge even with the failed required windows builds?

landam:
Yes, I bypassed the requirements exceptionally. Failing CI was not related to this PR.

PR #3590, opened 2024-04-13
v.surf.rst: Cross-validation OpenMP support

cyliang368:
This finishes the implementation of OpenMP support for cross-validation (#2644). A critical block is added to `segmen2d_parallel.c` to make the writing to the database in serial. `main.c` is modified to allow parallel computing with nprocs>1. Two simple test scripts are used to run cross-validation with nprocs=1 and nprocs>1. The results of these tests are the same.


lib/rst/interp_float/segmen2d_parallel.c
```diff
@@ -369,8 +369,12 @@ int IL_interp_segments_2d_parallel(
                         b[tid][0] = 0.;
                         G_lubksb(matrix[tid], data_local[tid]->n_points,
                                  indx[tid], b[tid]);
-                        params->check_points(params, data_local[tid], b[tid],
-                                             ertot, zmin, dnorm, skip_point);
+#pragma omp critical
```
HuidaeCho, line 372: I wouldn't consider a critical block parallel because it restricts execution of them to one thread at a time and performance won't be much different from serial. Have you checked speedup and efficiency? Any plots to prove performance improvement?

cyliang368: While doing cross-validation, file writing conflicts happen in the `check_points`  function. The critical block can avoid these conflicts. This function is not the biggest bottleneck of the whole process. The most computational time is caused by the LU decomposition in `create_matrix,` which is O(n^3) complexity. Running through every point by for loop, the complexity is O(n^4).

I did a simple speedup and efficiency analysis, and the results are shown in the figures below. Cross-validation speeds up and is much more efficient than one-time interpolation. Serializing the `check_points` function does not lead to significant overheads. Instead, parallelizing the most computational parts greatly speeds up the process.

efficiency
speedup

HuidaeCho: Got it. I can see `Vect_write*` and `db_execute*` in `segment2d_parallel.c`. `params->grid_calc()` uses the same pattern.

vector/v.surf.rst/test_suite/profile.sh
```diff
@@ -5,7 +5,7 @@ OUTFILE=callgrind.out.1196
 g.copy vect=elev_lid792_bepts,lidar_test
 v.build lidar_test
 g.region n=221230 s=220816 w=639090 e=639530 res=0.5 -p
-CMD="v.surf.rst in=lidar_test elev=omp_test.rst --o"
+CMD="v.surf.rst in=lidar_test elev=omp_test.rst nproc=8 --o"
```
echoix, line 8: We might not be able to test this well, in CI, where we do not have 8 cores available.

vector/v.surf.rst/test_suite/profile_cv.sh
```diff
@@ -0,0 +1,21 @@
...
+g.copy vect=elev_lid792_bepts,lidar_test
+v.build lidar_test
+g.region n=221230 s=220816 w=639090 e=639530 res=0.5 -p
+CMD="v.surf.rst in=lidar_test cvdev=cv_test -c --overwrite"
+time $CMD
+
+valgrind --tool=callgrind --callgrind-out-file=$OUTFILE --trace-children=yes $CMD
+
+# http://kcachegrind.sourceforge.net
+kcachegrind $OUTFILE
```
echoix, line 14: Does this mean valgrind and kcachegrind is needed just to run the tests, in CI or others? Does it work on macOS and Windows?

vector/v.surf.rst/test_suite/profile_ompcv.sh
```diff
@@ -0,0 +1,21 @@
+# nc2008/grass7 mapset
+
+OUTFILE=callgrind.out.1196
+
+g.copy vect=elev_lid792_bepts,lidar_test
+v.build lidar_test
+g.region n=221230 s=220816 w=639090 e=639530 res=0.5 -p
+CMD="v.surf.rst in=lidar_test cvdev=omp_cv_test nprocs=20 -c --overwrite"
```
echoix, line 8: What happens if we do not have 20 cores available?

vector/v.surf.rst/test_suite/profile_ompcv.sh
```diff
@@ -0,0 +1,21 @@
...
+time $CMD
+
+valgrind --tool=callgrind --callgrind-out-file=$OUTFILE --trace-children=yes $CMD
+
+# http://kcachegrind.sourceforge.net
+kcachegrind $OUTFILE
```
echoix, line 14: Same as above

vector/v.surf.rst/test_suite/profile_ompcv.sh
```diff
@@ -0,0 +1,21 @@
...
+
+exit 0
+
+time $CMD
+=========
+
+--raster map not found?
```
echoix, line 21: Does this mean the code underneath exit 0 doesn't run at all? Otherwise, it seems like invalid syntax.

vector/v.surf.rst/test_suite/profile.sh
cyliang368: I removed the change in this file.

vector/v.surf.rst/test_suite/profile_cv.sh
cyliang368: I removed [vector/v.surf.rst/test_suite/profile_cv.sh](https://github.com/OSGeo/grass/pull/3590/files/e49e73eb93639ecda321cab92c92ff9aa88e0e3e#diff-3a590f19bacbec9faefc9dda34dbbb6ea3963487cb9ef6537184fc6dffcb87e1) and [vector/v.surf.rst/test_suite/profile_ompcv.sh](https://github.com/OSGeo/grass/pull/3590/files/e49e73eb93639ecda321cab92c92ff9aa88e0e3e#diff-731c175ba9c7b535713b4124c5d920712aacd05598da96c8c755e273fb7c7a39). The unit tests are mainly done by the Python script in vector/v.surf.rst/testsuite. I just used these files to test the speedups and efficiencies.

lib/rst/interp_float/segmen2d_parallel.c
cyliang368, line 1: I added a critical block to the other `check_points`, too. This fixes the writing conflicts when calculating deviations.

vector/v.surf.rst/test_suite/profile.sh
echoix: If it is possible to use the amount of cores available, or at least 2, it might better test a non-serial invocation

petrasovaa:
Compilation errors:

```
2024-04-15T17:35:03.7536227Z point2d.c: In function ‘IL_check_at_points_2d_cvdev’:
2024-04-15T17:35:03.7537488Z point2d.c:213:19: error: unused variable ‘mm’ [-Werror=unused-variable]
2024-04-15T17:35:03.7538241Z   213 |     int /* n1, */ mm, m;
2024-04-15T17:35:03.7538696Z       |                   ^~
2024-04-15T17:35:03.7539546Z point2d.c:210:12: error: unused variable ‘north’ [-Werror=unused-variable]
2024-04-15T17:35:03.7540477Z   210 |     double north = data->ymax;
2024-04-15T17:35:03.7541011Z       |            ^~~~~
2024-04-15T17:35:03.7541851Z point2d.c:208:12: error: unused variable ‘east’ [-Werror=unused-variable]
2024-04-15T17:35:03.7542805Z   208 |     double east = data->xmax;
2024-04-15T17:35:03.7543408Z       |            ^~~~
2024-04-15T17:35:03.7664436Z cc1: all warnings being treated as errors
```

```
2024-04-15T17:37:05.2813633Z main.c: In function ‘main’:
2024-04-15T17:37:05.2814492Z main.c:466:5: error: too few arguments to function ‘IL_init_func_2d’
2024-04-15T17:37:05.2815104Z   466 |     IL_init_func_2d(&params, IL_grid_calc_2d, IL_matrix_create,
2024-04-15T17:37:05.2815541Z       |     ^~~~~~~~~~~~~~~
2024-04-15T17:37:05.2815830Z In file included from main.c:43:
2024-04-15T17:37:05.2816586Z /home/runner/work/grass/grass/dist.x86_64-pc-linux-gnu/include/grass/interpf.h:153:6: note: declared here
2024-04-15T17:37:05.2817450Z   153 | void IL_init_func_2d(struct interp_params *, grid_calc_fn *, matrix_create_fn *,
2024-04-15T17:37:05.2817964Z       |      ^~~~~~~~~~~~~~~
2024-04-15T17:37:05.2846707Z make[4]: *** [../../include/Make/Compile.make:32: OBJ.x86_64-pc-linux-gnu/main.o] Error 1
```
```
2024-04-15T17:38:28.9886495Z main.c:660:12: error: comparison between pointer and integer [-Werror]
2024-04-15T17:38:28.9887414Z   660 |     if (cv != NULL || create_devi != NULL) {
2024-04-15T17:38:28.9888056Z       |            ^~
2024-04-15T17:38:28.9888873Z main.c:660:35: error: comparison between pointer and integer [-Werror]
2024-04-15T17:38:28.9889786Z   660 |     if (cv != NULL || create_devi != NULL) {
2024-04-15T17:38:28.9890443Z       |                                   ^~
2024-04-15T17:38:29.0160707Z cc1: all warnings being treated as errors
2024-04-15T17:38:29.0177415Z make[3]: *** [../../include/Make/Compile.make:32: OBJ.x86_64-pc-linux-gnu/main.o] Error 1
```

nilason:
@cyliang368 Good to see you working on this! May I recommend using [pre-commit](https://github.com/OSGeo/grass/blob/main/doc/development/submitting/submitting.md#use-pre-commit), which will take care of formatting and other useful checks locally.

nilason:
You may also turn on `-Wall -Wextra -Wpedantic -Wvla` compiler flags as well (those are checked on some of the CI runners), thus saving some CI time.

cyliang368:
@petrasovaa @nilason I didn't notice that warnings are treated as errors and that I should use pre-commit. Thanks for the advice!

petrasovaa:
Before this can be merged, we need a test, at least just testing the cross-validation results are the same as before (reference based on results before this change) and that they are the same for different number of cores.

cyliang368:
> Before this can be merged, we need a test, at least just testing the cross-validation results are the same as before (reference based on results before this change) and that they are the same for different number of cores. The test can be added to v.surf.rst/testsuite/test_vsurfrst.py.

@petrasovaa I added tests for parallelized cross-validation and deviation-creation. I used the results from `IL_interp_segments_2d`, which I didn't change,  to check the results from `IL_interp_segments_2d_parallel`. The results are consistent, no matter whether single thread or multi-threads is used. Please check https://github.com/OSGeo/grass/pull/3590/commits/d4b8da7122c31dda61f592a2166aecb608b0899e.

HuidaeCho:
> > I wouldn't consider a critical block parallel because it restricts execution of them to one thread at a time and performance won't be much different from serial. Have you checked speedup and efficiency? Any plots to prove performance improvement?
> > Ideally, it would be best to create a parallel version of `lib/rst/interp_float/point2d.c` just like `lib/rst/interp_float/segment2d_parallel.c`.
> 
> @HuidaeCho I split the `IL_check_at_points_2d` function into two functions. One is used for calculation, whereas the other is used to write values into the map. Now, when running cross-validation or creating deviation, the parallel for loop in `lib/rst/interp_float/segment2d_parallel.c` computes everything in parallel except for file writing.
> 
> With this implementation, we get higher speedup and efficiency (green lines in the figures below). I modified some functions, files, and the structure of loops for this implementation. Please review them. Thanks!
> 
> speedup
> 
> efficiency

Looks great! I would keep the original `point2d.c` and copy it to `point2d_parallel.c` just like `segment2d_parallel.c` for consistency and *just in case*. This PR now breaks backward compatibility: 1) adding new init functions (new arguments; `IL_write_point_2d` is new, but is it necessary?), 2) changing pass-by-value to pass-by-reference. Are these all necessary and can we avoid doing these?

The code for `IL_write_point_2d` repeats three times in `point2d.c`.

lib/rst/interp_float/segmen2d_parallel.c
```diff
@@ -292,88 +293,123 @@ int IL_interp_segments_2d_parallel(
...
+                     * region*/
+                    if (xmm >= data_local[tid]->x_orig + params->x_orig &&
+                        xmm <= data_local[tid]->xmax + params->x_orig &&
+                        ymm >= data_local[tid]->y_orig + params->y_orig &&
+                        ymm <= data_local[tid]->ymax + params->y_orig) {
+                        /* vect append, count, vect_write, db_execute will have
+                         * conflicts between threads */
+#pragma omp critical
+                        {
+                            params->wr_point(target_point, err);
```
HuidaeCho, line 407: I would also check if `params->wr_point` is not `NULL` because it's set only for cv calculation.

lib/rst/interp_float/point2d.c
```diff
@@ -193,3 +193,83 @@ int IL_check_at_points_2d(struct interp_params *params,
 
     return 1;
 }
+
```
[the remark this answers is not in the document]
cyliang368: I added documentation in the new commit.

lib/rst/interp_float/interpf.h
```diff
@@ -50,7 +50,7 @@ typedef int matrix_create_fn(struct interp_params *, struct triple *, int,
                              double **, int *);
 
 typedef int check_points_fn(struct interp_params *, struct quaddata *, double *,
-                            double *, double, double, struct triple);
+                            double *, double, double, struct triple *);
```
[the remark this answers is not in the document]
cyliang368: I checked the `check_points_fn` and `IL_check_at_points_2d` functions. They are only used in the `rst` module. The value is computed, and the results are written in the database. Replacing pass-by-value with pass-by-reference will not change the original behavior. However, in my implementation, pass-by-reference is necessary to take the result out for the parallelization.

lib/rst/interp_float/point2d.c
```diff
@@ -193,3 +193,83 @@ int IL_check_at_points_2d(struct interp_params *params,
...
+    /* take the values out*/
+    target_point->x = xmm;
+    target_point->y = ymm;
+    target_point->z = err;
+
+    return 1;
+}
+
+int IL_write_point_2d(struct triple point, double err)
```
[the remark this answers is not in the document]
cyliang368: I added documentation in the new commit. Since this function only needs to write out values instead of updating or returning them, pass-by-value is used here, as `IL_check_at_points_2d` did before this PR.

lib/rst/interp_float/init2d.c
```diff
@@ -111,7 +111,8 @@ void IL_init_func_2d(
     secpar_fn *secp_f,         /*!< calculates aspect, slope, curvature */
     interp_fn *interp_f,       /*!< radial basis function */
     interpder_fn *interpder_f, /*!< derivatives of radial basis function */
-    wr_temp_fn *temp_f         /*!< writes temp files */
+    wr_temp_fn *temp_f,        /*!< writes temp files */
+    wr_point_fn *wr_point_f    /*!< writes temp files */
```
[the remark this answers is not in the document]
cyliang368: I removed it in the new commit.

lib/rst/interp_float/segmen2d_parallel.c
cyliang368: I removed it in the new commit.

cyliang368:
> > > I wouldn't consider a critical block parallel because it restricts execution of them to one thread at a time and performance won't be much different from serial. Have you checked speedup and efficiency? Any plots to prove performance improvement?
> > > Ideally, it would be best to create a parallel version of `lib/rst/interp_float/point2d.c` just like `lib/rst/interp_float/segment2d_parallel.c`.
> > 
> > 
> > @HuidaeCho I split the `IL_check_at_points_2d` function into two functions. One is used for calculation, whereas the other is used to write values into the map. Now, when running cross-validation or creating deviation, the parallel for loop in `lib/rst/interp_float/segment2d_parallel.c` computes everything in parallel except for file writing.
> > With this implementation, we get higher speedup and efficiency (green lines in the figures below). I modified some functions, files, and the structure of loops for this implementation. Please review them. Thanks!
> > speedup
> > efficiency
> 
> Looks great! I would keep the original `point2d.c` and copy it to `point2d_parallel.c` just like `segment2d_parallel.c` for consistency and _just in case_. This PR now breaks backward compatibility: 1) adding new init functions (new arguments; `IL_write_point_2d` is new, but is it necessary?), 2) changing pass-by-value to pass-by-reference. Are these all necessary and can we avoid doing these?
> 
> The code for `IL_write_point_2d` repeats three times in `point2d.c`. Given the fact that it already repeats twice before this PR in `point2d.c`, I would say, we don't even need to add this new function and argument to `IL_init_func_2d`. Why and when would we pass another function than this code? Have you considered calling this function within the library from the three places without adding it as an argument to `IL_init_func_2d`?

Thanks for reviewing my PR. These comments gave me some ideas on how to improve it. Based on these two questions, I pushed new commits, and I wrote down my thoughts here.

### 1.  adding new init functions (new arguments; `IL_write_point_2d` is new, but is it necessary?

It is not necessary to add a new init function and argument. However, to make them clear, I had thought I shouldn't make a function do two tasks. `IL_check_at_points_2d` did two things: compute the deviations and write values to the database. Therefore, I split it into two functions.

In the new commit, I included `IL_write_point_2d` into `IL_write_point_2d_parallel` so that we don't need to change the definition and usage of `IL_init_func_2d`.

### 2.  changing pass-by-value to pass-by-reference. Are these all necessary and can we avoid doing these?

This is necessary in my implementation. This question is related to the previous question. `IL_check_at_points_2d` function actually does two things. Computing should be parallelized while writing out should be in serial. I can only figure out three ways to save the outputs: (1) make this function return values, (2) pass-by-reference and save the values by this pointer, and (3) directly write them out to the database. Since we want to parallelize the function, we cannot use (3). (1) does not seem to be preferred, so I chose (2).

This is my first time contributing to open-source code, so I am unsure whether these are preferred practices. Please review them and give me advice. Thanks a lot.

lib/rst/interp_float/ressegm2d.c
```diff
@@ -190,7 +190,7 @@ int IL_resample_interp_segments_2d(
         b[0] = 0.;
         G_lubksb(matrix, m1 + 1, indx, b);
 
-        params->check_points(params, data, b, ertot, zmin, *dnorm, triple);
+        params->check_points(params, data, b, ertot, zmin, *dnorm, &triple);
```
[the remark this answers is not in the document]
cyliang368: Please see https://github.com/OSGeo/grass/pull/3590#issuecomment-2068171992.

vector/v.surf.rst/benchmark/benchmark_v_surf_rst.py
```diff
@@ -0,0 +1,51 @@
...
+    generate_map(npoints=npoints, rows=size, cols=size, fname=reference)
+
+    module = Module(
+        "v.surf.rst",
+        input=reference,
+        npmin=100,
+        elevation=output,
+        stdout_=DEVNULL,
+        overwrite=True,
+        c=True,
```
petrasovaa, line 36: This flag probably shouldn't be here, no? It doesn't run...
```suggestion
```

cyliang368: `c` is a flag to specify running for cross-validation. Without it, it will just run interpolation. It should be kept.

vector/v.surf.rst/benchmark/benchmark_v_surf_rst_cv.py
```diff
@@ -0,0 +1,51 @@
...
+from subprocess import DEVNULL
+import math
+
+
+def main():
+    results = []
+    # Users can add more or modify existing reference maps
+    raster_cells = [1e5, 2e5, 4e5, 8e5]  # 0.1M, 0.2M, 0.4M, 0.8M cells
+    for ncells in raster_cells:
+        benchmark(math.sqrt(ncells), f"v.surf.rst_cv_{int(ncells / 1e6)}M", results)
```
petrasovaa, line 18: Just a small change needed here, the result label ends us "v.surf.rst_cv_0M", so the number needs to be formatted in a better way.

petrasovaa (changes requested):
Let's avoid all the unnecessary changes in v.surf.rst.html, no?

vector/v.surf.rst/v.surf.rst.html
```diff
@@ -339,6 +339,26 @@ <h3>Usage of the where parameter</h3>
 d.vect elevrand where="value > 94.9"
 </pre></div>
 
+<h3>Performance</h3>
+To enable parallel processing, the user can specify the number of threads to be
+used with the <b>nprocs</b> parameter (default 1). The <b>memory</b> parameter
```
petrasovaa, line 344: v.surf.rst doesn't have memory parameter

vector/v.surf.rst/v.surf.rst.html
```diff
@@ -339,6 +339,26 @@ <h3>Usage of the where parameter</h3>
...
+<div align="center" style="margin: 10px">
+  <img src="vsurfrst_benchmark.png" alt="benchmark for v.surf.rst" border="0">
+  <br>
+  <i>Figure 1 Benchmark shows execution time for different
+    number of cells.
+</div>
+<div align="center" style="margin: 10px">
+    <img src="vsurfrst_cv_benchmark.png" alt="benchmark for cross-validation of v.surf.rst" border="0">
+    <br>
+    <i>Figure 2 Benchmark shows execution time for running cross-validation on different number of cells.
```
petrasovaa, line 357:
```suggestion
  <i>Figure 1: Benchmark shows execution time for different
    number of cells.
</div>
<div align="center" style="margin: 10px">
    <img src="vsurfrst_cv_benchmark.png" alt="benchmark for cross-validation of v.surf.rst" border="0">
    <br>
    <i>Figure 2: Benchmark shows execution time for running cross-validation on different number of cells.
```

vector/v.surf.rst/v.surf.rst.html
```diff
@@ -339,6 +339,26 @@ <h3>Usage of the where parameter</h3>
 d.vect elevrand where="value > 94.9"
 </pre></div>
 
+<h3>Performance</h3>
```
petrasovaa, line 342: This section should be probably in Notes, not Examples section.

vector/v.surf.rst/v.surf.rst.html
```diff
@@ -293,6 +293,22 @@ <h3>Cross validation procedure</h3>
 with lost details and fluctuations) or when significant noise is
 present that needs to be smoothed out.
 
+<h3>Performance</h3>
+To enable parallel processing, the user can specify the number of threads to be
+used with the <b>nprocs</b> parameter (default 1). Figure 1 and 2 provides benchmarks of the parallelizations running on Intel® Core™ i5-10210U CPU @ 1.60GHz × 8. See benchmark scripts in the source code for more details.
+
```
petrasovaa, line 299:
```suggestion
used with the <b>nprocs</b> parameter (default 1). Figures 1 and 2 show benchmark results
running on Intel® Core™ i5-10210U CPU @ 1.60GHz × 8. See benchmark scripts
in the source code for more details.

```

petrasovaa:
In this case I am fine with breaking the backwards compatibility, the API is not widely used, so I don't expect any problems.

vector/v.surf.rst/v.surf.rst.html
```diff
@@ -293,6 +293,24 @@ <h3>Cross validation procedure</h3>
...
+<div align="center" style="margin: 10px">
+  <img src="vsurfrst_benchmark.png" alt="benchmark for v.surf.rst" border="0">
+  <br>
+  <i>Figure 1: Benchmark shows execution time for different
+    number of cells (1M, 2M, 4M, and 8M).
+</div>
+<div align="center" style="margin: 10px">
+    <img src="vsurfrst_cv_benchmark.png" alt="benchmark for cross-validation of v.surf.rst" border="0">
+    <br>
+    <i>Figure 2: Benchmark shows execution time for running cross-validation on different number of cells (100k, 200k, 400k, and 800k).
```
neteler, line 311: Please break long lines in the HTML file as per our style guide:

https://github.com/OSGeo/grass/blob/main/doc/development/style_guide.md#markup-style-guide

More notes on markup:
- In order to minimize potential git merge conflicts, please break a line at approximately 70-80 chars.

petrasovaa (approved):
This is ready from my side. @HuidaeCho do you want to review this?

vector/v.surf.rst/v.surf.rst.html
cyliang368: fixed in https://github.com/OSGeo/grass/pull/3590/commits/c7cbf386c97a2a60f0b7319addc54d811e5b5a16

HuidaeCho:
> (2) pass-by-reference and save the values by this pointer

@cyliang368 Does this still apply? I cannot find anywhere in `IL_check_at_points_2d()` that updates `skip_point`.
```c
            xx = points[m - 1].x - skip_point->x;           
            yy = points[m - 1].y - skip_point->y;            
...
        zz = skip_point->z + zmin;        
...
        xmm = skip_point->x * dnorm + params->x_orig + west;
        ymm = skip_point->y * dnorm + params->y_orig + south;
```
`skip_point` seems read-only now? Have things changed?

HuidaeCho:
> > (2) pass-by-reference and save the values by this pointer
> 
> @cyliang368 Does this still apply? I cannot find anywhere in `IL_check_at_points_2d()` that updates `skip_point`.
> 
> ```c
>             xx = points[m - 1].x - skip_point->x;           
>             yy = points[m - 1].y - skip_point->y;            
> ...
>         zz = skip_point->z + zmin;        
> ...
>         xmm = skip_point->x * dnorm + params->x_orig + west;
>         ymm = skip_point->y * dnorm + params->y_orig + south;
> ```
> 
> `skip_point` seems read-only now? Have things changed?

Never mind. You use it in `IL_check_at_points_2d_cvdev()`. ;-)

echoix:
I'm running a new run of CI with the last 7 days of linter upgrades made, then merge

PR #3605, opened 2024-04-16
v.in.ogr: fix escape table column name during creating table

### LINKED ISSUE #3604: [Bug] v.in.ogr: Link to database table is broken after importing a new .shp file
**Describe the bug**
Vector .shp files can't be imported with attribute table. If they are imported, database connection breaks.

**To Reproduce**

1. Go to ['...'](https://www.hydrosheds.org/products/hydrobasins) and download any standard region.
2. Try to import this vector file into GRASS with v.in.ogr without -t flag and without any settings in attribute section.
3. See error "Can't create attribute table".
2a. Try to import with -t flag - the result will be the same.
4. Now take one watershed by your choice and export it with QGIS without any columns, but as automatically it does - with FID column.
5. Download lake polygons from here https://www.hydrosheds.org/products/hydrolakes
6. Try to import lakes to GRASS with v.in.ogr telling key=Hylak_id
7. Check if lakes have attribute table created.
8. Import now your single watershed with key=FID.
9. Try to select lakes with this layer.
10. See the error "No such table HydroLAKES_polys_v10"
11. You can also check with db.tables, that no tables you have in the database. But
12. In the dbln file the link exists.

**Expected behavior**
I expected any vector to be imported without any supplemental setting with autocreation of an attribute table. Also I expected that tables links are stable, otherwise GRASS can't be used as a database.

I make all operations in the same LOCATION with docker container, then I exit and enter from Windows GUI to see results.
- Operating System: Windows 10/Linux with GRASS docker container on Debian.
- GRASS GIS version 8.4.0dev

**Additional context**
When I imported lakes and a watershed (without a key column) I tried to select lakes with the watershed in Windows GUI, and I managed. Then from docker I failed, and now from GUI I also don't see any table and can't select in the same way (because of unlinked vector).

tmszi:
Fixes #3604. Allow create DB table column name with reserved keyword.

### DIFF
--- a/vector/v.in.ogr/main.c
+++ b/vector/v.in.ogr/main.c
@@ -1145,12 +1145,12 @@ int main(int argc, char *argv[])
 
             /* Create table */
             i = 0;
-            sprintf(buf, "create table %s (%s %s", Fi->table, col_info[i].name,
-                    col_info[i].type);
+            sprintf(buf, "create table %s ('%s' %s", Fi->table,
+                    col_info[i].name, col_info[i].type);
             db_set_string(&sql, buf);
 
             for (i = 1; i < ncols_out; i++) {
-                sprintf(buf, ", %s %s", col_info[i].name, col_info[i].type);
+                sprintf(buf, ", '%s' %s", col_info[i].name, col_info[i].type);
                 db_append_string(&sql, buf);
             }
 
--- a/vector/v.in.ogr/main.c
+++ b/vector/v.in.ogr/main.c
@@ -1145,12 +1145,12 @@ int main(int argc, char *argv[])
 
             /* Create table */
             i = 0;
-            sprintf(buf, "create table %s ('%s' %s", Fi->table,
+            sprintf(buf, "create table %s (\"%s\" %s", Fi->table,
                     col_info[i].name, col_info[i].type);
             db_set_string(&sql, buf);
 
             for (i = 1; i < ncols_out; i++) {
-                sprintf(buf, ", '%s' %s", col_info[i].name, col_info[i].type);
+                sprintf(buf, ", \"%s\" %s", col_info[i].name, col_info[i].type);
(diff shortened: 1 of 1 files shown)


marisn (dismissed):

marisn:
On the other hand – shouldn't it be " (double quote) according to SQL standard instead of a single quote?

tmszi:
> On the other hand – shouldn't it be " (double quote) according to SQL standard instead of a single quote?

Yes you are right, but it is not default option across all supported DB backends:

1. DBASE/DBF  use [ ]
2. SQLite, PostgreSQL, Oracle use " "
3. MySQL use ` ` but it is possible change it to " "

marisn:
> > On the other hand – shouldn't it be " (double quote) according to SQL standard instead of a single quote?
> 
> Yes you are right, but it is not default option across all supported DB backends:
> 
>     1. DBASE/DBF  use [ ]
> 
>     2. SQLite, PostgreSQL, Oracle use " "
> 
>     3. MySQL use backtick but it is possible change it to " "

And this is the reason why we should transition to prepared statements, but that is out of scope of this PR.

Lets go with double quote as it is according to the standard. We should just switch MySQL to ANSI_QUOTES in its driver (could you create a separate PR for this?). That only leaves DBF, but string quotation is the smallest of problems when dealing with it ;-)

marisn (approved):

tmszi:
> We should just switch MySQL to ANSI_QUOTES in its driver (could you create a separate PR for this?). 

Yes, I'm going to look at it.

tmszi:
Should I backport this fix into realeasebranch_8_3 branch?

marisn:
> Should I backport this fix into realeasebranch_8_3 branch?

Yes, this looks like a fix worth backporting. Although I'm afraid there are more places where proper quotations are required. Lets do one at a time (a.k.a. better than nothing).

PR #3606, opened 2024-04-16
CI: configure Coverity Scan for larger scans and artifact

jadenabrams100:
Changed upload process to match the process for Scans for which the scan results (ie the `cov-int` directory) is over 500MB in size.

Also added a step for saving the Scan output as an artifact.

Attempt to fix Coverity Scan.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -118,7 +118,7 @@ jobs:
 
       - name: Upload the tarball to the Cloud
         run: |
-          export COV_RES_PATH=`pwd`/grass.tgz
+          export COV_RES_PATH="$(pwd)/grass.tgz"
           curl -X PUT \
             --header 'Content-Type: application/json' \
             --upload-file $COV_RES_PATH \
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -22,9 +22,10 @@ jobs:
       - name: Get dependencies
         run: |
           sudo apt-get update -y
-          sudo apt-get install -y wget git gawk findutils
+          sudo apt-get install -y wget git gawk findutils jq
           xargs -a <(awk '! /^ *(#|$)/' ".github/workflows/apt.txt") -r -- \
               sudo apt-get install -y --no-install-recommends --no-install-suggests
+              
       - name: Create installation directory
         run: |
           mkdir $HOME/install
@@ -37,6 +38,7 @@ jobs:
           tar xzf cov-analysis-linux64.tar.gz --strip 1 -C cov-analysis-linux64
         env:
           TOKEN: ${{ secrets.COVERITY_PASSPHRASE }}
+          
       - name: Set number of cores for compilation
         run: |
           echo "MAKEFLAGS=-j$(nproc)" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/coverity.yml
```diff
@@ -77,22 +80,54 @@ jobs:
...
       - name: Build with cov-build
         run: |
           pwd
           export PATH=`pwd`/cov-analysis-linux64/bin:$PATH
           cov-build --dir cov-int make
-      - name: Submit to Coverity Scan
-        run: |
+
+      - name: Put results into Tarball
+        run: |
```
nilason, line 91: Trailing whitespace
```suggestion
        run: |
```

.github/workflows/coverity.yml
```diff
@@ -77,22 +80,54 @@ jobs:
...
+          echo "BUILD_ID=$(jq -r '.build_id' response)" >> $GITHUB_ENV
+
+      - name: Upload the tarball to the Cloud
+        run: |
+          export COV_RES_PATH=`pwd`/grass.tgz
+          curl -X PUT \
+            --header 'Content-Type: application/json' \
+            --upload-file $COV_RES_PATH \
+            $UPLOAD_URL
+
```
nilason, line 126: Whitespaces
```suggestion

```

.github/workflows/coverity.yml
```diff
@@ -77,22 +80,54 @@ jobs:
...
+        uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # v4.3.1
+        with:
+          name: grass.tgz
+          path: grass.tgz
+
+      - name: Initialize Build in Coverity Cloud
+        run: |
+          curl -X POST \
+          -d version="main" \
+          -d description="`git rev-parse --abbrev-ref HEAD` `git rev-parse --short HEAD`" \
```
nilason, line 104:
```suggestion
          -d description="$(git rev-parse --abbrev-ref HEAD) $(git rev-parse --short HEAD)" \
```

.github/workflows/coverity.yml
```diff
@@ -77,22 +80,54 @@ jobs:
...
           EMAIL: ${{ secrets.COVERITY_USER }}
+
+      - name: Save Upload URL and Build ID from Initialization Response
+        run: |
+          echo "UPLOAD_URL=$(jq -r '.url' response)" >> $GITHUB_ENV
+          echo "BUILD_ID=$(jq -r '.build_id' response)" >> $GITHUB_ENV
+
+      - name: Upload the tarball to the Cloud
+        run: |
+          export COV_RES_PATH=`pwd`/grass.tgz
```
nilason, line 121:
```suggestion
          export COV_RES_PATH="$(pwd)/grass.tgz"
```

nilason (changes requested):
Minor suggestions.

jadenabrams100:
> Minor suggestions.

Thank you for the feedback! I have added the requested changes to the PR!

nilason (approved):
Let's try this out.

I have my doubts regarding the size issue (it _does_ work on my fork), but the debugging parts will probably be helpful.
We should remove the GH artefact part once we get this working.

jadenabrams100:
So, it appears that the server gave us a null build ID, and the server returned a 500 Internal Server Error when we asked it to start building. I will email the admin again about why this happened. The only thing I can think that could have caused it is the file name being in quotes.

I also downloaded the scan output and it extracted to about 95MB, so not sure if the size is an issue either.

nilason:
> So, it appears that the server gave us a null build ID, and the server returned a 500 Internal Server Error when we asked it to start building. I will email the admin again about why this happened. The only thing I can think that could have caused it is the file name being in quotes.
> 
> I also downloaded the scan output and it extracted to about 95MB, so not sure if the size is an issue either.

I’m quite sure the size is not the issue. Just started “manually” online by entering the url.

PR #3614, opened 2024-04-17
v.db.select: fix escaping SQL columns names with SQL standard double quotes

tmszi:
Fix escaping SQL columns names with SQL standard double quotes if v.db.select module `columns` param arg is used.

[Example](https://github.com/OSGeo/grass/issues/3604#issuecomment-2058805737) for testing purpose.

v.db.select module with `columns` param arg should sucessfully return columns data in the case when column name **order** is DB reserved keyword:

```
GRASS hybas/PERMANENT:~ > v.db.select hybas_lake_eu_lev01_v1c column=cat,order
cat|order
1|0
```

### DIFF
--- a/vector/v.db.select/main.c
+++ b/vector/v.db.select/main.c
@@ -323,7 +323,7 @@ int main(int argc, char **argv)
         char *buf = NULL;
 
         buf = G_malloc((strlen(options.group->answer) + 8));
-        sprintf(buf, " GROUP BY %s", options.group->answer);
+        sprintf(buf, " GROUP BY \"%s\"", options.group->answer);
         db_append_string(&sql, buf);
         G_free(buf);
     }
--- a/vector/v.db.select/main.c
+++ b/vector/v.db.select/main.c
@@ -289,17 +289,19 @@ int main(int argc, char **argv)
 
     if (options.cols->answer) {
         char col_sep[] = ",";
-        char *token = strtok(options.cols->answer, col_sep);
+        char **tokens;
+        int i, ntokens;
 
         sprintf(query, "SELECT ");
-
-        while (token != NULL) {
-            sprintf(query + strlen(query), "\"%s\"", token);
-            token = strtok(NULL, col_sep);
-            if (token)
-                sprintf(query + strlen(query), "%c", *col_sep);
+        tokens = G_tokenize(options.cols->answer, col_sep);
+        ntokens = G_number_of_tokens(tokens);
+        for (i = 0; i < ntokens; i++) {
+            if (i < ntokens - 1)
+                sprintf(query + strlen(query), "\"%s\"%c", tokens[i], *col_sep);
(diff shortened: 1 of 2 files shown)


vector/v.db.select/main.c
```diff
@@ -287,8 +287,21 @@ int main(int argc, char **argv)
                       Fi->database, Fi->driver);
     db_set_error_handler_driver(driver);
 
-    if (options.cols->answer)
-        sprintf(query, "SELECT %s FROM ", options.cols->answer);
+    if (options.cols->answer) {
+        char col_sep[] = ",";
+        char *token = strtok(options.cols->answer, col_sep);
```
wenzeslaus, line 292: Any opinions on using strtok versus [G_tokenize](https://grass.osgeo.org/programming8/defs_2gis_8h.html#af6abbab981ffd86ea461668f6f5c8310)?

vector/v.db.select/main.c
```diff
@@ -287,8 +287,21 @@ int main(int argc, char **argv)
...
+
+        sprintf(query, "SELECT ");
+
+        while (token != NULL) {
+            sprintf(query + strlen(query), "\"%s\"", token);
+            token = strtok(NULL, col_sep);
```
wenzeslaus, line 298: Passing NULL is not clear to me here.

wenzeslaus (changes requested):
This is a issue in couple places and I'm happy to see this addressed at least here. On the other hand, I'm not sure how the different systems handle quoting. It seems that while standard is a double quote, what is enabled by default and what double quote actually means may differ. More analysis may be required.

The previous implementation was passing the values as is, so one could pass the quoting characters and it would work. This will not work with the current code in this PR.

A test would be really good.

marisn:
> On the other hand, I'm not sure how the different systems handle quoting. It seems that while standard is a double quote, what is enabled by default and what double quote actually means may differ. More analysis may be required.

Most of systems GRASS supports are SQL standards compliant. MySQL compliance is addressed in #3612. This PR also needs an update to documentation – mention that quoting column parameter is not necessary any more.

At the moment only DBF driver is broken – it needs some magick in sqlp.l file (any lex experts reading this?)

vector/v.db.select/main.c
tmszi: Fixed with `G_tokenize()` func.

tmszi:
> A test would be really good.

Test has been added.

tmszi:
With this changes another v.dissolve module pytests fails. According my analysis, with automatically enclose column names with double quotes when v.db.select module param `columns` and `group` is used, it will require much more code changes (handle SQL func, module JSON output format) and increase complexity of code.

If v.db.select require to use `columns` or `group` param with column name which is DB backend reserved keyword it will be much easier enclose this column with double quotes on the param arg level as before.

PR #3617, opened 2024-04-17
libvector/diglib: fix memory leaks

nilason:
Fix some memory leaks in `lib/vector/diglib` exposed by static analysis.

### DIFF
--- a/lib/vector/diglib/plus_area.c
+++ b/lib/vector/diglib/plus_area.c
@@ -208,8 +208,10 @@ int dig_add_area(struct Plus_head *plus, int n_lines, plus_t *lines,
     if (Area == NULL)
         return -1;
 
-    if (dig_area_alloc_line(Area, n_lines) == -1)
+    if (dig_area_alloc_line(Area, n_lines) == -1) {
+        dig_free_area(Area);
         return -1;
+    }
 
     for (i = 0; i < n_lines; i++) {
         line = lines[i];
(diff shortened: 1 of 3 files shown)


metzm (approved):
I don't like `goto` in C code, but the usage here seems ok:

Have you observed actual memory leaks that are fixed by this PR?

marisn:
> I don't like `goto` in C code, but the usage here seems ok:

This code is a correct reason why `goto` should exist and should be used.

wenzeslaus:
> This code is a correct reason why `goto` should exist and should be used.

I reluctantly agree. It seems that `goto` was overused and the legitimate usages disappeared from our minds. I did a quick search earlier and internet thinks usage like this is okay (e.g. for Linux kernel if I recall correctly).

echoix (approved):
I was already pretty sure of this, but with the discussion here, my last doubts are addressed.

nilason:
> I don't like `goto` in C code, but the usage here seems ok:
> 
> Have you observed actual memory leaks that are fixed by this PR?

This is as far as I personally would go to use goto: simple memory clean up operations. It has the advantage to make sure everything get released, without repeating code or creating additional clean-up functions (for every function).

nilason:
> Have you observed actual memory leaks that are fixed by this PR?

The leaks are reported by Coverity. The issues were fixed according to the testing of my fork against Coverity.
While GRASS has a liberal attitude to leaks in modules due to their short-lived runs, I specifically aim to address those reported in the libraries.

PR #3622, opened 2024-04-18
wxGUI/history: Add time period branching to history browser tree

lindakladivova:
This PR adds a new node level to the history browser tree which is the part of the new History tab dealing with command history management.

It adds five nodes which are then populated by commands:
Today, Yesterday, This week, Older than week and Missing info.

The hierarchy is as follows:
Screenshot from 2024-04-18 12-04-29

Missing info is a special node used for situations when history log is in the plain text format (in this case we do not have any info when the command was run).

### DIFF
--- a/gui/wxpython/core/treemodel.py
+++ b/gui/wxpython/core/treemodel.py
@@ -239,6 +239,50 @@ class DictNode:
         return False
 
 
+class DictFilterNode(DictNode):
+    """Node which has data in a form of dictionary and can be filtered."""
+
+    def __init__(self, data=None):
+        super().__init__(data=data)
+
+    def match(self, method="exact", **kwargs):
+        """Method used for searching according to given parameters.
+
+        :param method: 'exact' for exact match or 'filtering' for filtering by type/name
+        :param kwargs key-value to be matched, filtering method uses 'type' and 'name'
+        """
+        if not kwargs:
+            return False
+
+        if method == "exact":
+            return self._match_exact(**kwargs)
+        elif method == "filtering":
+            return self._match_filtering(**kwargs)
+
+    def _match_exact(self, **kwargs):
+        """Match method for exact matching."""
+        for key, value in kwargs.items():
+            if not (key in self.data and self.data[key] == value):
+                return False
+        return True
+
+    def _match_filtering(self, **kwargs):
+        """Match method for filtering."""
+        if (
+            "type" in kwargs
+            and "type" in self.data
+            and kwargs["type"] != self.data["type"]
+        ):
+            return False
+        if (
+            "name" in kwargs
+            and "name" in self.data
+            and not kwargs["name"].search(self.data["name"])
+        ):
+            return False
+        return True
+
+
 class ModuleNode(DictNode):
     """Node representing module."""
 
--- a/gui/wxpython/core/treemodel.py
+++ b/gui/wxpython/core/treemodel.py
@@ -248,8 +248,10 @@ class DictFilterNode(DictNode):
     def match(self, method="exact", **kwargs):
         """Method used for searching according to given parameters.
 
-        :param method: 'exact' for exact match or 'filtering' for filtering by type/name
+        :param str method: 'exact' for exact match or
+                           'filtering' for filtering by type/name
         :param kwargs key-value to be matched, filtering method uses 'type' and 'name'
+        :return bool: True if an entry matching given parameters was found
         """
         if not kwargs:
             return False
--- a/gui/wxpython/history/tree.py
+++ b/gui/wxpython/history/tree.py
@@ -32,7 +32,7 @@ from core.utils import (
     split,
 )
 from gui_core.forms import GUI
-from core.treemodel import TreeModel, DictNode
+from core.treemodel import TreeModel, DictFilterNode
 from gui_core.treeview import CTreeView
 from gui_core.wrap import Menu
 
@@ -41,7 +41,7 @@ from grass.pydispatch.signal import Signal
 from grass.grassdb import history
 
 
-class HistoryBrowserNode(DictNode):
+class HistoryBrowserNode(DictFilterNode):
     """Node representing item in history browser."""
 
     def __init__(self, data=None):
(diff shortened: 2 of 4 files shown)


gui/wxpython/core/treemodel.py
```diff
@@ -239,6 +239,50 @@ def match(self, key, value):
...
+class DictFilterNode(DictNode):
+    """Node which has data in a form of dictionary and can be filtered."""
+
+    def __init__(self, data=None):
+        super().__init__(data=data)
+
+    def match(self, method="exact", **kwargs):
+        """Method used for searching according to given parameters.
+
+        :param method: 'exact' for exact match or 'filtering' for filtering by type/name
```
tmszi, line 251: missing param type `:param str method` and return value/type

petrasovaa:
I like the idea of using today, yesterday etc, but I would still group them by day, so I wouldn't use 'older than week', because then everything will eventually end up there. Maybe something like 'April 22 (today)', 'April 21 (yesterday)', 'April 20 (this week)'  April 5 (this month)'  and use years of it's not this year: 'December 21, 2023'. Look at how Github shows commits (see commits tab in a PR), but this can get complicated, so maybe a simpler solution would be better.

petrasovaa:
> > I like the idea of using today, yesterday etc, but I would still group them by day, so I wouldn't use 'older than week', because then everything will eventually end up there. Maybe something like 'April 22 (today)', 'April 21 (yesterday)', 'April 20 (this week)' April 5 (this month)' and use years of it's not this year: 'December 21, 2023'. Look at how Github shows commits (see commits tab in a PR), but this can get complicated, so maybe a simpler solution would be better.
> 
> I can try to do it.. First, I did it similarly as you suggested - particularly Today, Yesterday, April 2024, March 2024, February 2024 etc. But your solution is more systematic as the grouping is always by the day. The reason why your and the original suggestion are more complex is that the node number is not fixed. Then there is a bit problem with sorting time nodes because the Today node has to be at the very top but at the same time, the Missing Info node has to be at the very bottom. When the number of nodes is fixed - it is easy, we can assign the number and sort them according to their numbers but when not, then it should be sorted based on time. Unfortunately, the Missing info node does not have any time info. So, the easiest thing how to solve it that came to my mind is to artificially assign some time from deep history to it.. I will try to implement it and we can then edit it.

I think using an old date is a good approach.

lindakladivova:
> > > I like the idea of using today, yesterday etc, but I would still group them by day, so I wouldn't use 'older than week', because then everything will eventually end up there. Maybe something like 'April 22 (today)', 'April 21 (yesterday)', 'April 20 (this week)' April 5 (this month)' and use years of it's not this year: 'December 21, 2023'. Look at how Github shows commits (see commits tab in a PR), but this can get complicated, so maybe a simpler solution would be better.
> > 
> > 
> > I can try to do it.. First, I did it similarly as you suggested - particularly Today, Yesterday, April 2024, March 2024, February 2024 etc. But your solution is more systematic as the grouping is always by the day. The reason why your and the original suggestion are more complex is that the node number is not fixed. Then there is a bit problem with sorting time nodes because the Today node has to be at the very top but at the same time, the Missing Info node has to be at the very bottom. When the number of nodes is fixed - it is easy, we can assign the number and sort them according to their numbers but when not, then it should be sorted based on time. Unfortunately, the Missing info node does not have any time info. So, the easiest thing how to solve it that came to my mind is to artificially assign some time from deep history to it.. I will try to implement it and we can then edit it.
> 
> I think using an old date is a good approach.

I implemented it using day grouping according to your suggestion. I think it looks good. (This week, Today, Yesterday) seems to me very useful for quick orientation, so I would keep it. Maybe I would remove (this month) label as it is clear that the month is the current one?

lindakladivova:
> > > > I like the idea of using today, yesterday etc, but I would still group them by day, so I wouldn't use 'older than week', because then everything will eventually end up there. Maybe something like 'April 22 (today)', 'April 21 (yesterday)', 'April 20 (this week)' April 5 (this month)' and use years of it's not this year: 'December 21, 2023'. Look at how Github shows commits (see commits tab in a PR), but this can get complicated, so maybe a simpler solution would be better.
> > > 
> > > 
> > > I can try to do it.. First, I did it similarly as you suggested - particularly Today, Yesterday, April 2024, March 2024, February 2024 etc. But your solution is more systematic as the grouping is always by the day. The reason why your and the original suggestion are more complex is that the node number is not fixed. Then there is a bit problem with sorting time nodes because the Today node has to be at the very top but at the same time, the Missing Info node has to be at the very bottom. When the number of nodes is fixed - it is easy, we can assign the number and sort them according to their numbers but when not, then it should be sorted based on time. Unfortunately, the Missing info node does not have any time info. So, the easiest thing how to solve it that came to my mind is to artificially assign some time from deep history to it.. I will try to implement it and we can then edit it.
> > 
> > 
> > I think using an old date is a good approach.
> 
> I implemented it using day grouping according to your suggestion. I think it looks good. (This week, Today, Yesterday) seems to me very useful for quick orientation, so I would keep it. Maybe I would remove (this month) label as it is clear that the month is the current one?

Screenshot from 2024-04-25 12-51-31

gui/wxpython/history/tree.py
```diff
@@ -40,6 +41,46 @@
...
+
+        if day == OLD_DATE:
+            return _("No time info")
+
+        if day == current_date:
+            return "{:%B %-d} (today)".format(day)
+        elif day == current_date - datetime.timedelta(days=1):
+            return "{:%B %-d} (yesterday)".format(day)
+        elif day >= current_date - datetime.timedelta(days=7):
+            return "{:%B %-d} (this week)".format(day)
```
petrasovaa, line 77: I think this week should refer to the Mon-Sun week, not past 7 days. Looks like iso standard takes Monday as the first day of the week (not Sunday as in the US).

petrasovaa: One other suggestion is to not show the year for current year to improve readability, but feel free to ignore this.

lindakladivova: > One other suggestion is to not show the year for current year to improve readability, but feel free to ignore this.

Yes, it is better not to show the current year. Btw I also removed the (this month) label as I suggested above - in this case the browser simply shows just the format 'December 21'.
Screenshot from 2024-05-02 09-19-57

lindakladivova: I think this PR is ready for the final review :-).

petrasovaa (approved):

petrasovaa:
I realized the ordering of commands - most recent first - would be probably better, since the day nodes are ordered like that as well, what do you think @lindakladivova ?

lindakladivova:
> I realized the ordering of commands - most recent first - would be probably better, since the day nodes are ordered like that as well, what do you think @lindakladivova ?

Yes, seems better. I will create PR for after merging open ones.

PR #3624, opened 2024-04-18
CI: reactivate Coverity workflow cron, remove artifact upload

nilason:
Now with Coverity Scan runner is working, we can reactivate the cron and remove the artifact upload part.


wenzeslaus (approved):

jadenabrams100:
Glad to see that the workflow is running now! What had to be changed to fix it?

nilason:
> Glad to see that the workflow is running now! What had to be changed to fix it?

Some considerable thinning out of my hair density and altering the COVERITY_USER, the reason for the need of the latter is unknown.

PR #3626, opened 2024-04-18
CI: improve Coverity build log

nilason:
This changes the runner log of Coverity, instead of printing URL encoded version string it now prints a human readable string.

before:

`Submitting Version%3A8.4.0dev%2C%20commit%3bb27c05%2C%20branch%3Amain.`

now:

`Submitting Version:8.4.0dev, commit:bb27c05, branch:main.`


wenzeslaus (approved):
I would say crude, but not worse what was there before.

PR #3637, opened 2024-04-21
db/drivers/odbc: catch and handle return value from setting MySQL SQL ANSI mode

tmszi:
Improving PR #3635.

### DIFF
--- a/db/drivers/odbc/db.c
+++ b/db/drivers/odbc/db.c
@@ -46,7 +46,11 @@ int db__driver_open_database(dbHandle *handle)
 
     if (strcmp((CHAR *)dbms_name, "MySQL") == 0 ||
         strcmp((CHAR *)dbms_name, "MariaDB") == 0) {
+        dbString sql;
         cursor *c;
+
+        db_init_string(&sql);
+        db_set_string(&sql, "SET SQL_MODE=ANSI_QUOTES;");
 
         c = alloc_cursor();
         if (c == NULL)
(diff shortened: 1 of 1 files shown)


db/drivers/odbc/db.c
```diff
@@ -46,20 +46,32 @@ int db__driver_open_database(dbHandle *handle)
 
     if (strcmp((CHAR *)dbms_name, "MySQL") == 0 ||
         strcmp((CHAR *)dbms_name, "MariaDB") == 0) {
+        dbString sql;
         cursor *c;
 
+        db_init_string(&sql);
+        db_set_string(&sql, "SET SQL_MODE=ANSI_QUOTES;");
```
marisn, line 53: Although most likely the rest of code is not consistent, it would be nice to call `db_free_string(&sql)` afterwards at certain points (before `return DB_FAILED`, before the end of block).

marisn (changes requested):

db/drivers/odbc/db.c
nilason: I was just about to suggest the same (`db_set_string()` allocates new string).
In addition, in case setting sql_mode fails, the database connection should be closed (see `db__driver_close_database()`).

nilason (changes requested):

db/drivers/odbc/db.c
tmszi: Fixed with commit https://github.com/OSGeo/grass/pull/3637/commits/36e68dc8b702a8c14680043eccc9780482d3fd72 and https://github.com/OSGeo/grass/pull/3637/commits/15cb4150e512e7e892d3e4a06b918c73649b96b8.

nilason (approved):

marisn (approved):

PR #3649, opened 2024-04-23
g.mkfontcap: Replace -o with --overwrite; Add a renamed flag warning

HuidaeCho:
This PR replaces `-o` with standard `--overwrite` and adds a warning.

### DIFF
--- a/binaryInstall.src
+++ b/binaryInstall.src
@@ -319,7 +319,7 @@ echo""
 
 echo "Generating display font configuration file..."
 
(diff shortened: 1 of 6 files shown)


petrasovaa (commented):
There seems to be the -o flag in several places:

./mswindows/osgeo4w/postinstall.bat
./mswindows/GRASS-Installer.nsi.tmpl
./lib/driver/font2.c
./binaryInstall.src

This should be probably replaced with --o, no?

HuidaeCho:
> There seems to be the -o flag in several places:
> 
> ./mswindows/osgeo4w/postinstall.bat ./mswindows/GRASS-Installer.nsi.tmpl ./lib/driver/font2.c ./binaryInstall.src
> 
> This should be probably replaced with --o, no?

Yes, better to do it now even though it's backward compatible.

echoix (approved):
Assuming there aren't missing places that also need this change, I agree with this!

PR #3657, opened 2024-04-23
Update bug report template

veroandreo:
Comment section descriptions so they are not part of the bug report, as it usually happens.


echoix (approved):
Maybe you'd be interested in creating issue forms too: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

PR #3708, opened 2024-05-14
r.flow: rename io.h file which causes conflict with MSYS2

nilason:
Rename `r.flow` file `io.h`, which causes conflict with MSYS2.

Closes: https://github.com/OSGeo/grass/issues/3700


echoix (approved):
Way better solution! I knew something was wrong and I didn't really believe that it was only a missing include after trying to create the error in Linux. I didn't think of that avenue yet.

PR #3714, opened 2024-05-16
wxGUI/history: update cross to green check after updating to current region

lindakladivova:
Fixes https://github.com/OSGeo/grass/issues/3713.
(besides that it also fixes the problem from here https://github.com/OSGeo/grass/pull/3712 and makes some mini code improvements).


petrasovaa (approved):
Seems to work, thank you!

PR #3718, opened 2024-05-18
CI: Print Python version first

wenzeslaus:
* Print first version of Python.
* Print also Git version which we use during build (when available and it is available in the CI).
* Print GRASS versions last because they actually need the build to be successful and this purposefully runs even for failures.


echoix (approved):
Even if these info exists in the CI logs already way up before, it's not a bad thing to bump it up earlier in that script. It's our tool, and what's the point of having this smoke test if we can't understand what is failing?

So, it's not a big deal here

PR #3727, opened 2024-05-21
d.redraw: Rewritten in C

HuidaeCho:
This PR rewrites the `d.redraw` Python script in C.

This PR replaces #3484. For details, see there.


neteler (approved):
I have tried  #3727 and #3500 together. The updated monitor seems to behave as expected (the test case above runs as expected for me).

PR #3752, opened 2024-05-30
DataCatalog: different message for moving and copying

### LINKED ISSUE #1095: [Bug] OnPasteMap in Data Catalog has likely wrong message
**Describe the bug**

The code in tree.py says:

```
                    if self.copy_mode:
                        self.showNotification.emit(message=_("g.copy completed"))
                    else:
                        self.showNotification.emit(message=_("g.copy completed"))
```

which is clearly wrong.

**Expected behavior**

Different messages for different cases.

**Additional context**

Identified when reviewing #994.

ninsbl:
Minor fix changing the messages for copying and moving, when pasting DataCatalog elements.

Fix #1095

### DIFF
--- a/gui/wxpython/datacatalog/tree.py
+++ b/gui/wxpython/datacatalog/tree.py
@@ -1339,9 +1339,7 @@ class DataCatalogTree(TreeView):
                 if pasted == 0:
                     Debug.msg(1, "COPIED TO: " + new_name)
                     if self.copy_mode:
-                        self.showNotification.emit(message=_("g.copy completed"))
-                    else:
-                        self.showNotification.emit(message=_("g.copy completed"))
+                        self.showNotification.emit(message=_("Copying completed"))
                     self._giface.grassdbChanged.emit(
                         grassdb=self.selected_grassdb[0].data["name"],
                         location=self.selected_location[0].data["name"],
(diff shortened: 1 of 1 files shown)


gui/wxpython/datacatalog/tree.py
```diff
@@ -1355,6 +1353,7 @@ def OnPasteMap(self, event):
                         self._removeMapAfterCopy(
                             self.copy_layer[i], self.copy_mapset[i], env2
                         )
+                        self.showNotification.emit(message=_("Moving completed"))
```
lindakarlovska, line 1356: Messages make sense, maybe just a small code suggestion - I would put    `else: self.showNotification.emit(message=_("Copying completed"))` below this and remove rows 1341-1344? 
We could also make the condition positive (instead of if not self.copy_mode:, use if self.copy_mode?)

wenzeslaus: +1 for some more shuffling here. `if self.copy_mode:` followed by `if not self.copy_mode:` is hard to understand.

ninsbl: You mean like in 
f2c7eff 
?

lindakarlovska (approved):

gui/wxpython/datacatalog/tree.py
lindakarlovska: Great :-)

Issue #3750, opened 2024-05-30
[Feat] i.group: print image group content with semantic_labels in JSON


ninsbl:
**Is your feature request related to a problem? Please describe.**
The concept of semantic labels got introduced with the temporal framework, were it`s original idea was to flag bands in e.g. satellite imagery. However, this concept can be very handy when handling all kinds of imagery data.

In order to make better use of it in the imagery group context, it would be great if `i.group` could print not only the maps in a group, but also their semantic labels. In scripts I use a python function to get the the semantic labels of maps in a group (see below), and I was thinking about adding it to the python library, However, looking at i.group, I think it would be more appropriate/convenient to extend the printing functionality of the i.group module.

**Describe the solution you'd like**
i.group allows printing of files in (sub) groups (_-l_) as well as subgroups (_-s_) in verbose and "shell script" style (`-g`), which actually is rather a plain list.
What I would like to see is:
a) either a new flag (e.g. _-S_) to request printing of semantic labels of the maps in the group or a _print_ option (instead of the _-l_ and _-s_ flags) and 
b) a  _G_OPT_F_FORMAT_ option to request different output formats (instead of the _-g_ flag).
 
Ideally, the current flags for printing get deprecated (but kept functional (with a warning to the user) until rempoval) in favor of a combination of _print_ and _format_ option

So, I would like to be able to do something like this:
```
# prepare input
import grass.script as gs
band_numbers = [1, 2, 3]
for band_number in band_numbers:
    gs.run_command("g.copy", raster=f"lsat7_2000_{band_number}0,lsat7_2000_{band_number}0")
    gs.run_command("r.support", raster="lsat7_2000_{band_number}0", semantic_label="L8_{band_number}")
gs.run_command("i.group", group="L8_group", input=",".join([f"lsat7_2000_{band_number}0" for band_number in band_numbers]))
```
And then
```
# Print group info (labels as keys, maps as values)
gs.run_command("i.group", group="L8_group", print="labels_and_maps", format="json")
```
Which should return something along the following lines:
```
{
    "L8_1": "lsat7_2000_10@user",
    "L8_2": "lsat7_2000_20@user",
    "L8_3": "lsat7_2000_30@user"
}
```
or
```
# Print group info (maps as keys, labels as values)
gs.run_command("i.group", group="L8_group", print="maps_and_labels", format="json")
```
Which should return something along the following lines:
```
{  
  "lsat7_2000_10@user", "L8_1",
[...]

PR #3756, opened 2024-05-31
library: Add imagery Python library module to grass.script

ninsbl:
This PR adds a new imagery module to grass.script. For the time being it has just one function, that returns a dictionary for an imagery group with the contained maps and metadata.

It addresses #3750


python/grass/script/imagery.py
```diff
@@ -0,0 +1,110 @@
...
+    env=None,
+):
+    """Create a dictionary to represent an imagery group, with it`s raster maps
+    and their associated semantic_labels. The dict_key option allows to choose
+    if the semantic_labels should be the keys in the dictionary (the default)
+    or the map IDs (full map names). The function can also operate on the level
+    of subgroups.
+    For raster maps in the imagery (sub-)group that do not have a semantic label
+    a warning is given.
+    If the imagery group is not found, an error message is printed and an empty
```
wenzeslaus, line 41: If these are supposed to be paragraphs, they need an empty line I think.

python/grass/script/imagery.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def group_to_dict(
+    imagery_group_name,
+    dict_key="semantic_label",
+    full_info=True,
+    subgroup=None,
+    env=None,
+):
+    """Create a dictionary to represent an imagery group, with it`s raster maps
```
wenzeslaus, line 34: We usually try to have a brief one-liner description first before the long doc.

python/grass/script/imagery.py
```diff
@@ -0,0 +1,110 @@
...
+    >>> group_to_dict("L8_group", full_info=False)  # doctest: +ELLIPSIS
+    {"L8_1": "lsat7_2000_10", ... "L8_3": "lsat7_2000_30"}
+    >>> run_command("g.remove", flags="f", type="group", name="L8_group")
+    0
+    >>> run_command("g.remove", flags="f", type="raster",
+    >>>             name="lsat7_2000_10,lsat7_2000_20,lsat7_2000_30")
+    0
+
+    :param str table: imagery_group_name: Name of the imagery group to process (or None)
+    :param str dict_key: What to use as key for dictionary "semantic_labels" or map ids
```
wenzeslaus, line 71: It is unclear what should be the other value.

wenzeslaus (changes requested):

python/grass/script/imagery.py
```diff
@@ -89,22 +85,37 @@ def group_to_dict(
             .strip()
             .split()
         )
-    except CalledModuleError:
-        return group_dict
+    except CalledModuleError as cme:
+        raise cme
```
wenzeslaus, line 98: The trend is to avoid names which are not words, i.e., no `strchr` :-)

```suggestion
    except CalledModuleError as error:
        raise error
```

python/grass/script/imagery.py
```diff
@@ -89,22 +85,37 @@ def group_to_dict(
...
+        elif dict_key == "semantic_labels":
+            key = raster_map_info["semantic_label"]
+            if not key or key == '"none"':
+                warning(
+                    _(
+                        "Raster map {m} in group <{g}> does not have a semantic label."
+                    ).format(m=raster_map, g=imagery_group_name)
+                )
+                if fill_semantic_label:
+                    key = str(idx + 1)
```
wenzeslaus, line 136: This is not what I meant. No warning should be issued when filling in is requested in the parameters.

I'm actually confused with this change because comparing to the previous code, it is filling only when set, what it is doing in the other case? My point was fill it silently when requested to do so. Fail when missing but filling is not requested.

python/grass/script/imagery.py
```diff
@@ -26,49 +26,45 @@
...
-    or the map IDs (full map names). The function can also operate on the level
-    of subgroups.
-    For raster maps in the imagery (sub-)group that do not have a semantic label
-    a warning is given.
-    If the imagery group is not found, an error message is printed and an empty
-    dictionary is returned. In case of non-existing subgroups no error message
-    is printed and an empty dictionary is returned.
+    """Create a dictionary to represent an imagery group with metadata.
+
+    Defined by the dict_key option, the dictionary uses either the names
```
wenzeslaus:
```suggestion
    Defined by the dict_key option, the returned dictionary uses either the names
```

python/grass/script/testsuite/test_imagery.py
```diff
@@ -0,0 +1,144 @@
...
+            self.group,
+            **{"dict_values": "invalid_dict_value"},
+        )
+        group_info = gs.imagery.group_to_dict(
+            self.group, subgroup="non_existing_subgroup"
+        )
+
+        # Check that an empty dict is returned
+        self.assertIsInstance(group_info, dict)
+        self.assertDictEqual(group_info, {})
```
echoix, line 68: My experience in other languages (not Python) is that these should be split as 5+ tests. It will really help pinpoint what is wrong when something fails (instead of 5+ possibilities). You already have it separated, and there isn't too much setup/parameters too, so I don't see why not.

echoix: One style is called AAA: Arrange, Act, and Assert. So one setup, one action, and the assert (some purists may say a unit test should only have one assert, but I don't think it's always applicable).

echoix (changes requested):

python/grass/script/testsuite/test_imagery.py
ninsbl: Thanks @echoix Good points. Now I did adjust the tests in [9e7923c](https://github.com/OSGeo/grass/pull/3756/commits/9e7923c9fc0213d5ca3259f5192643bdfaff469e).

python/grass/script/imagery.py
ninsbl: Addressed in [85d36a0](https://github.com/OSGeo/grass/pull/3756/commits/85d36a0c83efe0eea2d386b8dd68c970fd326f45)

python/grass/script/imagery.py
ninsbl: Addressed in [85d36a0](https://github.com/OSGeo/grass/pull/3756/commits/85d36a0c83efe0eea2d386b8dd68c970fd326f45)

python/grass/script/imagery.py
```diff
@@ -0,0 +1,155 @@
...
+                subgroup=subgroup,
+                flags="g",
+                quiet=True,
+                env=env,
+            )
+            .strip()
+            .split()
+        )
+    except CalledModuleError as error:
+        raise error
```
wenzeslaus, line 98: If you are not doing anything with the error, you can just not use try-except. This would be good to raise a different type of exception with raise-from.

ninsbl: So, you mean dropping try-except and just running the module and rely on the error message there?

ninsbl: I removed the ty-except now...

echoix:
One last question about this one: does it wire up everything in order for the sphinx docs to be generated correctly for it? For the makefiles it is probably enough, I didn't search more.

python/grass/script/imagery.py
wenzeslaus: Removing is fine with me. Just to be clear, if you do catch in except and raise you do almost the same traceback as if there is no raise. except + raise would make sense if you would be adding more info and/or change the type of the exception.

wenzeslaus (dismissed):

python/grass/script/imagery.py
```diff
@@ -0,0 +1,150 @@
+"""
+Imagery related functions to be used in Python scripts.
+
+Usage:
+
+::
+
+    import grass.script as gs
+
+    gs.imagery.group_to_dict(imagery_group)
```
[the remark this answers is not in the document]
echoix: We should tend to avoid adding new star imports

python/grass/script/__init__.py
```diff
@@ -3,6 +3,7 @@
 
 from .core import *
 from .db import *
+from .imagery import *
```
echoix, line 6: It's not needed if you want to dot access. This only "copies" (makes available) all from the imagery module (the `__init__.py`) inside of this namespace. You also need to make sure to avoid introducing new circular imports, as Pylint 3.x complains already at multiple places (before this PR)

python/grass/script/imagery.py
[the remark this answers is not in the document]
wenzeslaus: > We should tend to avoid adding new star imports

In general, of course, but here in `__init__.py` and next to `from .raster import *`?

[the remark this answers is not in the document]
ninsbl: The reason I included the star import is do have it consistent with the other modules...

And with that: `gs.imagery.group_to_dict(imagery_group)` does work.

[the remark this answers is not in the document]
ninsbl: That said, `gs.group_to_dict(imagery_group)` works too. I do what you prefer regarding the star-import. Not being able to do `gs.group_to_dict(imagery_group)` without the star-import would be fine with me...
Let me know what to do to merge it...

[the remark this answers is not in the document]
echoix: We have the possibility to do it right before making it public.

I started exploring the time taken for importing modules, with `-X importtime` https://docs.python.org/3/using/cmdline.html#cmdoption-X
Apart from having lib/db python wrapper being extremely slow, grass.script is already big (I'll try to dig it back soon).

So, does group_to_dict() have any sense to be used on anything else of imagery?
If we wish to have some functions available at the grass.script scope (polluting the scope for every import importing grass.script), we can select one by one the functions that we declare public, and that are "part of" grass.script. 

Doing imports in `__init__.py` isn't required have the function importable in a Python script. It only adds an shortcut of that function/constant/object when importing the package. Using the full name (dotted) as an import is always possible, if it is a package (the folder contains an `__init__.py`)

python/grass/script/__init__.py
ninsbl: I removed the line again by reverting the commit...

python/grass/script/imagery.py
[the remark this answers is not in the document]
ninsbl: Star import of imagery module has been removed.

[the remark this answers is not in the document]
ninsbl: Hm... tests fail without the import of `imagery` in `__init__.py`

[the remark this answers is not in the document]
echoix: Isn't it similar to https://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-submodules-are-added-to-the-package-namespace-trap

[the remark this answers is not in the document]
echoix: Or https://stackoverflow.com/a/40823467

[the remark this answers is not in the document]
echoix: Of course the long way, import grass.script.imagery will work.
if you add it, then gs.imagery should work too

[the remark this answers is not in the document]
wenzeslaus: I'm confused. Why it works right now? I thought `from . import imagery` is needed for `gs.imagery.group_to_dict` to work. I think I had to do it that way for `from . import setup` and `gs.setup.init`.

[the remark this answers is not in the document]
ninsbl: The imagery module is now added to __init__.py like the other modules in _python/grass/script_

ninsbl:
@wenzeslaus would you mind having a look at this if your requested changes are properly addressed and approve or comment?

echoix (approved):

PR #3757, opened 2024-06-01
testsuite: test_pansharpen

manan-T:
test case file for image fusion b/w multispectral & high-res pan image

### DIFF
--- a/testsuite/test_pansharpen.py
+++ b/testsuite/test_pansharpen.py
@@ -7,16 +7,29 @@ class TestPansharpeningAlgorithms(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


wenzeslaus (changes requested):
I like the overall structure, but the test does not run in the current form. The file needs to be in `testsuite` not `testsuite/tests` for the CI to  pick it up.

We recently added a [Style Guide](https://github.com/OSGeo/grass/blob/main/doc/development/style_guide.md) which has instructions on using pre-commit which is not required but may reduce the number of comments you get from the reviewdog bot who is watching our code for the style.

veroandreo:
Closing, see https://github.com/OSGeo/grass/pull/3764#issuecomment-2168683907

Issue #3655, opened 2024-04-23
[Bug] r.proj from EPSG:3031 to EPSG:4326 returns one grid cell


mankoff:
**Describe the bug**

I'm trying to reproject an Antarctic dataset on EPSG:3031 to EPSG:4326. I can successfully do Greenland EPSG:3413 to EPSG:4326, but the same commands for Antarctica return 1 grid cell and report 

**To Reproduce**

1. Run the following commands

Set up 4326 mapset

```bash
grass --text -c EPSG:4326 ./G_4326
g.region res=1 -pa w=-180 e=180 s=-90 n=90 -pas # 1° x 1°
exit
```

3031 mapset

```bash
grass -c EPSG:3031 ./G_3031
# 500 m resolution, matching BedMachine boundaries
bound=3333250
g.region res=500 w=-${bound} e=${bound} s=-${bound} n=${bound} -ps
r.mapcalc "foo = 42"
exit
```

Reproject

```bash
grass ./G_4326/PERMANENT
r.proj location=./G_3031 mapset=PERMANENT input=foo # output: 1 column!
```

**Expected behavior**

The input dataset spans all longitudes and covers Antarctica. I'd expect the output dataset to do this, reprojected into EPSG:4326.

**Screenshots**

Screenshots come from earlier (see first version of this issue under the `edit` item above) using BedMachine. The current `g.region` matches that from BedMachine.

3031:

![image](https://github.com/OSGeo/grass/assets/145117/a868441e-5c7f-42ee-9a47-9790561230cd)

4326:

![image](https://github.com/OSGeo/grass/assets/145117/24491629-4497-4635-9118-f648f625cf26)

**System description (please complete the following information):**

```
t480:~ $ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy
t480:~ $ uname -a
Linux t480 5.15.0-97-generic #107-Ubuntu SMP Wed Feb 7 13:26:48 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
t480:~ $ ```


```
version=8.3.2
date=2024
revision=exported
build_date=2024-03-08
build_platform=x86_64-pc-linux-gnu
build_off_t_size=8
libgis_revision=8.3.2
libgis_date=2024-03-08T09:26:59+00:00
proj=9.3.1
gdal=3.8.4
geos=3.12.1
sqlite=3.37.2
```

[...]

PR #3758, opened 2024-06-01
r.proj: clarify -n flag use case

mankoff:
Addresses #3655

I'm not sure this PR should be accepted. It doesn't look like a lot of other people have run into this issue. It is already documented under the 'Notes' section and this may all be a failure of mine to read the documentation sufficiently. Still, if people think it would be helpful to call out the need to use `-n` under certain situations, this may help.


echoix (approved):
Doc only changes (in the modules parameter description), I don't see why it could be problematic :)

PR #3772, opened 2024-06-04
doc: Rename location to project in database image within docs

veroandreo:
Fixes both svg and png files, addressing #3770.


echoix (approved):
I looked at the png, I assume th svg is the source of it.

wenzeslaus (approved):
After couple clicks to wake it up, the Onion Skin diff view of images worked really well on desktop.

PR #3785, opened 2024-06-07
r.texture: refactor

cyliang368:
This PR tries to parallelize `r.texture` with OpenMP. However, the original codes were written just for serial computation. Before starting to implement parallelization, I largely refactored the scripts.

The main problems of the original codes:

1. The computing results are stored in global variables (matrices and vectors), which will cause conflicts in multi-thread computing. I created a struct to store these results and moved the functions related to matrices and vectors from `h_measure.c`  to a new file, `matvec.c`.
2. The most heavy computing part is for loops running through every cell with a window in `main.c`. To make it clearer and distinguishable from the parallel version, I moved that part to the other file (`execute.c`) and replaced it with a function call in `main.c`. A parallelized version will be written in `execute_parallel.c` (I haven't created it).

I haven't finished the parallelization, but this is a big change in this module, so I created this PR. Please review these changes and give me any suggestions. Thanks.


wenzeslaus (commented):
Thanks for sharing that early. The global to struct move is similar to what @petrasovaa recently did for r.horizon and I did for r.sim.water some time ago. I would say it is a good and no-nonsense  approach which improves the code overall.

If you are thinking about it in steps, do you want to do that as two different PRs? Most of the refactoring in the first (variables, files) and focus on parallelization (and perhaps couple code adjustments) in the second one? It might be little harder for you to work on the parallelization before the refactoring is merged, though.

In any case, I would do the files little differently. You are extending h_measure.h while deleting the corresponding .c file. What about having .c and .h files in pairs? Not required by C but sufficiently common to be immediately understood.

I checked the tests for this. They are okay, but a more thorough test would be better. The current tests mostly test that result falls into an expected range.

cyliang368:
@wenzeslaus Thanks for suggestions. I hope this PR can be merged before I submit another PR for parallelization. I will update the files in the next commit, according to your sugguestions. 

I updated the test for this module and submitted it as another PR (#3786). Please consider merging #3786 first.

raster/r.texture/Makefile
```diff
@@ -2,8 +2,10 @@ MODULE_TOPDIR = ../..
 
 PGM = r.texture
 
-LIBES = $(RASTERLIB) $(GISLIB) $(MATHLIB)
+LIBES = $(RASTERLIB) $(GISLIB) $(MATHLIB) $(OPENMP_LIBPATH) $(OPENMP_LIB)
 DEPENDENCIES = $(RASTERDEP) $(GISDEP)
+EXTRA_CFLAGS = $(OPENMP_CFLAGS)
+EXTRA_INC = $(OPENMP_INCPATH)
```
wenzeslaus, line 8: Remove OPENMP

raster/r.texture/h_measure.h
```diff
@@ -22,9 +21,43 @@
...
+#define F4  "Variance              "
+#define F5  "Inverse Diff Moment   "
+#define F6  "Sum Average           "
+#define F7  "Sum Variance          "
+#define F8  "Sum Entropy           "
+#define F9  "Entropy               "
+#define F10 "Difference Variance   "
+#define F11 "Difference Entropy    "
+#define F12 "Measure of Correlation-1 "
+#define F13 "Measure of Correlation-2 "
```
wenzeslaus, line 39: All are unused, remove.

wenzeslaus (changes requested):

wenzeslaus (approved):
Tests pass in the fork.

PR #3786, opened 2024-06-09
r.texture: update tests

cyliang368:
`r.texture` module has thirteen measure functions, but only five are tested. I added tests for all these functions. They are checked by min, max, mean, variance, and cells instead of just ranges of outputs.

### DIFF
--- a/raster/r.texture/testsuite/test_texture.py
+++ b/raster/r.texture/testsuite/test_texture.py
@@ -42,147 +42,147 @@ class TestRasterreport(TestCase):
     def test_asm(self):
         """Testing method asm"""
         asm_ASM = "asm_ASM"
-        values = """min=-0.0455724261701107
-        max=0.0380486063659191
-        mean=-0.000136686790876467
-        variance=0.000158102
+        values = """min=0.104167
+        max=1
+        mean=0.112871783083256
+        variance=0.000158101620160753
         cells=1000246"""
         self.assertModule("r.texture", input=self.input, output="asm", method="asm")
-        self.assertRasterFitsUnivar(asm_ASM, reference=values, precision=2)
+        self.assertRasterFitsUnivar(asm_ASM, reference=values, precision=1e-2)
 
     def test_contrast(self):
         """Testing method contrast"""
         corr_output = "contrast_Contr"
         values = """min=0
-        max=14680.8
-        mean=176.569
-        variance=185767
+        max=14680.771484375
+        mean=176.569350873014
+        variance=185767.349118713
         cells=1000246"""
         self.assertModule(
             "r.texture", input=self.input, output="contrast", method="contrast"
         )
-        self.assertRasterFitsUnivar(corr_output, reference=values, precision=2)
+        self.assertRasterFitsUnivar(corr_output, reference=values, precision=1e-2)
 
     def test_corr(self):
         """Testing method corr"""
         corr_output = "corr_Corr"
-        values = """min=-0.533554
-        max=0.443134
-        mean=-0.0511982
-        variance=0.030665
+        values = """min=-0.533553719520569
+        max=0.443134188652039
+        mean=-0.0511981885382368
+        variance=0.0306650179549719
         cells=1000246"""
         self.assertModule("r.texture", input=self.input, output="corr", method="corr")
-        self.assertRasterFitsUnivar(corr_output, reference=values, precision=2)
+        self.assertRasterFitsUnivar(corr_output, reference=values, precision=1e-2)
 
     def test_var(self):
         """Testing method var"""
         corr_output = "var_Var"
         values = """min=0
-        max=8776.79
-        mean=90.3219
-        variance=58439.1
+        max=8776.787109375
+        mean=90.3218668390769
+        variance=58439.1071892438
         cells=1000246"""
         self.assertModule("r.texture", input=self.input, output="var", method="var")
-        self.assertRasterFitsUnivar(corr_output, reference=values, precision=2)
+        self.assertRasterFitsUnivar(corr_output, reference=values, precision=1e-2)
 
     def test_idm(self):
         """Testing method idm"""
         corr_output = "idm_IDM"
-        values = """min=0.000490662
+        values = """min=0.000490661768708378
         max=1
-        mean=0.123085
-        variance=0.0049711
+        mean=0.12308521457551
+        variance=0.00497110161251246
         cells=1000246"""
         self.assertModule("r.texture", input=self.input, output="idm", method="idm")
-        self.assertRasterFitsUnivar(corr_output, reference=values, precision=2)
+        self.assertRasterFitsUnivar(corr_output, reference=values, precision=1e-2)
 
     def test_sa(self):
         """Testing method sa"""
         sa_SA = "sa_SA"
-        values = """min=95.9583
-        max=2942.79
-        mean=742.368
-        variance=31616.4
+        values = """min=95.9583358764648
+        max=2942.79150390625
+        mean=742.367934271396
+        variance=31616.3527726298
         cells=1000246"""
         self.assertModule("r.texture", input=self.input, output="sa", method="sa")
-        self.assertRasterFitsUnivar(sa_SA, reference=values, precision=2)
+        self.assertRasterFitsUnivar(sa_SA, reference=values, precision=1e-2)
 
     def test_sv(self):
         """Testing method sv"""
-        sa_SA = "sv_SV"
+        sv_SV = "sv_SV"
         values = """min=0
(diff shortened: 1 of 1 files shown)


raster/r.texture/testsuite/test_texture.py
```diff
@@ -24,59 +25,164 @@ def setUpClass(cls):
...
         cls.runModule("g.remove", flags="f", type="raster", name="idm_IDM")
+        cls.runModule("g.remove", flags="f", type="raster", name="var_Var")
+        cls.runModule("g.remove", flags="f", type="raster", name="sa_SA")
+        cls.runModule("g.remove", flags="f", type="raster", name="sv_SV")
+        cls.runModule("g.remove", flags="f", type="raster", name="se_SE")
+        cls.runModule("g.remove", flags="f", type="raster", name="entr_Entr")
+        cls.runModule("g.remove", flags="f", type="raster", name="dv_DV")
+        cls.runModule("g.remove", flags="f", type="raster", name="de_DE")
+        cls.runModule("g.remove", flags="f", type="raster", name="moc1_MOC-1")
+        cls.runModule("g.remove", flags="f", type="raster", name="moc2_MOC-2")
```
wenzeslaus, line 40: The original code was not ideal, g.remove can take a list of rasters: name is defined as `name=string[,string,...]`, i.e., multiple values are allowed. (alternatively, you could just loop though them, but that's likely a bit slower)

raster/r.texture/testsuite/test_texture.py
```diff
@@ -24,59 +25,164 @@ def setUpClass(cls):
...
+        max=0
+        mean=-0.78294
+        variance=0.00736236
+        cells=1000246"""
+        self.assertModule("r.texture", input=self.input, output="moc1", method="moc1")
+        self.assertRasterFitsUnivar(sa_SA, reference=values, precision=2)
+
+    def test_moc2(self):
+        """Testing method moc2"""
+        sa_SA = "moc2_MOC-2"
```
wenzeslaus, line 178: This is confusing. I realize this comes from the original text and also the difficulty that r.texture generates its own names from a basename, but can you do this:

```python
basename = "..."
method = "..."
output = f"{basename}_{method}"
self.assertModule("r.texture", input=self.input, output=basename, method=method)
self.assertRasterFitsUnivar(output, reference=values, precision=2)
```

or simply without variable looks okay too given that the code is short:

```python
self.assertModule("r.texture", input=self.input, output="basename", method="method")
self.assertRasterFitsUnivar("basename_method", reference=values, precision=2)
```

raster/r.texture/testsuite/test_texture.py
```diff
@@ -24,59 +25,164 @@ def setUpClass(cls):
...
+        self.assertRasterFitsUnivar(asm_ASM, reference=values, precision=2)
+
+    def test_contrast(self):
+        """Testing method contrast"""
+        corr_output = "contrast_Contr"
+        values = """min=0
+        max=14680.8
+        mean=176.569
+        variance=185767
+        cells=1000246"""
```
wenzeslaus, line 60: Better than cells is probably n (which is count). There are strong guarantees in GRASS GIS (or at least very high chance) that a tool will generate the right size of a raster, so there is not a great need to test cell count (but you can since you have it there already). However, n tests the non-null cells which tells us if the tool is creating some unexpected nulls in the computation.

raster/r.texture/testsuite/test_texture.py
wenzeslaus: The duplication of names with the test methods is not idea, but I don't have a _good_ suggestion for that.

wenzeslaus (changes requested):
Thank you, this is helpful in general and extremely helpful for your work.

With couple minor changes, we can merge this any time.

raster/r.texture/testsuite/test_texture.py
cyliang368: Solved in the new commit.

raster/r.texture/testsuite/test_texture.py
cyliang368: I added `basename`, `method`, and `output`. Thanks for the suggestions.

raster/r.texture/testsuite/test_texture.py
```diff
@@ -23,60 +24,192 @@ def setUpClass(cls):
...
+            "sv_SV",
+            "se_SE",
+            "entr_Entr",
+            "dv_DV",
+            "de_DE",
+            "moc1_MOC-1",
+            "moc2_MOC-2",
+        ]
+        maps_string = ",".join(maps)
+        cls.runModule("g.remove", flags="f", type="raster", name=maps_string)
```
wenzeslaus, line 43: The underlying functions should deal with a list and do the join for the command line interface automatically:

```suggestion
        cls.runModule("g.remove", flags="f", type="raster", name=maps)
```

wenzeslaus (commented):
Just one detail to get not just a working tests but also a good example test file for the future.

raster/r.texture/testsuite/test_texture.py
```diff
@@ -23,60 +24,178 @@ def setUpClass(cls):
...
+        output = f"{method}_{basename}"
+        values = """min=95.9583358764648
+        max=2942.79150390625
+        mean=742.367934271396
+        variance=31616.3527726298
+        n=996244"""
+        self.assertModule("r.texture", input=self.input, output=method, method=method)
+        self.assertRasterFitsUnivar(output, reference=values, precision=1e-2)
+
+    def test_sv(self):
```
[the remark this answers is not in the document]
echoix: We had two bugs when switching to macOS arm runners (that would occur on M1 and later Macs anyways). 

See https://github.com/OSGeo/grass/issues/3398 and https://github.com/OSGeo/grass/issues/3397, that occurred in the context of this PR https://github.com/OSGeo/grass/pull/3395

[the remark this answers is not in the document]
wenzeslaus: The test can use _skipif_ decorator with _sys.platfrom.startswith_. There can be multiple tests for different platforms. Ideally, there would be no difference, but that would be out of scope of this PR.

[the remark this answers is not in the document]
nilason: Is this deviation caused by `r.texture` or `r.univar`? Either way, it would be good to create an issue for this, so it won't be forgotten. I'm curious whether these deviations are present with non-Apple ARM architectures too. My hunch is that this might be caused by something "trivial" as implicit rounding (via cast), or the like.

raster/r.texture/testsuite/test_texture.py
```diff
@@ -23,60 +28,192 @@ def setUpClass(cls):
...
+        output = f"{method}_{basename}"
+        values = """min=95.9583358764648
+        max=2942.79150390625
+        mean=742.367934271396
+        variance=31616.3527726298
+        n=996244"""
+        self.assertModule("r.texture", input=self.input, output=method, method=method)
+        self.assertRasterFitsUnivar(output, reference=values, precision=1e-2)
+
+    @unittest.skipIf(IS_MAC, reason="Fails on macOS")
```
wenzeslaus, line 126: One more thing, it will be better if this says "Produces a slightly different result on macOS"

echoix: Could we have a second test only for macOS+arm with a different precision until we find the issue? (Testing the same thing)

raster/r.texture/testsuite/test_texture.py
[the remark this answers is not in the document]
cyliang368: Thanks for answering. I created an issue #3796 for this.

raster/r.texture/testsuite/test_texture.py
cyliang368: I think using a very large precision would be confusing. Maybe we could use the other set of values for MacOS. @wenzeslaus Do you think we should have another test for macOS or just skip it now?

raster/r.texture/testsuite/test_texture.py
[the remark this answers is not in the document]
nilason: Thanks for the report!

raster/r.texture/testsuite/test_texture.py
wenzeslaus: No strong opinion, either one is good with me, having a test on macOS, i.e., having two tests is better simply because of higher coverage.

wenzeslaus (approved):

echoix:
CI successful on mirrored PR here: https://github.com/cyliang368/grass/pull/3, can be merged overriding the status checks requirements.

PR #3795, opened 2024-06-11
contributing: Add more known authors to release notes

wenzeslaus:
To match the commits to their GitHub names, we need to translate some of the names and emails to GitHub names. Some of these were missing. Now the script also generates a list of authors who need that update.

There is still some names which are not recognized and we need to have another mechanism to ignore bots, so this is a draft.


wenzeslaus:
The bots are resolved in another PR. The missing author is resolved.

neteler (approved):

PR #3824, opened 2024-06-15
contributing: Add API to check PR titles

wenzeslaus:
This extends the API of the release notes generator to check a string (title) provided in command line against the categories provided in utils/release.yml. If there is no match, a message is printed and the process exits with non-zero return code. There is category "other changes", so non-zero may or may not mean a failure.

This does not do the checking, only provides the API to implement it.

### DIFF
--- a/utils/generate_release_notes.py
+++ b/utils/generate_release_notes.py
@@ -284,10 +284,13 @@ def main():
         epilog="Run in utils directory to access the helper files.",
     )
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


echoix (approved):
Apart from the wording of the help string of the new command that sounds unnatural to me, it is good! But I don't think it's necessary to adjust it, it is more of an internal tool and the PR in this state does its job correctly!

PR #3829, opened 2024-06-15
contributing: Hide bots from release notes

wenzeslaus:
This hides all changes by recognized bots from release notes. The changes may still be relevant, e.g., change of version of dependency in a Docker image, so they are hidden but discoverable using Markdown details HTML tag.

Additionally, to make the release notes shorter, this shows only 20 items and hides the rest together with the changes from bots. The overflowing changes are printed first, then the ones from bots. Each group is sorted alphabetically.

### DIFF
--- a/utils/generate_release_notes.py
+++ b/utils/generate_release_notes.py
@@ -7,6 +7,7 @@ Needs PyYAML, Git, and GitHub CLI.
 
 import argparse
 import csv
+import itertools
 import json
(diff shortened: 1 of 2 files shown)


wenzeslaus:
An example how this looks like:

image

neteler (approved):
Great approach!

echoix (approved):

PR #3846, opened 2024-06-16
CI: Remove build tests for C99, C++11, C++14

wenzeslaus:
Remove tests for old C and C++ standards, aligning with the spirit of _RFC 7: Language Standards Support_.

Using (GNU) C17 and C++17, aligning with GDAL RFC 98 which introduces C++ >= 17 for v3.9.

### DIFF
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -21,12 +21,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - c: gnu99
-            cpp: c++11
-          - c: gnu11
-            cpp: c++11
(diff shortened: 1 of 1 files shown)


wenzeslaus:
@veroandreo Please check that this is aligned enough :-)

echoix (approved):
I like that it will allow removing 13-15 minutes of CI, on pull requests and on pushes to main!

Adding support for newer C/C++ language standards (i.e., C++23) should be added later when building with a newer compiler toolchain is possible (like GCC 14 with ubuntu 24.04)

PR #3853, opened 2024-06-16
lib: Accept more newline styles in G_getl (#3853)

wenzeslaus:
Use G_getl2 in G_getl to have everywhere the same behavior of supporting all newlines. Originally, G_getl2 was created to keep the behavior of G_getl. However, now, we want G_getl2 behavior everywhere.

Keeping G_getl2 for compatibility. Code can later be clean up to use only G_getl, probably at the time when G_getl2 is removed (v9).

The new test fails without the change in the library for CR and CRLF for G_getl, but passes with the change (at least on Linux).


marisn:
I created a LXC Ubuntu container and also observe test failure within the container. At the moment I have no idea of its cause, but there is a segfault when run under GDB and test runs fine under valgrind. (Use of uninitialized / unallocated memory?) @nilason maybe you see the error instantly?

wenzeslaus:
@marisn or @nilason, if you agree that the issue is the test, I can split this into two PRs to get the new code in now and add the test hopefully later.

nilason:
> @marisn or @nilason, if you agree that the issue is the test, I can split this into two PRs to get the new code in now and add the test hopefully later.

I'll take a look on the test. Probably easier to track down locally.

lib/gis/testsuite/test_gis_lib_getl.py
```diff
@@ -0,0 +1,76 @@
...
+from grass.gunittest.case import TestCase
+from grass.gunittest.main import test
+
+
+class TestNewlinesWithGetlFunctions(TestCase):
+    """Test C functions G_getl() and G_getl2() from gis library"""
+
+    @classmethod
+    def setUpClass(cls):
+        cls.libc = ctypes.cdll.LoadLibrary(ctypes.util.find_library("c"))
```
nilason, line 20:
```suggestion
        cls.libc = ctypes.cdll.LoadLibrary(ctypes.util.find_library("c"))
        FILE_ptr = ctypes.POINTER(libgis.FILE)
        cls.libc.fopen.restype = FILE_ptr
        cls.libc.fopen.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
```

nilason (changes requested):
The suggested change prevents the segfault at least.

lib/gis/testsuite/test_gis_lib_getl.py
wenzeslaus: Thank you. Is this something which will work across platforms or should I put that for macOS only? In other words, should I apply this as is?

nilason: This should work cross platform, I'm surprised it worked at all without.

wenzeslaus:
I would perhaps just ignore it since it is CR, although it seems that according to a [SE comment](
https://superuser.com/questions/439440/did-mac-os-lion-switch-to-using-line-feeds-lf-n-for-line-breaks-instead-of), still in 2012, you could get CR:

> The problem is with Microsoft Excel. When you save a CSV file on a Mac, it uses CR.

lib/gis/testsuite/test_gis_lib_getl.py
wenzeslaus: I see. Maybe just on Linux. ctypes is always more difficult then I expect.

nilason: The ubuntu based test also failed with similar result as the mac.

nilason:
> I would perhaps just ignore it since it is CR, although it seems that according to a [SE comment](https://superuser.com/questions/439440/did-mac-os-lion-switch-to-using-line-feeds-lf-n-for-line-breaks-instead-of), still in 2012, you could get CR:
> 
> > The problem is with Microsoft Excel. When you save a CSV file on a Mac, it uses CR.

It looks like the function never really worked as intended, just never was up to the test. It parses `\r\n` just fine.

wenzeslaus:
I did not see the issue with G_getl2 in #3850 and I don't see with a quick look now. The idea was to merge this PR before #3850 to have the test in place, but I got stuck with the library issues in this PR. (If I recall correctly, we knew #3850 helps on Windows with CRLF, but I don't see that explicitly mentioned in the PR.)

wenzeslaus:
I agree that in must be _fgets_. This is out of scope of this PR anyway, so I'm setting the tests to expect failure. Tests should be done soon. Please, review.

nilason (approved):
Great!

PR #3880, opened 2024-06-18
grass.jupyter: Add LayerControl to InteractiveMap by default

chaedri:
Previously, users needed to add a layer control to an `InteractiveMap` with `add_layer_control()`.

This PR changes this so that `add_layer_control()` is called when the users calls `InteractiveMap.show()` by default if the user has not already called it.


petrasovaa (changes requested):
Perhaps clarify that it's added by default in the add_layer_control documentation, so this can be used to change the default position. Not sure whether we need to change anything in the notebooks.

petrasovaa (approved):

PR #3883, opened 2024-06-18
contributing: Add monthly financial supporters to release notes

wenzeslaus:
The generated release notes now contain members of supportes tier on Open Collective. We want to highlight financial contributions in the release notes. Monthly supporters seem like a fitting category in context of release. Current (draft) release notes are already updated.

The list is randomized when generated. Images are not included because that would require generating content for the missing images and making them circular (more work and circular did not work for me).

They are included under a heading at the end of highlights section, so not at the end, but not at the very beginning either.

### DIFF
--- a/utils/generate_release_notes.py
+++ b/utils/generate_release_notes.py
@@ -9,12 +9,14 @@ import argparse
 import csv
 import itertools
 import json
+import random
 import re
 import subprocess
 import sys
 from collections import defaultdict
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


veroandreo (dismissed):

veroandreo (approved):

neteler:
@wenzeslaus backport candidate? We'll have for sure also 8.4.1 one day...

PR #3889, opened 2024-06-18
packaging: Add Nix files for creating development environment and the package

imincik:
Add couple of Nix files which will allow following:

* Create full development environment using single command
* Run GRASS from any GitHub revision

## Automatic development environment activation using Nix

- Install Nix
  [(learn more about this installer)](https://zero-to-nix.com/start/install)

```bash
curl --proto '=https' --tlsv1.2 -sSf \
    -L https://install.determinate.systems/nix \
    | sh -s -- install
```

- Run following command in source code root directory

```bash
nix develop
```

- Optionally, use `direnv` to activate the environment automatically when
  moving to the source code directory

```bash
echo "use flake" > .envrc
direnv allow
```

## Running GRASS directly from the source code 

- Run latest version from `main` branch

```bash
nix run github:OSGeo/grass#grass -- --text
```

- Run version from specific Git branch

```bash
nix run github:OSGeo/grass/<GIT-BRANCH>#grass -- --text
```

- Run version at specific Git revision

```bash
nix run github:OSGeo/grass/<GIT-REVISION>#grass -- --text
```

Usage documentation will be added to https://grasswiki.osgeo.org/wiki/Compile_and_Install

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /emacs.desktop
 *~
 *.lock
+!flake.lock
 *.pyc
 OBJ.*
 locale/scriptstrings/*
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,57 @@
+{
+  "nodes": {
+    "flake-parts": {
+      "inputs": {
+        "nixpkgs-lib": "nixpkgs-lib"
+      },
+      "locked": {
+        "lastModified": 1717285511,
+        "narHash": "sha256-iKzJcpdXih14qYVcZ9QC9XuZYnPc6T8YImb6dX166kw=",
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "rev": "2a55567fcf15b1b1c7ed712a2c6fadaec7412ea8",
+        "type": "github"
+      },
+      "original": {
+        "id": "flake-parts",
+        "type": "indirect"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1718428119,
+        "narHash": "sha256-WdWDpNaq6u1IPtxtYHHWpl5BmabtpmLnMAx0RdJ/vo8=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "e6cea36f83499eb4e9cd184c8a8e823296b50ad5",
+        "type": "github"
+      },
+      "original": {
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,6 +1,5 @@
 /emacs.desktop
 *~
-*.lock
```
echoix, line 3: I understand that this nix PR needs a lock file. Was the existing line there to prevent committing some files?

imincik: This `*.lock` ignore was preventing creation of `flake.lock` file which happens automatically during the first execution, but maybe it is not needed anymore. Good point. Thanks.

imincik:
@neteler , I think, this PR is good to go now.

.gitignore
echoix: I wasn't thinking about removing the change, as committing changes to the flake.lock file might probably be ignored or hard to commit.
If the `*.lock` rule serves a purpose, then maybe adding a negation pattern for `flake.lock` should be used.
If the `*.lock` rules serves no purpose, then removing it makes sense.

imincik: Updated with adding `!flake.lock` . Thanks.

echoix (dismissed):

echoix (approved):

echoix:
@wenzeslaus what PR prefix is appropriate? Should this one be added too? Nix? Build?

imincik:
I changed commit prefix to `packaging`.

echoix:
Perfect!

echoix (approved):

imincik:
And here we go, nix run grass latest dev version on my Ubuntu machine:

grass-nix-run

/cc @landam @wenzeslaus @neteler

echoix:
Isn't the `(exported)` in the version expected to be filled in from git info since a couple of months ago? I think this is the fallback.

imincik:
> Isn't the `(exported)` in the version expected to be filled in from git info since a couple of months ago? I think this is the fallback.

I think this kind of version detection from git can't work for Nix package.

echoix:
If git and the git repo is available at configure time, it should be detected. Otherwise there were some changes made this winter/early spring for helping with reproducible builds

imincik:
> If git and the git repo is available at configure time

That's the issue. Git is not available during the build.

PR #3891, opened 2024-06-18
contributing: Improve message for failing PR title check

wenzeslaus:
The message now includes the expressions (except the excluded ones typically used only by maintainers) and examples (newly added).


echoix (approved):

utils/release.yml
```diff
@@ -3,42 +3,55 @@ notes:
   categories:
     - title: Modules
```
veroandreo, line 4: Shall we start using Tools instead?

veroandreo (approved):

utils/release.yml
wenzeslaus: I'm sorry. I just added that to #3883.

PR #3899, opened 2024-06-19
lib/init: add GRASS_CONFIG_DIR environment variable

imincik:
GRASS_CONFIG_DIR variable allows to set alternative root path for
`.grass<VERSION>` configuration directory. If not set, $HOME directory
is used.

The main use case of this PR is to be able to run multiple isolated instances of grass. Using $HOME or  $XDG_CONFIG_HOME variables to force grass to store configuration in alternative directory is not a good idea, because this will also have an impact on other tools relying on those variables. That's why I introduced a new grass specific environment variable.

/cc @landam


nilason:
Related #3153

imincik:
> > Would it be possible to, instead of introducing another GRASS env var, adopt our code to respect (if set) $XDG_CONFIG_HOME (see https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html)?
> 

But I agree, that usage of XDG_CONFIG_HOME instead of HOME is a good fix for https://github.com/OSGeo/grass/issues/3153 .

nilason:
> The main use case of this PR is to be able to run multiple isolated instances of grass. Using $HOME or $XDG_CONFIG_HOME variables to force grass to store configuration in alternative directory is not a good idea, because this will also have an impact on other tools relying on those variables. That's why I introduced a new grass specific environment variable.

I see.

lib/init/grass.py
```diff
@@ -394,6 +395,7 @@ def help_message(default_gui):
             mapset=_("initial GRASS mapset"),
             full_mapset=_("fully qualified initial mapset directory"),
             env_vars=_("Environment variables relevant for startup"),
+            config_dir_var=_("set root path for configuation directory"),
```
nilason, line 398:
```suggestion
            config_dir_var=_("set root path for configuration directory"),
```

lib/init/variables.html
```diff
@@ -124,6 +124,13 @@ <h2>List of selected (GRASS related) shell environment variables</h2>
     available are RLE, ZLIB, and LZ4. The compressors BZIP2 and ZSTD
     must be enabled when configuring GRASS for compilation.</dd>
 
+  <dt>GRASS_CONFIG_DIR</dt>
+  <dd>[grass startup script]<br>
+    specifies root path for GRASS configuation directory. If not specified,
```
nilason, line 129:
```suggestion
    specifies root path for GRASS configuration directory. If not specified,
```

nilason (changes requested):
some typos

nilason (dismissed):

nilason:
I will not be able to look at and test this this week, but please hold on, I see there are Mac specific changes now that need hands-on testing/review.

landam:
> I will not be able to look at and test this this week, but please hold on, I see there are Mac specific changes now that need hands-on testing/review.

Sure. This PR should change behaviour compared to #3694.

python/grass/app/runtime.py
```diff
@@ -29,15 +29,34 @@ def get_grass_config_dir(major_version, minor_version, env):
...
+        config_path = os.path.join(env.get(env_dirname), "Library")
     else:
         config_dirname = f".grass{major_version}"
-        return os.path.join(env.get("HOME"), config_dirname)
+        config_path = os.path.join(env.get(env_dirname), config_dirname)
+
+    if env.get("GRASS_CONFIG_DIR"):
+        # override config_path if GRASS_CONFIG_DIR environmental
+        # variable is defined
+        config_path = os.path.join(env.get("GRASS_CONFIG_DIR"), config_dirname)
```
wenzeslaus, line 57: This overrides, so it can be first and no tests of base env vars or existence of paths from them is needed. (Current code would raise if `HOME` does not exist even when `GRASS_CONFIG_DIR` is set.)

python/grass/app/runtime.py
```diff
@@ -29,15 +29,34 @@ def get_grass_config_dir(major_version, minor_version, env):
...
     """
     # The code is in sync with grass.app.runtime (but not the same).
+    env_dirname = "APPDATA" if WINDOWS else "HOME"
+    if env.get(env_dirname) is None:
+        raise KeyError(
+            f"The {env_dirname} variable is not set, ask your operating"
+            " system support"
+        )
+    if not os.path.exists(env.get(env_dirname)):
+        raise IsADirectoryError(
```
wenzeslaus, line 39: Raising [IsADirectoryError](https://docs.python.org/3/library/exceptions.html) when it is confusing:

> exception IsADirectoryError
>
> Raised when a file operation (such as [os.remove()](https://docs.python.org/3/library/os.html#os.remove)) is requested on a directory. Corresponds to errno [EISDIR](https://docs.python.org/3/library/errno.html#errno.EISDIR).

[NotADirectoryError](https://docs.python.org/3/library/exceptions.html#NotADirectoryError) seems to be appropriate, but in that case, the test should be [isdir](https://docs.python.org/3/library/os.path.html#os.path.isdir) not [exists](https://docs.python.org/3/library/os.path.html#os.path.exists).

python/grass/app/runtime.py
```diff
@@ -29,15 +29,34 @@ def get_grass_config_dir(major_version, minor_version, env):
     Determines path of GRASS GIS user configuration directory.
     """
     # The code is in sync with grass.app.runtime (but not the same).
+    env_dirname = "APPDATA" if WINDOWS else "HOME"
+    if env.get(env_dirname) is None:
+        raise KeyError(
```
wenzeslaus, line 34: I can't quite see what exception would make sense here. `KeyError` is more related to the underlying dictionary rather than the state the error message is trying to describe. Consider [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError).

lib/init/grass.py
```diff
@@ -379,50 +381,34 @@ def help_message(default_gui):
     sys.stderr.write(s)
 
 
-def get_grass_config_dir():
-    """Get configuration directory
+def create_grass_config_dir():
+    """Create configuration directory
```
wenzeslaus, line 385: Without examining the code more closely, I don't see where the creation happens, but it is not here. I think "creation" is expected to refer to creation of the actual directory rather than construction of the path which is referred to by the simple "get".

wenzeslaus (changes requested):
This seems like a reasonable feature, although I'm wondering whether other (all?) software allows the same, i.e., having its own variable to set config dir just for that software.  

As for the implementation, there is couple of other places which may need fixing. g.extension needs to do something, but I didn't examine what. `lib/gis/home.c` worries me a little bit as it seems to figure out the path as well, at this point duplicating what is in Python which is more complex code with this PR and the anticipated XDG change.

python/grass/app/runtime.py
landam: Changed in 1ea52eb40f.

python/grass/app/runtime.py
landam: Make sense to me. Changed in 94a5369627.

lib/init/grass.py
landam: Creation happens here: https://github.com/imincik/grass/blob/config-dir-env-variable/lib/init/grass.py#L404

landam:
> As for the implementation, there is couple of other places which may need fixing. g.extension needs to do something, but I didn't 

@wenzeslaus  I just found this cb2d32cc93:

The installation of addons seems to respect GRASS_CONFIG_DIR variable already:

```
rm -rf /tmp/.grass8/
GRASS_CONFIG_DIR=/tmp bin.x86_64-pc-linux-gnu/grass ~/grassdata/world_latlong_wgs84/PERMANENT/ --exec g.extension r.hydrodem
ls /tmp/.grass8/addons/bin/
```

and

```
rm -rf ~/.grass8/
bin.x86_64-pc-linux-gnu/grass ~/grassdata/world_latlong_wgs84/PERMANENT/ --exec g.extension r.hydrodem
ls ~/.grass8/addons/bin/
```

Result in the both cases is:

```
r.hydrodem
```

landam:
@nilason Currently MacOS CI fails with

```
ERROR: Failed to create configuration directory '/Users/runner/Library/GRASS/8.5' with error: No such file or directory
```

which is triggered by https://github.com/imincik/grass/blob/config-dir-env-variable/lib/init/grass.py#L404

The path (`/Users/runner/Library/GRASS/8.5`) seems to be correct according grass.py@main: https://github.com/OSGeo/grass/blob/main/lib/init/grass.py#L413

Any idea why CI is failing?

landam:
> examine what. `lib/gis/home.c` worries me a little bit as it seems to figure out the path as well, at this point duplicating what is in Python which is more complex code with this PR and the anticipated XDG change.

@wenzeslaus This issue should be solved by ed4b4a8b11. 

Right, the `G_config_path()` should be also synced with Python code (especially related to MacOS - but this is not related to this PR and should be solved by a new separated PR).

Tested by `db.login` which uses `G_config_path()`:

```
bin.x86_64-pc-linux-gnu/grass ~/grassdata/world_latlong_wgs84/PERMANENT/ --exec db.login use=x pass=y --overwrite
The password was stored in file (/home/martin/.grass8/dblogin)
```

```
GRASS_CONFIG_DIR=/tmp bin.x86_64-pc-linux-gnu/grass ~/grassdata/world_latlong_wgs84/PERMANENT/ --exec db.login use=x pass=y --overwrite
The password was stored in file (/tmp/.grass8/dblogin)
```

lib/gis/home.c
```diff
@@ -100,15 +100,20 @@ const char *G_config_path(void)
...
+    if (!config_dir)
 #ifdef __MINGW32__
-    sprintf(buf, "%s%c%s", getenv("APPDATA"), HOST_DIRSEP, CONFIG_DIR);
+        config_dir = getenv("APPDATA");
 #else
-    sprintf(buf, "%s%c%s", G_home(), HOST_DIRSEP, CONFIG_DIR);
+        config_dir = G_home();
 #endif
+
+    sprintf(buf, "%s%c%s", config_dir, HOST_DIRSEP, CONFIG_DIR);
```
nilason, line 116:
```suggestion
    snprintf(buf, GPATH_MAX, "%s%c%s", config_dir, HOST_DIRSEP, CONFIG_DIR);
```

nilason (changes requested):
Sorry for the delay!

This actually addresses regression on Mac with #3694. I have encountered, probably unrelated, GUI problems so I can't test fully at the moment. I'm looking at this GUI related issue, hopefully coming up with a solution.

Meanwhile, I have a minor suggestion to change sprintf to snprintf (cf. #2767).

nilason:
> @nilason Currently MacOS CI fails with
> 
> ```
> ERROR: Failed to create configuration directory '/Users/runner/Library/GRASS/8.5' with error: No such file or directory
> ```
> 
> which is triggered by https://github.com/imincik/grass/blob/config-dir-env-variable/lib/init/grass.py#L404
> 
> The path (`/Users/runner/Library/GRASS/8.5`) seems to be correct according grass.py@main: https://github.com/OSGeo/grass/blob/main/lib/init/grass.py#L413
> 
> Any idea why CI is failing?

It works with: 

```diff
diff --git a/lib/init/grass.py b/lib/init/grass.py
index 244786d891..8fc6009198 100755
--- a/lib/init/grass.py
+++ b/lib/init/grass.py
@@ -401,7 +401,7 @@ def create_grass_config_dir():
 
     if not os.path.isdir(directory):
         try:
-            os.mkdir(directory)
+            os.makedirs(directory)
         except OSError as e:
             # Can happen as a race condition
             if not e.errno == errno.EEXIST or not os.path.isdir(directory):
```

`os.makedirs` is needed because the `/Users/runner/Library/GRASS` also must be created with a fresh installation.

nilason (approved):
> This actually addresses regression on Mac with https://github.com/OSGeo/grass/pull/3694. I have encountered, probably unrelated, GUI problems so I can't test fully at the moment. I'm looking at this GUI related issue, hopefully coming up with a solution.

Manually adding this PR onto 3f30f1bb59f79d18074a3527d9101200c0975214, solved the regression regarding the PATH refactoring and GRASS_CONFIG_DIR seems to work as intended on Mac as well.

(The GUI issue I encountered is most likely unrelated to this PR and I'll file an Issue or PR to address that).

nilason:
> (The GUI issue I encountered is most likely unrelated to this PR and I'll file an Issue or PR to address that).

Now I see I'm not alone, already reported with #4091.

landam:
@wenzeslaus if you have no objections, we can merge this PR. Please can you update your review?

wenzeslaus (approved):
This looks good to me now. Thanks everyone involved. This is a a sensitive topic not only because of the new variable introduced but also because so many things related startup rely on this to work.

nilason:
This seems to have broken the Docker workflow https://github.com/OSGeo/grass/actions/runs/10218758009.

landam:
> This seems to have broken the Docker workflow https://github.com/OSGeo/grass/actions/runs/10218758009.

See https://github.com/OSGeo/grass/pull/4130

PR #3901, opened 2024-06-19
tools: Use 4 cores as default for parallelized tools

wenzeslaus:
Given the common requests for parallelization by default and 4 cores (threads, processes) being most efficient, this is changing the default for nprocs for all tools to 4.

The value can be changed in GUI settings or using g.gisenv using the NPROCS variable.

This will require review of all test of tools with nprocs to make sure they are using nprocs=1 and not the default when parallelization is not tested. (This is in context of #3879.)

### DIFF
--- a/lib/gis/parser_standard_options.c
+++ b/lib/gis/parser_standard_options.c
@@ -3,7 +3,7 @@
 
    \brief GIS Library - Argument parsing functions (standard options)
 
-   (C) 2001-2019 by the GRASS Development Team
+   (C) 2001-2024 by the GRASS Development Team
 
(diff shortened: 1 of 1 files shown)


ldesousa (approved):
Even my cellphone has more than 4 cores. The only context in which this may be a nuisance is in HPC, where usually each process is assigned only one compute core. However, since the default behaviour is what newer users will have contact with I think this change is useful. In a HPC context more advanced users should be expected, aware of how tweak this setting.

ecodiv:
Setting a mask will disable use of multiple processes if I understand well. Something easy to overlook as an end user.

PR #3902, opened 2024-06-19
v.external: delete duplicated layer requirement definition

pesekon2:
`    options->layer->required = NO;` is contained just a few rows above


echoix (approved):
Is there a way to do a search automatically in all the code base for issues like this?
Good catch!

PR #3928, opened 2024-06-25
Checks: fix flake8 E741 remainder

mshukuno:
Reminder of E741 fixes for the "locale", "python", and "scripts" directories. The majority of fixes involved renaming variable names in iterables. The "noqa: E741" was added to "scripts/d.rast.leg/d.rast.leg.py" since it didn't make sense to change just the "l".

### DIFF
--- a/python/grass/pygrass/raster/rowio.py
+++ b/python/grass/pygrass/raster/rowio.py
@@ -18,17 +18,17 @@ CMPFUNC = ctypes.CFUNCTYPE(
 )
 
 
-def getmaprow_CELL(fd, buf, row, l):
+def getmaprow_CELL(fd, buf, row):
     librast.Rast_get_c_row(fd, ctypes.cast(buf, ctypes.POINTER(librast.CELL)), row)
     return 1
 
 
-def getmaprow_FCELL(fd, buf, row, l):
+def getmaprow_FCELL(fd, buf, row):
     librast.Rast_get_f_row(fd, ctypes.cast(buf, ctypes.POINTER(librast.FCELL)), row)
     return 1
 
 
-def getmaprow_DCELL(fd, buf, row, l):
+def getmaprow_DCELL(fd, buf, row):
     librast.Rast_get_d_row(fd, ctypes.cast(buf, ctypes.POINTER(librast.DCELL)), row)
     return 1
 
--- a/scripts/db.univar/db.univar.py
+++ b/scripts/db.univar/db.univar.py
@@ -313,23 +313,23 @@ def main():
         pval[i] = 0
 
     inf = open(tmp + ".sort")
-    l = 1
+    ln = 1
     for line in inf:
         line = line.rstrip("\r\n")
         if len(line) == 0:
             continue
-        if l == q25pos:
+        if ln == q25pos:
             q25 = float(line)
-        if l == q50apos:
+        if ln == q50apos:
             q50a = float(line)
-        if l == q50bpos:
+        if ln == q50bpos:
             q50b = float(line)
-        if l == q75pos:
+        if ln == q75pos:
             q75 = float(line)
         for i in range(len(ppos)):
-            if l == ppos[i]:
+            if ln == ppos[i]:
                 pval[i] = float(line)
-        l += 1
+        ln += 1
 
     q50 = (q50a + q50b) / 2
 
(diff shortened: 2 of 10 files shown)


python/grass/pygrass/raster/rowio.py
```diff
@@ -18,17 +18,17 @@
 )
 
 
-def getmaprow_CELL(fd, buf, row, l):
+def getmaprow_CELL(fd, buf, row):
```
echoix, line 21: Does this affect compatibility with the C-libraries or its Python wrappers (like ctypes)
Who can answer this?

scripts/db.univar/db.univar.py
```diff
@@ -313,23 +313,23 @@ def main():
         pval[i] = 0
 
     inf = open(tmp + ".sort")
-    l = 1
+    ln = 1
```
echoix, line 316: Is there a better name for this one? What was the reasoning for it (does it come from a formula or something)?

I'm reading it like the natural logarithm (ln), like in maths, but doesn't make sense.

echoix (commented):
I have two questions in two files, all the others are fine as is with me.

scripts/db.univar/db.univar.py
mshukuno: The "line_number" is what I meant.  The logarithm didn't occur to me, but you're right. Would "line_number" or "counter" make sense?

echoix: line_number sounds better than counter for me.

echoix: Also, a note for other readers: I don't know if this part of code is critical or not, but that loop and double loop is probably quite inefficient. Probably doing a "in" check (with a pre-created local tuple) once per loop, and in the 4+pops times it matches, then find the correct assignment.

It's not something to do with this PR.

python/grass/pygrass/raster/rowio.py
echoix: That's the only point that needs to be answered by someone else

echoix: If you aren't able to get feedback and really want to get it finished, I'm comfortable with the remaining changes, if you take the changes of the function signature out for another PR.

echoix: @nilason Do you have an idea on the impact of this, as it seems to be related to the python-c library interfacing.

nilason: [Rast_get_c_row()](https://grass.osgeo.org/programming8/raster_2get__row_8c.html#a7879236341a1d0ad962ba82e82576093) takes three parameters, and `l` was never used so this shouldn't cause any trouble. It is probably a leftover from earlier implementations.

nilason: The unused `l` was there from the start https://github.com/OSGeo/grass/commit/0982f55fa6b3b31e4577bacfdab7d2bbea8c18d3

neteler: FWIW, a historical pointer here is 

"List of functions moved from GIS to Raster library"
https://trac.osgeo.org/grass/wiki/Grass7/RasterLib/ListOfFunctions

echoix (approved):

python/grass/pygrass/raster/rowio.py
echoix: Little update here, it seems that ctypes has a problem that seems to be related to this, on windows:

IMG_7696
IMG_7697
 
 
It’s from a run on main from today: https://github.com/OSGeo/grass/actions/runs/10012925978/job/27679432818#step:11:1495

echoix (commented):
There seems to be a regression possibly caused by a change in the number of arguments in the getmaprow functions in this PR, see https://github.com/OSGeo/grass/pull/3928#discussion_r1685058517

mshukuno:
The "l" parameter should be reinserted? If so

- My lack of C language experience leads me to have trouble finding the name of the "l" parameter. Can you help me with renaming "l"? 
- Can I directory PR for this or should I address the issue and then PR?

neteler:
Does @nilason have a ctypes related hint here? (if it applies)

echoix:
If we make a draft PR re-adding the same numbers of parameters, we'll see right away if there is the same error.

PR #3973, opened 2024-07-02
i.segment: Fix uninitialized struct member 'ngbr_rc.next' in mean_shift.c

ShubhamDesai:
This pull request fixes an issue in `mean_shift.c` where the struct member `ngbr_rc.next` was used without being initialized. The change initializes `ngbr_rc.next` to `NULL` to ensure it is properly set before use.

**Changes Made:**
- Initialize `ngbr_rc.next` to `NULL` 

This fix resolves the cppcheck warning related to the uninitialized variable.

### DIFF
--- a/imagery/i.segment/mean_shift.c
+++ b/imagery/i.segment/mean_shift.c
@@ -457,6 +457,7 @@ static int find_best_neighbour(struct globals *globals, int row, int col,
     visited = pavl_create(cmp_rc, NULL);
     ngbr_rc.row = row;
     ngbr_rc.col = col;
+    ngbr_rc.next = NULL;
     pngbr_rc = G_malloc(sizeof(struct rc));
     *pngbr_rc = ngbr_rc;
     pavl_insert(visited, pngbr_rc);


echoix (commented):
Seems fine to me, but I'm surprised that only one needs to be initialized. If I read correctly, there are 3 variables of type struct rc (one with *, so pointer?), and I would have expected the variable (struct of type rc) named `next` to need it too. C/C++ is very far for me, I'm letting someone else judge for this. I also changed the title so it won't fail the title check (for release notes categorization), so that PR would be ready to merge, once I get a confirmation that it is enough.

nilason:
> Seems fine to me, but I'm surprised that only one needs to be initialized. If I read correctly, there are 3 variables of type struct rc (one with *, so pointer?), and I would have expected the variable (struct of type rc) named `next` to need it too. C/C++ is very far for me, I'm letting someone else judge for this. I also changed the title so it won't fail the title check (for release notes categorization), so that PR would be ready to merge, once I get a confirmation that it is enough.

Member `next` of struct `rs` a pointer, which may need to be initialised with `NULL`.

nilason (approved):

PR #3975, opened 2024-07-02
lib: Fix uninitialized variable 'encoding' in parser_interface.c

ShubhamDesai:
This pull request fixes an issue in `parser_interface.c` where the variable encoding was used without being initialized. The change initializes encoding to NULL to ensure it is properly set before use.

**Changes Made:**

- Initialize encoding to NULL


lbartoletti (approved):

echoix (approved):
Looks fine to me, and proxying @lbartoletti's approval

PR #4003, opened 2024-07-08
grass.jupyter: Allow users to draw simple geometries and save it as vector

29riyasaxena:
Hi there,

Here’s my basic idea for implementing a feature that allows users to draw simple geometries to add as a GRASS native vector map. I’d love to hear your suggestions on how to improve it.


petrasovaa:
Right now it's not creating a GRASS vector, it's just adding  geojson layer to the map. The vector needs to be imported, I think v.import should do that as we discussed before.

29riyasaxena:
> should do

Hi Anna,
Yes, I remember. I hadn't thought of saving them as vectors. Sorry about that. I've pushed the changes now.

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +347,100 @@ def add_layer_control(self, **kwargs):
...
+            def save_geometries(b):
+                name = name_input.value
+                if name and drawn_geometries:
+                    for geometry in drawn_geometries:
+                        geometry["properties"]["name"] = name
+                    geo_json = {
+                        "type": "FeatureCollection",
+                        "features": drawn_geometries,
+                    }
+                    geojson_filename = f"/tmp/{name}.geojson"
```
petrasovaa, line 411: This won't work on all operating systems, use NamedTemporaryFile

petrasovaa:
This is not working, in the jupyterlab log I get:
```
Failed to import geometries into GRASS GIS: a bytes-like object is required, not 'str'
```

Please test more. Run it in the notebook and check the output is correctly written as a vector map.

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +347,110 @@ def add_layer_control(self, **kwargs):
...
+                        ) as temp_file:
+                            temp_filename = temp_file.name
+                            json.dump(geo_json, temp_file)
+                        save_vector(temp_filename, name)
+                        print(f"Imported geometry with name '{name}' into GRASS GIS.")
+                        geo_json_layer = self._ipyleaflet.GeoJSON(
+                            data=geo_json, name=name
+                        )
+                        self.map.add_layer(geo_json_layer)
+                    except Exception as e:
```
petrasovaa, line 427: What is this trying to catch specifically? Pylint doesn't like generic Exception.

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +347,110 @@ def add_layer_control(self, **kwargs):
...
+                        print(f"Failed to import geometries into GRASS GIS: {e}")
+                    geo_json_layer = self._ipyleaflet.GeoJSON(data=geo_json, name=name)
+                    self.map.add_layer(geo_json_layer)
```
petrasovaa, line 430: Isn't this adding it twice?

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +347,110 @@ def add_layer_control(self, **kwargs):
...
+            def save_geometries(b):
+                name = name_input.value
+                if name and drawn_geometries:
+                    for geometry in drawn_geometries:
+                        geometry["properties"]["name"] = name
+                    geo_json = {
+                        "type": "FeatureCollection",
+                        "features": drawn_geometries,
+                    }
+                    geojson_filename = f"/tmp/{name}.geojson"
```
petrasovaa, line 412: Please clean this up.

python/grass/jupyter/utils.py
```diff
@@ -202,6 +202,16 @@ def get_rendering_size(region, width, height, default_width=600, default_height=
     return (default_width, round(default_width * region_height / region_width))
 
 
+def save_vector(geojson_filename, name):
```
petrasovaa, line 205: This is rather trivial function, perhaps this could include the temporary file creation (and removal).

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +347,110 @@ def add_layer_control(self, **kwargs):
...
 
+    def setup_drawing_interface(self):
+        """
+        Allow Users to draw Geometry and Save them.
+        """
+        import ipywidgets as widgets
+        from IPython.display import display
+        import tempfile
+
+        draw_control = self._ipyleaflet.DrawControl()
```
petrasovaa, line 358: I mentioned this before, but here it is again from ipyleaflet documentation:
_The DrawControl is deprecated and will be removed in a future release. Please use GeomanDrawControl instead._

python/grass/jupyter/interactivemap.py
29riyasaxena: If the user adds an extra space after typing the name of the vector (`vector `) or any space in the name (`vector 1`), it will show the error to the user.

python/grass/jupyter/utils.py
29riyasaxena: To clarify, should I add a button to delete the the drawn vector?

python/grass/jupyter/interactivemap.py
29riyasaxena: Hi Anna,

Sorry, it slipped out of my mind. I have made the changes and cannot hide the controls. Could you please help me with this?

python/grass/jupyter/utils.py
petrasovaa: I am talking about moving some of the code handling temporary file creation to this function. The temporary file needs to be removed once it's not needed.

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +346,190 @@ def add_layer_control(self, **kwargs):
...
+                geoman_control.polygon = {"pathOptions": {}}
+                geoman_control.circlemarker = {"pathOptions": {}}
+                geoman_control.drag = False
+                geoman_control.cut = False
+                geoman_control.rotate = False
+                geoman_control.edit = False
+                geoman_control.remove = False
+                self.map.add_control(geoman_control)
+            else:
+                self.map.remove_control(geoman_control.polyline)
```
petrasovaa, line 520: Shouldn't you remove `geoman_control` instead of `geoman_control.polyline`? I haven't actually tested it, it's just a guess.

petrasovaa: BTW, `remove_control` is obsolete, use `remove`, see docs.

petrasovaa: This is probably some limitation on their side. Let's go back to DrawControl then.

29riyasaxena:
Hi Anna, I can successfully add vectors and save them, but I am stuck on the part when deleting them. I tried clearing all the geometries and removing layers when clicking the delete button, but things aren't working. Could you please share your opinions on this?

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +347,89 @@ def add_layer_control(self, **kwargs):
         else:
             self.layer_control_object = self._ipyleaflet.LayersControl(**kwargs)
 
+    def setup_drawing_interface(self):
+        """
+        Allow Users to draw Geometry and Save/Delete them.
+        """
+        import ipywidgets as widgets  # pylint: disable=import-outside-toplevel
+
+        draw_control = self._ipyleaflet.DrawControl()
```
petrasovaa, line 356: I would disable editing and deleting (`DrawControl(edit=False, remove=False)`) to make things simpler.

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +347,89 @@ def add_layer_control(self, **kwargs):
...
+                self.map.remove_control(draw_control)
+                hide_interface()
+
+        draw_button.observe(toggle_geometry, names="value")
+
+        def show_interface():
+            nonlocal save_button_control
+
+            name_input = widgets.Text(description="Name:")
+            save_button = widgets.Button(
```
petrasovaa, line 396: Could you put the button to the right from the text box, not below it? Also, I don't think we need the button_style here.

python/grass/jupyter/interactivemap.py
29riyasaxena: So, aren't we given the option to delete the drawn geometries?

petrasovaa: Exactly, this should be really simple interaction, no moving or deleting, you can always scratch it and start again. Later, we can think if we want to add this functionality, but for now, stick with the simplest.

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +347,95 @@ def add_layer_control(self, **kwargs):
...
+            name_input = widgets.Text(
+                description="Name:",
+                style={"description_width": "initial"},
+                layout=widgets.Layout(
+                    width="50%",
+                    margin="1px 1px 1px 1px",
+                ),
+            )
+            save_button = widgets.Button(
+                description="Save Geometries",
```
petrasovaa, line 398:
```suggestion
                description="Save",
```

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +347,95 @@ def add_layer_control(self, **kwargs):
...
+                self.map.remove_control(draw_control)
+                hide_interface()
+
+        draw_button.observe(toggle_geometry, names="value")
+
+        def show_interface():
+            nonlocal save_button_control
+
+            name_input = widgets.Text(
+                description="Name:",
```
petrasovaa, line 390:
```suggestion
                description="New vector map name:",
```

python/grass/jupyter/interactivemap.py
```diff
@@ -346,6 +347,95 @@ def add_layer_control(self, **kwargs):
...
+                    for geometry in drawn_geometries:
+                        geometry["properties"]["name"] = name
+                    geo_json = {
+                        "type": "FeatureCollection",
+                        "features": drawn_geometries,
+                    }
+                    save_vector(name, geo_json)
+                    geo_json_layer = self._ipyleaflet.GeoJSON(data=geo_json, name=name)
+                    geo_json_layers[name] = geo_json_layer
+                    self.map.add_layer(geo_json_layer)
```
petrasovaa, line 414: When you add the drawn geometries as geojson layer, it should remove the drawings, otherwise it shows up twice.

python/grass/jupyter/interactivemap.py
29riyasaxena: This takes half of the space of the displayed widget. I have added an icon for the save button. Should I move the save button, or does the icon work?

python/grass/jupyter/utils.py
```diff
@@ -202,6 +202,28 @@ def get_rendering_size(region, width, height, default_width=600, default_height=
...
+    param name: name with which vector should be saved
+    """
+    import tempfile
+    import json
+
+    try:
+        with tempfile.NamedTemporaryFile(
+            suffix=".geojson", delete=False, mode="w"
+        ) as temp_file:
+            temp_filename = temp_file.name
```
petrasovaa: When importing, it's saving some junk about style into the database, so I suggest removing properties from json:
```
        for each in geo_json["features"]:
            each["properties"].clear()
```

python/grass/jupyter/utils.py
```diff
@@ -201,6 +204,23 @@ def get_rendering_size(region, width, height, default_width=600, default_height=
...
 
 
+def save_vector(name, geo_json):
+    """
+    Saves the user drawn vector.
+
+    param geojson_filename: name of the geojson file to be saved
+    param name: name with which vector should be saved
+    """
+    with tempfile.NamedTemporaryFile(
```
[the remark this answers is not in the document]
29riyasaxena: I am not sure why and when to delete the temp file.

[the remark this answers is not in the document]
echoix: For the why, its because it is not deleted automatically with delete=False. Take a look at the section here: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile, it goes through the behavior of NamedTemporaryFile vs other tempfile classes.

petrasovaa (approved):

PR #4041, opened 2024-07-12
d.labels: Fix buffer overflow issues in do_labels.c

ShubhamDesai:
This pull request addresses multiple warnings identified by cppcheck related to potential buffer overflow issues.

**Issues:**
1. do_labels.c:87:13: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
            sscanf(text, "%*s %s", buff);
2. do_labels.c:97:13: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
            sscanf(text, "%*s %s", buff);
3. do_labels.c:101:13: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
            sscanf(text, "%*s %s", buff);
4. do_labels.c:105:13: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
            sscanf(text, "%*s %s", buff);
5. do_labels.c:118:17: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
            if (sscanf(text, "%*s %s", font) != 1 || !strcmp(font, "standard"))
6. do_labels.c:126:13: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
            sscanf(text, "%*s %s", buff);
7. do_labels.c:455:13: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
    switch (sscanf(buf, "%s%s", word1, word2)) 

**Changes Made:**
 Added field width specifiers to the sscanf calls to prevent buffer overflows.

The field width is one less than the buffer size.
buff[128] -> %127s
font[256] -> %255s
word1[50], word2[50] -> %49s

### DIFF
--- a/display/d.labels/do_labels.c
+++ b/display/d.labels/do_labels.c
@@ -465,7 +465,6 @@ int scan_ref(char *buf)
         if (buf[i] >= 'A' && buf[i] <= 'Z')
             buf[i] += 'a' - 'A';
     xref = yref = CENT;
-
     switch (sscanf(buf, word_fmt, word1, word2)) {
     case 2:
         if (!(xmatch(word2) || ymatch(word2)))
--- a/display/d.labels/do_labels.c
+++ b/display/d.labels/do_labels.c
@@ -84,7 +84,7 @@ int do_labels(FILE *infile, int do_rotation)
         else if (!strncmp(text, "yof", 3))
             sscanf(text, "%*s %d", &yoffset);
         else if (!strncmp(text, "col", 3)) {
-            sscanf(text, "%*s %s", buff);
+            sscanf(text, "%*s %127s", buff);
             set_RGBA_from_str(&color, buff);
         }
         else if (!strncmp(text, "siz", 3))
@@ -94,15 +94,15 @@ int do_labels(FILE *infile, int do_rotation)
         else if (!strncmp(text, "wid", 3))
             sscanf(text, "%*s %lf", &width);
         else if (!strncmp(text, "bac", 3)) {
-            sscanf(text, "%*s %s", buff);
+            sscanf(text, "%*s %127s", buff);
             set_RGBA_from_str(&background, buff);
         }
         else if (!strncmp(text, "bor", 3)) {
-            sscanf(text, "%*s %s", buff);
+            sscanf(text, "%*s %127s", buff);
             set_RGBA_from_str(&border, buff);
         }
         else if (!strncmp(text, "opa", 3)) {
-            sscanf(text, "%*s %s", buff);
+            sscanf(text, "%*s %127s", buff);
             if (!strncmp(buff, "YES", 3))
                 opaque = YES;
             else
@@ -115,7 +115,7 @@ int do_labels(FILE *infile, int do_rotation)
             }
         }
         else if (!strncmp(text, "fon", 3)) {
-            if (sscanf(text, "%*s %s", font) != 1 || !strcmp(font, "standard"))
+            if (sscanf(text, "%*s %255s", font) != 1 || !strcmp(font, "standard"))
                 strcpy(font, std_font);
         }
         else if (!strncmp(text, "rot", 3)) {
@@ -123,7 +123,7 @@ int do_labels(FILE *infile, int do_rotation)
                 sscanf(text, "%*s %lf", &rotation);
         }
         else if (!strncmp(text, "hco", 3)) {
-            sscanf(text, "%*s %s", buff);
+            sscanf(text, "%*s %127s", buff);
             set_RGBA_from_str(&highlight_color, buff);
         }
         else if (!strncmp(text, "hwi", 3))
@@ -452,7 +452,7 @@ int scan_ref(char *buf)
         if (buf[i] >= 'A' && buf[i] <= 'Z')
             buf[i] += 'a' - 'A';
     xref = yref = CENT;
-    switch (sscanf(buf, "%s%s", word1, word2)) {
+    switch (sscanf(buf, "%49s%49s", word1, word2)) {
     case 2:
         if (!(xmatch(word2) || ymatch(word2)))
             return 0;
--- a/display/d.labels/do_labels.c
+++ b/display/d.labels/do_labels.c
@@ -79,7 +79,8 @@ int do_labels(FILE *infile, int do_rotation)
 
     snprintf(buff_fmt, sizeof(buff_fmt), "%%*s %%%ds", BUFFSIZE - 1);
     snprintf(font_fmt, sizeof(font_fmt), "%%*s %%%ds", FONTSIZE - 1);
-    snprintf(word_fmt, sizeof(word_fmt), "%%%ds %%%ds", WORDSIZE - 1, WORDSIZE - 1);
+    snprintf(word_fmt, sizeof(word_fmt), "%%%ds %%%ds", WORDSIZE - 1,
+             WORDSIZE - 1);
 
     initialize_options();
 
--- a/display/d.labels/do_labels.c
+++ b/display/d.labels/do_labels.c
@@ -129,7 +129,6 @@ int do_labels(FILE *infile, int do_rotation)
         }
         else if (!strncmp(text, "fon", 3)) {
             if (sscanf(text, font_fmt, font) != 1 || !strcmp(font, "standard"))
-
                 strcpy(font, std_font);
         }
         else if (!strncmp(text, "rot", 3)) {
--- a/display/d.labels/do_labels.c
+++ b/display/d.labels/do_labels.c
@@ -20,6 +20,10 @@
 #define YES   1
 #define NO    0
 
+#define BUFFSIZE 128
+#define FONTSIZE 256
+#define WORDSIZE 50
+
 static double east;
 static double north;
 static int xoffset;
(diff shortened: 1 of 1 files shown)


nilason:
Please consider using [pre-commit](https://github.com/OSGeo/grass/blob/main/doc/development/style_guide.md#using-pre-commit), which will solve e.g. clang-format issues before going to the CIs.

display/d.labels/do_labels.c
```diff
@@ -452,7 +452,7 @@
         if (buf[i] >= 'A' && buf[i] <= 'Z')
             buf[i] += 'a' - 'A';
     xref = yref = CENT;
-    switch (sscanf(buf, "%s%s", word1, word2)) {
+    switch (sscanf(buf, "%49s%49s", word1, word2)) {
```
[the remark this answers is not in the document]
ShubhamDesai: added case EOF in switch statement

display/d.labels/do_labels.c
```diff
@@ -84,7 +84,7 @@ int do_labels(FILE *infile, int do_rotation)
         else if (!strncmp(text, "yof", 3))
             sscanf(text, "%*s %d", &yoffset);
         else if (!strncmp(text, "col", 3)) {
-            sscanf(text, "%*s %s", buff);
+            sscanf(text, "%*s %127s", buff);
```
lbartoletti, line 87: You could use number (SCHAR_MAX, UCHAR_MAX) from `limits.h`  instead of using "magic number" for 127, 255, etc

ShubhamDesai: so you are suggesting to change 127 to SCHAR_MAX and 255 to UCHAR_MAX.
And in that case should i leave 49 and 49 for word1 and word2 as it is?

nilason: I agree it would be a good idea to avoid "magic numbers", but we may just as well define them ourself (e.g. `#define BUFFSIZE 128`) – no need for limits.h constants which are not really related. To make full use of the defined constants, we must also dynamically create the formats given to `sscanf()` (e.g. `snprintf(buff_fmt, sizeof(buff_fmt), "%%*s %%%ds", BUFFSIZE - 1);`).
We can use `GNAME_MAX` (defined in gis.h) instead of the value `256`.

display/d.labels/do_labels.c
```diff
@@ -463,6 +464,8 @@ int scan_ref(char *buf)
         FALLTHROUGH;
     default:
         return 0;
+    case EOF:
+        return 0;
```
nilason, line 468: `case EOF` should be the placed as the first case after the switch statement. Alternatively, before `default`, but with `FALLTHROUGH;` instead of `return 0;`

The `default` would be the outcome even without a specific `case EOF`, so this change only clarifies the intention.

ShubhamDesai: did changes but CI check is failing

display/d.labels/do_labels.c
ShubhamDesai: did changes but CI check is failing

display/d.labels/do_labels.c
```diff
@@ -68,6 +72,13 @@ int initialize_options(void)
 int do_labels(FILE *infile, int do_rotation)
 {
     char buff[128];
```
lbartoletti, line 74:
```suggestion
    char buff[BUFFSIZE];
```

display/d.labels/do_labels.c
```diff
@@ -67,7 +71,15 @@ int initialize_options(void)
 
 int do_labels(FILE *infile, int do_rotation)
 {
-    char buff[128];
+    char buff[BUFFSIZE];
+    char buff_fmt[10];
+    char font_fmt[10];
+    char word_fmt[10];
```
nilason, line 77: To avoid using "magic numbers" you may (re-)use `WORDSIZE` here.

To enable the scope of `word_fmt` to `scan_ref()`, you should declare the variable as `static char` for file scope around line 40. (For symmetry, you may do the same for buff_fmt and font_fmt too).

display/d.labels/do_labels.c
```diff
@@ -67,7 +71,15 @@ int initialize_options(void)
...
+
+    snprintf(buff_fmt, sizeof(buff_fmt), "%%%ds", BUFFSIZE - 1);
+    snprintf(font_fmt, sizeof(font_fmt), "%%%ds", FONTSIZE - 1);
```
nilason, line 80: `"%%%ds"` will not produce the intended format.

display/d.labels/do_labels.c
```diff
@@ -452,7 +465,11 @@ int scan_ref(char *buf)
         if (buf[i] >= 'A' && buf[i] <= 'Z')
             buf[i] += 'a' - 'A';
     xref = yref = CENT;
-    switch (sscanf(buf, "%s%s", word1, word2)) {
+    char word_fmt[10];
+    snprintf(word_fmt, sizeof(word_fmt), "%%%ds%%%ds", WORDSIZE - 1,
+             WORDSIZE - 1);
```
nilason, line 470: No need for this if word_fmt is declared in file scope.

display/d.labels/do_labels.c
```diff
@@ -461,6 +478,8 @@ int scan_ref(char *buf)
         if (xmatch(word1) || ymatch(word1))
             return 1;
         FALLTHROUGH;
+    case EOF:
+        FALLTHROUGH;
```
nilason, line 482: Please remove the line with `FALLTHROUGH`, this is what caused the CI failure and it is not needed.

nilason (changes requested):
I added some minor suggestions below.

Also, please add the new size constant to the `font[]` variable at line 40.

display/d.labels/do_labels.c
nilason: > did changes but CI check is failing

You can always look at the "Details" of each CI runner to see what caused the failure in the build log.

display/d.labels/do_labels.c
ShubhamDesai: will "%%ds" work

nilason: Take a good look at the original format. (You may also test this kind of changes in a test file, which can be used as "playground").

ShubhamDesai: It was %127s.
So now it should be %%%ds right
first %% will be %
%d for BUFFSIZE - 1
s for character

nilason: It was `"%*s %s"`

ShubhamDesai: ok so will it be this "%%*s"

i am using snprintf first time so little confused trying to understand from documentation

nilason: It was `"%*s %s"`, you originally (correctly) wanted to change this to `"%*s %127s"`, so how will the format be to produce this format dynamically?

Formatting with [printf() and co](https://en.cppreference.com/w/c/io/fprintf) is very useful, so I recommend you study it and practice with a single file.

ShubhamDesai: snprintf(buff_fmt, sizeof(buff_fmt), "%%*s %%%ds", BUFFSIZE - 1);
sscanf(text, buff_fmt, buff);

this is working. Is it correct?

nilason: > snprintf(buff_fmt, sizeof(buff_fmt), "%%*s %%%ds", BUFFSIZE - 1); sscanf(text, buff_fmt, buff);
> 
> this is working. Is it correct?

Yes!

ShubhamDesai: Thanks for providing good documentation. I was reading wrong examples

display/d.labels/do_labels.c
ShubhamDesai: Ok

display/d.labels/do_labels.c
ShubhamDesai: done

display/d.labels/do_labels.c
ShubhamDesai: done

display/d.labels/do_labels.c
ShubhamDesai: done

display/d.labels/do_labels.c
ShubhamDesai: done

display/d.labels/do_labels.c
nilason: You may change these to`char buff_fmt[WORDSIZE];` etc. The `10` would have been enough, but to avoid another magic number that would be sufficient.

display/d.labels/do_labels.c
```diff
@@ -115,15 +128,16 @@ int do_labels(FILE *infile, int do_rotation)
             }
         }
         else if (!strncmp(text, "fon", 3)) {
-            if (sscanf(text, "%*s %s", font) != 1 || !strcmp(font, "standard"))
+            if (sscanf(text, font_fmt, font) != 1 || !strcmp(font, "standard"))
+
```
nilason, line 132: Unrelated added empty line.

```suggestion
```

display/d.labels/do_labels.c
```diff
@@ -452,7 +466,8 @@ int scan_ref(char *buf)
         if (buf[i] >= 'A' && buf[i] <= 'Z')
             buf[i] += 'a' - 'A';
     xref = yref = CENT;
-    switch (sscanf(buf, "%s%s", word1, word2)) {
+
```
nilason, line 469: Unrelated added empty line.

```suggestion
```

nilason (changes requested):

display/d.labels/do_labels.c
ShubhamDesai: done all 3 of them changed to WORDSIZE

nilason (approved):

PR #4056, opened 2024-07-14
g.findfile: copy into fixed size buffer issue

ShubhamDesai:
This pull request resolves a buffer overflow issue detected by Coverity Scan **(CID 1208235)** in the g.findfile module. The problem arises when the file_opt->answer string is copied into a fixed-size buffer without validating its length, potentially causing the buffer to be overrun if the source string is excessively long.

**Changes Made:**
Replaced the use of strcpy with strncpy to copy file_opt->answer into the name buffer.
Ensured that the name buffer is null-terminated by explicitly setting the last character to '\0'.

### DIFF
--- a/general/g.findfile/main.c
+++ b/general/g.findfile/main.c
@@ -104,8 +104,11 @@ int main(int argc, char *argv[])
             strcpy(name, file_opt->answer);
         G_free_tokens(map_mapset);
     }
-    else
-        strcpy(name, file_opt->answer);
+    else {
+    strncpy(name, file_opt->answer, sizeof(name) - 1);
+    name[sizeof(name) - 1] = '\0';  // Ensure null-termination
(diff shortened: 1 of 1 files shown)


general/g.findfile/main.c
```diff
@@ -104,8 +104,10 @@ int main(int argc, char *argv[])
             strcpy(name, file_opt->answer);
         G_free_tokens(map_mapset);
     }
-    else
-        strcpy(name, file_opt->answer);
+    else {
+        strncpy(name, file_opt->answer, sizeof(name) - 1);
```
nilason, line 108: Better use the new `G_strlcpy()` function.

ShubhamDesai: Done changes. Could you please review it once

PR #4060, opened 2024-07-15
i.smap: fix sizeof not portable issue

ShubhamDesai:
**Issue:** Sizeof not portable issue identified by coverity scan

This pull request resolves a sizeof mismatch issue identified by Coverity Scan (CID 1208214) in the memory allocation logic of the multialloc function.

**Changes Made:**
Replaced the use of sizeof(char **) with sizeof(char *) to ensure correct and portable memory allocation.


nilason (approved):
Thanks!

~~(In fact, this was the original implementation in https://engineering.purdue.edu/~bouman/software/segmentation/, must have been changed in the way in to GRASS GIS).~~ edit: wrong, I misread the lines

PR #4151, opened 2024-08-07
r.clump: Fix unchecked return value from lseek

ShubhamDesai:
This pull request addresses the issue identified by coverity scan (CID : 1415709) [unchecked return value from library]
The checks are added for lseek.


raster/r.clump/clump.c
```diff
@@ -92,7 +94,9 @@ static CELL do_renumber(int *in_fd, DCELL *rng, int nin, int diag, int minsize,
             G_percent(row, nrows, 2);
 
             coffset = (off_t)row * csize;
-            lseek(cfd, coffset, SEEK_SET);
+            if (lseek(cfd, coffset, SEEK_SET) == -1) {
+                G_fatal_error(_("Seek error: %s"), strerror(errno));
```
wenzeslaus, line 98: We need to come up with a better message here. "Seek error" sounds like "Seek help for an error" not "error of seek function" which is all what we are saying here.

"Unable to seek in a temp file: %s" - maybe not the best one possible but good at least in consistency with other error messages here and messages in `lib/segment/format.c`.

wenzeslaus (changes requested):

raster/r.clump/clump.c
ShubhamDesai: done

neteler: We have some similar (?) messages in ´lib/`:

```bash
ag seek | grep G_fatal
gis/seek.c:54:        G_fatal_error(_("Unable to seek: %s"), strerror(errno));
gis/seek.c:59:        G_fatal_error(_("Seek offset out of range"));
gis/seek.c:61:        G_fatal_error(_("Unable to seek: %s"), strerror(errno));
raster/get_row.c:911:        G_fatal_error(_("Error seeking null row %d for <%s>"), R, fcb->name);
segment/seek.c:43:        G_fatal_error("Segment seek: %s", strerror(errno));
segment/seek.c:52:        G_fatal_error("Segment seek: %s", strerror(errno));
```

It might be a good idea to streamline them (also good for the translations to homogenize them).

ShubhamDesai: Sure i will format them as well.

neteler: Thanks (just to clarify: I meant to use the same wording for identical errors).

nilason: May I suggest to change the message to `G_fatal_error(_("Unable to seek: %s"), strerror(errno));`, reusing the existing phrase from `G_fseek()` wrapper function. The `strerror()` message should make it clear what "seek" is about.

echoix (commented):
I think the requested changes by @wenzeslaus and @nilason were addressed correctly. I think the PR is in good shape too, do any of you two can take a last look?

nilason:
> I think the requested changes by @wenzeslaus and @nilason were addressed correctly. I think the PR is in good shape too, do any of you two can take a last look?

Not quite yet. See https://github.com/OSGeo/grass/pull/4151#discussion_r1708877726.

raster/r.clump/clump.c
ShubhamDesai: @nilason so you are suggesting to keep the error message as "Unable to seek" in all files right.

nilason: As this PR aims for a fix in r.clump, we should constrain the scope to just this file.

ShubhamDesai: ok then in that case i will modify the statements to this G_fatal_error(_("Unable to seek: %s"), strerror(errno));

nilason (approved):

wenzeslaus (approved):

PR #4153, opened 2024-08-07
i.segment: Move unused variable value into a code comment

ShubhamDesai:
This pull request addresses the issue identified by coverity scan (CID : 1415633).
The variable hspec is getting overwritten and only last assigned value is being updated.
To resolve this, the redundant assignments have been commented out rather than removing completely.

### DIFF
--- a/imagery/i.segment/mean_shift.c
+++ b/imagery/i.segment/mean_shift.c
@@ -179,8 +179,9 @@ int mean_shift(struct globals *globals)
 
     hspec = globals->hr;
     if (hspec < 0 || hspec >= 1) {
-        // hspec = sqrt(avgdiffavg / 10.0);
-        // hspec = avgdiffavg;
+        // Other ideas how to compute this are:
+        // sqrt(avgdiffavg / 10.0)
+        // avgdiffavg (as is)
         hspec = mindiffzeroavg;
 
         if (do_progressive)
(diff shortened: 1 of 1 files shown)


wenzeslaus:
Looks like "previous ideas". The computation was added by @metzm in 6d56fca.

imagery/i.segment/mean_shift.c
```diff
@@ -179,8 +179,8 @@ int mean_shift(struct globals *globals)
 
     hspec = globals->hr;
     if (hspec < 0 || hspec >= 1) {
-        hspec = sqrt(avgdiffavg / 10.0);
-        hspec = avgdiffavg;
+        // hspec = sqrt(avgdiffavg / 10.0);
+        // hspec = avgdiffavg;
```
wenzeslaus, line 183:
```suggestion
        // Other ideas how to compute this are:
        // sqrt(avgdiffavg / 10.0)
        // avgdiffavg (as is)
```

wenzeslaus (changes requested):
Just commenting this out creates ugly code, but I didn't dive into the code to be confident in just leaving this in the file history as opposed to having it in the code itself. There was no other feedback. Hence, I would just explicitly specify this in the code.

wenzeslaus (approved):

PR #4175, opened 2024-08-13
style: use overwrite and not 'o' flag in Python scripts

neteler:
Fixes messages like

```bash
WARNING: Please update the usage of <v.db.connect>: flag <o> has been
         renamed to <--overwrite>
```

in modules.
Also updates several testsuite files, accordingly.

Note that in some places `overwrite=...` was already present (duplication detected by `pre-commit`).


neteler:
Ok, this went completely wrong. Closing.

PR #3890, opened 2024-06-18
r.horizon: support parallel computing by OpenMP

cyliang368:
This PR parallelizes `r.horizon` module by OpenMP and creates parallelization benchmarks. It was tested on my fork repository (https://github.com/cyliang368/grass/pull/10).

PR #4213, opened 2024-08-22
r.horizon: Support parallel computing for the point mode by OpenMP

cyliang368:
This PR continues parallelizing the `r.horizon` module by OpenMP. [#3890](https://github.com/OSGeo/grass/pull/3890) parallelized the raster mode, while this PR tried to parallelize the point mode.


cyliang368:
The point mode has much fewer computations than the raster mode. Additionally, the results are directly printed out, which is very expensive in terms of runtime. The runtime of the whole process is dominated mainly by printing out the results, but it cannot be parallelized. As a result, the efficiency does not look good if we directly run the benchmark. Overall, parallelization can save some running time but not too much. @petrasovaa, please review and check this PR. If you think we should move on, I will update the documentation.

The two figures below are overall speedup and efficiency.
runtime
efficiency

The two figures below are speedup and efficiency if we let the tool compute but don't print anything.
speedup
efficiency

echoix:
A little unrelated, but did you ever hear of the tool "hyperfine" for benchmarking? It is known for doing stats and reporting outliers, and adjusting the number of runs, doing warmup, ignoring the shell startup time, parametrized sweeping, etc. It's quite useful to know if you're benchmarking in correct conditions (it will report if the noise/outliers may be caused by a busy system), and tell you what is better between these two (or more) alternatives.

https://github.com/sharkdp/hyperfine

cyliang368:
@echoix Thanks for sharing. I will check it.

echoix:
Seeing the plots in the PR that had discontinuities (ups and downs) made me think of it, and how many runs each data point had (and what was the distribution of these, were there some outliers?).

petrasovaa:
This benchmark print the results to json:
r horizon_point_time

petrasovaa:
Larger raster:
r horizon_point_time_1000000000 0
r horizon_point_speedup_1000000000 0
r horizon_point_efficiency_1000000000 0

cyliang368:
It seems that r.horizon point mode does not benefit from this parallelization implementation. Let's close this PR.

PR #4214, opened 2024-08-22
r.out.png: fix consecutive fclose calls on same pointer

ymdatta:
This patch fixes two issues:

1. In one of the code paths, we are calling fclose on a file pointer which could potentially be NULL. Doing that would lead to undefined behavior. Check if a file pointer is NULL before closing it.

2. If we call fclose on same file pointer twice, in the second instance we could be closing file descriptor allocated to some other file, which typically happens to a freed descriptor.

This issue was found by using cppcheck tool.

Additional information:

2. Tool used in finding this error: cppcheck

image

Tool's output after applying the patch

image

3. Reproduction rate: 100%, reproducible every time.
4. Architecture: Not specific to any underlying architecture.

### DIFF
--- a/raster/r.out.png/main.c
+++ b/raster/r.out.png/main.c
@@ -63,7 +63,7 @@ int main(int argc, char *argv[])
     int png_compr, /* ret, */ do_alpha;
     struct Cell_head win;
     FILEDESC cellfile = 0;
-    FILE *fp;
+    FILE *fp = NULL;
 
     /* now goes from pnmtopng.c* -A.Sh */
     /*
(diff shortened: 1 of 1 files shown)


echoix (commented):
I agree with the PR. I would want some feedback on if it is the correct pattern to be autonomous on future reviews of this kind.

wenzeslaus (approved):
Here I my thoughts:

- [x] Default value is set.
- [x] NULL is the right default for FILE.
- [x] fclose requires a check.

What Cppcheck does not see is that _G_fatal_error_ does not return, so the later _fclose_ calls will not happen. I'm not sure what is better here, try to configure, try to provide attributes to the function, or simply apply the best practice (close and set to NULL) regardless of the fact that it is not need. I lean towards the best practice as it is seems the simplest, however providing more info  about the function to the compilers or the linters seems good long-term strategy.

nilason:
Missed this by the second. The only place where it would have made any (if ever so little) sense to reset `fp = NULL` was
https://github.com/OSGeo/grass/blob/917ba5890cf34cef9dcbf1477339c1b0caaaaf24/raster/r.out.png/main.c#L372-L373.

Other than that note, this was good to go.

wenzeslaus:
The last `fp = NULL` would indeed fit the pattern.

ymdatta:
I will push another small patch adding `fp = NULL` at the missed place.

PR #4216, opened 2024-08-23
r.info: Output comments as one string in JSON

wenzeslaus:
The part of history which appears under the comments key can be multiple lines and is stored in a format which limits line length. This is basically how r.info shows it by default in the plain output for humans. Long lines can be split and continuation is marked using a backslash.

This adds a function which puts the continued lines back together and adds newlines between the other lines. There is no newline at the end. This string is then serialized in JSON instead of the original list of strings.

This also avoids serializing the history twice when the h flag is not used, but it does change the behavior for what keys are included.


wenzeslaus:
I assume that following #4304, I should replace strcat by G_strlcat here, no?

nilason:
> I assume that following #4304, I should replace strcat by G_strlcat here, no?

I think that would be good, and replace `strncat()` too. If you __know__ it will fit into dest buffer, just suppress return value with `(void)`.

ninsbl:
Nice. I just noticed that the parsed history may contain empty lines. See e.g.:
```
print(json.loads(gs.read_command("r.info", format="json", map="basin_50K"))["comments"])
```
No problem, but probably worth removing?

wenzeslaus:
Why to remove the empty line? That's what was stored. Why to remove it? See the results with this PR:

```bash
$ grass-dev ~/grassdata/nc_spm_08_grass7/ --tmp-mapset --exec python
```

```python
>>> import grass.script as gs
>>> import json
>>> print(json.loads(gs.read_command("r.info", format="json", map="basin_50K"))["comments"])
Processing mode: All in RAM

r.watershed elevation="elevation" threshold=50000 accumulation="accum_50K" drainage="draindir_50K" basin="basin_50K"
>>> print(gs.parse_command("r.info", format="json", map="basin_50K")["comments"])
Processing mode: All in RAM

r.watershed elevation="elevation" threshold=50000 accumulation="accum_50K" drainage="draindir_50K" basin="basin_50K"
```

wenzeslaus:
What do you think about showing it in JSON just as it is stored in terms of newlines? Can you review this @ninsbl?

ninsbl (approved):
Sorry for the late reply. No problem to keep history as is in terms of newline...

Issue #4230, opened 2024-08-26
[Bug] GDAL VRT with Int8 data type fail to import


ninsbl:
## Describe the bug
`r.in.gdal` and `r.external` fail to import GDAL Virtual Rasters (VRT) with _Int8_ datatype.
_Int8_ was relatively recently introduced to GDAL (3.7) and then made available in VRT datasets.
While GRASS GIS reads _Int8_ data in GeoTiff format, it fails with that datatype in VRT format.

## To reproduce
```
# Create a Int8 raster in GeoTiff
gdal_create -ot Int8 -of GTiff -outsize 24 24 -burn 1 -a_srs EPSG:25833 -a_ullr 0 24 24 0 ./test.tif
# Create a VRT from it
gdal_translate -of VRT ./test.tif ./test.vrt
# Import the GeoTiff (works)
r.in.gdal -o input=test.tif output=test.tif
# Try to import the VRT (fails)
r.in.gdal -o input=test.vrt output=test.vrt
r.external -o input=test.vrt output=test.vrt
```
The last two commands fail with:
```
ERROR 1: Invalid dataType = Int8
ERROR: Unable to open datasource <test.vrt>
```
## Expected behavior
GRASS GIS should import VRT with Int8 data.

## System description
- Ubuntu 22.04
- GRASS GIS version: 8.4
- GDAL 3.7

## Additional context
GRASS GIS does not seem to be the only one here:
https://community.safe.com/data-7/int8-gdal-vrt-raster-not-supported-by-fme-18505

neteler:
I guess it is "easy" to fix by adding `GDT_Int8` support here (possibly with a test for the minimum required GDAL version, i.e. 3.7):

https://github.com/OSGeo/grass/blob/4aa4abf43f69d23ecdea932f3e550b5479df428f/raster/r.external/link.c#L17

nilason:
Is this solved by #4256?

nilason:
#4256 is merged and backported to 8.4, @ninsbl can this be closed?

ninsbl:
> #4256 is merged and backported to 8.4, @ninsbl can this be closed?

Yes. Thanks for backporting @nilason

nilason:
Greit!

PR #4256, opened 2024-08-29
r.in.gdal/r.external: add basic support for GDALs GDT_Int8

ninsbl:
This PR adds basic support for importing GDT_Int8 datatype via GDAL.

I am no C-programmer, so I would be happy if a knowledgeable person could look into this. I am especially unsure about the _#ifndef_ part (note that _14_ is the constant representing GDT_Int8).

Addresses #4230


rouault:
> #if GDAL_VERSION_NUM >= 3000000

Better use
``
#if GDAL_VERSION_NUM >= GDAL_COMPUTE_VERSION(3,7,0)
``

lib/raster/get_row.c
```diff
@@ -368,6 +372,9 @@ static void gdal_values_int(int fd, const unsigned char *data,
         case GDT_Byte:
             c[i] = *(GByte *)d;
             break;
+        case GDT_Int8:
+            c[i] = *(GInt16 *)d;
```
rouault, line 376: I'm not familiar at all with that code, but I believe this should be:
```suggestion
            c[i] = *(signed char *)d;
```

ninsbl: Included in bc9f6a89890e882a19d8a0e35fb92346d8183c78

ninsbl:
Thanks @neteler and @rouault ! Your feedback is very much appreciated. I tried to implement yor suggestions in the latest commits...

lib/raster/get_row.c
nilason: Why not `c[i] = *(GInt8 *)d;` to use the GDAL API and for consistency in the file?

rouault: yes, although I'm not  big fan of our custom type . Using C99's int8_t would probably be better

nilason: Indeed, standard fixed sized integers for fixed sized data types are certainly better.

nilason:
Stefan, you should also add `#include <stdint.h>` (which is where the C99 fixed width integer types are defined, [see more](https://en.cppreference.com/w/c/types/integer)) to `get_row.c`.

ninsbl:
> Stefan, you should also add `#include <stdint.h>` (which is where the C99 fixed width integer types are defined, [see more](https://en.cppreference.com/w/c/types/integer)) to `get_row.c`.

Thanks for the hint. Added in [1565b1d](https://github.com/OSGeo/grass/pull/4256/commits/1565b1de59640450ca4177adaac2f15eff550fe1)

I also added a test for this case.

nilason (dismissed):

ninsbl:
Thanks for your approval @nilason . I had to add a switch to skip the test for systems with GDAL < 3.7 (like Ubuntu 22.04, where tests failed). That commit dismissed your review, it would be great if you could re-approve...

nilason (approved):

nilason:
Should this be backported?

nilason:
Backported to 8.4 with c4b2304.

PR #4260, opened 2024-08-30
d.mon: Fix Unbounded Source Buffer in main.c file of d.mon module

ShubhamDesai:
This Pull Request addresses the unbounded source buffer issue identified by Coverity Scan (CID: 1270265).

**Changes Implemented:**

- Monitor Name Length Check:

- Defined a maximum length for the monitor name (mon) as 256 characters using the MONITOR_NAME_LIMIT macro.

- Introduced checks to ensure that the mon variable, which stores the monitor name, does not exceed this defined length.

- The check if (mon && strlen(mon) < MONITOR_NAME_LIMIT) ensures that mon is not NULL and that its length does not exceed MONITOR_NAME_LIMIT.

### DIFF
--- a/display/d.mon/list.c
+++ b/display/d.mon/list.c
@@ -12,7 +12,7 @@
 char *get_path(const char *name, int fpath)
 {
     char tmpdir[GPATH_MAX];
-    int available_space;
+    size_t available_space;
 
     G_temp_element(tmpdir);
     available_space = GPATH_MAX - strlen(tmpdir) - 1;
@@ -157,7 +157,7 @@ void list_files(const char *name, FILE *fd_out)
     char tmpdir[GPATH_MAX], mon_path[GPATH_MAX];
     struct dirent *dp;
     DIR *dirp;
-    int available_space;
+    size_t available_space;
 
     G_temp_element(tmpdir);
     available_space = GPATH_MAX - strlen(tmpdir) - 1;
--- a/display/d.mon/list.c
+++ b/display/d.mon/list.c
@@ -21,7 +21,7 @@ char *get_path(const char *name, int fpath)
             _("Insufficient space to append /MONITORS to the path for <%s>."),
             name);
     }
-    strncat(tmpdir, "/MONITORS", available_space);
+    strcat(tmpdir, "/MONITORS");
     available_space -= strlen("/MONITORS");
     if (name) {
         if (available_space < strlen("/")) {
@@ -29,7 +29,7 @@ char *get_path(const char *name, int fpath)
                 _("Insufficient space to append / to the path for <%s>."),
                 name);
         }
-        strncat(tmpdir, "/", available_space);
+        strcat(tmpdir, "/");
         available_space -= strlen("/");
 
         if (available_space < strlen(name)) {
@@ -37,7 +37,7 @@ char *get_path(const char *name, int fpath)
                             "<%s> to the path."),
                           name);
         }
-        strncat(tmpdir, name, available_space);
+        strcat(tmpdir, name);
         available_space -= strlen(name);
     }
 
@@ -159,14 +159,14 @@ void list_files(const char *name, FILE *fd_out)
             _("Insufficient space to append /MONITORS to the path for <%s>."),
             name);
     }
-    strncat(tmpdir, "/MONITORS", available_space);
+    strcat(tmpdir, "/MONITORS");
     available_space -= strlen("/MONITORS");
 
     if (available_space < strlen("/")) {
         G_fatal_error(_("Insufficient space to append / to the path for <%s>."),
                       name);
     }
-    strncat(tmpdir, "/", available_space);
+    strcat(tmpdir, "/");
     available_space -= strlen("/");
 
     if (available_space < strlen(name)) {
@@ -174,7 +174,7 @@ void list_files(const char *name, FILE *fd_out)
                         "the path."),
                       name);
     }
-    strncat(tmpdir, name, available_space);
+    strcat(tmpdir, name);
     available_space -= strlen(name);
 
     G_file_name(mon_path, tmpdir, NULL, G_mapset());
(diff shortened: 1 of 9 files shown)


wenzeslaus (commented):
While this may address the unbounded source buffer issue, it seems to me that it will just lead to another warning. The mon string goes into functions like list_cmd and somewhere there it is put to another buffer which may or may not be large enough. Maybe the issue needs to be addressed there. I checked this issue in Coverity Scan and it seems to be able to take you there.

ShubhamDesai:
> While this may address the unbounded source buffer issue, it seems to me that it will just lead to another warning. The mon string goes into functions like list_cmd and somewhere there it is put to another buffer which may or may not be large enough. Maybe the issue needs to be addressed there. I checked this issue in Coverity Scan and it seems to be able to take you there.

To address this, I have now added two checks:

Primary Check in the Main File: I implemented a length check for mon directly in the main file to ensure it is within safe limits before being passed to other functions.

Additional Checks in Individual Functions: In each of the individual functions, the length of name is also checked to provide an extra layer of safety.

For these checks, instead of using a random number, I used GPATH_MAX as the baseline. The MONITOR_NAME_LIMIT is specifically calculated as GPATH_MAX - strlen("/MONITORS/") - 1 because the "/MONITORS/" directory path is used in one of the functions to create a temporary directory. This calculation ensures that the actual monitor name length is within safe limits, considering the directory structure.


So if the checks pass in the main file still there will be additional checks inside the functions

display/d.mon/stop.c
```diff
@@ -14,6 +14,10 @@ static int stop(const char *);
 
 int stop_mon(const char *name)
 {
+    if (strlen(name) >= GPATH_MAX) {
+        G_fatal_error(_("Monitor name <%s> is too long."), name);
+    }
+
```
wenzeslaus, line 20: Here, it is not constructing any strings, so a check is not needed.

display/d.mon/main.c
```diff
@@ -176,7 +178,7 @@ int main(int argc, char *argv[])
         if (list_flag->answer)
             G_warning(_("Flag -%c ignored"), list_flag->key);
         mon = G_getenv_nofatal("MONITOR");
-        if (mon) {
+        if (mon && strlen(mon) < MONITOR_NAME_LIMIT) {
```
wenzeslaus, line 181: There is actually `GNAME_MAX`, so you that.

display/d.mon/list.c
```diff
@@ -105,6 +105,9 @@ int check_mon(const char *name)
 /* list related commands for given monitor */
 void list_cmd(const char *name, FILE *fd_out)
 {
+    if (strlen(name) >= GPATH_MAX) {
+        G_fatal_error(_("Monitor name <%s> is too long."), name);
+    }
```
wenzeslaus, line 110: The check should be done in `get_path` where the strings are concatenated. See also below for the issue with the check itself.

display/d.mon/list.c
```diff
@@ -136,6 +139,9 @@ void list_files(const char *name, FILE *fd_out)
     strcat(tmpdir, "MONITORS");
     strcat(tmpdir, "/");
     strcat(tmpdir, name);
+    if (strlen(tmpdir) >= GPATH_MAX) {
+        G_fatal_error(_("The path for monitor <%s> is too long."), name);
+    }
```
wenzeslaus, line 144: This type of check actually does not work. The point is not to limit the length of the string just because. The point is to not overflow the allocated buffer. You need to compute the length before using strcat, for strings which are too long, the behavior is undefined ([see strcat doc](https://en.cppreference.com/w/c/string/byte/strcat)).

wenzeslaus (changes requested):

display/d.mon/stop.c
ShubhamDesai: Removed. Done

display/d.mon/main.c
ShubhamDesai: Done

display/d.mon/list.c
ShubhamDesai: I initialized the available_space variable with GPATH_MAX - strlen(tmpdir) - 1, ensuring that it accounts for the initial size of tmpdir and the null terminator.
After each concatenation (such as "/", "MONITORS", and the monitor name), I subtracted the length of the appended string from available_space and checked whether there was enough space for the next operation.
This ensures that each part of the path can be appended safely without overflowing the buffer.

display/d.mon/list.c
ShubhamDesai: same usage of variable available_space for calculating the buffer availability

nilason (changes requested):
I have planned to create a new `G_strlcat()` (local wrapper/implementation of [strlcat()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/strlcat.html), to accompany `G_strlcpy()`), but have not got the time for that yet.
That would make this fix much less complicated. Please hold on for that (or implement for the time being with posix `strlcat()` if possible).

nilason:
> I have planned to create a new `G_strlcat()` (local wrapper/implementation of [strlcat()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/strlcat.html), to accompany `G_strlcpy()`)

Addressed by #4304.

ShubhamDesai:
> > I have planned to create a new `G_strlcat()` (local wrapper/implementation of [strlcat()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/strlcat.html), to accompany `G_strlcpy()`)
> 
> Addressed by #4304.

Thanks. I will use it

nilason:
G_strlcat is now in place.

ShubhamDesai:
> G_strlcat is now in place.

Done.

ShubhamDesai:
> ` G_temp_element() + "/" + "MONITORS" + "/"` is guaranteed to fit into a buffer of size GPATH_MAX, so you can (void) the return value of the three first G_strlcat (in both get_path and list_files). The `name` variable is coming from user, so this should be checked (and G_fatal_error if truncated). I don't see the need for the change of main.c.

changes done.

PR #4262, opened 2024-08-30
v.in.dwg: Avoid using same variable as parameter and destination in sprintf

ymdatta:
Currently, one instance of sprintf has same variable as parameter and destination in sprintf. This scneario leads to undefined behavior in C.

```
sprintf(buf, _("%s Cannot open %s"), buf, path);
```

Modify the code to:

1. Write initial error string using snprintf() onto the buffer. Using snprintf() makes sure that we stay within the buffer size and avoid overflow errors.
2. Use snprintf() again to write another error string at the end of previous error string in the same buffer. We again use snprintf() to make sure we are not overflowing the buffer with data.

Prior to fix:

image

After fix:

image

Additional information:

3. Architecture: Not specific to any underlying architecture.

### DIFF
--- a/vector/v.in.dwg/main.c
+++ b/vector/v.in.dwg/main.c
@@ -66,7 +66,8 @@ int main(int argc, char *argv[])
     struct GModule *module;
     struct Option *out_opt, *in_opt;
     struct Flag *z_flag, *circle_flag, *l_flag, *int_flag;
-    char buf[2000];
+    const size_t BUFSIZE = 2000;
+    char buf[BUFSIZE];
 
     /* DWG */
     char path[2000];
@@ -135,10 +136,13 @@ int main(int argc, char *argv[])
     /* Init OpenDWG */
     sprintf(path, "%s/etc/adinit.dat", G_gisbase());
     if (!adInitAd2(path, &initerror)) {
-        sprintf(buf, _("Unable to initialize OpenDWG Toolkit, error: %d: %s."),
-                initerror, adErrorStr(initerror));
+        snprintf(buf, BUFSIZE,
+                 _("Unable to initialize OpenDWG Toolkit, error: %d: %s."),
+                 initerror, adErrorStr(initerror));
+        size_t buflen = strlen(buf);
         if (initerror == AD_UNABLE_TO_OPEN_INIT_FILE)
-            sprintf(buf, _("%s Cannot open %s"), buf, path);
+            snprintf(buf + buflen, BUFSIZE - buflen, _(" Cannot open %s"),
+                     path);
         G_fatal_error(buf);
     }
     adSetupDwgRead();


ymdatta:
Note: I couldn't compile the files in this package as I was not able to download the OpenDWG toolkit headers and libraries from [http://www.opendwg.org](http://www.opendwg.org) from instructions in [README](https://github.com/OSGeo/grass/blob/main/vector/v.in.dwg/README). I believe the link is dead. So, I'm a bit uneasy about these changes. But, it's still a problem at hand as current code can cause undefined behavior.

I want to get inputs from you all on this.

vector/v.in.dwg/main.c
```diff
@@ -137,8 +137,11 @@ int main(int argc, char *argv[])
     if (!adInitAd2(path, &initerror)) {
         sprintf(buf, _("Unable to initialize OpenDWG Toolkit, error: %d: %s."),
                 initerror, adErrorStr(initerror));
-        if (initerror == AD_UNABLE_TO_OPEN_INIT_FILE)
-            sprintf(buf, _("%s Cannot open %s"), buf, path);
+        size_t buflen = strlen(buf);
+        if (initerror == AD_UNABLE_TO_OPEN_INIT_FILE) {
+            snprintf(buf + buflen, 2000 - buflen - 1, (" Cannot open %s"),
```
nilason, line 142: I recommend you take a closer look at the `snprintf()` manual or https://en.cppreference.com/w/c/io/fprintf to see how it works and specifically what arguments it can receive.
Furthermore, make a small test file and try it out to get to know it better.

ymdatta: Thanks for the review @nilason!

I have made the following changes after looking at the `snprintf()` manual and testing few things out:

1. Using a variable of `size_t` datatype for `bufsz` argument in the function call.
2. Using `BUFSIZE - buflen` rather than `2000 - buflen - 1` as snprintf guarantees that it only writes `BUFSIZE-buflen-1` characters onto the buffer leaving space for null character.
3. There is a chance that prior to this `snprintf()` call, buffer can overflow from previous `sprintf()`, as we do not have boundary checks. I have modified to use `snprintf()` over there as well.

     Now, by the time we reach second `snprintf()` we are guaranteed that the buffer hasn't overflowed and we are writing at a valid space.
     
I checked with a small test [example](https://pastebin.com/GuCz39nF) that if we do this, we would be writing one string after another in the same buffer, which is what we are trying to achieve here.

Let me know what do you think and if there is something I missed. Thanks!

neteler:
> I was not able to download the OpenDWG toolkit headers and libraries from http://www.opendwg.org from instructions in [README](https://github.com/OSGeo/grass/blob/main/vector/v.in.dwg/README). I believe the link is dead. So, I'm a bit uneasy about these changes

I have search around and the OpenDWG page went offline around 2011. Perhaps we should rather retire this code?

nilason:
> > I was not able to download the OpenDWG toolkit headers and libraries from http://www.opendwg.org from instructions in [README](https://github.com/OSGeo/grass/blob/main/vector/v.in.dwg/README). I believe the link is dead. So, I'm a bit uneasy about these changes
> 
> I have search around and the OpenDWG page went offline around 2011. Perhaps we should rather retire this code?

I think we should. We have [v.in.redwg](https://github.com/OSGeo/grass-addons/tree/grass8/src/vector/v.in.redwg) as replacement.

ymdatta:
That makes sense.

By retiring the code, do we stop pushing new commits to this module? If so, I can close this PR.

wenzeslaus:
@nilason if your comment was resolved, I suggest merging and then removing the disabled code in another PR. If it is not resolved, I suggest closing and, again, removing.

vector/v.in.dwg/main.c
nilason: Now the code looks good!

nilason:
> @nilason if your comment was resolved, I suggest merging and then removing the disabled code in another PR. If it is not resolved, I suggest closing and, again, removing.

@ymdatta did resolve the changes of the code and came up a nice string catenation solution using snprinft() (strlen() was the key).
However, if we are going to remove the module, why push any updates to it?

neteler:
The update is appreciated and then we may move it to add-ons, in a good shape.

nilason:
> The update is appreciated and then we may move it to add-ons, in a good shape.

Move is something else than remove :-) In not sure about the good shape, this has probably not been tested (or even used) in quite a while (depending on proprietary SDK), although it will certainly be better!

wenzeslaus:
> However, if we are going to remove the module, why push any updates to it?

Just because @ymdatta already created the changes. I would not be suggesting that otherwise.

I'm saying, let's merge this (rather than close), and then remove the code (directory and Makefile entry).

nilason (approved):

marisn (approved):

wenzeslaus:
@ymdatta, can you please create a PR to remove v.in.dwg?

PR #4263, opened 2024-08-31
lib/display: D_show_conversions() %.1f to %g where values likely to be smaller than 0.0

HamishB:
[display lib], along with minor debug message whitespace tweaks

Display pixels or raster region rows,cols are likely be on the order of 500, while projected map units are often in the millions, so pixels/meter factor can end up as a very small number. This patch edits an unused in the production code debug helper function to not round that division into a confusing 0.0, as well as some whitespace tweaks to make it more readable.

### DIFF
--- a/lib/display/cnversions.c
+++ b/lib/display/cnversions.c
@@ -113,15 +113,9 @@ void D_show_conversions(void)
     fprintf(stderr, " U_w %10.1f  U_e %10.1f  U_s %10.1f  U_n %10.1f\n\n",
             U.west, U.east, U.south, U.north);
 
-    fprintf(stderr,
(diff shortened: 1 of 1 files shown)


nilason:
@HamishB Please format the code with ClangFormat v18 (see https://github.com/OSGeo/grass/blob/main/doc/development/style_guide.md#c-and-c), the easiest and recommended way is by using `pre-commit` (see further down on the same link).

See also https://github.com/OSGeo/grass/blob/main/doc/development/github_guide.md#pr-title) regarding PR titel, in this case it should be prefixed with `lib/display:`.

nilason (approved):

PR #4268, opened 2024-09-02
docker: fix missing libraries in ubuntu wxgui Dockerfile

neteler:
Add `libgtk-3-dev` to address:

```
...
checking for GTK+ - version >= 3.0.0... Package gtk+-3.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `gtk+-3.0.pc'
to the PKG_CONFIG_PATH environment variable
No package 'gtk+-3.0' found
```

For the usage, also install `python3-wxgtk4.0`.


nilason (dismissed):
Looks reasonable to me!

nilason (approved):

PR #4302, opened 2024-09-11
lib/db: remove deprecated implementation of db_get_login() and replace it with that of db_get_login2()

ymdatta:
`db_get_login()` is deprecated in 7.8.0 release with da399c52637ee5c81228459fa374a114c8fef06c. Change its function signature and implementation to match that of the currently preferred `db_get_login2()`.

Modify `db_get_login2()` to internally call `db_get_login()` function call. In the future major version, `db_get_login2()` would be removed to have only one method to get login details for simplicity.

### DIFF
--- a/lib/db/dbmi_base/login.c
+++ b/lib/db/dbmi_base/login.c
@@ -346,22 +346,20 @@ static int get_login(const char *driver, const char *database,
 
    If driver/database is not found, output arguments are set to NULL.
 
-   \deprecated Use db_set_login2() instead.
-
-   \todo: GRASS 8: to be replaced by db_set_login2().
-
    \param driver driver name
    \param database database name (can be NULL)
    \param[out] user name
    \param[out] password string
+   \param[out] host name
+   \param[out] port
 
    \return DB_OK on success
    \return DB_FAILED on failure
  */
-int db_get_login(const char *driver, const char *database, const char **user,
-                 const char **password)
+int db_get_login2(const char *driver, const char *database, const char **user,
+                  const char **password, const char **host, const char **port)
 {
-    return get_login(driver, database, user, password, NULL, NULL);
+    return db_get_login(driver, database, user, password, host, port);
 }
 
 /*!
@@ -379,8 +377,8 @@ int db_get_login(const char *driver, const char *database, const char **user,
    \return DB_OK on success
    \return DB_FAILED on failure
  */
-int db_get_login2(const char *driver, const char *database, const char **user,
-                  const char **password, const char **host, const char **port)
+int db_get_login(const char *driver, const char *database, const char **user,
+                 const char **password, const char **host, const char **port)
 {
     return get_login(driver, database, user, password, host, port);
 }
--- a/db/drivers/mysql/db.c
+++ b/db/drivers/mysql/db.c
@@ -51,7 +51,7 @@ int db__driver_open_database(dbHandle *handle)
                 connpar.host, connpar.port, connpar.dbname, connpar.user,
                 connpar.password);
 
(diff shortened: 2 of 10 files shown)


nilason:
`db_get_login()` was deprecated (in favour of `db_get_login2()`) in 2015 with the 7.8.0 release: https://github.com/OSGeo/grass/commit/da399c52637ee5c81228459fa374a114c8fef06c

I think this (8.5) may the moment to drop it.

nilason:
(The CodeQL CI failure is unrelated to changes of this PR.)

lib/db/dbmi_base/login.c
```diff
@@ -341,29 +341,6 @@ static int get_login(const char *driver, const char *database,
...
-   \param[out] password string
-
-   \return DB_OK on success
-   \return DB_FAILED on failure
- */
-int db_get_login(const char *driver, const char *database, const char **user,
-                 const char **password)
-{
-    return get_login(driver, database, user, password, NULL, NULL);
-}
```
wenzeslaus, line 365: I suggests to rename this to db_get_login2 and call db_get_login. To keep semantic versioning, we should not remove db_get_login2. We can remove db_get_login, but I agree that renaming db_get_login2 to db_get_login might be better way forward. We need to keep db_get_login2 in the API, though. How important it is? I don't know, I don't see it used in grass-addons.

ymdatta: I agree with your point on semantic versioning, as this change is not backward compatible.

Renaming the above function to `db_get_login2` is also not possible as both of these have different function signatures:

`int db_get_login(const char *, const char *, const char **, const char **)`

vs

`int db_get_login2(const char *, const char *, const char **, const char **, const char **, const char **)`

Maybe we can keep this on hold till next major release by tagging it?

nilason: If we are the least worried to break anything, we should only remove the deprecated `db_get_login()` and leave the `db_get_login2()` as is.

wenzeslaus: My thinking was that we should be also cleaning these "version 2" functions and allowing for the use of non-versioned ones. Old _db_get_login_ will be removed, new db_get_login_ with different parameters will be introduced.

ymdatta: @nilason : I agree with your point. I am just concerned that removing `db_get_login()` in its current form, even if its deprecated warrants a bump to MAJOR version as its still a public API (+ deprecated).

@wenzeslaus : I think changing the function signatures (essentially removing old `db_get_login()` and introducing new one) would still cause the semantic versioning problem as it's a backward incompatible change and needs a major version bump.

Quoting [semver 2.0.0](https://semver.org/#how-should-i-handle-deprecating-functionality) with my emphasis: "_Before you completely remove the functionality in a new major release_ there should be at least one minor release that contains the deprecation so that users can smoothly transition to the new API."

wenzeslaus: I think our (project) understanding is that if we document it as deprecated in one major version, we can remove it (or replace it or change it) in the next one. The function was documented as deprecated for v7.8.0 in https://github.com/OSGeo/grass/commit/da399c52637ee5c81228459fa374a114c8fef06c, hence we can remove in v8.

ymdatta: Ohh, that's my mistake. I assumed that it was deprecated in v8. I'll make the changes. Thanks!

ymdatta: I have reintroduced `db_get_login2()` API, changed its function signature and implementation to internally call `db_get_login()` function call. This ensures we still have `db_get_login2()` in the API list.

wenzeslaus (approved):

PR #4332, opened 2024-09-17
lib/gis: match prototype with declaration for G_strlcat and G_strlcpy

nilason:
Make local implementations of strlcpy() and strlcat() optimizable by using restrict type qualifier, while keeping public API for G_strlcat and G_strlcpy available to C++ code and having identical prototype and declaration.

Currently the prototypes look like:

```c
size_t G_strlcat(char *, const char *, size_t);
```
but declared as:

```c
size_t G_strlcat(char *restrict dst, const char *restrict src, size_t dsize) { ...}
```

which the compilers seem to accept, but was probably not the best choice.

### DIFF
--- a/lib/gis/strlcat.c
+++ b/lib/gis/strlcat.c
@@ -28,6 +28,9 @@
 #include <stddef.h>
 #include <string.h>
 
+static size_t G__strlcat(char *restrict dst, const char *restrict src,
+                         size_t dsize);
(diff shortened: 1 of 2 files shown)


wenzeslaus (approved):
That makes sense and even somewhat looks like the other wrappers we have elsewhere!

PR #4346, opened 2024-09-19
lib/gmath: properly guard BLAS and LAPACK dependent code

nilason:
Properly guard BLAS and LAPACK dependent code.

### DIFF
--- a/.github/workflows/macos_install.sh
+++ b/.github/workflows/macos_install.sh
@@ -47,6 +47,8 @@ CONFIGURE_FLAGS="\
   --with-bzlib-libs=${CONDA_PREFIX}/lib \
   --with-bzlib-includes=${CONDA_PREFIX}/include \
   --with-netcdf=${CONDA_PREFIX}/bin/nc-config \
+  --with-blas=openblas \
+  --with-lapack=openblas \
   --with-netcdf=${CONDA_PREFIX}/bin/nc-config \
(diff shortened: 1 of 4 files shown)


neteler:
Side-question concerning addons:

In my fork I get
https://github.com/neteler/grass-addons/actions/runs/10937266666/job/30362726188

` checking for lapacke.h... no configure: error: *** Unable to find "lapacke.h" with "-I/usr/include/x86_64-linux-gnu/openblas-pthread/  -g -O2 ". Error: Process completed with exit code 1. `

Would that be addressed here as well?

nilason:
https://github.com/OSGeo/grass-addons/pull/1206 adds liblapacke-dev which is enough.

echoix (dismissed):

nilason:
I have now updated this to work as it did before d5bb442. Namely, to only compile lib/gmath/la.c (and being able to include la.h) IFF compiled with BLAS and LAPACK.

The first solution (https://github.com/OSGeo/grass/pull/4346/commits/02b0d9690ff1eeb8ce19fa48d4d83552015e8dec) made available all parts which weren't directly using BLAS or LAPACK (only a couple of functions out of 30+) even without compiling with BLAS and LAPACK.

I'd say we follow the original logic and go for the present state of this PR. If need arises in the future, we may revise this.

PR #4351, opened 2024-09-20
lib/gmath: rename ATLAS_wrapper to CBLAS_wrapper

nilason:
Rename file to reflect changes made with
https://github.com/OSGeo/grass/commit/d5bb442d7b121861928a37842e05e5265a980427


echoix (approved):
I'm surprised nothing more is needed, and that if it was needed by the previous merged commit, it didn't fail then

wenzeslaus:
@echoix The GNU Autotools Makefiles are set up to pick all files based on file extension. C itself looks at function names.

PR #4353, opened 2024-09-20
doc: Fix Flake8 issues in example files

arohanajit:
This PR fixes the flake8 error in `doc/python`. It addresses mainly `F403`, `F405` and `E501`. These are mostly related to wild card imports and `max-line-length` errors. I haven't fixed `F401` in `dialogs.py` yet as that will require more comprehensive changes

### DIFF
--- a/doc/python/m.distance.py
+++ b/doc/python/m.distance.py
@@ -25,7 +25,7 @@
 
 # %module
 # % label: Finds the distance between two or more points.
-# % description: If the projection is latitude-longitude, this distance is measured along the geodesic.
+# % description: If the projection is lat-long, distance measured along the geodesic.
 # % keyword: miscellaneous
 # % keyword: distance
 # % keyword: measure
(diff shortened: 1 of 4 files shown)


doc/python/m.distance.py
```diff
@@ -25,7 +25,7 @@
 
 # %module
 # % label: Finds the distance between two or more points.
-# % description: If the projection is latitude-longitude, this distance is measured along the geodesic.
+# % description: If the projection is lat-long, distance measured along the geodesic.
```
wenzeslaus, line 28: Please keep the original line and open an issue for this line. This is a general issue and we can't resolve it by less clear documentation which this shortening approach may eventually lead to.

arohanajit: Got it, this is also something which I highlighted here in the existing `flake8` discussion: https://github.com/OSGeo/grass/issues/1522#issuecomment-2364560761
For this merge, I'll roll back this change

arohanajit: Updated commit to remove this

wenzeslaus (approved):
These need more work to be easily testable. Let's merge this as is relying on the static checks making sure that the PR does not make the code worse.

PR #4356, opened 2024-09-21
CI: Updated parser doc in scripts/d.wms.py to fix E501

arohanajit:
Fixed `flake8: E501` in `scripts/d.wms.py`. I just split description into multiple lines to comply with `max-line-length`

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -25,7 +25,6 @@ per-file-ignores =
     doc/python/m.distance.py: E501
     doc/gui/wxpython/example/dialogs.py: F401
     locale/grass_po_stats.py: E122, E128, E231, E401, E722
-    gui/scripts/d.wms.py: E501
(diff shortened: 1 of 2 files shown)


neteler:
Does the parser accept multiple description lines?

arohanajit:
If I am being honest I don't really know. I was hoping to rely on github actions to highlight any potential errors. If you could tell me how can I check for this (I am new to this repo) that would be really helpful! Considering how common this error is

wenzeslaus:
For now, just don't touch these #% lines. We need to figure out what we
actually want to do with them.

neteler:
> I was hoping to rely on github actions to highlight any potential errors.

(In this case there is AFAIK no test implemented, hence the github actions can not catch any potential error.)

PR #4359, opened 2024-09-21
g.proj: Fix copy into fixed size buffer issue in input.c file

ShubhamDesai:
This pull request resolves a buffer overflow issue detected by Coverity Scan (CID 1501203).
strcpy is replaced with G_strlcpy.
The build was successful after the make process


echoix (approved):
Assuming the string should never appear unless there is an exceptional condition that would have been a bug before, and that it follows the same pattern of the other PRs, I think it's fine to go right away :)

PR #4367, opened 2024-09-23
checks: Replace bare except (Flake8 E722) in wxGUI/core

arohanajit:
### **Description:**

This PR addresses the PEP8 style error `E722` by specifying exception types in all bare `except` clauses found throughout the codebase. By replacing bare `except:` statements with specific exception types, we improve code clarity, prevent unintended exception handling, and adhere to PEP8 guidelines.

---

### **Changes Made:**

1. **Replaced bare `except:` with `except Exception:` where appropriate.**

   - **Affected Code Blocks:**
     ```python
     # Example 1
     if len(command) == 1 and not skipInterface:
         try:
             task = gtask.parse_interface(command[0])
         except Exception:
             task = None

2. **Replaced bare `except:` with specific exceptions to catch only expected errors.**

   - **Changed to `except OSError:` when handling OS-related exceptions.**

     ```python
     # Example 1
     try:
         self.locs = os.listdir(os.path.join(os.environ["GISBASE"], "locale"))
         self.locs.append("en")  # GRASS doesn't ship EN po files
         self.locs.sort()
         self.locs.insert(0, "system")
     except OSError:
         # No NLS
         self.locs = ["system"]

### **Rationale:**

- **Explicit Exception Handling:**
  - **Clarity:** Specifying the exception type makes it clear which exceptions are expected and handled, improving code readability.

- **Preventing Unintended Behavior:**
  - **Selective Exception Catching:** By catching only the exceptions we expect, we prevent the accidental suppression of errors that could lead to inconsistent states or make debugging more difficult.

---

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -25,12 +25,6 @@ per-file-ignores =
     doc/python/m.distance.py: E501
     doc/gui/wxpython/example/dialogs.py: F401
     gui/scripts/d.wms.py: E501
-    gui/wxpython/core/gconsole.py: E722
-    gui/wxpython/core/render.py: E722
-    gui/wxpython/core/settings.py: E722
-    gui/wxpython/core/toolboxes.py: E722
-    gui/wxpython/core/utils.py: E722
-    gui/wxpython/core/workspace.py: E722
     gui/wxpython/datacatalog/tree.py: E731, E402
     gui/wxpython/dbmgr/base.py: E722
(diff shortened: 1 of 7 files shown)


wenzeslaus (changes requested):
Two cases with Exception need to be changed to use grass.script.CalledModuleError. The others will need to be changed later (not in this PR).

PR #4400, opened 2024-09-27
wxGUI: Fixed unused 'wizard' variable - F841 in image2target/

arohanajit:
Since `wizard` isn't being used anywhere, it looks like it's only purpose is to trigger `GCPWizard` constructor. However, instead of deleting it, I've just commented it our rn incase it needs to be used in future (just as a previous version of it was) - which I'll remove if need be.
I'm triggering the constrcutor standalone without any variable which fixes `F841`


petrasovaa (dismissed):
Yes, just remove it, no need to keep it commented.

echoix (approved):

echoix (approved):

petrasovaa (approved):

PR #4418, opened 2024-09-30
wxGUI: Fixed F841 in ii2t_manager.py

arohanajit:
FIxed `F841` by following changes
- Removed `busy`  variable and calling the function by itself since we don't need the return value
- Removed `point` variable which I assume was used to get cursor position but isn't used anywhere. I am not sure if this function has some inbuilt settings which are called with the constructor - in which case, we can just call it without getting the return values
- Removed `diff` and `reload_target` - which were just being reassigned without being used or passed on


petrasovaa (changes requested):
See https://docs.wxpython.org/wx.BusyInfo.html to use BusyInfo with context manager

petrasovaa (approved):

PR #4422, opened 2024-09-30
v.generalize: Fix Resource Leak issue in displacement.c

ShubhamDesai:
This pull request addresses resource leak issue identified by coverity scan (CID : 1207602)
Used existing function Vect_destroy_cats_struct() to fix the memory leak issue.


nilason (changes requested):
While you’re at it, you can destroy `Points` too.

ShubhamDesai:
> While you’re at it, you can destroy `Points` too.

Done

nilason (approved):
Thanks!

PR #4476, opened 2024-10-08
grass.jupyter: importing display function from IPython.display

### LINKED ISSUE #4340: [Feat] grass.jupyter: map.show
GRASS Jupyter currently returns an `IPython.display` `Image` object when displaying a map image with a Jupyter notebook. As shown here.

```python
def show(self):
        """Displays a PNG image of map"""
        # Lazy import to avoid an import-time dependency on IPython.
        from IPython.display import Image  # pylint: disable=import-outside-toplevel

        return Image(self._filename)
```
This is fine for many use cases but does not work when the image needs to be displayed inline with various Jupyter frontends (JupyterHub, Colab, VSCode, etc..). My specific use-case is using quarto (.qmd) to create figure layouts with VSCode that are rendered to html, pdf, or latex formats. Using the the current `grass.jupyter api`, you are unable to use the layout functionality of quarto without calling `map.show()` within the `IPython.display`  `display` function like  `display(map.show()`.

_Example using quarto layouts_
```python
#| label: fig-dsm-map
#| fig-cap-location: bottom
#| fig-cap: 1m DEMs
#| fig-subcap: 
#|      - "DSM"
#|      - "DTM"
#| column: body
#| output: true
#| layout-ncol: 2
#| echo: false

from IPython.display import display

map_dsm = gj.Map(width=250, use_region=False)
map_dsm.d_rast(map=dsm)
map_dsm.d_legend(raster=dsm, at="7,35,2,5", flags="b", unit="m")
map_dsm.d_barscale(at=(1,5), flags="n")
display(map_dsm.show())

map_dtm = gj.Map(width=250, use_region=False)
map_dtm.d_rast(map=dtm)
map_dtm.d_legend(raster=dtm, at="7,35,2,5", flags="b", unit="m")
map_dtm.d_barscale(at=(1,5), flags="n")
display(map_dtm.show())
```

To address this we should use the `IPython.display`  `display` function to display the map image so that we let the Jupyter frontends  decide which representation is used and how.

```python
def show(self):
        """Displays a PNG image of map"""
        # Lazy import to avoid an import-time dependency on IPython.
        from IPython.display import Image, display  # pylint: disable=import-outside-toplevel

        display(Image(self._filename))
```

Read more here:
[https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.display](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.display)

rohannallamadge:
importing display function from IPython.display to display the map image

Fixes #4340.


python/grass/jupyter/map.py
```diff
@@ -188,9 +188,9 @@ def wrapper(**kwargs):
     def show(self):
         """Displays a PNG image of map"""
         # Lazy import to avoid an import-time dependency on IPython.
-        from IPython.display import Image  # pylint: disable=import-outside-toplevel
+        from IPython.display import Image, display  # pylint: disable=import-outside-toplevel
 
-        return Image(self._filename)
+        return display(Image(self._filename))
```
wenzeslaus, line 193: Is return needed here? What does IPython.display.display function return?

python/grass/jupyter/map.py
```diff
@@ -188,9 +188,9 @@ def wrapper(**kwargs):
     def show(self):
         """Displays a PNG image of map"""
         # Lazy import to avoid an import-time dependency on IPython.
-        from IPython.display import Image  # pylint: disable=import-outside-toplevel
+        from IPython.display import Image, display  # pylint: disable=import-outside-toplevel
```
wenzeslaus, line 191: I think the line is too long and that's what triggering the CI failure for quality and style. Run recent Black locally on the file or install pre-commit to get the right formatting easily.

wenzeslaus (changes requested):
Thank you. This is great. It looks like it is addressing the original idea in #4340.

I left two comments and I also modified the title to make it valid (there was a leading space) and I moved the referenced issue to the text.

python/grass/jupyter/map.py
rohannallamadge: sure sir, i will do

rohannallamadge: sir : [Errno 9] Bad file descriptor 
Screenshot 2024-10-16 233753
and   tempcoderunnerfile.py is created automatically 
Screenshot 2024-10-16 234140
what changes i have to do

petrasovaa: This doesn't seem to be related to GRASS.

cwhite911: > This doesn't seem to be related to GRASS.

Agreed. Looks like an issue with VSCode.

cwhite911 (changes requested):
Nice job working on this PR.

python/grass/jupyter/map.py
echoix: > sir : [Errno 9] Bad file descriptor 
> 
> Screenshot 2024-10-16 233753
> 
> and   tempcoderunnerfile.py is created automatically 
> 
> Screenshot 2024-10-16 234140
> 
> what changes i have to do
> 
> 

This isn't specific to grass at all, but you shouldn't have your cloned git repo in a OneDrive folder. First, all file accesses will be slow (including all the thousands of files inside .git), as it will need to pass through the Cloud Filter API. Then, files that are locked will lag the sync process. I wouldn't be surprised if some antivirus would be interfering with some Python tools requesting a file that might not really exist yet (a placeholder, not hydrated), or that is currently being synced.

Try again with a fresh clone of the repo in a local folder instead. If you insist on not using pre-commit, make sure you have black installed with Jupyter support. You can also configure your IDE to use black to format, but we are only able to have something working uniformly with pre-commit.

rohannallamadge: Thank you for the suggestion! I'll try using a fresh clone of the repo in a local folder 
Thanks again for the guidance!

rohannallamadge:
- Added _repr_pretty_ method to provide a custom pretty representation in IPython environments 
-  Ensures proper handling of recursive cycles by displaying "Map(...)" as a fallback

rohannallamadge:
Removed the return statement from the `show()` method
Now using `display()` directly to handle image rendering inline, without returning an `Image` object

echoix:
Seems like some basic code formatting and checking is still missing. Let us know once ready ;)

rohannallamadge:
sir, error 
Screenshot 2024-10-19 234659

.pylintrc
```diff
@@ -652,5 +652,6 @@ min-public-methods=1
 
 # Exceptions that will emit a warning when being caught. Defaults to
 # "BaseException, Exception".
-overgeneral-exceptions=BaseException,
-                       Exception
+overgeneral-exceptions=BaseException,Exception
```
petrasovaa, line 655: I don't see a need for this change

.pylintrc
```diff
@@ -654,3 +654,5 @@ min-public-methods=1
 # "BaseException, Exception".
 overgeneral-exceptions=BaseException,
                        Exception
+
```
cwhite911, line 657: Remove extra lines.

.pylintrc
```diff
@@ -654,3 +654,4 @@ min-public-methods=1
...
```
cwhite911, line 657: No change occurred in this file other than the extra line getting included. Please fix this and the PR will be ready to approve.

cwhite911 (changes requested):

PR #4494, opened 2024-10-11
checks: Remove extra semicolons in r.terraflow

wenzeslaus:
With GCC and -Werror=pedantic, the r.terraflow C++ code produces errors because extra semicolons are present after some function definitions. This removes the semincolons and I no longer get 'error: extra ;'.


nilason (approved):
One would think `-Wpedantic -Werror ` is the same as  `-Werror=pedantic` …

PR #4507, opened 2024-10-13
lib/vector/Vlib: Fix Resource leak issue in geos.c

ShubhamDesai:
This pull request addresses resource leak issue identified by coverity scan (CID : 1208050, 1208051, 1208052)
Used G_free for fixing resource leak

### DIFF
--- a/lib/vector/Vlib/geos.c
+++ b/lib/vector/Vlib/geos.c
@@ -353,8 +353,6 @@ GEOSCoordSequence *V1_read_line_geos(struct Map_info *Map, long offset,
     G_debug(3, "    n_points = %d dim = %d", n_points,
             (Map->head.with_z) ? 3 : 2);
 
-    pseq = GEOSCoordSeq_create(n_points, (Map->head.with_z) ? 3 : 2);
-
     x = (double *)G_malloc(n_points * sizeof(double));
     y = (double *)G_malloc(n_points * sizeof(double));
     if (Map->head.with_z)
@@ -363,24 +361,20 @@ GEOSCoordSequence *V1_read_line_geos(struct Map_info *Map, long offset,
         z = NULL;
 
     if (0 >= dig__fread_port_D(x, n_points, &(Map->dig_fp))) {
-        GEOSCoordSeq_destroy(pseq);
-        pseq = NULL;
         goto free_return;
     }
 
     if (0 >= dig__fread_port_D(y, n_points, &(Map->dig_fp))) {
-        GEOSCoordSeq_destroy(pseq);
-        pseq = NULL;
         goto free_return;
     }
 
     if (Map->head.with_z) {
         if (0 >= dig__fread_port_D(z, n_points, &(Map->dig_fp))) {
-            GEOSCoordSeq_destroy(pseq);
-            pseq = NULL;
             goto free_return;
         }
     }
+
+    pseq = GEOSCoordSeq_create(n_points, (Map->head.with_z) ? 3 : 2);
 
     for (i = 0; i < n_points; i++) {
         GEOSCoordSeq_setX(pseq, i, x[i]);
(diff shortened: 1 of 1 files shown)


lib/vector/Vlib/geos.c
```diff
@@ -362,15 +362,27 @@ GEOSCoordSequence *V1_read_line_geos(struct Map_info *Map, long offset,
     else
         z = NULL;
 
-    if (0 >= dig__fread_port_D(x, n_points, &(Map->dig_fp)))
+    if (0 >= dig__fread_port_D(x, n_points, &(Map->dig_fp))) {
+        G_free((void *)x);
+        G_free((void *)y);
+        G_free((void *)z);
```
nilason, line 368: `pseq` should also be free'd, adding code. In this case I'd suggest to avoid some repeated code by using goto to clean up before return.

```suggestion
        GEOSCoordSeq_destroy(pseq);
        pseq = NULL;
        goto free_return;
```

The above should be added to all three cases.

And add the goto label `free_return:` before the line with `G_free((void *)x);` at the end of the function.

nilason (changes requested):

lib/vector/Vlib/geos.c
ShubhamDesai: done changes

lib/vector/Vlib/geos.c
```diff
@@ -362,15 +362,24 @@ GEOSCoordSequence *V1_read_line_geos(struct Map_info *Map, long offset,
     else
         z = NULL;
 
-    if (0 >= dig__fread_port_D(x, n_points, &(Map->dig_fp)))
-        return NULL; /* end of file */
+    if (0 >= dig__fread_port_D(x, n_points, &(Map->dig_fp))) {
+        GEOSCoordSeq_destroy(pseq);
```
nilason, line 366: If you move the `pseq = GEOSCoordSeq_create…`to just before `pseq` is used, then no need to destroy and set to NULL. Just need to initialise to NULL.

ShubhamDesai: done

nilason: > Just need to initialise to NULL.

ShubhamDesai: done

nilason (approved):

lib/vector/Vlib/geos.c
```diff
@@ -353,26 +353,29 @@ GEOSCoordSequence *V1_read_line_geos(struct Map_info *Map, long offset,
...
     y = (double *)G_malloc(n_points * sizeof(double));
     if (Map->head.with_z)
         z = (double *)G_malloc(n_points * sizeof(double));
     else
         z = NULL;
 
-    if (0 >= dig__fread_port_D(x, n_points, &(Map->dig_fp)))
-        return NULL; /* end of file */
+    if (0 >= dig__fread_port_D(x, n_points, &(Map->dig_fp))) {
+        goto free_return;
```
nilason, line 364:
```suggestion
        goto free_return; /* end of file */
```

lib/vector/Vlib/geos.c
```diff
@@ -353,26 +353,29 @@ GEOSCoordSequence *V1_read_line_geos(struct Map_info *Map, long offset,
...
+    }
 
-    if (0 >= dig__fread_port_D(y, n_points, &(Map->dig_fp)))
-        return NULL; /* end of file */
+    if (0 >= dig__fread_port_D(y, n_points, &(Map->dig_fp))) {
+        goto free_return;
```
nilason, line 368:
```suggestion
        goto free_return; /* end of file */
```

lib/vector/Vlib/geos.c
```diff
@@ -353,26 +353,29 @@ GEOSCoordSequence *V1_read_line_geos(struct Map_info *Map, long offset,
...
+    }
 
     if (Map->head.with_z) {
-        if (0 >= dig__fread_port_D(z, n_points, &(Map->dig_fp)))
-            return NULL; /* end of file */
+        if (0 >= dig__fread_port_D(z, n_points, &(Map->dig_fp))) {
+            goto free_return;
```
nilason, line 373:
```suggestion
            goto free_return; /* end of file */
```

nilason (changes requested):

nilason (approved):

PR #4508, opened 2024-10-14
v.net.timetable: Fix Resource Leak issue in main.c

ShubhamDesai:
This pull request addresses resource leak issue identified by coverity scan (CID : 1208053, 1208054)
Used existing function Vect_destroy_line_struct(), Vect_destroy_list() to fix the memory leak issue.

### DIFF
--- a/vector/v.net.timetable/main.c
+++ b/vector/v.net.timetable/main.c
@@ -176,19 +176,18 @@ void write_subroute(struct segment *seg, struct line_pnts *line, int line_id)
     struct line_cats *Cats;
     struct ilist *list;
 
-    Points = Vect_new_line_struct();
     Cats = Vect_new_cats_struct();
-    list = Vect_new_list();
     r = seg->route;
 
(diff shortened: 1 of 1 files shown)


vector/v.net.timetable/main.c
```diff
@@ -185,6 +185,8 @@ void write_subroute(struct segment *seg, struct line_pnts *line, int line_id)
 
     if (r < 0) {
         Vect_write_line(&Out, GV_LINE, line, Cats);
+        Vect_destroy_list(list);
+        Vect_destroy_line_struct(Points);
```
nilason, line 189: Wouldn't it be better to create `Points` and `list` _after_ this if statement?

nilason (changes requested):

vector/v.net.timetable/main.c
ShubhamDesai: yes you are right.
changes done

nilason (approved):

PR #4512, opened 2024-10-14
r.mask.status: fix null pointer dereference and false positive string overflow

nilason:
Fixes two new issues reported by Coverity Scan:

- Handles case of unsuccessful creation of json string
- Silences false positive issue for string copy operation to buffer of size GMAPSET_MAX from G_mapset()


wenzeslaus (approved):
Thanks for fixing it! I'm happy the detection works.

PR #4535, opened 2024-10-16
GUI: EPSG code source: replace epsg.io with spatialreference.org

neteler:
As epsg.io is partially outdated (see [source](https://github.com/maptiler/epsg.io/issues/171) and related) and PROJ is our engine anyway, this PR switches the references in the GUI to https://spatialreference.org.

image

### DIFF
--- a/lib/init/helptext.html
+++ b/lib/init/helptext.html
@@ -62,7 +62,7 @@ projections.
 <h3>Creating a New project with the Project Wizard</h3>
 <p>
 If you know the CRS of your data or study area,
-you can fill <a href="https://epsg.io">EPSG code</a>
+you can fill <a href="https://spatialreference.org/">EPSG code</a>
 or description and Project Wizard finds appropriate CRS from a predefined list
 of projections.
 
@@ -97,4 +97,4 @@ See <a href="grass.html">examples</a> of running GRASS GIS from a command line.
 </em>
 
 <p>
- <a href="https://epsg.io/">List of EPSG codes</a> (Database of worldwide coordinate systems)
+ <a href="https://spatialreference.org/">List of EPSG codes</a> (Database of worldwide coordinate systems)
--- a/lib/init/helptext.html
+++ b/lib/init/helptext.html
@@ -97,4 +97,6 @@ See <a href="grass.html">examples</a> of running GRASS GIS from a command line.
 </em>
 
 <p>
- <a href="https://spatialreference.org/">List of EPSG codes</a> (Database of worldwide coordinate systems)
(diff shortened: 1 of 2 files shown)


lib/init/helptext.html
```diff
@@ -97,4 +97,4 @@ <h2>See also</h2>
 </em>
 
 <p>
- <a href="https://epsg.io/">List of EPSG codes</a> (Database of worldwide coordinate systems)
+ <a href="https://spatialreference.org/">List of EPSG codes</a> (Database of worldwide coordinate systems)
```
wenzeslaus, line 100: Is epsg.io that bad we can't even have it in a _See also_ section? Even the linked issue https://github.com/maptiler/epsg.io/issues/171 says:

> When presenting objects like CRS, it is probably OK, ...

CRS is what we link to even in the GUI.

wenzeslaus (commented):
Do you have a reference for spatialreference.org being actually better than epsg.io or are we replacing one outdated source by another?

lib/init/helptext.html
nilason: Maybe https://crs-explorer.proj.org/

jjimenezshaw: https://spatialreference.org/ is open source, now maintained by the PROJ team (currently me). It is updated on every PROJ release, every 3 months. It was completely refactored last December.

https://crs-explorer.proj.org/ is also under the PROJ umbrella (also me ;) It is older than the refactorization of SR, but SR is better known because started a long time ago (2007).

A few weeks ago SR was improved with the main functionality of crs-explorer: more complex filter, including the map.

rouault: 100% agreeing with @jjimenezshaw . Also to mention that spatialreference.org comes with no ads and no tracking, contrary to epsg.io   spatialreference.org also covers "odd" use cases such as IAU planetary CRS.  Basically it is the web view of PROJ database.

hellik (approved):

nilason (dismissed):

lib/init/helptext.html
wenzeslaus: Thank you @jjimenezshaw and @rouault. That is the info I was asking for!

wenzeslaus: Extending my original comment, should the "See also" section link also https://crs-explorer.proj.org/ and/or https://epsg.org and/or https://epsg.io/?

neteler: (Partially) done in 78d8a8f685b8629b3944b1ca89e9f98cab1d1032

wenzeslaus (approved):
Thank you. This looks good to me now!

PR #4541, opened 2024-10-17
wxGUI: Fixed F405 in psmap/

arohanajit:
Following changes were made to fix F405 in `psmap/`
- removed `* imports`
- ran `isort` to sort all the imports
- `cmp` function was not working post doing explicit import with the error `gui/wxpython/psmap/dialogs.py:4127:47: F821 undefined name 'cmp'` - On looking more at this, I found this was removed in python3 and required an explicit comparison function. Hence I defined an explicit comparison function and referenced that instead.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -28,7 +28,7 @@ per-file-ignores =
     gui/wxpython/nviz/*: F841, E266, E722, F403, F405
     gui/wxpython/photo2image/*: F841, E722, E265
     gui/wxpython/photo2image/g.gui.photo2image.py: E501, F841
-    gui/wxpython/psmap/*: F841, E266, E722, F403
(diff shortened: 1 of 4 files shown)


petrasovaa (changes requested):
The cmp can be resolved simply by:
```
--- a/gui/wxpython/psmap/dialogs.py
+++ b/gui/wxpython/psmap/dialogs.py
@@ -53,7 +53,7 @@ else:
 
 import grass.script as gs
 
-from core.utils import PilImageToWxImage
+from core.utils import PilImageToWxImage, cmp
 from dbmgr.vinfo import VectorDBInfo
 from gui_core.gselect import Select
 from core.gcmd import RunCommand, GError, GMessage
```

petrasovaa (approved):

PR #4549, opened 2024-10-18
docs: Fixed E721 in gunittest/

arohanajit:
Fixed `E721` by using `isinstance()` for type checks instead of `==`
Also I went through some entries on `.flake8` which were in the file but weren't detected on the running the `flake8` check on there so I removed those as well


petrasovaa (changes requested):
Do you need to remove the invoker.py line from .flake8?

arohanajit:
Yes, I have updated the `.flake8`

petrasovaa (approved):

PR #4567, opened 2024-10-21
wxGUI: Fixed F405 in nviz/

arohanajit:
Fixed `F405` by specifying imports in `nviz/`


echoix (approved):
This looks pretty equivalent to the initial code, minus the star import. It will help static analysis (pyright for sure and maybe even CodeQL), especially since it's harder for them for anything that comes from the generated c types Python code, combined with start imports.

Issue #4218, opened 2024-08-23
[Feat] Add JSON output for columns in v.info


wenzeslaus:
**Is your feature request related to a problem? Please describe.**

The use of _c_ flag for _v.info_ should produce JSON with `format=json`:

```bash
grass --tmp-mapset ~/grassdata/nc_spm_08_grass7/ --exec v.info roadsmajor format=json -c
```

**Describe the solution you'd like**

JSON output which contains column names and metadata for these columns. Type is already in the current output, but there is more what could be included: other types of types (e.g., C versus SQL: [db_sqltype_to_Ctype](https://grass.osgeo.org/programming8/sqlCtype_8c.html#a1c02fe09d3c3d7632e4498494f3a1efc)) or some info on whether the column represents a number or whether it needs to be quoted in SQL (e.g., #1110).

**Describe alternatives you've considered**

The output could be a dictionary (mapping) or a list. In any case, the format should be extensible so more info can be included in the future. If a single column is a dictionary, then there is no problem with extending that. The question is if the list of columns should be in a dictionary.

Taking route of maximum flexibility:

```json
{
  "columns": [
    {
      "name": "cat",
      "sql_type": "INTEGER",
    },
    {
      "name": "MAJORRDS_",
      "sql_type": "DOUBLE PRECISION",
    }
  ]
}
```

Flexible only for the individual columns (simpler, but (perhaps?) less common list as a first level):

```json
[
  {
    "name": "cat",
    "sql_type": "INTEGER",
  },
  {
    "name": "MAJORRDS_",
    "sql_type": "DOUBLE PRECISION",
  }
]
```

Shortest possible:

```json
[
  {
    "name": "cat",
    "type": "INTEGER",
  },
  {
    "name": "MAJORRDS_",
    "type": "DOUBLE PRECISION",
  }
]
```

I don't think we should do this:

```json
{
  "names": ["cat", "MAJORRDS_"],
  "types": ["INTEGER", "DOUBLE PRECISION"]
}
```

**Additional context**

The output should be similar, compatible and/or in sync with _v.db.select_:

```sh
$ grass --tmp-mapset ~/grassdata/nc_spm_08_grass7/ --exec v.db.select roadsmajor format=json | jq .info
```

```json
{
  "columns": [
    {
      "name": "cat",
      "sql_type": "INTEGER",
      "is_number": true
    },
    {
      "name": "MAJORRDS_",
      "sql_type": "DOUBLE PRECISION",
      "is_number": true
[...]

PR #4590, opened 2024-10-25
v.info: add json output for columns

NishantBansal2003:
Use parson to add json support to v.info for columns. The JSON output looks like as follows:
```json
{
    "columns": [
        {
            "name": "cat",
            "sql_type": "INTEGER",
            "is_number": true
        },
        {
            "name": "MAJORRDS_",
            "sql_type": "DOUBLE PRECISION",
            "is_number": true
        },
        {
            "name": "ROAD_NAME",
            "sql_type": "CHARACTER",
            "is_number": false
        },
        {
            "name": "MULTILANE",
            "sql_type": "CHARACTER",
            "is_number": false
        },
        {
            "name": "PROPYEAR",
            "sql_type": "INTEGER",
            "is_number": true
        },
        {
            "name": "OBJECTID",
            "sql_type": "INTEGER",
            "is_number": true
        },
        {
            "name": "SHAPE_LEN",
            "sql_type": "DOUBLE PRECISION",
            "is_number": true
        }
    ]
}
```
fixes: https://github.com/OSGeo/grass/issues/4218

### DIFF
--- a/vector/v.info/print.c
+++ b/vector/v.info/print.c
@@ -208,29 +208,22 @@ void print_columns(struct Map_info *Map, const char *input_opt,
     driver = db_start_driver(fi->driver);
     if (driver == NULL) {
         Vect_close(Map);
-        const char *driver_name = fi->driver;
-        Vect_destroy_field_info(fi);
-        G_fatal_error(_("Unable to open driver <%s>"), driver_name);
+        G_fatal_error(_("Unable to open driver <%s>"), fi->driver);
     }
     db_init_handle(&handle);
     db_set_handle(&handle, fi->database, NULL);
     if (db_open_database(driver, &handle) != DB_OK) {
-        const char *database_name = fi->database;
-        const char *driver_name = fi->driver;
-        Vect_destroy_field_info(fi);
         db_shutdown_driver(driver);
         Vect_close(Map);
         G_fatal_error(_("Unable to open database <%s> by driver <%s>"),
-                      database_name, driver_name);
+                      fi->database, fi->driver);
     }
     db_init_string(&table_name);
     db_set_string(&table_name, fi->table);
     if (db_describe_table(driver, &table_name, &table) != DB_OK) {
-        const char *table_name = fi->table;
-        Vect_destroy_field_info(fi);
         db_close_database_shutdown_driver(driver);
         Vect_close(Map);
-        G_fatal_error(_("Unable to describe table <%s>"), table_name);
+        G_fatal_error(_("Unable to describe table <%s>"), fi->table);
     }
 
     JSON_Value *root_value = NULL, *columns_value = NULL, *column_value = NULL;
@@ -286,7 +279,6 @@ void print_columns(struct Map_info *Map, const char *input_opt,
         serialized_string = json_serialize_to_string_pretty(root_value);
         if (serialized_string == NULL) {
             json_value_free(root_value);
-            Vect_destroy_field_info(fi);
             db_close_database_shutdown_driver(driver);
             Vect_close(Map);
             G_fatal_error(_("Failed to initialize pretty JSON string."));
--- a/vector/v.info/print.c
+++ b/vector/v.info/print.c
@@ -189,7 +189,7 @@ void print_columns(struct Map_info *Map, const char *input_opt,
     num_dblinks = Vect_get_num_dblinks(Map);
 
     if (num_dblinks <= 0) {
-        Vect_close(&Map);
+        Vect_close(Map);
         G_fatal_error(
             _("Database connection for map <%s> is not defined in DB file"),
             input_opt);
@@ -200,21 +200,21 @@ void print_columns(struct Map_info *Map, const char *input_opt,
               field_opt);
 
     if ((fi = Vect_get_field2(Map, field_opt)) == NULL) {
-        Vect_close(&Map);
+        Vect_close(Map);
         G_fatal_error(
             _("Database connection not defined for layer <%s> of <%s>"),
             field_opt, input_opt);
     }
     driver = db_start_driver(fi->driver);
     if (driver == NULL) {
-        Vect_close(&Map);
+        Vect_close(Map);
         G_fatal_error(_("Unable to open driver <%s>"), fi->driver);
     }
     db_init_handle(&handle);
     db_set_handle(&handle, fi->database, NULL);
     if (db_open_database(driver, &handle) != DB_OK) {
         db_shutdown_driver(driver);
-        Vect_close(&Map);
+        Vect_close(Map);
         G_fatal_error(_("Unable to open database <%s> by driver <%s>"),
                       fi->database, fi->driver);
     }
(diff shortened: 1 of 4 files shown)


vector/v.info/print.c
```diff
@@ -215,13 +215,58 @@ void print_columns(struct Map_info *Map, const char *input_opt,
     if (db_describe_table(driver, &table_name, &table) != DB_OK)
         G_fatal_error(_("Unable to describe table <%s>"), fi->table);
 
+    JSON_Value *root_value = json_value_init_object();
+    JSON_Object *root_object = json_object(root_value);
+
+    JSON_Value *columns_value = json_value_init_array();
+    JSON_Array *columns_array = json_array(columns_value);
```
wenzeslaus, line 222: Init these to NULL and set them up only if needed after line 224.

vector/v.info/print.c
```diff
@@ -215,13 +215,58 @@ void print_columns(struct Map_info *Map, const char *input_opt,
...
+                        db_get_column_sqltype(db_get_table_column(table, col))),
+                    db_get_column_name(db_get_table_column(table, col)));
+        }
+    }
+
+    if (format == JSON) {
+        char *serialized_string = NULL;
+        serialized_string = json_serialize_to_string_pretty(root_value);
+        if (serialized_string == NULL) {
+            G_fatal_error(_("Failed to initialize pretty JSON string."));
```
wenzeslaus, line 263: Looking at #4533 and #4507, I wonder if more careful resource handling is needed in these cases. It is G_fatal_error, so basically exit for v.info, so I'm fine with merging this as is, but perhaps we want to set up some best practices here. @nilason and @aajit232, can you please comment? Feel free to add your 2 cents, @NishantBansal2003.

wenzeslaus (changes requested):
Looks good and it has a test for most of the functionality. Thanks!

vector/v.info/print.c
NishantBansal2003: I agree with your suggestion for improved resource handling. I think we should free the resources to prevent any potential leaks. If you'd like, I can modify the code to free resources upon encountering a `G_fatal_error`.

wenzeslaus: Sure, sounds good!

vector/v.info/print.c
```diff
@@ -198,30 +199,92 @@ void print_columns(struct Map_info *Map, const char *input_opt,
...
+        json_object_set_value(root_object, "columns", columns_value);
+    }
 
     ncols = db_get_table_number_of_columns(table);
-    for (col = 0; col < ncols; col++)
-        fprintf(stdout, "%s|%s\n",
-                db_sqltype_name(
-                    db_get_column_sqltype(db_get_table_column(table, col))),
+    for (col = 0; col < ncols; col++) {
+        if (format == JSON) {
```
nilason, line 244: Please consider using `switch` here and in similar cases. Especially, when we have an enum already in `format`. It leads in my opinion to clearer code and have the advantage in that compilers may detect flaws (eg. -Wswitch-enum).

vector/v.info/print.c
```diff
@@ -198,30 +199,92 @@ void print_columns(struct Map_info *Map, const char *input_opt,
...
+        Vect_close(Map);
         G_fatal_error(_("Unable to open database <%s> by driver <%s>"),
                       fi->database, fi->driver);
+    }
     db_init_string(&table_name);
     db_set_string(&table_name, fi->table);
-    if (db_describe_table(driver, &table_name, &table) != DB_OK)
+    if (db_describe_table(driver, &table_name, &table) != DB_OK) {
+        db_close_database(driver);
+        db_shutdown_driver(driver);
```
nilason, line 225: This can be simplified by  [db_close_database_shutdown_driver()](https://grass.osgeo.org/programming8/defs_2dbmi_8h.html#a671c7691d0fc81620a1831b6a5684360).

vector/v.info/print.c
```diff
@@ -198,30 +199,92 @@ void print_columns(struct Map_info *Map, const char *input_opt,
                 "layer <%s>:"),
               field_opt);
 
-    if ((fi = Vect_get_field2(Map, field_opt)) == NULL)
+    if ((fi = Vect_get_field2(Map, field_opt)) == NULL) {
```
nilason, line 202: `fi` is allocated and needs to be freed, before return after this if statement.

nilason (commented):
Some minor comments.

vector/v.info/print.c
nilason: In general I'd say it is good to closing database connections and similar (?!) before exit. Otherwise (strings etc.) not necessary.

nilason: > and similar (?!)

which includes closing open files.

nilason: Reading my above comments, it was perhaps not clearly expressed. In the case of `G_fatal_error()` it is definitely better to let the system free any in-memory objects like strings, structs etc. but close any file or db connections.  In this particular case only  `db_close_database_shutdown_driver()` and `Vect_close()` should be called.

Freeing in-memory objects before an `exit(EXIT_FAILURE)` which is the result of a call to G_fatal_error, only leads to unnecessary complicated code. (And competent static analysers do not complain on these cases either).

Another problem, if we take the following as an example:
```c
    const char *driver_name = fi->driver;
    Vect_destroy_field_info(fi);
    G_fatal_error(_("Unable to open driver <%s>"), driver_name);
```

the string `driver_name` points to will be freed by `Vect_destroy_field_info` causing the inclusion in the G_fatal_error message to fail.

NishantBansal2003: Got it, done...

vector/v.info/print.c
```diff
@@ -198,32 +199,97 @@ void print_columns(struct Map_info *Map, const char *input_opt,
...
+    }
 
     ncols = db_get_table_number_of_columns(table);
-    for (col = 0; col < ncols; col++)
-        fprintf(stdout, "%s|%s\n",
-                db_sqltype_name(
-                    db_get_column_sqltype(db_get_table_column(table, col))),
+    for (col = 0; col < ncols; col++) {
+        switch (format) {
+        case SHELL:
```
nilason, line 244: Wasn’t `SHELL` output originally the same as `PLAIN`?

NishantBansal2003: Previously, there was no output format defined for columns. I assumed the default output to be `PLAIN`, but if you would like the `SHELL` output format to be the same as `PLAIN`, I can add that as well?

nilason: Ok, there is no shell variant of column output.

nilason:
As `print_columns()` creates its own json objects, the following part in main:

https://github.com/OSGeo/grass/blob/24469d8dcb73fdcec1c59713de4a8c82ce3eea3e/vector/v.info/main.c#L57-L60

may be moved down, before:

https://github.com/OSGeo/grass/blob/24469d8dcb73fdcec1c59713de4a8c82ce3eea3e/vector/v.info/main.c#L95-L96

nilason:
Looks good to me now.

NishantBansal2003:
Could we move forward with this PR?
CC: @cwhite911 

Any suggestions are welcome.

petrasovaa (approved):

PR #4612, opened 2024-10-29
lib/vector/Vlib: Fix Resource Leak issue in break_polygons.c

ShubhamDesai:
This pull request fixes issue identified by Coverity scan (CID: 1207832).
**Changes Made:**
- Each call to open(filename, O_RDWR | O_CREAT | O_EXCL, 0600) now checks if the file descriptor (fd or xpntfd) was successfully created.
- If file creation fails, G_fatal_error() is invoked, and the temporary filename is freed before the function exits.
- If xpntfd creation fails, it also cleans up by destroying the RTree and closing fd.
- Also there is a check for remove(filename). I have used G_warning over there to print message. Not sure whether to use G_fatal or G_warning.

### DIFF
--- a/lib/vector/Vlib/break_polygons.c
+++ b/lib/vector/Vlib/break_polygons.c
@@ -20,6 +20,7 @@
 #include <unistd.h>
 #include <math.h>
 #include <errno.h>
+#include <string.h>
 #include <grass/vector.h>
 #include <grass/glocale.h>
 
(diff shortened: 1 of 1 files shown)


lib/vector/Vlib/break_polygons.c
```diff
@@ -130,12 +131,30 @@ void Vect_break_polygons_file(struct Map_info *Map, int type,
...
+    }
+    G_free(filename);
 
     filename = G_tempfile();
     xpntfd = open(filename, O_RDWR | O_CREAT | O_EXCL, 0600);
-    remove(filename);
+    if (xpntfd < 0) {
+        RTreeDestroyTree(RTree);
+        close(fd);
+        G_free(filename);
```
nilason, line 150:
```suggestion
        close(RTree->fd);
```

No need to free in-memory objects in case of G_fatal_error, the system will take care of it.

lib/vector/Vlib/break_polygons.c
```diff
@@ -130,12 +131,30 @@ void Vect_break_polygons_file(struct Map_info *Map, int type,
...
+    if (fd < 0) {
+        G_free(filename);
+        G_fatal_error(_("Failed to create temporary file: %s"),
+                      strerror(errno));
+    }
     RTree = RTreeCreateTree(fd, 0, 2);
-    remove(filename);
+    if (remove(filename) != 0) {
+        G_warning(_("Failed to remove temporary file: %s"), filename);
+    }
```
nilason, line 142:
```suggestion
    (void)remove(filename);
```

As the above open may fail, this may too.

nilason (changes requested):
Some suggestions.
And please address CID 1207831 too with this PR.

ShubhamDesai:
> Some suggestions. And please address CID 1207831 too with this PR.

Done

lib/vector/Vlib/break_polygons.c
```diff
@@ -130,12 +131,30 @@ void Vect_break_polygons_file(struct Map_info *Map, int type,
 
     filename = G_tempfile();
     fd = open(filename, O_RDWR | O_CREAT | O_EXCL, 0600);
+    if (fd < 0) {
+        G_free(filename);
+        G_fatal_error(_("Failed to create temporary file: %s"),
+                      strerror(errno));
+    }
```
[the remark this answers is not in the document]
ShubhamDesai: Done

lib/vector/Vlib/break_polygons.c
ShubhamDesai: Done

lib/vector/Vlib/break_polygons.c
ShubhamDesai: Done

nilason (approved):
Thanks!

PR #4618, opened 2024-10-30
v.report: Updated instance checks

arohanajit:
used `isinstance()` instead of `==` for type checks


echoix (approved):
The change is fine, it's correct and better Python than before, but I'm not sure why the code looked only for the first item if it is an int and boxes it into a list.

PR #4650, opened 2024-11-05
lib/gis: Fix out of scope memory access error in file_name function call

ymdatta:
When execution takes else path, pname, a pointer, is set to point to a local variable array which has limited scope. This same pointer is accessed outside of the block containing the local variable, essentially creating a scenario where we are accessing memory outside its score, which is undefined behavior.

Move the variable array out of the loop, so that it has the same scope as pname.

This was found using cppcheck static analysis tool.

### DIFF
--- a/lib/gis/file_name.c
+++ b/lib/gis/file_name.c
@@ -161,13 +161,13 @@ char *file_name(char *path, const char *dir, const char *element,
                 const char *name, const char *mapset, const char *base)
 {
     const char *pname = name;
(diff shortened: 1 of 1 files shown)


nilason:
Seems to me that moving the declaration `char xname[GNAME_MAX];` to the head of the function solves all related issues.

ymdatta:
Thanks for the suggestion @nilason. It's much cleaner and effective way to solve the issue. I didn't think of it at all! :)

nilason (approved):

PR #4659, opened 2024-11-06
CI: create versioned directory name of extracted tarball

nilason:
Extracting the tarball will with this create a containing directory named 'grass-x.y.z', where x, y, z stands for major, minor, micro/patch version.

Previously, the tarball created by CI was extracted to a directory named 'grass', which was a regression to accustomed and expected behaviour.

### DIFF
--- a/.github/workflows/create_release_draft.yml
+++ b/.github/workflows/create_release_draft.yml
@@ -63,7 +63,8 @@ jobs:
         run: |
           cd ..
           tar -cvf ${{ env.OUT_DIR }}/${{ env.GRASS }}.tar \
-            --exclude=".gi*" --exclude=".tr*" grass
+            --exclude=".gi*" --exclude=".tr*" \
+            --transform s/grass/${{ env.GRASS }}/ grass
           cd ${{ env.OUT_DIR }}
           gzip -9k ${{ env.GRASS }}.tar
           md5sum ${{ env.GRASS }}.tar.gz > ${{ env.GRASS }}.tar.gz.md5


nilason:
Now, I'm not sure whether this need to be backported at all. Isn't it the file in main that is triggered by addition of a tag?

neteler:
> Extracting the tarball

Just for clarification: where (in the CI?) is the tarball created? Because the release tarballs contain the version in the directory name.

nilason:
> > Extracting the tarball
> 
> Just for clarification: where (in the CI?) is the tarball created? Because the release tarballs contain the version in the directory name.

The tarball is created in CI and added to the Assets section of a release (e.g. https://github.com/OSGeo/grass/releases/tag/8.4.0). That tarball is/was named correctly, e.g. `grass-8.4.0.tar.gz`,  but the same tarball extracted a containing directory named only `grass` (because of the git repo name is "grass"). With this change the container directory will be named `grass-8.4.0`.

This/these tarballs may (or rather should, as `include/VERSION_GIT` file is generated into it) be uploaded to the osgeo download server manually (in the future possibly automatically).

See discussion in [GRASS-dev ML](https://lists.osgeo.org/pipermail/grass-dev/2024-July/096320.html).

neteler:
> See discussion in [GRASS-dev ML](https://lists.osgeo.org/pipermail/grass-dev/2024-July/096320.html).

Oh, I had completely forgotten about this one... and it would affect the upcoming GRASS GIS 8.4.1.

nilason:
> > See discussion in [GRASS-dev ML](https://lists.osgeo.org/pipermail/grass-dev/2024-July/096320.html).
> 
> Oh, I had completely forgotten about this one... and it would affect the upcoming GRASS GIS 8.4.1.

I didn't :-), and yes, that's why I hurried up with this now.

neteler (approved):
I cannot really judge the code but full trust.

nilason:
Backported to 8.4 with eb9cd95.

PR #4678, opened 2024-11-10
ps.map: Fix Resource Leak issue in ps_vlegend.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207603, 1207606).

### DIFF
--- a/ps/ps.map/ps_vlegend.c
+++ b/ps/ps.map/ps_vlegend.c
@@ -280,6 +280,14 @@ int PS_vlegend(void)
 
     if (PS.min_y > y)
         PS.min_y = y;
-
+    if (nvec)
+        G_free(nvec);
+    if (vec) {
+        for (i = 0; i < vector.count; i++) {
+            if (vec[i])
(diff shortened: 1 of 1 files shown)


nilason (changes requested):
No need to check for `NULL` before `free()`, in case of NULL pointer free() does nothing.

See e.g. https://medium.com/@Code_Analysis/shall-we-check-pointer-for-null-before-calling-free-function-63a27bb2e75e for a summary.

ShubhamDesai:
> No need to check for `NULL` before `free()`, in case of NULL pointer free() does nothing.
> 
> See e.g. https://medium.com/@Code_Analysis/shall-we-check-pointer-for-null-before-calling-free-function-63a27bb2e75e for a summary.
> 
> (Update: the same goes for C++ https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null).

Done

nilason (approved):

echoix (approved):

PR #4687, opened 2024-11-11
r.watershed: Fix Resource Leak issue

ShubhamDesai:
This pull request fixes issue identified by coverity scan (CID : 1207844)
Used G_free() to fix the issue.


nilason (dismissed):
I think this aims to address CID: 1207847, but please also address here 1207844 and 1207845.

metzm (approved):

PR #4702, opened 2024-11-14
lib/gis: Refactor va_list usage in debug module

ymdatta:
lib/gis: Refactor va_list usage in debug module

va_list() macro initializes va_list structure before it's usage,
and each va_list() call has to be accompanied by corresponding
va_end() macro call on the same va_list structure.

By having these macros directly before and after the va_list
structure usage, reduces the number of places we need to keep
track of whether the va_list structure was properly inintialized
or closed.

### DIFF
--- a/lib/gis/debug.c
+++ b/lib/gis/debug.c
@@ -72,8 +72,6 @@ int G_debug(int level, const char *msg, ...)
     G_init_debug();
 
     if (grass_debug_level >= level) {
-        va_start(ap, msg);
-
         filen = getenv("GRASS_DEBUG_FILE");
         if (filen != NULL) {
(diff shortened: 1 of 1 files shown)


metzm:
Start ap as late as possible and close it as early as possible ?
```
        va_start(ap, msg);
        vfprintf(fd, msg, ap);
        va_end(ap);
```
i.e. move `va_start()` and `va_end()` to be directly around `vfprintf()`

See also [error.c](https://github.com/OSGeo/grass/blob/main/lib/gis/error.c)

ymdatta:
@metzm : Thanks for the suggestion! I think that's a really clean and better way. Made the changes.

echoix (approved):

PR #4706, opened 2024-11-15
wxGUI: Fixed PEP8 errors in web_services/

arohanajit:
- FIxed `E402` rebasing module import to the top
- Fixed `F841` removing unused empty list


petrasovaa (commented):
Didn't test, but the imports likely won't work

arohanajit:
> Didn't test, but the imports likely won't work

I have imported modules through a function which allows me to set the path again without raising the flake8 warning

petrasovaa (commented):
I think I would prefer # noqa instead of the workaround here. Note that Signal can be imported normally, it's not part of r.in.wms.

petrasovaa (approved):

echoix (approved):

PR #4733, opened 2024-11-21
r.clump: added tests for r.clump module

Shreshth-Malik:
This PR introduces `pytest`-based tests for the `r.clump` module to validate its functionality. The tests cover:

Basic Functionality: Verifies that `r.clump` correctly identifies clumped regions in a raster map.
Diagonal Connectivity: Tests the `-d` flag for identifying clumps with diagonal connectivity.

The tests use temporary regions and custom raster maps.


raster/r.clump/tests/test_clump.py
```diff
@@ -0,0 +1,70 @@
...
+
+def test_clump_basic(setup_map):
+    """Test basic clumped map."""
+
+    run_command("r.clump", input="custom_map", output="clumped_map", overwrite=True)
+
+    output_maps = parse_command("g.list", type="raster")
+    assert "clumped_map" in output_maps, "Output raster map 'clumped_map' should exist"
+
+    category_output = read_command("r.category", map="clumped_map").strip().split("\n")
```
petrasovaa, line 41: r.category has json output, you can use parse_command to get the list

raster/r.clump/tests/test_clump.py
```diff
@@ -0,0 +1,70 @@
...
+    actual_categories = {
+        int(line.split("\t")[0]): line.split("\t")[1].strip() if "\t" in line else ""
+        for line in category_output
+    }
+
+    expected_categories = {1: "", 2: "", 3: ""}
```
petrasovaa, line 47: why dict with empty keys?

petrasovaa (changes requested):
Please also add test for minsize and threshold

raster/r.clump/tests/test_clump.py
Shreshth-Malik: The output for r.category came in the form of {'1': None, '2': None, '3': None}, the categories did not have any labels so i kept the expected categories values empty as well.

Shreshth-Malik:
@petrasovaa I was trying to test the `threshold` functionality with cells values `5` and `5.5` and keeping the `threshold 0.6`. From my understanding if the absolute difference the cell values is `<=threshold` value then it clumps them together, but in my case it is not doing so. Is my understanding wrong?

My map layout looks something like:
  `5`       `5.5`    NULL
 NULL    `5`      NULL
 NULL   `5.5`    `5.5`

echoix:
The error here, now that the test is actually picked up, has to do that a grass session wasn't set. Try to look at `with gs.setup.init(project, env=os.environ.copy()) as session:`, that you even used in one of your other PRs here: #4755. It should do the trick. Otherwise, if the env from the session can't be passed and is really needed, use the workaround I explained in #4784

raster/r.clump/tests/test_clump.py
petrasovaa: because r.category already has format option, the call is more complicated:
```
import json
gs.parse_command("r.category", map="clumped_map", output_format="json", parse=(json.loads, {}))
```

raster/r.clump/tests/test_clump.py
petrasovaa: Then you should test that the category label is empty as well.

petrasovaa (approved):

raster/r.clump/tests/test_clump.py
petrasovaa: This still needs to be done.

raster/r.clump/tests/test_clump.py
```diff
@@ -0,0 +1,117 @@
...
+        input="custom_map",
+        output="clumped_map",
+        overwrite=True,
+        env=session.env,
+    )
+
+    output_maps = gs.parse_command("g.list", type="raster", env=session.env)
+    assert "clumped_map" in output_maps, "Output raster map 'clumped_map' should exist"
+
+    category_output = gs.parse_command("r.category", map="clumped_map", env=session.env)
```
petrasovaa, line 27: See the note about parse_command

petrasovaa (changes requested):

petrasovaa:
@Shreshth-Malik this still needs to be finished, thanks!

raster/r.clump/tests/conftest.py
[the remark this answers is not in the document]
Shreshth-Malik: For this test, i won't be adding any more files, i kept those separate to maintain a clean structure but I can merge the files as well. Do you want me to do this for this test as well or just from next time?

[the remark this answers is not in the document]
echoix: If there's value in sharing the fixtures, placing them a conftest is fine. All conftest.py files in the hierarchy are loaded, and if the same fixtures are duplicated, the closest takes precedence.

The advantages I see in having it in the same file, apart from being easier for reviewers (to not have to open the file structure to know if there's another file to change), is for the import time (one less to import), but also to have more help from static checking tools. The conftest mechanism in pytest is kinda "magic" as our own code doesn't import it. These conftest files are more useful when we have to tweak pytest's behavior (advanced or custom collection) and use things like autouse fixture with non-pytest files that can't be changed (like what I'm tidying up recently for gunittest tests running from pytest; they still have to work the old way too).


So, if under these considerations, a fixture makes better sense in the same file as it is used and made for, then yes, in the future in the same file is the good choice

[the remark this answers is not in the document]
echoix: (This is when it for when it is single-purpose fixtures)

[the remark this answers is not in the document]
Shreshth-Malik: This helps a ton, I will keep this in mind and use fixtures accordingly. Maybe in the meantime I will try to experiment around with them understand them a bit better.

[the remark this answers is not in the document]
echoix: If Anna thinks that the contents of the tests are good, I don't mind at all to keep conftest though, if there were no changes left to do.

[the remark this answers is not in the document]
petrasovaa: Let's do it as you suggest, I didn't give it much thought.

petrasovaa (approved):

PR #4800, opened 2024-12-03
r.out.gdal: raster attribute table export enhancement

metzm:
This PR solves an old TODO by writing both labels and colors to a raster attribute table (RAT) with `r.out.gdal -t`.

Newer QGIS versions can now also read raster attribute tables and will use the labels and/or colors listed in a RAT.

Test with
```
g.region -p rast=landcover_1m
r.out.gdal in=landcover_1m out=landcover_1m.tif format=GTiff createopt="COMPRESS=DEFLATE" -t
gdalino landcover_1m.tif
```
in the NC sample dataset.

### DIFF
--- a/raster/r.out.gdal/attr.c
+++ b/raster/r.out.gdal/attr.c
@@ -50,7 +50,7 @@ int export_attr(GDALDatasetH hMEMDS, int band, const char *name,
      */
 
     if (cats.ncats > 0 && rcount > 0) {
-        int use_minmax = 0;
+        bool use_minmax = false;
         int r1, g1, b1, r2, g2, b2;
 
         /* merge categories and color rules:
@@ -62,8 +62,11 @@ int export_attr(GDALDatasetH hMEMDS, int band, const char *name,
         if (maptype == CELL_TYPE) {
             for (i = 0; i < cats.ncats; i++) {
                 label = Rast_get_ith_c_cat(&cats, i, &CellMin, &CellMax);
+                if (!label || !*label) {
+                    G_fatal_error(_("No label for category entry no %d "), i);
+                }
                 if (CellMin != CellMax) {
-                    use_minmax = 1;
+                    use_minmax = true;
                     break;
                 }
             }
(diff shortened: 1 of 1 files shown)


mmacata (approved):
Very cool, thanks!

neteler:
Works smoothly! I have tested it with a different dataset, both in GRASS GIS and QGIS:

image

neteler (approved):

raster/r.out.gdal/attr.c
```diff
@@ -47,12 +49,149 @@ int export_attr(GDALDatasetH hMEMDS, int band, const char *name,
...
+
+        /* merge categories and color rules:
+         * go through categories and fetch corresponding color
+         * write out value, label, red, green, blue
+         * or the minmax variant
+         */
+
+        if (maptype == CELL_TYPE) {
+            for (i = 0; i < cats.ncats; i++) {
+                label = Rast_get_ith_c_cat(&cats, i, &CellMin, &CellMax);
```
nilason, line 64: This looks very good to me, and a great addition!

I have only minor suggestions: (1) `label` which is set on L64 and L73 (and L199 and L208) are never used: the return value may just as well be ignored on these lines; (2) `use_minmax` would be better represented by a bool.

metzm: (1) done and (2) done. OK?

metzm: I understand, but with `(void)Rast_get_ith_c_cat()` you just trick the format checker to not complain about unused return codes. Therefore I think my proposed solution where the code now also checks if the call to `Rast_get_ith_c_cat()` was successful (there is someting in label) is the formally correct solution.

nilason: (This came up with `scan-build`, as "unused assignment" or something similar. Occasionally, I run it on new code or substantial updates).

I didn't check for possible erroring of `Rast_get_ith_c_cat()`, so with your update there's actually a code improvement (not only for better "code reading").

nilason (approved):
Thanks!

PR #4813, opened 2024-12-07
CQ: optimise if conditions in gmodeler.canvas

pesekon2:
no need to run the same `if` condition twice when the interluding one could be inserted in


echoix (approved):
Yep, it is equivalent ;)

PR #4865, opened 2024-12-20
manual: conversion of all HTML manual pages to markdown

neteler:
Test submission of conversion of all HTML manual pages to markdown using the `pandoc` based converter script (see #4620). Submitted without `pre-commit` to not be blocked.

For figure code conversion issues, see #4864


wenzeslaus:
Current state of conversion:

* #5057 does the adding of files now with the two-commit rename strategy from #4620#issuecomment-2465555707, so the actual Markdown update will be file modification.
* #5054 and #5057 modify filters so that the CI does not fail. #5054 improves the Markdown itself to pass some of the tests.

neteler:
@wenzeslaus shall this PR be closed in favor of #5057?

Issue #4593, opened 2024-10-25
[Feat] db.in.ogr should recognize column data types when importing CSV files


cmbarton:
**The problem**
CSV is one of the most commonly and widely used formats for exchanging tabular data across platforms. However, it is difficult to import csv files into GRASS. Currently, db.in.ogr imports csv files but transforms all columns to text. The suggested work around is to create an accompanying *.csvt file that specifies the data types for each column. This is very cumbersome, especially for files with many columns. But OGR will automatically recognize column data types if a file open option is set
```-oo AUTODETECT_TYPE=YES```
This makes importing csv files much easier. This can be manually entered into the gdal_oo argument in db.in.ogr but this is not mentioned in the manual. It would be even better if this is the default behavior of db.in.ogr

**Proposed solution**
-oo AUTODETECT_TYPE=YES should be the default for importing a csv file using db.in.ogr
a "csvt" flag should be added to enable a user to create a csvt file to override this default behavior

PR #4883, opened 2024-12-25
manual: db.in.ogr add column type detection for CSV

### LINKED ISSUE #4674: [Feat] Update db.in.ogr manual to improve import CSV files
### The problem
CSV is one of the most commonly and widely used formats for exchanging tabular data across platforms. However, it is difficult to import csv files into GRASS. Currently, db.in.ogr imports csv files but transforms all columns to text. In the current manual text, the suggested work around is to create an accompanying *.csvt file that specifies the data types for each column. This is very cumbersome, especially for files with many columns. But OGR will automatically recognize column data types if a file open option is set
`-oo AUTODETECT_TYPE=YES`
This makes importing csv files much easier. This can be manually entered into the **_gdal_oo_** argument in db.in.ogr but this is not mentioned in the current manual. 

### Proposed solution
While -oo AUTODETECT_TYPE=YES should be the default for importing a csv file using db.in.ogr (see #4593), until that change can be made, it would be very helpful to describe in the manual how this argument can be implemented as a workaround for now. The information in the current manual about using a *.csvt file should also be maintained for finer manual control. I propose the manual should be updated as follows: 

**Current:**
Import CSV file

Limited type recognition can be done for Integer, Real, String, Date, Time and DateTime columns through a descriptive file with same name as the CSV file, but .csvt extension (see details here).

NOTE: create koeppen_gridcode.csvt first for automated type recognition
```
db.in.ogr input=koeppen_gridcode.csv output=koeppen_gridcode
db.select table=koeppen_gridcode
```

**New:**
Import CSV file

db.in.ogr will attempt to automatically read input data types if the gdal_doo flag is set to 
AUTODETECT_TYPE=YES.
```
db.in.ogr input=koeppen_gridcode.csv output=koeppen_gridcode gdal_doo=AUTODETECT_TYPE=YES
db.select table=koeppen_gridcode
```
Users can also specify data types for CSV columns using a type definition file with same name as the CSV file, but *.csvt extension (see details here). Columns can be defined as Integer, Real, String, Date, Time and DateTime in this way.

NOTE: create koeppen_gridcode.csvt first for automated type recognition
```
db.in.ogr input=koeppen_gridcode.csv output=koeppen_gridcode
db.select table=koeppen_gridcode
```

neteler:
Improve CSV import example with GDAL related column type autodetection (suggested by @cmbarton).

Fixes #4674 (see also related issue #4593).

Additionally:
- minor HTML fixes in some manual pages

### DIFF
--- a/imagery/imageryintro.html
+++ b/imagery/imageryintro.html
@@ -209,11 +209,13 @@ GRASS supports the following methods:
   <li> Supervised classification (<a href="i.gensigset.html">i.gensigset</a>,
    <a href="i.smap.html">i.smap</a>)</li>
   </ul>
+  </li>
 <li> Object-oriented classification:
   <ul>
   <li> Unsupervised classification (segmentation based:
        <a href="i.segment.html">i.segment</a>)</li>
   </ul>
+  </li>
 </ul>
 
 Kappa statistic can be calculated to validate the results
--- a/imagery/imageryintro.html
+++ b/imagery/imageryintro.html
@@ -48,7 +48,7 @@ influences the signal as recorded at the sensor. This atmospheric
 interaction with the sun energy reflected back into space by
 ground/vegetation/soil needs to be corrected. The need of
 removing atmospheric artifacts stems from the fact that the
-atmosphericic conditions are changing over time. Hence, to gain
+atmospheric conditions are changing over time. Hence, to gain
 comparability between Earth surface images taken at different
 times, atmospheric need to be removed converting at-sensor values
 which are top of atmosphere to surface reflectance values.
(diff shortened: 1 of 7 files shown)


imagery/imageryintro.html
```diff
@@ -209,11 +209,13 @@ <h3>Image classification</h3>
   <li> Supervised classification (<a href="i.gensigset.html">i.gensigset</a>,
    <a href="i.smap.html">i.smap</a>)</li>
   </ul>
+  </li>
```
echoix, line 212: "Radiometric classification:" doesn't close it's li inside the top ul. I double checked it with indented formatting to find what was missing,

image

So, total 3 added lines are needed, we've got 2: 
image

echoix:
Otherwise it is good and ready to go

imagery/imageryintro.html
echoix: There's also a typo in that file I think in "atmosphericic" ->"atmospheric"?

neteler: Fixed in 0e9e3849b33a538c4621021b3896df9da9604fda

neteler: The page now passes https://validator.w3.org/

echoix (approved):

PR #4936, opened 2025-01-10
packaging: rename macos/app dir to macos/files

nilason:
Follow up to #4935, and final preparatory PR.


echoix (approved):
All these folders don't seem used yet, I'm eager to see the new workflow that you're migrating here ;)

nilason:
> All these folders don't seem used yet, I'm eager to see the new workflow that you're migrating here ;)

This part hasn't been used in a long time. There are some files I'd like to keep though. That's the reason for renaming the folders to keep their history.

PR #4957, opened 2025-01-15
v.to.rast: Fix Resource Leak issue in do_areas.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207750)
Used Vect_destroy_cats_struct() to fix this issue.

### DIFF
--- a/vector/v.to.rast/do_areas.c
+++ b/vector/v.to.rast/do_areas.c
@@ -94,14 +94,13 @@ int sort_areas(struct Map_info *Map, struct line_pnts *Points, int field,
     CELL cat;
 
     G_begin_polygon_area_calculations();
-    Cats = Vect_new_cats_struct();
 
     /* first count valid areas */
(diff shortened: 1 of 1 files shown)


vector/v.to.rast/do_areas.c
```diff
@@ -98,8 +98,10 @@ int sort_areas(struct Map_info *Map, struct line_pnts *Points, int field,
 
     /* first count valid areas */
     nareas = Vect_get_num_areas(Map);
-    if (nareas == 0)
+    if (nareas == 0) {
+        Vect_destroy_cats_struct(Cats);
```
nilason, line 102: If you move line 97 after the if statement, there will be no need to destroy here.

ShubhamDesai: Done

nilason (approved):

PR #4959, opened 2025-01-16
d.barscale: Fix Resource Leak issue in draw_scale.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1415718)

### DIFF
--- a/display/d.barscale/draw_scale.c
+++ b/display/d.barscale/draw_scale.c
@@ -634,8 +634,10 @@ int draw_scale(double east, double north, int length, int seg, int units,
     }
     D_stroke();
 
-    if (fontsize < 0)
+    if (fontsize < 0) {
+        G_free(label);
         return 0;
+    }    
 
     /* draw the distance + units text */
 
@@ -676,6 +678,7 @@ int draw_scale(double east, double north, int length, int seg, int units,
         D_pos_abs(x_pos + 5 - (tr - tl), y_pos + ysize / 2 + (tt - tb) / 2);
         D_text(label);
     }
+    G_free(label);
 
     return 0;
 }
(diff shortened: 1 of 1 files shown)


display/d.barscale/draw_scale.c
```diff
@@ -676,6 +678,7 @@ int draw_scale(double east, double north, int length, int seg, int units,
         D_pos_abs(x_pos + 5 - (tr - tl), y_pos + ysize / 2 + (tt - tb) / 2);
         D_text(label);
     }
+    G_free(label);
```
nilason, line 681: At line 99: `if (length == 0) {` is true then `label = scales[incr].symbol;`. Thus, in that case`label` is not "owned" by this function and shouldn't be freed.

ShubhamDesai: ok in that case i kept it only where it is highlighted in Coverity Scan. Thanks for pointing the mistake

nilason: They should be freed in both cases if `if (length != 0)`, because then it was malloced with `label = G_malloc(GNAME_MAX);`.

ShubhamDesai: i added the condition before end of function, is it correct now?

display/d.barscale/draw_scale.c
```diff
@@ -634,8 +634,10 @@ int draw_scale(double east, double north, int length, int seg, int units,
     }
     D_stroke();
 
-    if (fontsize < 0)
+    if (fontsize < 0) {
+        G_free(label);
```
nilason, line 638: You need to check `length` here too.

ShubhamDesai: Done

nilason (approved):

PR #4960, opened 2025-01-16
lib/vector/Vlib: Fix Resource Leak issue in open_ogr.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1588911)
Used fclose() to fix this issue

### DIFF
--- a/lib/vector/Vlib/open_ogr.c
+++ b/lib/vector/Vlib/open_ogr.c
@@ -148,13 +148,13 @@ int V2_open_old_ogr(struct Map_info *Map)
 
     G_debug(3, "V2_open_old_ogr(): name = %s mapset = %s", Map->name,
             Map->mapset);
-    char *map_name = NULL;
+    const char *map_name = NULL;
 
     if (Vect_open_fidx(Map, &(Map->fInfo.ogr.offset)) != 0) {
         map_name = Vect_get_full_name(Map);
         G_warning(_("Unable to open feature index file for vector map <%s>"),
                   map_name);
-        G_free(map_name);
+        G_free((void *)map_name);
         G_zero(&(Map->fInfo.ogr.offset), sizeof(struct Format_info_offset));
     }
 
@@ -271,10 +271,10 @@ int Vect_open_fidx(struct Map_info *Map, struct Format_info_offset *offset)
     dig_file_init(&fp);
     fp.file = G_fopen_old(elem, GV_FIDX_ELEMENT, Map->mapset);
     if (fp.file == NULL) {
-        char *map_name = Vect_get_full_name(Map);
+        const char *map_name = Vect_get_full_name(Map);
         G_debug(1, "unable to open fidx file for vector map <%s>",
                 map_name);
-        G_free(map_name);
+        G_free((void *)map_name);
         return -1;
     }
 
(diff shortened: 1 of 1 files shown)


nilason:
Please address CID 1207813 and 1207814 here too.

ShubhamDesai:
> Please address CID 1207813 and 1207814 here too.

Done addressed.

lib/vector/Vlib/open_ogr.c
```diff
@@ -148,10 +148,13 @@ int V2_open_old_ogr(struct Map_info *Map)
 
     G_debug(3, "V2_open_old_ogr(): name = %s mapset = %s", Map->name,
             Map->mapset);
+    const char *map_name = NULL;
 
     if (Vect_open_fidx(Map, &(Map->fInfo.ogr.offset)) != 0) {
+        map_name = Vect_get_full_name(Map);
```
nilason, line 154: I think it would be better here to declare `map_name` and initialise it at the same time (as you did at L274), thus saving a line of code and limit the scope.

lib/vector/Vlib/open_ogr.c
```diff
@@ -148,10 +148,13 @@ int V2_open_old_ogr(struct Map_info *Map)
...
         G_warning(_("Unable to open feature index file for vector map <%s>"),
-                  Vect_get_full_name(Map));
+                  map_name);
+        G_free((void *)map_name);
```
nilason, line 157: No need to cast here, is there?

lib/vector/Vlib/open_ogr.c
```diff
@@ -268,14 +271,17 @@ int Vect_open_fidx(struct Map_info *Map, struct Format_info_offset *offset)
     dig_file_init(&fp);
     fp.file = G_fopen_old(elem, GV_FIDX_ELEMENT, Map->mapset);
     if (fp.file == NULL) {
-        G_debug(1, "unable to open fidx file for vector map <%s>",
-                Vect_get_full_name(Map));
+        const char *map_name = Vect_get_full_name(Map);
+        G_debug(1, "unable to open fidx file for vector map <%s>", map_name);
+        G_free((void *)map_name);
```
nilason, line 276: Ditto (cast?).

nilason (changes requested):
Looks good to me, just some minor suggestions.

lib/vector/Vlib/open_ogr.c
ShubhamDesai: done

lib/vector/Vlib/open_ogr.c
ShubhamDesai: without casting it failed so i referred previous PRs in which CIs passed https://github.com/OSGeo/grass/pull/4533/files

lib/vector/Vlib/open_ogr.c
ShubhamDesai: same

lib/vector/Vlib/open_ogr.c
nilason: I assume it is the `const` causing warnings without cast. Good, just wanted to be sure, thanks!

nilason (approved):

PR #4965, opened 2025-01-18
v.to.db: Fix Resource Leak issue in lines.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207675, 1207676, 1207677, 1207678, 1207679)
Used Vect_destroy_line_struct(), Vect_destroy_cats_struct() to fix this issue.

### DIFF
--- a/vector/v.to.db/lines.c
+++ b/vector/v.to.db/lines.c
@@ -264,6 +264,11 @@ int read_lines(struct Map_info *Map)
         }
         G_percent(line_num, nlines, 2);
     }
+    Vect_destroy_line_struct(EndPoints);
+    Vect_destroy_cats_struct(RCats);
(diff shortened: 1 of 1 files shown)


metzm:
`read_lines()` is called only once in `main()`, thus there is no resource leak (danger of OOM errors).

ShubhamDesai:
> `read_lines()` is called only once in `main()`, thus there is no resource leak (danger of OOM errors).

I understand that read_lines() is only called once in main(), so there’s no immediate OOM risk. However, I included the memory deallocation to ensure clean code and prevent potential issues if this function is ever used in a loop or called multiple times in the future.

nilason (approved):
While absolutely true, there was no current risk of resource leak(s), this is correct and eliminates issues reported with static analysers.

PR #4969, opened 2025-01-21
r.sim.water: build topology for vector output

metzm:
Topology was not built for the vector output `walkers_output`. This is within GRASS not a problem, but the QGIS interface fails because a vector without topology can not be exported with `v.out.ogr`, consequently QGIS complains about a missing output. With this PR, topology is built for the vector output which should not have any effects otherwise because it is only a points vector with simple topology.


neteler:
FYI
At time our CI fails due to "Bitbucket is experiencing an incident" leading the step `scripts/g.extension/testsuite/test_addons_download.py` to fail.

neteler (approved):
Looks good to me as it fixes a compatibility issue to other software.

PR #4977, opened 2025-01-22
lib/ogsf: Fix Resource Leak issue in gp3.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207703, 1207704, 1415564).
Used G_free(), db_close_database_shutdown_driver(), Vect_destroy_field_info() to fix the issue.


lib/ogsf/gp3.c
```diff
@@ -295,5 +297,7 @@ int Gp_load_sites_thematic(geosite *gp, struct Colors *colors)
             _("%d points without category. "
               "Unable to determine color rules for features without category."),
             nskipped);
+    db_close_database_shutdown_driver(driver);
+    Vect_destroy_field_info(Fi);
```
metzm, line 301:
```suggestion
    if (driver) {
        db_close_database_shutdown_driver(driver);
        Vect_destroy_field_info(Fi);
    }
```

The db driver might not be available, see L202.

Also, the calls to `db_select_value()` above need to be protected with if statements if there is no database connection (lines 235ff).

lib/ogsf/gp3.c
```diff
@@ -241,13 +243,15 @@ int Gp_load_sites_thematic(geosite *gp, struct Colors *colors)
                                 ((int)((blu) << 16) & BLU_MASK);
         }
         if (gp->tstyle->color_column) {
-            nvals = db_select_value(driver, Fi->table, Fi->key, cat,
+            if (driver) {
```
metzm, line 246: `driver` must be initialized to NULL, e.g. with
```
dbDriver *driver = NULL;
```
at L184. There is no guarantee that with `dbDriver *driver;` `driver` is set to NULL.

ShubhamDesai: done

nilason:
There are also related CID:1207706 and 1207707 for this file, which would be good to address here.

(You can filter issues by file location under "Edit settings" (cogwheel icon), add e.g. /lib/ogsf/gp3.c` for "File".)

lib/ogsf/gp3.c
nilason: > Also, the calls to db_select_value() above need to be protected with if statements if there is no database connection (lines 235ff).

- If `Fi` fails to be set (L201ff), then `driver` is not set either: both remain NULL.
- If `Fi` is successfully set, but `driver` fails to open leads to G_fatal_error.

So,  both `Fi` and `driver` are either NULL or !NULL from L211.

With `driver` and `Fi` uninitiated (NULL), `db_select_value` returns  `-1` and continues the loop:

```c
            nvals = db_select_value(driver, Fi->table, Fi->key, cat,
                                    gp->tstyle->color_column, &value);
            if (nvals < 1)
                continue;
```

Checking for `driver` is not necessary as it works as is. There's no doubt this function may be improved for clarity, but that change – as far as I understand – is not needed for the initial aim of this PR.

lib/ogsf/gp3.c
```diff
@@ -295,5 +304,9 @@ int Gp_load_sites_thematic(geosite *gp, struct Colors *colors)
             _("%d points without category. "
               "Unable to determine color rules for features without category."),
             nskipped);
+    if (driver) {
+        db_close_database_shutdown_driver(driver);
+        Vect_destroy_field_info(Fi);
```
nilason, line 309: The call of `Vect_destroy_field_info(Fi);` can be lifted out of the `if (driver)` statement.
(If `Fi` is NULL Vect_destroy_field_info() silently returns, no need to check).

ShubhamDesai: done changes

PR #4980, opened 2025-01-23
wxGUI: Fixed bare except clause in mapwin/base.py

arohanajit:
Added general `Exception` since `handler` can raise multiple different exceptions


echoix (approved):

arohanajit:
> What about the unrelated changes?

I am not sure why `wxdisplay` is here, I will delete that file

PR #4991, opened 2025-01-25
v.net: Fix Resource Leak issue in arcs.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207757)
Removed the overwritten variable points.


metzm (approved):
Thanks for the fix!

PR #4992, opened 2025-01-25
lib/gmath: Fix Resource Leak issue in solvers_classic_iter.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207773, 1207774 )


metzm (approved):
Thanks for the fix!

PR #4993, opened 2025-01-25
r.water.outlet: Fix Resource Leak issue in over_cells.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207897)


metzm (approved):
Thanks for the fix!

PR #5003, opened 2025-01-27
init: Use mask, not MASK in shell prompt

wenzeslaus:


echoix (approved):
We're also way overdue of regenerating the translation files, plus unblocking weblate (as these new changed strings won't be propagated in other languages, it would use the fallback in English).

PR #5012, opened 2025-01-30
v.in.dxf: Fix Resource Leak issue in read_dxf.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207848)


nilason (approved):
In my opinion a bit too defensive to check for results of G_malloc and G_store here, but not wrong.

PR #5024, opened 2025-02-02
Vcluster

manavkamdar27:


neteler:
Hi @manavkamdar27 

It is highly recommended to install and use `pre-commit` before submitting any new or modified code or any other content. The pre-commit uses Git hooks to check validity and executes automated formatting for a range of file formats, including C/C++ and Python. Pre-commit installs all necessary tools in a virtual environment upon first use.

See: https://github.com/OSGeo/grass/blob/main/doc/development/style_guide.md#using-pre-commit

PR #5032, opened 2025-02-03
lib/nviz: Fix Resource Leak issue in exag.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207695)

### DIFF
--- a/lib/nviz/exag.c
+++ b/lib/nviz/exag.c
@@ -63,6 +63,8 @@ int Nviz_get_exag_height(double *val, double *min, double *max)
 
     G_debug(1, "Nviz_get_exag_height(): value = %f min = %f max = %f", *val,
             min ? *min : 0.0, max ? *max : 0.0);
+    if (nsurfs > 0)
+        G_free(surf_list);
 
     return 1;
 }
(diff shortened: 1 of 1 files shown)


lib/nviz/exag.c
```diff
@@ -63,6 +63,8 @@ int Nviz_get_exag_height(double *val, double *min, double *max)
 
     G_debug(1, "Nviz_get_exag_height(): value = %f min = %f max = %f", *val,
             min ? *min : 0.0, max ? *max : 0.0);
+    if (nsurfs > 0)
```
nilason, line 66: No need for this check, but you should (for good measure) explicitly initialise `surf_list` to NULL.
(If surf_list is NULL, G_free does nothing.)

ShubhamDesai: done changes.

lib/nviz/exag.c
```diff
@@ -63,6 +64,8 @@ int Nviz_get_exag_height(double *val, double *min, double *max)
...
+        G_free(surf_list);
```
nilason, line 68: This is what I meant:

```suggestion
    G_free(surf_list);
```

ShubhamDesai: sorry i got confused with the word "should".
changes done in both the functions

nilason: No worries, maybe I wasn't clear.
(GS_get_surf_list returns malloced array only if `nsurfs > 0`, otherwise NULL, so the check was really about that. Remember it is safe to free() a NULL value.)

nilason (approved):

PR #5042, opened 2025-02-04
g.proj: fix resource leaks

nilason:
Fix resource leaks in `g.proj` as reported by Coverity Scan.


echoix (approved):
I assume G_free is used correctly, with the members of dstruct individually. Are all these the correct patterns I should learn from?

nilason:
> I assume G_free is used correctly, with the members of dstruct individually. Are all these the correct patterns I should learn from?

In this case the dstruct is allocated on stack (and freed when going out of scope), but (some of) its members are malloced strings, which should be explicitly freed.

PR #5051, opened 2025-02-05
utils: Fixed ruff errors in utils/

arohanajit:
Fixed multiple ruff errors across all the files in `utils/`


echoix (approved):
I assume there are no changes other than the indentation in gitlog2changelog

echoix (approved):
I double-checked on a computer and I don't need to assume anything to approve :) Thanks!

PR #5054, opened 2025-02-05
docs: HTML to Markdown lua filters

cwhite911:
This PR is a continuation of #4620.

It's adding additional pandoc lua filters to address remaining issues caught by the markdownlint-cli.

Markdownlint Rules Docs: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md
Pandoc Lua Filters Docs: https://pandoc.org/lua-filters.html#module-pandoc.utils


petrasovaa:
This doesn't seem to address the problem with figures (#4864).
Also d.graph seemed to get broken:
```
## COMMANDS

The graphics language is simple, and uses the following commands:

\[ [\#](#comment) | [move](#move) | [draw](#draw) | [polygon](#polygon) |
[polyline](#polyline) | [color](#color) | [text](#text) | [size](#size) |
[symbol](#symbol) | [rotation](#rotation) | [icon](#icon) | [width](#width) \]

- # comment A line of comment which is ignored in the processing.
- move xpos ypos The current location is updated to xpos ypos. Unless the -m flag is used, values are stated as a
-percent of the active display frame’s horizontal (xpos) and vertical (ypos) size, and may be floating point values.
-Values are between 0-100. Note. A space must separate xpos and ypos.
- draw xpos ypos A line is drawn in the current color from the current location to the new location xpos ypos, which
-then becomes the current location. Unless the -m flag is used, values are stated as a percent of the active display
-frame’s horizontal (xpos) and vertical (ypos) size, and may be floating point values. Values are between 0-100. Note.
-A space must separate xpos and ypos.
```

wenzeslaus:
What is the diff between files generated from main and files generated from this branch? Use of Meld or diff should point out issues like the one with d.graph.

cwhite911:
I've updated the filters. However, we still have issues with some html element types. The two major ones are

- `<table>` is kind of hit or miss if it works correctly. (i.atcorr, r.slope.aspect)
- `<dl>` description lists will also most likely need to be manually fixed.

We also have some markdownlint warnings we can choose to address or ignore.

- [MD051 - Link fragments should be valid](https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md051.md)
- [MD026 - Trailing punctuation in heading](https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md026.md)
- [MD035 - Horizontal rule style](https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md035.md)
- [MD009 - Trailing spaces](https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md009.md)

wenzeslaus:
I merged the main branch here, bringing the HTML .md files. There was a conflict in the linting rules. I removed those here. They are refined for the converted files, but this is not converting them yet, so I'm just keeping the version from main. The config will be used later and for convenience is here:

```yaml
line-length:
  code_blocks: false
  tables: false
  line_length: 120

no-inline-html:
  allowed_elements: [sup]
```

wenzeslaus:
This removed for me an error from mkdocs output. Strangely, I have not seen it before, but anyway, I see it now without this conversion update and I don't see it with it. 

```
Error in wrapping img tag with anchor tag: expected string or bytes-like object, got 'NoneType' <a
href="r_carve_dem_orig_shaded.png"><img
src="r_carve_dem_orig_shaded.png" data-border="0" width="300"
height="321" alt="r.carve example: original DEM shaded" /></a>
Error in wrapping img tag with anchor tag: expected string or bytes-like object, got 'NoneType' <a
href="r_carve_dem_carved_shaded.png"><img
src="r_carve_dem_carved_shaded.png" data-border="0" width="300"
height="321" alt="r.carve example: carved DEM shaded" /></a>
Error in wrapping img tag with anchor tag: expected string or bytes-like object, got 'NoneType' <a
href="r_carve_dem_orig_accum.png"><img src="r_carve_dem_orig_accum.png"
data-border="0" width="300" height="321"
alt="r.carve example: original DEM flow accumulated" /></a>
Error in wrapping img tag with anchor tag: expected string or bytes-like object, got 'NoneType' <a
href="r_carve_dem_carved_accum.png"><img
src="r_carve_dem_carved_accum.png" data-border="0" width="300"
height="321"
alt="r.carve example: carved DEM flow accumulation" /></a>
```

wenzeslaus:
As for the images, it really does make some images worse in the final mkdocs output, but maybe it does not matter, because manual fixes are needed anyway, both for v.fill.holes and r.carve.

## Before

Screenshot from 2025-02-07 11-42-28
Screenshot from 2025-02-07 11-42-15

## After

Screenshot from 2025-02-07 11-35-52
Screenshot from 2025-02-07 11-35-31

cwhite911:
Here is the `.markdownlint.yml` I used to run markdownlint (v0.44.0)

```bash
markdownlint **/**/*.md --ignore node_modules --ignore venv --fix
```

```yaml

  default: true
  
  # Fix any fixable errors (depending on the markdownlint wrapper tool used)
  fix: true
  
  MD041: false # first-line-h1
  line-length:
    code_blocks: false
    tables: false
    # line_length: 120
    line_length: 300
    # imagery/i.atcorr/i.atcorr.md:285:121 MD013/line-length Line length [Expected: 120; Actual: 294]
  
  
  # HTML is coming over from HTML table elements
  no-inline-html:
    allowed_elements: [sup, sub, table, span, p, br, img, colgroup, col, tbody, tr, td, img, em, strong, thead, th]
  
  ul-indent:
    start_indented: true
    indent: 2
  
  no-trailing-punctuation: false
  
  # Fix v.transform, v.dssovle, r3.in.ascii, i.vi, d.mon
  no-duplicate-heading: false 
  
  # Fix v.net.iso, v.label
  heading-increment: false
  
  no-emphasis-as-heading: false
  
  # Fix v.random, r.surf.idw
  link-fragments: false
```

wenzeslaus (approved):
This at the end only changes shell to sh and removes the div tags (which is a nice cleanup) in comparison to the original. For that, this is too much code. However, I want to proceed with this as we can't clearly get any further with automatic conversion. The Markdown files in #5064 are based on this script with additional changes and it seems to work on an acceptable level.

I didn't used the markdownlint config from here, but went with a simpler version with some additional ignores.

PR #5064, opened 2025-02-07
doc: Convert HTML to Markdown

wenzeslaus:
This converts HTML markup in .md files to Markdown using the ./utils/grass_html2md.sh script.

This is committed with --no-verify. There are new or different Markdown linting rules broken and trailing spaces are not allowed.

How do we want to do it with the trailing spaces? Our pre-commit checks don't allow them, but they are used in Markdown extensively now to accommodate the various `<br>` used for formatting and it is used to mimic `<dt>`.


wenzeslaus:
The conversion from HTML to Markdown together with the automatic fixes removed a ton of warnings. I removed the relevant ignore and added three which did not occur with the HTML. I added `.md` to exclude for trailing spaces.

petrasovaa (approved):

PR #5080, opened 2025-02-10
lib/vector/Vlib: Fix Resource Leak issue in open_nat.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207740)
Used G_free() to fix this issue


nilason (approved):
Thanks!

(Let's keep issues under `/lib/*` be addressed per file, but modules can be addressed as per directory).

PR #5098, opened 2025-02-12
doc: Redefine r.mapcalc build deps for color table thumbnails

wenzeslaus:
The original dependency handling between r.color and r.mapcalc from fd7d0ece694b9abea6a36c496c8939249adf4029 and 9e56f6c48d3e43b759455d47dbb0a101e076a3dd uses dependency on binaries and builds r.mapcalc ahead of time. The new build code explicitly declares the dependency between r.color and r.mapcalc subdir phony targets.

This works for me locally with (-j1, -j2, -j4, -j10, -j30):

```
make distclean -j30 && ./configure.sh && make -j30
```

This fixes the new issue which arose in #5048 with the Markdown documentation where r.colors build failed because r.mapcalc gave permission denied. I temporarily fixed the issue in #5048 by removing r.mapcalc from SUBDIRS. However, the change broke the recompilation with distclean not tested in CI). The new approach with explicit subdir phony target dependency avoids both issues.


wenzeslaus:
@tmszi Does this fix also #3038 by any chance?

petrasovaa (approved):
Let's give this a try (depending on the CI results)

nilason:
See also #3038.

wenzeslaus:
> See also #3038.

I see the images and I didn't notice the issue lately, but I asked @tmszi above for feedback on that. Anyway, this PR does not aim at that, so I'm planning to merge it whether it makes a difference or not.

wenzeslaus:
The CI looks good but needs update to #5102 for stability.

PR #5106, opened 2025-02-13
docker: build gdal-grass with CMake

nilason:
After gdal-grass' drop of Autotools support with
https://github.com/OSGeo/gdal-grass/commit/fa4b0450d26f6f4791e3eca0332b1650b400e00b.

I can't test these changes, but it should be good. Either someone could test it manually, or we merge and see...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -297,11 +297,9 @@ RUN ./configure $GRASS_CONFIG \
 # Build the GDAL-GRASS plugin
 RUN git clone https://github.com/OSGeo/gdal-grass \
     && cd "gdal-grass" \
-    && ./configure \
-      --with-gdal=/usr/bin/gdal-config \
-      --with-grass=/usr/local/grass85 \
-    && make -j $NUMTHREADS \
-    && make install -j $NUMTHREADS \
+    && cmake -B build -DAUTOLOAD_DIR=/usr/lib/gdalplugins -DBUILD_TESTING=OFF \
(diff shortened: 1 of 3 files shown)


neteler:
Thanks! To [make G84 happy as well](https://github.com/OSGeo/grass/actions/runs/13303348655/job/37148780927) we'll need to backport this PR once merged.

wenzeslaus:
Fixes #5107

wenzeslaus:
We need to have a representative the Docker build (e.g. the one from root dir Dockerfile) build for PRs for clarity. It may also be a good idea to not use latest version in the clone, but a release instead.

wenzeslaus (approved):
Well, the build is broken and the CI tests that only on main and we have already bunch of commits there failing with this error. Let's merge it.

nilason:
>  It may also be a good idea to not use latest version in the clone, but a release instead.

These changes should work for latest release too, if we were to change version.

nilason:
Ahh, Alpine didn't install cmake. I'll fix that.

neteler:
@nilason shall we keep the hardcoded-version workaround for G84 or shall anything from main be backported?

nilason:
> @nilason shall we keep the hardcoded-version workaround for G84 or shall anything from main be backported?

I don't know. I never tested if gdal-grass driver actually worked before without hardcoded paths. But backporting that part wouldn't hurt either.

PR #5109, opened 2025-02-13
lib/vector/Vlib: fix segfault introduced with #5038

metzm:
Fixing a small resource leak in #5038 did not test if there is actually some memory allocated, introducing a segmentation fault. This PR fixes #5038.

### DIFF
--- a/lib/vector/Vlib/net_build.c
+++ b/lib/vector/Vlib/net_build.c
@@ -60,12 +60,12 @@ int Vect_net_ttb_build_graph(struct Map_info *Map, int ltype, int afield,
     dglGraph_s *gr;
     dglInt32_t opaqueset[16] = {360000, 0, 0, 0, 0, 0, 0, 0,
                                 0,      0, 0, 0, 0, 0, 0, 0};
-    struct field_info *Fi;
+    struct field_info *Fi = NULL;
     dbDriver *driver = NULL;
     dbDriver *ttbdriver = NULL;
     dbHandle handle;
     dbString stmt;
-    dbColumn *Column;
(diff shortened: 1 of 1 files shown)


neteler:
(@ShubhamDesai please have a look)

wenzeslaus:
I don't see this in Coverity Scan or CodeQL, so it seems neither of them caught it. (Coverity Scan reports additional resource leak issues at the same location and null pointer dereferences at other places.) I'm guessing you have some code which triggered it. Do you want to create a test? CI tests seem to be a crucial part of the mix even with all the static analysis.

wenzeslaus (approved):

ShubhamDesai:
> (@ShubhamDesai please have a look)

actually there were two lines for the check (muliple times whenever the memory is allocated)
 Fi = Vect_get_field(Map, tfield);
    if (Fi == NULL)
        G_fatal_error(_("Database connection not defined for layer %d"),
                      tfield);
this makes sure Fi is not null


if (db_get_column(driver, Fi->table, afcol, &Column) != DB_OK)
            G_fatal_error(_("Column <%s> not found in table <%s>"), afcol,
                          Fi->table);
also this would have made ensured Column is not null

there were checks for both so in any case the memory would be allocated otherwise program would exit. So i directly used free without check

But in anyways adding additional checks is good.

metzm:
@ShubhamDesai  These tests are only performed `if (afcol != NULL)` thus unconditional calling of `Vect_destroy_field_info(Fi);` and `db_free_column(Column);` can lead to segmentation faults.

You must check if a resource (memory) is really used before freeing it.

PR #5112, opened 2025-02-13
lib/vector/Vlib: free allocated field_info and dbColumn objects (net_build.c)

nilason:
I was just about to check #5109 when it was merged. So I figure I'll give my review in this form :).

`Fi` always get set by `Vect_get_field()`, and its result is always checked and is G_fatal_error()'ed if NULL. So we can always expect Fi to be not NULL. Furthermore, `Vect_destroy_field_info()` is safe with a NULL as argument (just like `free()`), but it mustn't be bogus address (it should be initialised to NULL at declaration). Thus, no need to check `if (Fi) Vect_destroy_field_info(Fi);`

`db_get_column()` allocates Column and G_fatal_errors if the call fails. So also here there is no real need for NULL check before `db_free_column()` _if freed at the right place_.

There are several calls to both Vect_get_field() and db_get_column(), which must _each_ be accompanied by freeing up allocated memory. I believe I got it right this time.

### DIFF
--- a/lib/vector/Vlib/net_build.c
+++ b/lib/vector/Vlib/net_build.c
@@ -167,6 +167,7 @@ int Vect_net_ttb_build_graph(struct Map_info *Map, int ltype, int afield,
                           tcols[i], Fi->table);
 
         tctype[i] = db_sqltype_to_Ctype(db_get_column_sqltype(Column));
+        db_free_column(Column);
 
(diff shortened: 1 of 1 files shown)


metzm (approved):
Thinking about my previous quick fix, your improvement is what I had in mind. Thanks!

PR #5132, opened 2025-02-17
lib/manage: Fix Resource Leak Issue in do_copy.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207901)

### DIFF
--- a/lib/manage/do_copy.c
+++ b/lib/manage/do_copy.c
@@ -37,18 +37,19 @@ int M_do_copy(int n, const char *old, const char *mapset, const char *new)
     int i, ret;
     char path[GPATH_MAX], path2[GPATH_MAX];
     int result = 0;
+    char *mname = G_fully_qualified_name(old, mapset);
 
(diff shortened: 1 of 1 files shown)


lib/manage/do_copy.c
```diff
@@ -64,8 +65,10 @@ int M_do_copy(int n, const char *old, const char *mapset, const char *new)
             }
             G_file_name(path2, list[n].element[i], new, G_mapset());
             if (G_recursive_copy(path, path2) == 1) {
+                char *mname2 = G_fully_qualified_name(old, mapset);
```
nilason, line 68: No need for new variable `nname2` here. `mname` is unchanged through this function. (Note: `old` and `mapset` are both const).

nilason (changes requested):

lib/manage/do_copy.c
ShubhamDesai: Done

nilason (approved):

PR #5152, opened 2025-02-19
CI: Add workflow to build mkdocs in GHA

wenzeslaus:
Build core and addons, then create mkdocs site as an artifact. Ignore errors with addons.

I'm trying to get a rough workflow where the main part works and the addons may work. I want to continue-on-error in this step and others because they are currently failing or at least not delivering what they should.

In the meantime, https://github.com/OSGeo/grass/pull/5177 and https://github.com/OSGeo/grass/pull/5175 are fixing the issues I identified so far and that should be enough to remove continue-on-error here. These are now sort of chicken and egg, the other PRs fix the build, but this PR actually builds in CI, possibly showing how the other two work. We can merge them in any order and I can then go back and improve this workflow. This was my original idea before I created the other PRs.


echoix:
Can it be made a bit general where we could build all our docs, if we manage to integrate the what currently is sphinx python reference, and (doxygen) C api docs? They would share the same triggers and purpose. The main reason of this workflow isn't really mkdocs, but the documentation/documentation site.

.github/workflows/mkdocs.yml
```diff
@@ -0,0 +1,116 @@
...
+        run: |
+          cd grass
+          eval $(./utils/update_version.py status --bash)
+          cd ..
+          export SITE_NAME="GRASS GIS $(VERSION) Reference Manual"
+          export COPYRIGHT="&copy; 2003-$(YEAR) GRASS Development Team, GRASS GIS $(VERSION) Reference Manual"
+          cd $MKDOCS_DIR
+          mkdocs build
+
+      - name: Make HTML test report available
```
echoix, line 111: Oups, not test report. (I know it's a draft)

wenzeslaus:
Any opinions on the script from addons? Shoud we just start from scratch here?

echoix:
> Any opinions on the script from addons? Shoud we just start from scratch here?

If we are to have this in the main repo, probably to build for main and the other release branch repeatedly

wenzeslaus:
> > Any opinions on the script from addons? Shoud we just start from scratch here?
> 
> If we are to have this in the main repo, probably to build for main and the other release branch repeatedly

While we want to include the addons, the doc for the main repo is the focus, so the build should happen here. My question if that infrastructure script for building addons should live here, continue living in addons, or if we should just start from scratch with new code (in YAML or a separate file).

.github/workflows/mkdocs.yml
wenzeslaus: Thanks, also the artifact name is fixed now.

wenzeslaus:
> Can it be made a bit general where we could build all our docs, if we manage to integrate the what currently is sphinx python reference, and (doxygen) C api docs?

They should definitively be integrated on the basic level of links (at least at this point) just like the HTML does:

https://grass.osgeo.org/grass-stable/manuals/

links to

https://grass.osgeo.org/grass84/manuals/libpython/index.html

and

https://grass.osgeo.org/programming8/

The artifact has actually already >100MB (>160MB unzipped), so the result will be large with all three, but then, building all here in CI would be great, so it really depends on how the upload to the documentation server is handled.

> They would share the same triggers and purpose.

Yes, on the other hand, they have different build times are requirements, so I guess you could argue both ways.

> The main reason of this workflow isn't really mkdocs, but the documentation/documentation site.

I'll rename it, but I'll keep the scope of the PR to mkdocs.

wenzeslaus:
This is actually in a pretty good shape. It even generates files for addons which is already better than I expected. The addon stuff is expected to fail now as it depends on updates to grass-addons, so that's allowed to fail without failing the workflow.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,146 @@
...
+          path: ${{ env.MKDOCS_DIR }}/site
+          retention-days: 3
+
+      - name: Make logs available
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+        continue-on-error: true
+        with:
+          name: grass-addon-build-logs
+          path: mv addons-build-dir/docs/logs
+          retention-days: 3
```
echoix, line 146: You might want to use if: always() if you want them even if you cancel a job, or if the job fails.

The continue-on-error will allow this step to not make the job fail if it fails here.

wenzeslaus:
> Another question: what is the time impact, and the impact of building addons on every PR?

2.5 min, total time of this job is 9 min

image
image

.github/workflows/documentation.yml
echoix: By default, upload artifact will warn if no files are found. See https://github.com/actions/upload-artifact?tab=readme-ov-file#customization-if-no-files-are-found, where it has the example:
```yaml
- uses: actions/upload-artifact@v4
  with:
    name: my-artifact
    path: path/to/artifact/
    if-no-files-found: error # 'warn' or 'ignore' are also available, defaults to `warn`
```

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,146 @@
...
+          name: mkdocs-site
+          path: ${{ env.MKDOCS_DIR }}/site
+          retention-days: 3
+
+      - name: Make logs available
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+        continue-on-error: true
+        with:
+          name: grass-addon-build-logs
+          path: mv addons-build-dir/docs/logs
```
echoix, line 145: What is the mv folder? (I think it is a copy-paste typo, but I'm asking the question as interpreted by the action)

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,146 @@
...
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y wget git gawk findutils
+          xargs -a <(awk '! /^ *(#|$)/' "grass/.github/workflows/apt.txt") -r -- \
+              sudo apt-get install -y --no-install-recommends --no-install-suggests
+
+      - name: Set up Python
+        uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38 # v5.4.0
+        with:
+          python-version: "3.12"
```
echoix, line 42: Do we want to keep the version updated like python-code-quality.yml?

See 
https://github.com/OSGeo/grass/blob/c112752ef1e070f6d202d262211e4275f7671a83/.github/workflows/python-code-quality.yml#L28-L30

https://github.com/OSGeo/grass/blob/c112752ef1e070f6d202d262211e4275f7671a83/.github/workflows/python-code-quality.yml#L58-L62

echoix (commented):
I'm still reviewing, but here's some to start with

.github/workflows/documentation.yml
echoix: With the default behavior, I think it is enough, and you could remove `continue-on-error`, as it may hide real problems that the actions will report. 

Is the addons directory really missing, or just misspecified?

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,146 @@
...
+      - name: Compile addons
+        continue-on-error: true
+        run: |
+          ./grass-addons/utils/cronjobs_osgeo_lxd/compile_addons_git.sh \
+            "$MAJOR" \
+            "$MINOR" \
+            $(pwd)/grass-addons/src \
+            $(grass --config path) \
+            $(pwd)/addons-build-dir \
+            grass
```
wenzeslaus, line 98: I'm trying to get a rough workflow where the main part works and the addons may work. I want to continue-on-error in this step and others because they are currently failing or at least not delivering what they should.

In the meantime, #5177 and #5175 are fixing the issues I identified so far and that should be enough to remove continue-on-error here. These are now sort of chicken and egg, the other PRs fix the build, but this PR actually builds in CI, possibly showing how the other two work. We can merge them in any order and I can then go back and improve this workflow. This was my original idea before I created the other PRs.

.github/workflows/documentation.yml
wenzeslaus, line 145:
```suggestion
          path: addons-build-dir/docs/logs
```

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,146 @@
...
+      - name: Make the result available
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+        with:
+          name: mkdocs-site
+          path: ${{ env.MKDOCS_DIR }}/site
+          retention-days: 3
+
+      - name: Make logs available
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+        continue-on-error: true
```
wenzeslaus, line 142: Thanks @echoix, if-no-files-found is much better way of handling this:

```suggestion
        if-no-files-found: warn
```

.github/workflows/documentation.yml
echoix: Good comment location: With the default behavior, I think it is enough, and you could remove `continue-on-error`, as it may hide real problems that the actions will report. 

Is the addons directory really missing, or just misspecified?

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,146 @@
...
+          cd ..
+          export SITE_NAME="GRASS GIS $VERSION Reference Manual"
+          export COPYRIGHT="&copy; 2003-$YEAR GRASS Development Team, GRASS GIS $VERSION Reference Manual"
+          cd $MKDOCS_DIR
+          mkdocs build
+
+      - name: Make the result available
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+        with:
+          name: mkdocs-site
```
wenzeslaus, line 136: As a next step, I will consider changing this to error and changing order.

```suggestion
          name: mkdocs-site
          if-no-files-found: warn
```

.github/workflows/documentation.yml
echoix: Could even be this:

```suggestion
```

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,146 @@
...
+
+jobs:
+  ubuntu:
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
+      cancel-in-progress: true
+
+    runs-on: ubuntu-22.04
+    env:
+      PYTHONWARNINGS: always
```
wenzeslaus, line 23:
```suggestion
      PYTHONWARNINGS: always
      # renovate: datasource=python-version depName=python 
      PYTHON_VERSION: "3.12" 
```

.github/workflows/documentation.yml
wenzeslaus, line 42:
```suggestion
          python-version: ${{ env.PYTHON_VERSION }}
```

.github/workflows/documentation.yml
wenzeslaus: Just typo, thanks.

.github/workflows/documentation.yml
wenzeslaus: Yes, updated. Done.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,149 @@
...
+          name: mkdocs-site
+          if-no-files-found: warn
+          path: ${{ env.MKDOCS_DIR }}/site
+          retention-days: 3
+
+      - name: Make logs available
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+        if-no-files-found: warn
+        with:
+          name: grass-addon-build-logs
```
wenzeslaus, line 147:
```suggestion
        with:
          name: grass-addon-build-logs
          if-no-files-found: warn
```

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,149 @@
...
+  ubuntu:
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
+      cancel-in-progress: true
+
+    runs-on: ubuntu-22.04
+    env:
+      PYTHONWARNINGS: always
+      # renovate: datasource=python-version depName=python 
+      PYTHON_VERSION: "3.12"
```
wenzeslaus, line 25:
```suggestion
      # renovate: datasource=python-version depName=python
      PYTHON_VERSION: "3.12"
```

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,149 @@
...
+
+# Builds Markdown documentation for core and addons.
+
+on:
+  push:
+    branches:
+      - main
+      - releasebranch_*
+  pull_request:
+
```
echoix, line 12: Would adding a workflow_dispatch trigger be useful for a first iteration?
```suggestion
on:
  push:
    branches:
      - main
      - releasebranch_*
  pull_request:
  workflow_dispatch:

```

.github/workflows/documentation.yml
wenzeslaus: > Is the addons directory really missing, or just misspecified?

The typo didn't help, but I actually didn't expect that to succeed - I just set all the addon-related things to continue-on-error. I plan to investigate more with the other PRs in place.

.github/workflows/documentation.yml
wenzeslaus: I can see that this may be useful later due to the grass-addons repo being a dependency of this workflow.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,149 @@
...
+          path: ${{ env.MKDOCS_DIR }}/site
+          retention-days: 3
+
+      - name: Make logs available
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+        with:
+          name: grass-addon-build-logs
+          if-no-files-found: warn
+          path: addons-build-dir/docs/logs
+          retention-days: 3
```
wenzeslaus, line 149:
```suggestion
      - name: Make logs available
        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
        with:
          name: grass-addon-build-logs
          if-no-files-found: warn
          path: addons-build-dir/docs/logs
          retention-days: 3

      - name: Make the result available
        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
        with:
          name: mkdocs-site
          if-no-files-found: error
          path: ${{ env.MKDOCS_DIR }}/site
          retention-days: 3
```

echoix:
Seems to have the same problem as https://github.com/conda-forge/sqlite-feedstock/issues/130

.github/workflows/documentation.yml
echoix: Do you want this PR to go through first?

echoix (approved):

.github/workflows/documentation.yml
wenzeslaus: Yes. The build of mkdocs works which was my goal here.

PR #3621, opened 2024-04-18
build: Add CMake build

HuidaeCho:
This PR replaces #3021. Branched from https://github.com/nilason/grass/tree/cmake_build_work. Thanks @nilason!

Current status: I can compile the entire `gisbase` using CMake 3.29.1 and GCC 13.2.0. Running `grass` directly from `build/gisbase` works fine.

## Windows Tests

For my records, message from `message(WIN32=${WIN32}, MSVC=${MSVC}, MSYS=${MSYS}, MINGW=${MINGW})` on Windows (all from `cmd` in `ssh`):
* Using MSYS cmake (`cmd` for MSVC): `WIN32=,MSVC=,MSYS=1,MINGW=` ([`WIN32`](https://cmake.org/cmake/help/latest/variable/WIN32.html#win32) not set?)
  * https://github.com/xmrig/xmrig/issues/950
  * https://developercommunity.visualstudio.com/t/win32-not-defined/722576
* Using MinGW-w64 cmake (`bash` for MinGW-w64 GCC): `WIN32=1,MSVC=,MSYS=,MINGW=1`

CMake variables:
* [WIN32](https://cmake.org/cmake/help/latest/variable/WIN32.html)
* [MSVC](https://cmake.org/cmake/help/latest/variable/MSVC.html)
* [MSYS](https://cmake.org/cmake/help/latest/variable/MSYS.html)
* [MINGW](https://cmake.org/cmake/help/latest/variable/MINGW.html)

## TODOs

* Windows
* macOS
* `locale`
* addon support
* ...

## Documentation

@Mahesh1998 and I are testing, fixing, and documenting this PR on Ubuntu. https://github.com/Mahesh1998/grass-docs/blob/main/cmake_build_docs/
* [Slackware with system libraries](https://github.com/Mahesh1998/grass-docs/blob/main/cmake_build_docs/slackware_sys_cmake.rst)
* [Ubuntu with system libraries](https://github.com/Mahesh1998/grass-docs/blob/main/cmake_build_docs/ubuntu_sys_cmake.rst)
* [Ubuntu with Conda libraries (without root permissions)](https://github.com/Mahesh1998/grass-docs/blob/main/cmake_build_docs/ubuntu_conda_cmake.rst)
* [Windows with Conda libraries](https://github.com/Mahesh1998/grass-docs/blob/main/cmake_build_docs/windows_conda_cmake.rst)

## Building on Windows

[...]

PR #5159, opened 2025-02-20
lib/gis: add support for PCRE

nilason:
Add support for PCRE, where/when GNU Regex cannot be used.

As this is mainly (perhaps only) needed on Win systems, `--without-pcre` is default. If both Regex and PCRE are available on the system, Regex is used by default.

### DIFF
--- a/configure
+++ b/configure
@@ -740,10 +740,11 @@ HISTORYLIB
 READLINELIB
 READLINELIBPATH
 READLINEINCPATH
-USE_REGEX
 REGEXLIB
 REGEXLIBPATH
 REGEXINCPATH
+USE_PCRE
+USE_REGEX
 PROJSHARE
 PROJLIB
 PROJINC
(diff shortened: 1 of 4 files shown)


wenzeslaus:
We are not testing this in the CI with the assumption that the code will be eventually tested with CMake on Windows, right?

Also, I think it does not need to go in before #3621.

nilason:
> We are not testing this in the CI with the assumption that the code will be eventually tested with CMake on Windows, right?

I tested it locally, configure and build, at least.
 
> Also, I think it does not need to go in before #3621.

Not necessarily, I'll only remove the code changes there. The rest can stay.

Probably easier than rebase/merge main on THAT one :-).

HuidaeCho:
> Rashad was the original author, wasn't he?

I think so.

HuidaeCho (approved):

PR #5160, opened 2025-02-20
build: Continue build when a keyword is empty

wenzeslaus:
This skips over empty keywords in tool documentation when building keyword index.

This is a draft. I'm actually not sure if this is a good idea because it does not bring the issues with keywords to light (one needs to inspect the log after noticing the issue in some way). I'll continue evaluating this.

### DIFF
--- a/man/build_keywords.py
+++ b/man/build_keywords.py
@@ -100,6 +100,13 @@ def build_keywords(ext):
                     break
 
         for key in keys:
+            if not key:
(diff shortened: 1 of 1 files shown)


wenzeslaus:
Replaced by a more comprehensive solution in #5175, but the idea stays the same. It just warns about empty or missing keywords for Markdown doc. The HTML doc just skips over it silently.

PR #5171, opened 2025-02-21
lib/init: remove old database image

cwhite911:
Removed the old grass_database.png version that was overwriting the new version when building mkdocs.

See issue https://github.com/OSGeo/grass/issues/5161 for more details.


echoix (approved):
I'm approving, but I didn't fully understand how the overwritten file happened, but you know more about the docs for now, you're on it :)

cwhite911:
> I imagine if the files get copied to a common directory it could happen.

Yes. When the docs build all of the images are flattened into a single directory.

PR #5175, opened 2025-02-21
docs: Add support for addon keywords to Markdown

wenzeslaus:
The modifies the build scripts to support combining addon documentation into the main documentation. Links to keyword index, topics, and tool family now work from addons. Keyword index contains keywords from addons and link to addons.

HTML_PAGE_FOOTER_PAGES_PATH build-only helper env var is used to provide a prefix for URL during the build of addons (this would be set to '../'). The already generated keyword section is replaced by a new one which has the links with a prefix. This copies how the HTML documentation deals with the keywords. The name HTML_PAGE_FOOTER_PAGES_PATH comes from the usage in the HTML documentation and does not apply for Markdown as the footer is not generated on the page level. Following the HTML documentation, the links are corrected during processing in Python rather than during the original creation in the command line parser C code.

Separate, yet related, issue is dealing with empty keywords in the tools. The original code just fails with addons which don't have any keywords. This addresses that issue by printing the message to stderr, but skipping the empty keyword. (The HTML does not have that issue presumably because the parsed list of keywords is empty while the Markdown parsing results in an one empty item in a list. This may change how this is handled in Markdown in the future depending on how we deal with custom keywords versus mkdoc tags.)


wenzeslaus:
The fixes of links are now visible in the CI (after #5152).

In CI Documentation, the number of lines in mkdocs output is reduced by roughly 800. The following warnings are present on main, but are not here as keyword links in addon files now correctly point to `../keywords.md`.

```
WARNING -  Doc file 'addons/v.profile.points.md' contains a link 'keywords.md#profile', but the target 'addons/keywords.md' is not found among documentation files.
WARNING -  Doc file 'addons/v.profile.points.md' contains a link 'keywords.md#transect', but the target 'addons/keywords.md' is not found among documentation files.
```

PR #5177, opened 2025-02-21
build: Allow separate build of Markdown and HTML keywords

wenzeslaus:
The original Markdown support from #3849 breaks generating addon keywords on the doc build server because the parameter for addons used in the build script was ignored. The addon keywords are currently integrated for 8.4, but are not integrated into the index for 8.5. This update should fix that.

The new code allows for the original usage of the script where documentation directories are supplied as parameters. This assumes HTML documentation and corresponds to the usage on the build server.

The intended usage from #3849 for the new build where both Markdown and HTML are build using the script without the build system providing parameters is preserved.

Finally, with parameters md <core-path> <addon-path>, the script allows for creating Markdown keyword list with optionally integrated addons using the paths provided. Unfortunately, ARCH_DISTDIR still needs to be set even in this case because it used directly a import time and a lot of other code uses that variable.

This also adds a main function to separate the variables and handle the parameters in one place. Additionally, it creates the dict for the first letter TOC only for HTML where it is used (which allows the script to execute without the fixes for empty keyword list from #5175).


cwhite911 (approved):

wenzeslaus:
The addon keywords are now part of the generated keyword index in mkdocs:

image

In the CI, the following, rather trivial, change is enabled by this PR:

```diff
- grass/man/build_keywords.py "$MKDOCS_DIR/source/addons"
+ grass/man/build_keywords.py md "$MKDOCS_DIR/source" "$MKDOCS_DIR/source/addons"
```

It should also bring back properly interlinked core and addon keywords in HTML docs on the server for the preview doc build which was broken after #3849 (initial Markdown build PR). The `man/build_keywords.py` should now again produce addon keywords in the keyword index in this line:

https://github.com/OSGeo/grass-addons/blob/grass8/utils/cronjobs_osgeo_lxd/cron_grass_preview_build_binaries.sh#L331

The sever impact is untested by me, but now the CI Documentation workflow builds with keywords and the mkdocs site can be downloaded as an artifact, and `keywords.html` can be examined.

petrasovaa (approved):

PR #5194, opened 2025-02-23
Windows: Wrapper header files for MSVC

HuidaeCho:
This PR introduces additional wrapper header files for MSVC:
* msvc/stdlib.h
* msvc/string.h
* msvc/sys/stat.h

It also `#define`s necessary constants, macros, and data types (e.g., `ssize_t`) to align with their definitions in Linux headers. The objective is to avoid `#ifdef _MSC_VER` in C files as much as possible by providing compatibility through these files.

With this PR, `cmake .. -DWITH_OPENGL=OFF` + `msbuild grass.sln` compiles without errors.

### DIFF
--- a/include/config.h.cmake.in
+++ b/include/config.h.cmake.in
@@ -409,17 +409,6 @@
 #define fstat _fstati64
 #endif /* _MSC_VER LFS */
 
-#define strncasecmp _strnicmp
-#define strcasecmp _stricmp
-#define fdopen _fdopen
-/* _open() and _creat() are wrapped by msvc/open.c and msvc/creat.c,
-* respectively, for compatible permission mode */
-#define read _read
-#define write _write
-#define close _close
-#define unlink _unlink
-#define getpid _getpid
-
 /* define gid_t type */
 typedef @GID_TYPE@ gid_t;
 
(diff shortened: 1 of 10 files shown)


nilason:
I never reflected much on the fact that `ssize_t` is a POSIX type. Introducing GRASS code to MSVS pinpoint this circumstance however.

I think that we should keep to C/C++ standard as long as we can. And there are only few cases where ssize_t is actually warranted for[^1]. A case like `(ssize_t)strlen(buf) > n` (where `n` is an int) is not one of them. There, the cast from size_t to ssize_t was made with the intent to make the return value a _signed_ value (to prevent compiler warnings on comparing signed and unsigned values).

_I think a better choice  in these cases would be to replace `ssize_t` with `intmax_t`._

We don't use very much of ssize_t in repo:
https://github.com/search?q=repo%3AOSGeo%2Fgrass%20ssize_t&type=code

[^1]: By using e.g. [read()](https://man7.org/linux/man-pages/man2/read.2.html) or [write()](https://man7.org/linux/man-pages/man2/write.2.html).

HuidaeCho:
> I think a better choice in these cases would be to replace `ssize_t` with `intmax_t`.

Oh! We have more `ssize_t`. How did they not fail... because they all include `grass/config.h` (directly/indirectly). I think it would be better to move some MSVC stuff from the global `grass/config.h` to individual `msvc/*` wrapper headers.

`intmax_t` requires an additional header `stdint.h` and I think `ssize_t` makes more sense semantically. What about adding `msvc/stdio.h`:
```c
#ifndef GRASS_MSVC_STDIO_H
#define GRASS_MSVC_STDIO_H

#include <../ucrt/stdio.h>

#include <BaseTsd.h>
#ifndef ssize_t_defined
typedef SSIZE_T ssize_t;
#define ssize_t_defined
#endif

#endif
```

nilason (approved):
Looks good to me!

I'm particularly happy to move things away from` config.h.cmake.in`!
(There are still things there which should be (re-)moved, but that is for another time).

PR #5203, opened 2025-02-25
r.kappa: Copy into fix buffer size issue

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1208250)
Used G_strlcpy() to fix this issue.


nilason (approved):

nilason (changes requested):
We have a similar issue in r.kappa with CID1208249, please address that here too.

nilason (approved):

PR #5204, opened 2025-02-25
wxGUI: Fix Unchecked return value in main.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan ( CID : 1501249)
If the return value == -1 then G_fatal_error

### DIFF
--- a/visualization/ximgview/main.c
+++ b/visualization/ximgview/main.c
@@ -120,11 +120,15 @@ static void redraw(void)
 {
     struct timeval tv0, tv1;
 
-    gettimeofday(&tv0, NULL);
+    if (gettimeofday(&tv0, NULL) == -1) {
+        G_fatal_error(_("gettimeofday failed: %s"), strerror(errno));
+    }
 
     draw();
 
-    gettimeofday(&tv1, NULL);
+    if (gettimeofday(&tv1, NULL) == -1) {
+        G_fatal_error(_("gettimeofday failed: %s"), strerror(errno));
+    }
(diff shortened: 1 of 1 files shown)


visualization/ximgview/main.c
```diff
@@ -120,11 +120,15 @@ static void redraw(void)
 {
     struct timeval tv0, tv1;
 
-    gettimeofday(&tv0, NULL);
+    if (gettimeofday(&tv0, NULL) == -1) {
+        G_fatal_error(_("gettimeofday failed: %s"), strerror(errno));
```
nilason, line 124: Let us remove the strerror(errno) part and check `gettimeofday(&tv0, NULL) != 0`.

POSIX [gettimeofday() ](https://pubs.opengroup.org/onlinepubs/9699919799/):
> The gettimeofday() function shall return 0 and no value shall be reserved to indicate an error.

nilason (changes requested):

visualization/ximgview/main.c
ShubhamDesai: done. Thanks for the documentation.
I referred this and thought it would be -1
https://github.com/OSGeo/grass/blob/e60ee63e15fcd27cf2d79fadc028fa6106eba040/utils/timer/main.c#L30

nilason (approved):

visualization/ximgview/main.c
ShubhamDesai: Ok i will create another pull request to fix this.

PR #5211, opened 2025-02-25
cmake: build 'r.mask.status'

nilason:
Build `r.mask.status` also with CMake.


wenzeslaus (approved):
Oh, little more code. I forgot about the dependencies.

PR #5216, opened 2025-02-26
lib/gis: fix potential buffer overflow issues

nilason:
Addresses issues of type [CWE-120](https://cwe.mitre.org/data/definitions/120.html): _Buffer Copy without Checking Size of Input_, reported by Coverity Scan.


echoix (approved):
All patterns of changing strcpy->G_strlcpy, strcat->G_strlcat are the same, and the few changes of sprintf->snprintf or adding snprintf look consistent together.

PR #5225, opened 2025-02-27
CI(macOS): only run 'macOS distribute app' on OSGeo repo

nilason:
Avoid attempts to run _macOS distribute app_ workflow on forks.


wenzeslaus (commented):
Looks good to me.

wenzeslaus (approved):

PR #5259, opened 2025-03-01
lib/psdriver: Fix printf format specifier issue

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1415568, 1415601, 1415712, 1415744, 1415767,1415773)
Some of the variables were declared as double but the format specifier use was of int.


nilason:
This is already addressed by #3551, with no consensus.

ShubhamDesai:
> This is already addressed by #3551, with no consensus.

Oh Ok in that case I will close this pull request.

PR #5260, opened 2025-03-01
v.colors: Fix Resource Leak issue in write_rgb.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207647,1207649)
Used Vect_destroy_field_info(), G_free() to fix this issue.


nilason (changes requested):
They need to be freed at the end of the function too.

ShubhamDesai:
> They need to be freed at the end of the function too.

Done

nilason:
Now I see, db_select_int may return without allocating for `pval`, we better initialise with NULL.

ShubhamDesai:
> Now I see, db_select_int may return without allocating for `pval`, we better initialise with NULL.

Done

nilason (approved):

PR #5273, opened 2025-03-03
d.vect.thematic: Fix Resource Leak issue in plot1.c

ShubhamDesai:
This pull request fixes issue identified by coverity scan (CID : 1207610, 1207611, 1207612, 1207613, 1207614)
Used `Vect_destroy_cats_struct(), Vect_destroy_line_struct(), G_free()` to fix this issue.

### DIFF
--- a/display/d.vect.thematic/plot1.c
+++ b/display/d.vect.thematic/plot1.c
@@ -196,7 +196,7 @@ int plot1(struct Map_info *Map, int type, int area UNUSED,
             line++;
             if (line > nlines) {
                 ret = 0;
-                goto cleanup;
+                goto cleanup_and_exit;
             }
             if (!Vect_line_alive(Map, line))
                 continue;
@@ -208,10 +208,10 @@ int plot1(struct Map_info *Map, int type, int area UNUSED,
             case -1:
                 fprintf(stderr, _("\nERROR: vector map - can't read\n"));
                 ret = -1;
-                goto cleanup;
+                goto cleanup_and_exit;
             case -2: /* EOF */
                 ret = 0;
-                goto cleanup;
+                goto cleanup_and_exit;
             }
         }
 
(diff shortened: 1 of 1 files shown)


nilason:
All the variables primary_color, fill_color, line_color, Points and Cats need to be freed before every exit (return) of the function. The while-statement (L153) is an infinite loop, exiting only in three cases in the if-else-statement of L194-211.

So, instead of adding 5x3 lines of repetitive code, let us add a `goto` label at the end of the function (adding a `ret` variable) and free them all there.

ShubhamDesai:
> All the variables primary_color, fill_color, line_color, Points and Cats need to be freed before every exit (return) of the function. The while-statement (L153) is an infinite loop, exiting only in three cases in the if-else-statement of L194-211.
> 
> So, instead of adding 5x3 lines of repetitive code, let us add a `goto` label at the end of the function (adding a `ret` variable) and free them all there.

done.

display/d.vect.thematic/plot1.c
```diff
@@ -193,8 +194,10 @@ int plot1(struct Map_info *Map, int type, int area UNUSED,
     while (1) {
         if (Vect_level(Map) >= 2) {
             line++;
-            if (line > nlines)
-                return 0;
+            if (line > nlines) {
+                ret = 0;
+                goto cleanup;
```
nilason, line 199: Considering this is a relatively long code, I would prefer a goto label that includes the information that it also exits the function ("cleanup" does not).

nilason (changes requested):

display/d.vect.thematic/plot1.c
nilason: Sounds good to me.

ShubhamDesai: done

nilason (approved):
Thanks!

PR #5274, opened 2025-03-03
v.surf.idw: Fix Resource Leak issue in read_sites.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207627)
Used `G_free()` to fix this issue.

### DIFF
--- a/vector/v.surf.idw/read_sites.c
+++ b/vector/v.surf.idw/read_sites.c
@@ -34,14 +34,14 @@ void read_sites(const char *name, const char *field_name, const char *col,
                                                  only when column is
                                                  not defined */
     if (!col) {
+        const char *mname = Vect_get_full_name(&Map);
         if (!with_z)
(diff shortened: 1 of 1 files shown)


nilason:
Shouldn’t `Fi`, `Points` and `Cats` also be freed?

ShubhamDesai:
> Shouldn’t `Fi`, `Points` and `Cats` also be freed?

yeah i also thought the same but coverity scan did not highlight and was waiting for the confirmation whether to add that as well.
I will add it.

ShubhamDesai:
> > Shouldn’t `Fi`, `Points` and `Cats` also be freed?
> 
> yeah i also thought the same but coverity scan did not highlight and was waiting for the confirmation whether to add that as well. I will add it.

Done added.

nilason (approved):

PR #5276, opened 2025-03-04
d.vect: Fix resource Leak Issue in shape.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1207875)
Used `Vect_destroy_field_info()` to fix this issue.

### DIFF
--- a/display/d.vect/shape.c
+++ b/display/d.vect/shape.c
@@ -37,7 +37,7 @@ int display_shape(struct Map_info *Map, int type, struct cat_list *Clist,
     if (open_db) {
         field = Clist->field > 0 ? Clist->field : 1;
         fi = Vect_get_field(Map, field);
-        if (!fi) {
+        if (fi == NULL) {
             G_fatal_error(_("Database connection not defined for layer %d"),
                           field);
         }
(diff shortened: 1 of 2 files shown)


nilason (changes requested):
Not strictly necessary, but please initialise `fi` to NULL at its definition.

display/d.vect/shape.c
```diff
@@ -37,7 +37,7 @@ int display_shape(struct Map_info *Map, int type, struct cat_list *Clist,
     if (open_db) {
         field = Clist->field > 0 ? Clist->field : 1;
         fi = Vect_get_field(Map, field);
-        if (!fi) {
+        if (fi == NULL) {
```
nilason, line 40: This shouldn't be necessary.

ShubhamDesai:
> Not strictly necessary, but please initialise `fi` to NULL at its definition.

done changes. Also since it was failing on Mac and ubuntu modified the if statement
from `!fi to fi == NULL`

display/d.vect/shape.c
ShubhamDesai: actually it failed on Mac

ShubhamDesai: oh okay. i will revert back to previous code

ShubhamDesai: done

nilason (approved):

PR #5277, opened 2025-03-04
docs: Added mkdocs feature content.tabs.link

cwhite911:
I've added the `content.tabs.link` feature to mkdocs.

> When enabled, all content tabs across the whole documentation site will be linked and switch to the same label when the user clicks on a tab. 

https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#linked-content-tabs


wenzeslaus:
Does this add the actual tabs? I thought we don't have them enabled yet.

cwhite911:
It makes it so that when a user clicks from the Python tab to the R tab all other tabs on the page switch to R as well.

wenzeslaus (changes requested):
As I mentioned, the actual content tabs are not enabled.

wenzeslaus (approved):
That should do it based on my experiment.

PR #5281, opened 2025-03-04
libgis: define new standardised option `G_OPT_M_SEED`

NishantBansal2003:
This PR implements the standard `G_OPT_M_SEED` option for the random number generator's seed.


lib/gis/parser_standard_options.c
```diff
@@ -786,6 +787,16 @@ struct Option *G_define_standard_option(int opt)
         Opt->description = _("Number of threads for parallel computing");
         break;
 
+    case G_OPT_M_SEED:
+        Opt->key = "seed";
+        Opt->type = TYPE_INTEGER;
+        Opt->required = NO;
+        Opt->label = _("Seed for random number generator");
```
wenzeslaus, line 794: Smooter wording:

```suggestion
        Opt->label = _("Seed value for the random number generator");
```

lib/gis/parser_standard_options.c
```diff
@@ -786,6 +787,16 @@ struct Option *G_define_standard_option(int opt)
...
+        Opt->description =
+            _("The same seed can be used to obtain same results"
+              " or random seed can be generated by other means.");
```
wenzeslaus, line 797: This leave out the allusion to generating seeds externally (which would need more explanation anyway) and adds more details on what happens without the explicit seed (assuming the auto-generated seed behavior as an alternative).

```suggestion
            _("Using the same seed ensures identical results, "
              "while a randomly generated seed produces different outcomes "
              "in each run.");
```

wenzeslaus (changes requested):
This is great!

I have a suggestion to refine the wording.

petrasovaa (approved):

wenzeslaus (approved):

PR #5297, opened 2025-03-04
cmake: fix Native Language Support (NLS) support

nilason:
Fix Native Language Support (NLS) support with CMake builds.

~~This is another reason to increase CMake required version to 3.22 (#5269). The version of FindIntl from 3.16 is not particularly good and we'll have to use a custom one if staying with 3.16.
I included here the version of 3.22, for "proof of concept".~~


nilason:
#5269 is merged, this is now ready too.

echoix (approved):

PR #5299, opened 2025-03-04
cmake: use imported PostgreSQL target

nilason:
Use imported PostgreSQL target.
This too would benefit from #5269, but in the meantime I added FindPostgreSQL from CMake 3.22.


wenzeslaus (approved):
Adding PostgreSQL, great.

Unrelated to this PR, but do you want to create a min and max configurations? (We sort of started that for Autotools.)

nilason:
> Adding PostgreSQL, great.
> 
> Unrelated to this PR, but do you want to create a min and max configurations? (We sort of started that for Autotools.)

At some point absolutely, now I think we’re good.

PR #5310, opened 2025-03-05
docs: Build Markdown addon index using core index tooling

wenzeslaus:
This builds addon index (for `addons` subdirectory) in CI using the tooling from core, specifically the script to build full tool index. The intro text is different for addons and for the core. The addons version is derived from the current addon index page.


wenzeslaus:
This the [index page for addons](https://grass.osgeo.org/grass-stable/manuals/addons/) now generated in the same way as the [full index page](https://grass.osgeo.org/grass-stable/manuals/full_index.html) for core tool. Result:

image

Available in the mkdocs-site asset of the Documentation workflow.

petrasovaa (approved):
There is old link to trac, but I assume this is the TODO mentioned in there.

PR #5311, opened 2025-03-06
cmake: use of imported SQLite target, drop custom FindSQLite

nilason:
Use imported SQLite target, drop custom FindSQLite.


echoix (approved):
Thanks very much! This I know worked on windows with the OSGeo4W dependencies

PR #5324, opened 2025-03-06
CI: Fetch all Git history for building docs

wenzeslaus:
Get all the history when cloning Git repos to have all commits when asking about the lastest change for the Source code section in the doc.

See #5323


echoix (approved):
I don't know if something better exists in some weird options to have enough info available to get the date for every file, but what is certain is that with a fetch-depth of 0, done here, will get all the commits and will work.

PR #5346, opened 2025-03-09
docs: fix typos

luzpaz:
Found via codespell

### DIFF
--- a/doc/examples/python/m.distance.py
+++ b/doc/examples/python/m.distance.py
@@ -68,7 +68,7 @@ def main():
     # calc distance
 
     proj_type = G_begin_distance_calculations()
-    # returns 0 if projection has no metrix (ie. imagery)
+    # returns 0 if projection is not metric (ie. imagery)
     # returns 1 if projection is planimetric
     # returns 2 if projection is latitude-longitude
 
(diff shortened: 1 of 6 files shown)


doc/examples/python/m.distance.py
```diff
@@ -68,7 +68,7 @@ def main():
     # calc distance
 
     proj_type = G_begin_distance_calculations()
-    # returns 0 if projection has no metrix (ie. imagery)
+    # returns 0 if projection has no metrics (ie. imagery)
```
echoix, line 71: I have a question here. Is it "doesn't have metrics", or "is not metric".
The source of [G_begin_distance_calculations](https://grass.osgeo.org/programming8/gis_2distance_8c.html#a988f3691493c06d4c7410d564c78c840)(void) writes the same:
image
But the 0 case would be returned by [G_database_units_to_meters_factor](https://grass.osgeo.org/programming8/defs_2gis_8h.html#a385919f1c6753d3274e44deae2f18f21)();
image

There, the documentation header is clearer. But maybe one or both of them could be wrong

echoix, line 71:
```suggestion
    # returns 0 if projection is not metric (ie. imagery)
```

echoix (commented):
Probably to apply to other cases too like #5363 lib/gis/distance.c

luzpaz:
squashed and rebased

echoix (approved):

PR #5378, opened 2025-03-13
pythonlib: Add missing docstring summary lines to fix D402 errors

arohanajit:
This error popped up post updating `ruff` to latest version


echoix (approved):
Thanks! It would have appeared this weekend indeed. They released a minor (aka major) version today, with breaking changes and new stabilized rules.

Since I'm probably not available this weekend, at least, not on my computer, do you want to take a look at what is needed to do to have the version change go through correctly? I'll trigger the PR so you can look at what's missing after this

PR #5386, opened 2025-03-14
cmake: Consider source path with underscores

HuidaeCho:
This PR fixes locale CMake build when the GRASS source path contains underscores. For example, `conda-build` creates a temporary build directory using timestamp with an underscore (e.g., `.../envs/grass-conda/conda-bld/grass_1741956650154/work/build`).

Needs testing on Windows...


nilason:
Perhaps we can extract the file name here with [get_filename_component](https://cmake.org/cmake/help/v3.22/command/get_filename_component.html)?

nilason:
> Perhaps we can extract the file name here with [get_filename_component](https://cmake.org/cmake/help/v3.22/command/get_filename_component.html)?

Here I'm referring to L17, where we only extract the locale.

HuidaeCho:
> > Perhaps we can extract the file name here with [get_filename_component](https://cmake.org/cmake/help/v3.22/command/get_filename_component.html)?
> 
> Here I'm referring to L17, where we only extract the locale.

Yes, just extracting filename components early would be ideal.

nilason:
Something like this might work, replacing L16:

```cmake
set(_locales)
foreach(_locale ${po_files})
  get_filename_component(_loc ${_locale} NAME)
  list(APPEND _locales ${_loc})
endforeach()
```

nilason:
Looks good to me!

(#5426 is a blocker for merging any PR currently.)

nilason (approved):

PR #5398, opened 2025-03-15
lib/raster: Fix Resource Leak issue in cats.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1270327)


nilason (changes requested):
There are a total of nine similar issues left in `/lib/raster/*`, please address them all here.

ShubhamDesai:
> There are a total of nine similar issues left in `/lib/raster/*`, please address them all here.

before this comment i had already opened PR for CIDs 1208145, 1208146, 1415666, 1415672 in #5407 
Remaining 5 are here 1270327, 1208045, 1208055, 1208056, 1299726

lib/raster/reclass.c
```diff
@@ -52,8 +52,11 @@ int Rast_is_reclass(const char *name, const char *mapset, char *rname,
 
     type = reclass_type(fd, &rname, &rmapset);
     fclose(fd);
-    if (type < 0)
+    if (type < 0) {
+        G_free(rname);
+        G_free(rmapset);
```
nilason, line 57: This one is trickier than first appears.

The typical (perhaps exclusive) use of `Rast_is_reclass()` is by passing (static) arrays for `rname` and `rmapset`:

```c
char rname[GNAME_MAX], rmapset[GMAPSET_MAX];
int is_reclass = (Rast_is_reclass(name, mapset, rname, rmapset) > 0);
```

However `reclass_type()`, unfortunately allows for using both a static and a dynamic array, by either strcpy to the static array or by allocating a new string with G_store. Only in the latter case, we are allowed to free the memory. Given that the latter case, currently never happens, we should skip CID 1208055 and 1208056 this time.

(A possible solution, for another PR, could be to update the documentation of `Rast_is_reclass()` to emphasise that `rname` and `rmapset` must be a pointer to an (allocated) character array; update `reclass_type()` according to this fact (remove G_store etc.); and make sure all calls to `Rast_is_reclass()` use static arrays which are NUL initialised.)

nilason (changes requested):
A few suggestions below. And I miss 1208045 from color_rule.c.

lib/raster/format.c
```diff
@@ -151,6 +151,7 @@ static int read_row_ptrs(int nrows, int old, off_t *row_ptr, int fd)
     return 1;
 
 badread:
+    G_free(buf);
```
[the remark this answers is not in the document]
ShubhamDesai: done

lib/raster/reclass.c
ShubhamDesai: yeah i thought the same but i was not sure if i do too many changes how to test the working so i tried in the same function.
yeah but for time being i am removing those 2 CIDs from PR and create another PR for those issues.

nilason: Only lib/raster/color_rule.c (CID1208045) remain.

ShubhamDesai: Done

nilason (approved):

PR #5399, opened 2025-03-15
adding blank line

shonilbhide:


wenzeslaus:
What are you trying to achieve by this? The branch name does not fit with the change and the commit message? Would you like to make a more meaningful contribution than one white space style change? For example, there is nothing wrong with a PR fixing typos. Actually, that's a great first contribution. However, fixing one single typo adds to repository traffic, increasing maintainer burden without a commensurate benefit to the project. You also need to provide rationale for the change in the description. I can actually find you in NC State directory, so I'm not closing this, but I'm not approving the CI run at this point.

PR #5403, opened 2025-03-17
style: Fixed noqa warnings inline with recent ruff update

arohanajit:


echoix (approved):
It's not that pretty, but not really important at all. Let's just go with this as is

PR #5407, opened 2025-03-17
lib/raster: Fix Resource Leak Issue in range.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1208145, 1208146, 1415666, 1415672)
Used G_free(), fclose() to fix this issue.

### DIFF
--- a/lib/raster/range.c
+++ b/lib/raster/range.c
@@ -205,7 +205,7 @@ int Rast_read_range(const char *name, const char *mapset, struct Range *range)
             else {
                 G_free(mname);
                 return -1;
-            }        
+            }
         }
         else
             Rast_quant_get_limits(&quant, &dmin, &dmax, &x[0], &x[1]);
--- a/lib/raster/range.c
+++ b/lib/raster/range.c
@@ -246,8 +246,7 @@ int Rast_read_range(const char *name, const char *mapset, struct Range *range)
     }
     else {
         /* "range" file does not exist */
-        G_warning(_("Missing range file for <%s> (run r.support -s)"),
-                  mname);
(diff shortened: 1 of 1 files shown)


lib/raster/range.c
```diff
@@ -168,14 +168,15 @@ int Rast_read_range(const char *name, const char *mapset, struct Range *range)
 
     Rast_init_range(range);
     fd = NULL;
+    char *mname = G_fully_qualified_name(name, mapset);
```
nilason, line 171: Same, zou need to free mname before every return.

lib/raster/range.c
```diff
@@ -288,6 +288,7 @@ int Rast_read_rstats(const char *name, const char *mapset,
     init_rstats(rstats);
 
     fd = -1;
+    char *mname = G_fully_qualified_name(name, mapset);
```
nilason, line 291: Move this line (L291) to after next if statement (->L296).
Same here, you need to free mname before every return.

nilason (changes requested):

lib/raster/range.c
```diff
@@ -96,20 +96,19 @@ int Rast_read_fp_range(const char *name, const char *mapset,
     }
 
     fd = -1;
+    char *mname = G_fully_qualified_name(name, mapset);
```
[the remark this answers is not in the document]
ShubhamDesai: done

lib/raster/range.c
ShubhamDesai: done

lib/raster/range.c
ShubhamDesai: done

nilason (approved):

PR #5427, opened 2025-03-21
docs: Remove old footer links from MkDocs site

### LINKED ISSUE #5314: Replace links in custom footer by MkDocs navigation
Replace the custom footer links to index pages by MkDocs nav (menu) items. This can be at the top or the bottom and it does not have to the exact links or wording, the point is to have whatever is the best, not necessarily the original.

The current footer has an issue that the links either don't work for addons which are in a subdir or they need to be absolute URLs, so they would link to a wrong version and always online. This is due to an MkDocs limitation (["it's not possible to use relative links at this point"](https://github.com/squidfunk/mkdocs-material/discussions/5362)). Rather than overcoming the issue, our time is probably better spent on creating a better navigation rather than maintaining a block of links which were replacing a missing navigation in the custom HTML doc.

This may actually remove the need to have the large partial footer file which does not seem to be something we want to maintain - there is a ton of general code which we don't customize and the only reason we customize it are the footer links:

```html
<center>
        <a href="{{ config.site_url }}index.html">Main index</a> | <a href="{{ config.site_url }}topics.html">Topics index</a> | <a href="{{ config.site_url }}keywords.html">Keywords index</a> | <a href="{{ config.site_url }}tags.html">Tags index</a> | <a href="{{ config.site_url }}graphical_index.html">Graphical index</a> | <a href="{{ config.site_url }}full_index.html">Full index</a><br>
</center>
```

### Current state

Image

wenzeslaus:
This is removing the series of links in the footer of the new documentation which go to the various index pages. These links to index pages served as a menu in the custom HTML documentation, but don't have much use in the new documentation which is using MkDocs navigation with a top bar menu being the main navigation place. In very special cases, the bottom bar contained extra link to GUI and the link to tool category would be generated dynamically. This has a limited use as menus are not expected to be changing, instead you expect to find things in the same place there. (So this is something which the new MkDocs menus don't have, but it is not a problem.) In the future, if anything from this footer links should be linked from menu, it should be added to the new menu. The full index having links to all tools is part of the menu.

Closes #5314

Before and after:

Screenshot from 2025-03-21 15-32-54

### DIFF
--- a/man/mkdocs/overrides/partials/footer.html
+++ b/man/mkdocs/overrides/partials/footer.html
@@ -86,9 +86,6 @@
 
   <!-- Further information -->
    <div class="md-footer-meta md-typeset">
-      <center>
-        <a href="{{ config.site_url }}index.html">Main index</a> | <a href="{{ config.site_url }}topics.html">Topics index</a> | <a href="{{ config.site_url }}keywords.html">Keywords index</a> | <a href="{{ config.site_url }}tags.html">Tags index</a> | <a href="{{ config.site_url }}graphical_index.html">Graphical index</a> | <a href="{{ config.site_url }}full_index.html">Full index</a><br>
-      </center>
      <div class="md-footer-meta__inner md-grid">
       {% include "partials/copyright.html" %}
       <!-- Social links -->


wenzeslaus:
I added "before and after" screenshot.

veroandreo (approved):

PR #5440, opened 2025-03-24
Handle null cases in assertRasterMinMax

### LINKED ISSUE #3241: [Bug] gunittest.TestCase.assertRasterMinMax TypeError if range is NULL
Tested with GRASS 8.4.0dev (2023)
```
r.info map=ccc.1 -r
min=NULL
max=NULL
```
```
File "etc/python/grass/gunittest/case.py", line 533, in assertRasterMinMax
    if refmin > actual["min"]:
       ^^^^^^^^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'int' and 'str'
```

kritibirda26:
assertRasterMinMax currently fails with a TypeError if either the provided reference values or the actual determined range values are NULL. Update the method to handle NULL cases correctly.

Fixes #3241

### DIFF
--- a/python/grass/gunittest/case.py
+++ b/python/grass/gunittest/case.py
@@ -524,7 +524,27 @@ class TestCase(unittest.TestCase):
         """
         stdout = call_module("r.info", map=map, flags="r")
         actual = text_to_keyvalue(stdout, sep="=")
-        if refmin > actual["min"]:
+        if actual["min"] == "NULL":
+            actual["min"] = None
+        if actual["max"] == "NULL":
+            actual["max"] = None
+
+        if actual["min"] is None and refmin is not None:
+            stdmsg = (
+                "The actual minimum is null for raster map {m}, "
+                "but a reference minimum ({r}) was provided".format(r=refmin, m=map)
+            )
+            self.fail(self._formatMessage(msg, stdmsg))
+        if actual["min"] is not None and refmin is None:
+            stdmsg = (
+                "The actual minimum is ({a}) for raster map {m}, "
+                "but the provided reference minimum was null".format(
+                    a=actual["min"], m=map
+                )
+            )
+            self.fail(self._formatMessage(msg, stdmsg))
+
+        if actual["min"] is not None and refmin is not None and refmin > actual["min"]:
             stdmsg = (
                 "The actual minimum ({a}) is smaller than the reference"
                 " one ({r}) for raster map {m}"
@@ -533,7 +553,24 @@ class TestCase(unittest.TestCase):
                 )
             )
             self.fail(self._formatMessage(msg, stdmsg))
-        if refmax < actual["max"]:
+
+        if actual["max"] is None and refmax is not None:
+            stdmsg = (
+                "The actual maximum is null for raster map {m}, "
+                "but a reference maximum ({r}) was provided".format(r=refmax, m=map)
+            )
+            self.fail(self._formatMessage(msg, stdmsg))
+
+        if actual["max"] is not None and refmax is None:
+            stdmsg = (
+                "The actual maximum is ({a}) for raster map {m}, "
+                "but the provided reference maximum was null".format(
+                    a=actual["max"], m=map
+                )
+            )
+            self.fail(self._formatMessage(msg, stdmsg))
+
+        if actual["max"] is not None and refmax is not None and refmax < actual["max"]:
             stdmsg = (
                 "The actual maximum ({a}) is greater than the reference"
                 " one ({r}) for raster map {m}"
--- a/python/grass/gunittest/testsuite/test_assertions.py
+++ b/python/grass/gunittest/testsuite/test_assertions.py
@@ -6,6 +6,7 @@ import os
 from pathlib import Path
 
 import grass.script.core as gcore
+from grass.script import tempname
 from grass.pygrass.modules import Module
 
 from grass.gunittest.case import TestCase
@@ -179,6 +180,16 @@ class TestRasterMapAssertions(TestCase):
     @classmethod
     def tearDownClass(cls):
         cls.del_temp_region()
+
+    def test_assertRasterMinMax(self):
+        test_map = tempname(10)
+        self.runModule("r.mapcalc", expression=f"{test_map} = null()")
+        with self.assertRaises(AssertionError) as context:
+            self.assertRasterMinMax(test_map, refmin=1, refmax=10)
+        self.assertEqual(
+            str(context.exception),
+            f"The actual minimum is null for raster map {test_map}, but a reference minimum (1) was provided",
+        )
 
     def test_assertRasterFitsUnivar(self):
         self.assertRasterFitsUnivar(


python/grass/gunittest/case.py
```diff
@@ -524,7 +524,27 @@ def assertRasterMinMax(self, map, refmin, refmax, msg=None):
         """
         stdout = call_module("r.info", map=map, flags="r")
         actual = text_to_keyvalue(stdout, sep="=")
-        if refmin > actual["min"]:
+        if actual["min"] == "NULL":
+            actual["min"] = None
+        if actual["max"] == "NULL":
+            actual["max"] = None
```
wenzeslaus, line 530: The correct solution at this point is to just use the JSON output with parse_command.

python/grass/gunittest/case.py
```diff
@@ -524,7 +524,27 @@ def assertRasterMinMax(self, map, refmin, refmax, msg=None):
...
+        if actual["min"] is not None and refmin is None:
+            stdmsg = (
+                "The actual minimum is ({a}) for raster map {m}, "
+                "but the provided reference minimum was null".format(
+                    a=actual["min"], m=map
+                )
+            )
+            self.fail(self._formatMessage(msg, stdmsg))
+
+        if actual["min"] is not None and refmin is not None and refmin > actual["min"]:
```
wenzeslaus, line 547: The current code is long and the conditions hard to navigate. Can you try switching the conditions so that they are not that long? For example, by testing the "both none" case explicitly as the first thing:

```python
if actual["min"] is None and refmin is None:
     test_min = False
# ...
if test_min and refmin > actual["min"]:
     # ...
```

Adding 4 more messages is also too much. While this test has two messages already, because it basically tests two values, other tests are fine with just one message - sort of "values differ ({values listed here})".

At minimum (without changing the rest of the function), what about something like:

```python
if test_min and (actual["min"] is None or refmin is None):
   stdmsg = (
    f"The actual minimum for raster map {m} is {...}, "
    f"but a reference minimum ({r}) was provided"
)
```

Also, here the strings can be f-strings.

wenzeslaus (changes requested):

python/grass/gunittest/case.py
kritibirda26: I updated the code to use `parse_command` but due to a design decision last year to represent numbers using strings for precision, I had to add a utility to make sure the numbers are parsed properly. For now, it might be useful to include that in the `parse_command` itself, under a flag maybe?

python/grass/gunittest/case.py
kritibirda26: I have updated the function to use a common message and also simplified the boolean checks.

python/grass/gunittest/case.py
wenzeslaus: Whatever was the discussion back then, r.info gives me floats now and I think it is right to do that:

```bash
grass --tmp-mapset ~/grassdata/nc_spm_08_grass7/ --exec r.info elevation format=json
```
```json
{
    "north": 228500,
    "south": 215000,
    "nsres": 10,
    "east": 645000,
    "west": 630000,
    "ewres": 10,
    "rows": 1350,
    "cols": 1500,
    "cells": 2025000,
    "datatype": "FCELL",
    "ncats": 255,
    "min": 55.578792572021484,
    "max": 156.32986450195312,
    "map": "elevation",
    "maptype": "raster",
    "mapset": "PERMANENT",
    ...
}
```
```bash
grass --tmp-mapset ~/grassdata/nc_spm_08_grass7/ --exec python -c 'import grass.script as gs; print(gs.parse_command("r.info", map="elevation", format="json"))'
```
```python
{'north': 228500, 'south': 215000, 'nsres': 10, 'east': 645000, 'west': 630000, 'ewres': 10, 'rows': 1350, 'cols': 1500, 'cells': 2025000, 'datatype': 'FCELL', 'ncats': 255, 'min': 55.578792572021484, 'max': 156.32986450195312, 'map': 'elevation', 'maptype': 'raster', 'mapset': 'PERMANENT', 'location': 'nc_spm_08_grass7', 'project': 'nc_spm_08_grass7', 'database': '/home/vpetras/grassdata', 'date': 'Tue Nov  7 01:09:51 2006', 'creator': 'helena'...}
```

kritibirda26: Oh that's great, my bad I should verified it myself again. Using parse_command directly now.

kritibirda26: I looked at the failing test on macOS and it seems to be there because of the precision of numbers issue. I did find a method to check equality with precision but not one to check greater/less than. I think to fix the issue we can add precision support to this method. Something like: `(refmin - actual['min']) > precision`?

wenzeslaus: Why this surfaced here? Why the test now fails and it didn't fail before?

python/grass/gunittest/case.py
```diff
@@ -522,24 +521,22 @@ def assertRasterMinMax(self, map, refmin, refmax, msg=None):
...
-                )
-            )
-            self.fail(self._formatMessage(msg, stdmsg))
-        if refmax < actual["max"]:
+        actual = parse_command("r.info", map=map, format="json")
+        min_mismatch = (
+            ((actual["min"] is None) != (refmin is None))
+            # previous check only passes if both values are None or not None
+            # at the same time, hence just checking one here is sufficient
+            or (actual["min"] is not None and (refmin - actual["min"] > precision))
```
wenzeslaus, line 529: This SO discussion has different opinion on making approximate (precision-dependent) comparisons other than equality (`>` and `<`).

https://stackoverflow.com/questions/54147658/pytest-approx-with-and

I find the current comparison little hard to navigate. Perhaps extent the reference by precision in the appropriate direction and then do `>`.

Also note the doc for this function. It is basically testing a range, not the specific values of min and max values.

wenzeslaus (changes requested):

python/grass/gunittest/case.py
wenzeslaus: This function tests if min and max fall into an interval - see docs. Do the new tests do that? Please check if that's why the tests fail.

python/grass/gunittest/case.py
wenzeslaus: The shorter code is indeed better.

python/grass/gunittest/case.py
kritibirda26: The data of the failing test is:
```
Actual: minimum=74.75373840332031, maximum=156.32986450195312
Reference: minimum=74.75374, maximum=156.3299
```
Testing for refmin <= actualmin: `74.75374 <= 74.75373840332031` results to False. This breaks now because the JSON output contains the numbers upto double precision whereas the text output contains it upto float precision.

If we relax the reference values in the tests slights, the tests should pass again.

python/grass/gunittest/case.py
kritibirda26: Sure, will do.

python/grass/gunittest/testsuite/test_assertions.py
```diff
@@ -180,6 +181,18 @@ def setUpClass(cls):
     def tearDownClass(cls):
         cls.del_temp_region()
 
+    def test_assertRasterMinMax(self):
```
wenzeslaus, line 184:
```suggestion
    def test_assertRasterMinMaxNulls(self):
```

python/grass/gunittest/testsuite/test_assertions.py
```diff
@@ -180,6 +181,18 @@ def setUpClass(cls):
...
+        test_map = tempname(10)
```
wenzeslaus, line 185: It does not matter here, but this gives a better, more robust example:

```suggestion
        test_map = append_uuid("tmp")
```

Everything is temporary here anyway. We made some effort in deleting what we create in grass.gunittest-based tests in tear down methods, but it may not be worth it. See other tests which have bigger tear down methods.

wenzeslaus (changes requested):
I agree with the logic in the assert method and the wider range in testing. I have some notes about the test.

raster/r.tile/testsuite/testrt.py
```diff
@@ -63,8 +63,8 @@ def test_overlap(self):
         )
         self.assertRasterMinMax(
             map=tile,
-            refmin=74.75374,
-            refmax=156.3299,
+            refmin=74.75373,
+            refmax=156.3300,
             msg="tile-000-000 in degrees must be between 74.75374 and 156.3299",
```
wenzeslaus, line 68: This needs an update, but actually, deleting should be fine too. The default message will do here. Same below.

wenzeslaus (changes requested):

python/grass/gunittest/case.py
wenzeslaus: I agree, given the higher precision, relaxing the tests makes sense.

PR #5446, opened 2025-03-25
docs: Replace keywords index by MkDocs tags

wenzeslaus:
This removes custom keywords index and replaces the keywords.html page by a MkDocs tags listing.

The default MkDocs tag anchor contains tag: prefix to help distinguishing tag headings from others, but we likely won't have a heading of the same name on the same page. However, we were using anchors without any prefix, so this will make most of the anchor links work (if there are any external ones anywhere). It also makes the anchor little simple to read and write if managed manually for any reason.

The check for anchor existence is done by MkDocs before the tags page is generated. This creates false positive for the Keywords section (to be removed) and for topics. To avoid this warning for topics, and perhaps to be little more helpful, instead of linking to the keyword, this new code uses MkDocs tags listing to list pages with that keyword right on the topic page. Unless MkDocs changes when or how the check is done, we will get a warning for manually included links to keywords (tags), e.g., from the See also section.

This disables custom keywords build and disables the addons keyword integration in the CI workflow. The Python code to generate the keywords in Markdown is still in place, but it can be removed in the future possibly together with the custom HTML code (they are mixed together).

## keywords.html: old doc - custom keyword index - MkDocs tag listing

image

(The difference in listed tools in the example is given by doc build with and without addons. The tag listing includes addons automatically since MkDocs processes all Markdown files at once in a separate build step.)

## Topic page

image

(We may remove or completely redo the topics in the future, but this keeps external links working if any and the resulting pages look okay.)

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -118,15 +118,13 @@ jobs:
           mkdir "$MKDOCS_DIR/source/addons"
           mv -v addons-build-dir/docs/md/source/* "$MKDOCS_DIR/source/addons"
 
-      - name: Build index and rebuild keywords
+      - name: Build index
         run: |
           export ARCH="$(grass --config arch)"
           export ARCH_DISTDIR="$(grass --config path)"
           export VERSION_NUMBER="$VERSION"
           grass --tmp-project XY --exec \
             python grass/man/build_full_index.py md index "$MKDOCS_DIR/source/addons" addons
-          grass --tmp-project XY --exec \
-            python grass/man/build_keywords.py md "$MKDOCS_DIR/source" "$MKDOCS_DIR/source/addons"
 
       - name: Copy shared files to addons
         run: |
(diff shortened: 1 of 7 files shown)


neteler:
I'd either have the tools under each keyword be comma separated (as it used to be) to avoid an endless keywords page or add the short description to each tool to use the currently white space.

wenzeslaus:
I added description for each tool and made the tool name in cursive which we tend to use in the doc. This provides details for each tool without need to visit the link. It is long and some descriptions are long, so even one keyword section can substantial, but it somewhat mirrors the format of the other tool listings which include descriptions. (Full and category index pages have descriptions, but currently use a table, no cursive, and don't repeat the listed tools.)

## In the current colors

image

## A possible style with green links

(not part of this PR)

image

neteler:
This looks great!

petrasovaa (approved):

PR #5493, opened 2025-04-02
CQ: remove misused gisprompt options

nilason:
Remove misused gisprompt option, which is expected to be a [comma separated three part value](https://grass.osgeo.org/programming8/gislib_cmdline_parsing.html#gisprompt_Member).

### DIFF
--- a/temporal/t.rast.accdetect/t.rast.accdetect.py
+++ b/temporal/t.rast.accdetect/t.rast.accdetect.py
@@ -94,7 +94,7 @@
 # % description: A numerical suffix separated by an underscore will be attached to create a unique identifier
 # % required: yes
 # % multiple: no
-# % gisprompt:
+# #% gisprompt:
 # %end
 
 # %option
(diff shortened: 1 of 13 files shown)


temporal/t.rast.accdetect/t.rast.accdetect.py
```diff
@@ -94,7 +94,7 @@
 # % description: A numerical suffix separated by an underscore will be attached to create a unique identifier
 # % required: yes
 # % multiple: no
-# % gisprompt:
+# #% gisprompt:
```
neteler, line 97: Also to be deleted like below?

temporal/t.rast.accumulate/t.rast.accumulate.py
```diff
@@ -95,7 +95,7 @@
 # % description: A numerical suffix separated by an underscore will be attached to create a unique identifier
 # % required: yes
 # % multiple: no
-# % gisprompt:
+# #% gisprompt:
```
neteler, line 98: Also to be deleted like below?

temporal/t.rast.accdetect/t.rast.accdetect.py
nilason: Oh, left over from testing, thought I removed them. Will do...

wenzeslaus (approved):

PR #5498, opened 2025-04-02
lib/gis/stac: Added G_OPT_STAC_API

cwhite911:
Added a new standard option **G_OPT_STAC_API**. The option provides a drop down list of 90 public STAC APIs to reduce friction finding STAC resources. The plan is to use the option with the **t.stac** addons and have it available for any other new related tools (e.g., gui/scripts/g.stac.py).

The list of URLs is read from the JSON file *stac_api.json* that was sourced from [https://stacindex.org/catalogs?access=public&type=api](https://stacindex.org/catalogs?access=public&type=api) on 2025-04-02. It is not exhaustive, only including public STAC APIs.


wenzeslaus:
Can you provide examples how this would look like from user perspective?

cwhite911:
@wenzeslaus If you click on the drop down it provides a bunch of additional urls. We have the name and description of each STAC API as well that is currently not displayed. So if you have any suggestions of how to incorporated them let me know. I tried using descriptions but it was way to busy.

Screenshot from 2025-04-02 20-35-57

petrasovaa:
I would need to think about this more, but I don't think this is a great solution, having a json that gets obsolete quickly in the library. And just selecting the catalog is not enough, more interactive selection of the assets is needed. A dedicated GUI might be better like with r.in.wms.

cwhite911:
Ok I'm going to take another route to achieve this from addons.

PR #5517, opened 2025-04-07
r.to.vect: Fix uninitialized variable issue in points.c

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1208461, 1208459)

### DIFF
--- a/raster/r.to.vect/points.c
+++ b/raster/r.to.vect/points.c
@@ -75,6 +75,8 @@ int extract_points(int z_flag UNUSED)
                     continue;
                 dval = dcellbuf[col];
                 break;
+            default:
+                G_fatal_error(_("Unsupported raster data type encountered."));
             }
 
             /* value_flag is used only for CELL type */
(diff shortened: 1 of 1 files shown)


nilason (changes requested):
This is now a false positive, the switch statement (L61) is currently exhausting the options. To be on the safe side (for future changes) we should add a `default` label, which `G_fatal_error` if reached.

ShubhamDesai:
> This is now a false positive, the switch statement (L61) is currently exhausting the options. To be on the safe side (for future changes) we should add a `default` label, which `G_fatal_error` if reached.

Done

nilason (changes requested):
It would be better to make an early exit, to put the default to the very first switch statement in the function.

ShubhamDesai:
> It would be better to make an early exit, to put the default to the very first switch statement in the function.

done shifted the default case to first switch statement

nilason (approved):
Thanks!

PR #5525, opened 2025-04-10
docs: Use Documentation instead of Reference Manual

wenzeslaus:
Use simpler Documentation instead of loaded Reference Manual. We call it both doc and ref man and using doc in the doc itself should lead to just one name although a less specific one. Doc is also just one work so it avoids title case versus sentence case inconsistencies.

This shuffles things around in a couple of the See also sections which were linking the index page which now is simply linked from the navigation, mainly it moves EPSG links from helptext aka quick start aka GUI intro) into projectionintro page.

## New

image

## Old

image

### DIFF
--- a/gui/wxpython/docs/wxgui_sphinx/conf.py
+++ b/gui/wxpython/docs/wxgui_sphinx/conf.py
@@ -22,7 +22,7 @@ footer_tmpl = string.Template(
 | <a href="../keywords.html">Keywords Index</a>
 | <a href="../full_index.html">Full Index</a></p>
 <p>&copy; 2003-${year} <a href="https://grass.osgeo.org">GRASS Development Team</a>,
-GRASS GIS ${grass_version} Documentation</p>
+GRASS ${grass_version} Documentation</p>
 {% endblock %}
 """
 )
--- a/gui/wxpython/docs/wxgui_sphinx/conf.py
+++ b/gui/wxpython/docs/wxgui_sphinx/conf.py
@@ -22,7 +22,7 @@ footer_tmpl = string.Template(
 | <a href="../keywords.html">Keywords Index</a>
 | <a href="../full_index.html">Full Index</a></p>
 <p>&copy; 2003-${year} <a href="https://grass.osgeo.org">GRASS Development Team</a>,
-GRASS GIS ${grass_version} Reference Manual</p>
+GRASS GIS ${grass_version} Documentation</p>
 {% endblock %}
 """
 )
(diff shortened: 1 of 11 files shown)


man/build_graphical_index.py
```diff
@@ -67,11 +67,10 @@ def main(ext):
...
 
     with open(os.path.join(man_dir, output_name + f".{ext}"), "w") as output:
-        output.write(
-            header1_tmpl.substitute(
-                title=f"GRASS {grass_version} Reference Manual - Graphical index"
-            )
-        )
+        title = "Graphical index"
+        if ext == "html":
+            title = f"GRASS {grass_version} Reference Manual - {title}"
```
petrasovaa, line 72: Should this be Documentation?

man/build_topics.py
```diff
@@ -108,11 +108,10 @@ def build_topics(ext):
...
 
     with Path(man_dir, f"topics.{ext}").open("w") as topicsfile:
-        topicsfile.write(
-            header1_tmpl.substitute(
-                title="GRASS %s Reference Manual - Topics index" % grass_version
-            )
-        )
+        title = "Topics index"
+        if ext == "html":
+            title = f"GRASS {grass_version} Reference Manual - {title}"
```
petrasovaa, line 113: The same here

petrasovaa (changes requested):

man/build_graphical_index.py
wenzeslaus: I'm not changing it for HTML (as opposed to Markdown/MkDocs), so I'll leave these as is.

gui/wxpython/docs/wxgui_sphinx/conf.py
```diff
@@ -22,7 +22,7 @@
 | <a href="../keywords.html">Keywords Index</a>
 | <a href="../full_index.html">Full Index</a></p>
 <p>&copy; 2003-${year} <a href="https://grass.osgeo.org">GRASS Development Team</a>,
-GRASS GIS ${grass_version} Reference Manual</p>
+GRASS GIS ${grass_version} Documentation</p>
```
wenzeslaus, line 25:
```suggestion
GRASS ${grass_version} Documentation</p>
```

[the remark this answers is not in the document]
wenzeslaus: Right, I was trying to get those too when on the same line. Otherwise, the GRASS GIS to GRASS change is still on the todo list, but in doc, much less prominent after this PR.

PR #5570, opened 2025-04-22
CQ: replace 'sprintf()` in raster and raster3d directories

nilason:
See: #5568.


echoix (approved):

nilason:
Coverity Scan reported a new issue with this PR:

> overlapping_copy: In the call to function snprintf, the object pointed to by argument c->suffix may overlap with the object pointed to by argument c->name

https://github.com/OSGeo/grass/blob/4144ba81fa125e1549fc52e1cf5feb4fa004be8b/raster/r.in.png/main.c#L84


I can't get my head around that one. False positive?

PR #5591, opened 2025-04-25
grass.script: Add locking to init

wenzeslaus:
The function _gs.setup.init_ can now optionally lock the mapset and unlock when exiting the context or with call to finish (both free function or bound method). This keeps the current state of not locking as a default, i.e., the locking is opt-in, so multiple session are responsibility of the user just like running multiple processes within one mapset.

Importantly, this brings feature parity with the grass command to Python, and makes it possible for users to use the mapset locking from Python.

The PID used is the the current process ID which is already used for variable GIS_LOCK in init. The underlying locking (but not unlocking) functionality is the same one as used for the main _grass_ command thanks to a previous refactoring of locking from lib/init/grass.py to _grass.app.data_.

The tests are included, but the thing about testing this is that when lock is found lib/init/lock.c (etc/lock) waits for a second before proceeding to check the file content. This makes each test with existing lock to take at least one second. There is currently six tests like that. The lock program wants to read the PID written to the lock file, so it waits for one second to ensure that another process wrote the content to the file. Without a lock preset, the operation is fast.

### DIFF
--- a/python/grass/app/data.py
+++ b/python/grass/app/data.py
@@ -172,7 +172,9 @@ class MapsetLockingException(Exception):
     pass
 
 
-def lock_mapset(mapset_path, force_lock_removal, message_callback):
+def lock_mapset(
+    mapset_path, force_lock_removal, message_callback, process_id=None, env=None
+):
     """Acquire a lock for a mapset and return name of new lock file
 
     Raises MapsetLockingException when it is not possible to acquire a lock for the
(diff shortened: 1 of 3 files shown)


joao-krauchenco-mottu (approved):

wenzeslaus:
@joao-krauchenco-mottu Thanks for the review. First in this repo from what I can tell. That's great. I appreciate any feedback.

echoix (approved):

PR #5601, opened 2025-04-27
r.plane: remove misused gisprompt option

nilason:
As the option does not contain a comma separated three part string, this causes heap buffer overflow. This is probably the cause of some random CI failures with `r.plane`.

Follow-up to #5493.

### DIFF
--- a/scripts/r.plane/r.plane.py
+++ b/scripts/r.plane/r.plane.py
@@ -27,7 +27,6 @@
 # %option
 # % key: dip
 # % type: double
-# % gisprompt: -90-90
 # % answer: 0.0
 # % description: Dip of plane in degrees
 # % required : yes
@@ -35,7 +34,6 @@
 # %option
 # % key: azimuth
 # % type: double
-# % gisprompt: 0-360
 # % answer: 0.0
 # % description: Azimuth of the plane in degrees
 # % required : yes
--- a/scripts/r.plane/r.plane.py
+++ b/scripts/r.plane/r.plane.py
@@ -27,6 +27,7 @@
 # %option
 # % key: dip
 # % type: double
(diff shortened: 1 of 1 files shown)


wenzeslaus:
It looks like "options" or whatever the allowed values are called.

nilason:
> It looks like "options" or whatever the allowed values are called.

They are passed to the module parser and G__split_gisprompt() assumes the option value is correct:

https://github.com/OSGeo/grass/blob/2e55b158af94127c520d384978405d1da561f534/lib/gis/parser.c#L1773


During compilation this is triggered by documentation generation (calling parser html|md).

wenzeslaus:
I don't doubt this fails. What I meant is that there is `options` and the current values look like the value to set there.

example: https://github.com/OSGeo/grass/blob/main/raster/r.viewshed/main.cpp#L509
doc: https://grass.osgeo.org/programming8/structOption.html#aaf64185e77299e332f9fcb4880a0507b

nilason:
> I don't doubt this fails. What I meant is that there is `options` and the current values look like the value to set there.
> 
> example: https://github.com/OSGeo/grass/blob/main/raster/r.viewshed/main.cpp#L509 doc: https://grass.osgeo.org/programming8/structOption.html#aaf64185e77299e332f9fcb4880a0507b

Aha, now I see what you mean, you suggest something like:

```diff
- # % gisprompt: -90-90
+ # % options: -90-90
```

wenzeslaus:
Yes!

nilason:
Now I changed to use "options".

wenzeslaus (commented):
Thanks!

wenzeslaus (approved):

PR #5602, opened 2025-04-27
r.li: fix resource leak issues

nilason:
Fix resource leak issues in `r.li[.*]` modules. In addition remove return after G_fatal_error().

This may look like a lot of changes in one PR, but the code in `r.li` is very repetitive (which even Coverity Scan have problem to handle).

(This module group could probably gain a lot by major refactoring and re-use of repetitive code).


wenzeslaus (approved):
This seems appropriate given the code. I don't know how is our test coverage on this, but the changes look safe in relation to the computational results. I didn't disect every each of them, though.

nilason:
> This seems appropriate given the code. I don't know how is our test coverage on this, but the changes look safe in relation to the computational results. I didn't disect every each of them, though.

As far as I could see, the existing test is not actively used in the test runs. I did however run the test manually.

PR #5603, opened 2025-04-28
wxGUI/gconsole: avoid infinite GUI-in-GUI loops when running python scripts with no required params

### LINKED ISSUE #5524: [Bug] GUI generated by Simple Python Editor reopens instead of executing the command if default settings are used
## Describe the bug
When using the Simple Python Editor to create a custom GRASS module with a GUI interface, there's an issue when running the module without changing any of the default values in the GUI. Instead of executing the command, the GUI simply reopens, and the command is not run.  If any parameter is changed, the module runs correctly.

## To reproduce
1. Create a module using Simple Python Editor.
2. Leave all parameters in their default state.
3. Run the module via GUI.
4. the GUI reopens instead of executing the command.

## Expected behavior
The module should execute with the default parameters, without requiring the user to change any values.

pesekon2:
fix #5524

### DIFF
--- a/gui/wxpython/core/gconsole.py
+++ b/gui/wxpython/core/gconsole.py
@@ -673,7 +673,9 @@ class GConsole(wx.EvtHandler):
                 return
 
             skipInterface = True
-            if os.path.splitext(command[0])[1] in {".py", ".sh"}:
+            if os.path.splitext(command[0])[1] in {".py", ".sh"} and not isinstance(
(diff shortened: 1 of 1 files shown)


petrasovaa (approved):

pesekon2:
Backported to releasebranch_8_4 in 9ab0505c6c7af227195836818dd2aa82d4a910e5

PR #5606, opened 2025-04-30
lib/vector/Vlib: Fix Resource Leak Issue in ascii.c

ShubhamDesai:
This pull request fixes issue identified by Coverity scan (CID: 1207967, 1207968, 1207969, 1207970, 1207971, 1207972, 1207973, 1207974, 1207975, 1207976, 1207977)
Used G_free(), db_close_database_shutdown_driver(), Vect_destroy_field_info(), Vect_destroy_list(), Vect_destroy_cats_struct()


nilason (changes requested):
In Vect_read_ascii() we should make use of goto, note that also `Points` and `Cats` need to be freed too before return.

ShubhamDesai:
> In Vect_read_ascii() we should make use of goto, note that also `Points` and `Cats` need to be freed too before return.

Done changes

lib/vector/Vlib/ascii.c
```diff
@@ -476,6 +489,8 @@ int Vect_write_ascii(FILE *ascii, FILE *att, struct Map_info *Map, int ver,
                             G_warning(_("Export cancelled"));
                             db_close_database(driver);
                             db_shutdown_driver(driver);
+                            Vect_destroy_field_info(Fi);
+                            G_free(columns);
```
nilason: `columns` is an array of array (or an array of strings, `char**`). Each string must also be freed.

There is another case, in line 524, which must be updated similarly.

lib/vector/Vlib/ascii.c
```diff
@@ -238,9 +246,12 @@ int Vect_read_ascii(FILE *ascii, struct Map_info *Map)
         G_warning(_("Vector map <%s> is 2D. %d 3D features (faces or kernels) "
                     "skipped."),
                   Vect_get_name(Map), nskipped_3d);
-
+cleanup:
```
nilason: I'd prefer a bit more descriptive label, e.g. 'free_exit' or 'cleanup_exit'.

lib/vector/Vlib/ascii.c
ShubhamDesai: Done

lib/vector/Vlib/ascii.c
ShubhamDesai: Done

PR #5608, opened 2025-04-30
v.vol.rst: Fix Uninitialized Variables

ShubhamDesai:
This pull request fixes issue identified by Coverity Scan (CID : 1208484)


nilason (changes requested):
There is a similar issue with file /vector/v.vol.rst/user3.c, please address that here too.

ShubhamDesai:
> There is a similar issue with file /vector/v.vol.rst/user3.c, please address that here too.

Done.

ShubhamDesai:
> > > There is a similar issue with file /vector/v.vol.rst/user3.c, please address that here too.
> > 
> > 
> > Done.
> 
> That one is a roughly similar issue (CID 1208515), but not the one I had in mind, which is CID1208514. We should, this time leave 1208515, because if we set `mmax` we might end up with a negative array index with `mmax - 1`.
> 
> (Take a look at the status for 1208515, you may notice it is "Triaged". Under the "Triage History" section you may see this issue may be cause of a real bug, which might need a proper (what ever that will mean) solution.)

Done for now i reverted back CID 1208515 and included CID 1208514

PR #5612, opened 2025-05-02
docs: adding new logo to docs

cwhite911:
I've added the new logo to the docs (See screenshot)

## House keeping

We have two files grasslogo_vector.{eps|pdf} files in `man` directory that are used in `utils/module_synopsis.sh` and `ps/ps.map/decorations`.

Are these files still in use or can they be safely removed?

Screenshot from 2025-05-01 21-05-00

## Note 
This PR does not address updating the logo in the GUI. That issue is getting addressed in PR https://github.com/OSGeo/grass/pull/5577


wenzeslaus:
Check also the Sphinx-generated doc for Python API. That has a separate logo. I had to update that when updating the build. You should be able to check the result in the downloaded artifact if you are not building that locally.

> We have two files grasslogo_vector.{eps|pdf} files in `man` directory that are used in `utils/module_synopsis.sh` and `ps/ps.map/decorations`.
> 
> Are these files still in use or can they be safely removed?

Good catch, but they are referenced by these two, so why to remove?

I was looking at `utils/module_synopsis.sh` and that looked like dead code, so perhaps you can see (let AI figure out) what it is doing and delete that in separate PR with the corresponding logo file.

`ps/ps.map/decorations` is not deprecated or anything, so that needs to be updated or "thoughtfully removed".

petrasovaa (changes requested):
Yes, please change also the logo in the python doc:
https://github.com/OSGeo/grass/blob/main/python/grass/docs/_static/grass_logo.svg

cwhite911:
I've updated the eps and pdf versions of the logo, so they can work with PostScript. 

@petrasovaa I'm going to address the logo for the python docs in another PR because I've updated them to MkDocs styling using the `sphinx_material`. So, this PR is good to go unless you have something else you want me to look at.

## Python Docs Styling

image

wenzeslaus:
> ...I've updated them to MkDocs styling using the `sphinx_material`. So, this PR is good to go unless you have something else you want me to look at.

That's awesome and it looks great.

wenzeslaus (approved):

petrasovaa (approved):

PR #5617, opened 2025-05-05
docs: convert Python API docs to Sphinx Material theme

cwhite911:
Converting the Python API documentation to use the [Sphinx Material](https://bashtage.github.io/sphinx-material/) theme styled to match the main GRASS documentation.

## Example View
image


petrasovaa:
The heading style causing "Script.Array Module" should be removed

cwhite911:
This PR is at a point where the bulk of the modifications that could be done are done. There are still some remaining issues see Issue #5619 that we can resolve moving forward as we set priorities.

Here are some screenshots to give you an example of the current state.

## Getting Started (index.html)

image

## Scripting API

image

## script package (script.html)

This one is really long, but I think we can improve it by tweaking some of the `sphinx-apidoc` settings where it will break the code base into smaller files with `-e, --separate        put documentation for each module on its own page`

Screenshot 2025-05-06 at 15-20-41 script package — GRASS Python Library Documentation

## PyGRASS API

image

## Introduction to Raster classes

image

## GRASS Jupyter API (grass.jupyter package)

image

python/grass/docs/conf.py
```diff
@@ -92,9 +92,9 @@
 today = date.today()
 
 copy("_templates/layout.html.template", "_templates/layout.html")
-with open("_templates/layout.html", "a") as f:
-    f.write(footer_tmpl.substitute(grass_version=grass_version, year=today.year))
-    f.close()
+# with open("_templates/layout.html", "a") as f:
+#     f.write(footer_tmpl.substitute(grass_version=grass_version, year=today.year))
+#     f.close()
```
wenzeslaus, line 97: Delete, not comment out.

.github/workflows/documentation.yml
```diff
@@ -151,7 +151,7 @@ jobs:
 
       - name: Build Sphinx documentation
         run: |
-          pip install sphinx sphinx-sitemap
+          pip install sphinx sphinx-sitemap sphinx-material
```
wenzeslaus, line 154: Pin the versions here? Or in a separate PR.

wenzeslaus (changes requested):

.github/workflows/documentation.yml
cwhite911: We could add it to the requirements.txt file.

petrasovaa (approved):

echoix:
I was curious on checking it out, but it still isn't live yet

PR #5620, opened 2025-05-07
docs: Fixed mkdocs fg-footer color

cwhite911:
The nav footer font color was set incorrectly to white when it should have been set as the secondary color (blue).

The change made the previous and next text visible.

Screenshot from 2025-05-06 22-41-14


wenzeslaus (approved):
I thought it just was not captured in the screenshots when it was in fact invisible. Ouch!

PR #5630, opened 2025-05-09
build: prepare for FHS, where GISBASE is not the absolute truth

nilason:
`GISBASE` is used as a means to retrieve paths to various resources in source code. With FHS (#3661) this is no longer possible. This adds a number of new environment variables to address this.

Currently this is addressed with following workaround:

https://github.com/OSGeo/grass/blob/5e998d33893867bf675a3c61bbf8cce07515cea2/CMakeLists.txt#L176-L194

~~This is work in progress.~~


general/g.mkfontcap/Makefile
```diff
@@ -17,6 +17,7 @@ default: cmd $(CAPFILE)
 endif
 
 $(CAPFILE): $(BIN)/$(PGM)$(EXE)
-	GISRC=junk GISBASE=$(RUN_GISBASE) \
+	GISRC=junk GISBASE=$(RUN_GISBASE) GRASS_SHARE_DIR=$(RUN_GISBASE) \
+	GRASS_LOCALEDIR=$(RUN_GISBASE)/locale \
```
marisn, line 21: Why `GRASS_LOCALEDIR` and `GRASS_SHARE_DIR`?

nilason:
Finally ready for review! Please test it out in every conceivable way (scripting, Jupiter ...), __in particular with the legacy Autotools build__. We don't want any regression. The CMake legacy build should work, but may also be tested of course. 

(The FHS build, is still not fully working – the Python GUI part (module) still needs some consideration.)

echoix:
~~If~~ When this gets merged, how does it play with addons? Are all the required changes made on our side, or some addons' makefiles will need to be adjusted? What about the rare addons not in the OSGeo/grass-addons repo?

wenzeslaus:
I updated the code to align with #6469 in terms of setting the variable explicitly, while keeping the idea that this may be used to set the variables.

The conflict which was already there in setup.py will be better to fix only with #6469 merged to also resolve some of the GISBASE setting and fallback inconsistencies between this and #6469. (for some of it, I can also update #6469 to reduce the changes here).

I have some small tests lined up which will test at least some the variables in their specific context, but most variables here are either already tested (because they are crucial) or they are used in context which is hard-to-test automatically (typically GUI).

@nilason I didn't merge main here to update, because you were doing rebase (not merge) on this branch, but it seems merge would be easier now with the multiple commits, no?

wenzeslaus:
#6482 will bring the main closer to this one, so it will be easier to merge, i.e., no need to attempt merging with main now. (#6482 needs changes to tests I'm working on, but don't have ready yet.)

python/libgrass_interface_generator/CMakeLists.txt
```diff
@@ -63,7 +63,6 @@ if(MSVC)
   string(APPEND _c_flags " -I${CMAKE_SOURCE_DIR}/msvc")
 elseif(APPLE AND CMAKE_OSX_SYSROOT)
   string(APPEND _c_flags " --sysroot ${CMAKE_OSX_SYSROOT}")
-  message("_c_flags")
```
wenzeslaus: Does not seem to be related here.

wenzeslaus:
I merged the main branch into this branch. Some of the new tests fail here on Windows. It seems to be related to creation of a project with g.proj, but I don't know beyond that.

wenzeslaus:
On Windows, g.proj is failing with a return code that may indicate a stack overflow:

```
grass.exceptions.ScriptError: g.proj failed with return code 3221225725, but no error message was produced
```

echoix:
> On Windows, g.proj is failing with a return code that may indicate a stack overflow:
> 
> ```
> grass.exceptions.ScriptError: g.proj failed with return code 3221225725, but no error message was produced
> ```

Is this a new failure? I think I remember seeing this kind of error quite often in the logs.

wenzeslaus:
This specific message is new, see the latest commit, but I don't see why it would surface here and not elsewhere.

echoix:
> This specific message is new, see the latest commit, but I don't see why it would surface here and not elsewhere.

I had in mind only the big number as return code

python/libgrass_interface_generator/CMakeLists.txt
nilason: It is an early debugging left-over, which shouldn't have remained at all, and doesn't deserve a PR in its own right.

nilason:
Thanks for updating to main! Still occupied the next couple of days, but will check out the failures as soon as possible.

wenzeslaus:
The Windows failure is a stack overflow caused by infinite recursion in `G_init_locale()`, introduced by this PR. The g.proj return code 3221225725 is `0xC00000FD` = `STATUS_STACK_OVERFLOW`, and the process dies before it can print anything, which is why no error message is produced.

The recursion chain:

1. `G_init_locale()` now calls `G_locale_dir()` to get the translations directory.
2. When `GRASS_LOCALEDIR` is not set, `get_g_env()` in `resource_dirs.c` calls `G_fatal_error(_("Incomplete GRASS session: ..."))`.
3. The `_()` macro is `G_gettext()`, whose first act is to call `G_init_locale()` again. The `initialized` flag is only set at the *end* of `G_init_locale()`, which is never reached, so the recursion is unbounded.

This explains why it shows up where it does:

- The failing tests create projects, and `grass.script` runs `g.proj` in a constructed minimal environment that has `GISBASE` but not `GRASS_LOCALEDIR`, so the first translated string triggers the recursion.
- It only fails on Windows because the affected code is inside `#if defined(HAVE_LIBINTL_H) && defined(USE_NLS)`: the OSGeo4W build configures with `--with-nls`, while the Ubuntu CI builds do not, so on Linux `G_init_locale()` never reaches this code.

Fix: read the variable directly in `G_init_locale()` and keep a missing locale directory non-fatal, exactly like the previous `GISBASE`-based code did. Locale initialization can never be allowed to fail fatally, because `G_fatal_error()` itself translates its message. Note that the current `if (localedir && *localedir)` check is dead code (`get_g_env()` either returns a value or exits), so the non-fatal behavior seems to have been the intent here anyway; this restores it:

```diff
--- a/lib/gis/locale.c
+++ b/lib/gis/locale.c
@@ -39,7 +39,12 @@ void G_init_locale(void)
 #ifdef LC_MESSAGES
     setlocale(LC_MESSAGES, "");
 #endif
-    const char *localedir = G_locale_dir();
+    /* Read the variable directly instead of using G_locale_dir():
+       G_locale_dir() calls G_fatal_error() when the variable is not set,
+       and G_fatal_error() translates its message, which re-enters
+       G_init_locale() and recurses without bound. Missing translations
+       must never be fatal here. */
+    const char *localedir = getenv("GRASS_LOCALEDIR");
 
     if (localedir && *localedir) {
         bindtextdomain("grasslibs", localedir);
```

I have this fix applied on a local branch based on this PR and can confirm the builds and session smoke tests pass with it.

A related, broader suggestion: the same "environment has `GISBASE` but not the new variables" situation can hit every `get_g_env()` call site (as `G_fatal_error` in C, or `KeyError`/`TypeError` in Python), for example with environments constructed by third-party launchers or older `grass.script` versions. On my local branch, `get_g_env()` takes a second argument with the legacy `GISBASE`-relative subdirectory and falls back to `$GISBASE/<subdir>` when the variable is unset, failing fatally only when neither source yields an existing path. That keeps `GISBASE`-only environments working with the legacy layout while enabling the new variables. Happy to submit either the minimal locale fix or the fallback version as a PR against this branch.

The diagnosis and this write-up were AI-assisted (Claude Code); the fix and its verification were reviewed and tested locally.

nilason:
Window still fails. Now with:

```
D:\a\grass\grass>call C:\OSGeo4W\opt\grass\grass86.bat --tmp-project EPSG:4326 --exec g.region res=0.1 -p 
ERROR: Incomplete GRASS session: Variable 'GRASS_ETCDIR' not set
```

Any suggestions?

echoix:
I'm currently solving conflicts for this one

echoix:
There's at least one place where some newly merged code should need to apply the same kind of changes that where applied here, but I wasn't confident enough so I didn't take the liberty to add it in a commit. So it is probably missing some changes before being able to merge.

PR #5681, opened 2025-05-19
Python: PyGRASS Module description property not defined always

landam:
The PyGRASS `Module` class sets properties dynamically from parsed XML description. In result the API is not really predictable:

```py
from grass.pygrass.modules import Module

for cmd in ['r.import', 'r.relief']:
    m = Module(cmd, run_=False)
    print(m.name, m.description)
```

Fails on `r.relief` which has only label defined (https://github.com/OSGeo/grass/blob/main/raster/r.relief/main.c#L109).
 
```
r.import Imports raster data into a GRASS raster map using GDAL library and reprojects on the fly.
  File "/home/martin/Downloads/test_module.py", line 5, in <module>
    print(m.name, m.description)
                  ^^^^^^^^^^^^^
AttributeError: 'Module' object has no attribute 'description'
```

This PR overcomes this issue by defining `Module` class `description` property if doesn't exists.


python/grass/pygrass/modules/interface/module.py
```diff
@@ -566,6 +566,16 @@ def __init__(self, cmd, *args, **kargs):
             if e.tag not in {"parameter", "flag"}:
                 self.__setattr__(e.tag, GETFROMTAG[e.tag](e))
 
+        # some modules may have a label defined: in this case stay
+        # consistent with other modules -> define 'description'
+        # property instead
+        if hasattr(self, "description") is False:
+            if hasattr(self, "label") is True:
```
wenzeslaus, line 573: Isn't `if hasattr(...)` and `if not hasattr(...)` more readable here than `if hasattr(...) is True/False`?

wenzeslaus (changes requested):
This is hiding the actual metadata of the underlying module. Why would you want to hide it here? If you consider Module as a simpler interface to the full underlying Module, than that's okay. If it is meant to simple present the module, it is little questionable. I don't have a strong opinion either way.

landam:
> This is hiding the actual metadata of the underlying module. Why would you want to hide it here? If you consider Module as a simpler interface to the full underlying Module, than that's okay. If it is meant to simple present the module, it is little questionable. I don't have a strong opinion either way.

@wenzeslaus Right, hiding `label` property is not correct. These properties are set dynamically from XML UI description. In the case of modules, the most of them have `description` defined, only few a `label` defined (eg. `r.relief`). Some of them probably both. By 00ce31ed70:

- `label` property only exists when defined by UI (no change)
- `description` property returns description if defined otherwise `label` (so the property is always available - new behaviour)

Note: 08799546a6 in the case that none of `description` and `label` are not defined `"?"` is returned

PR #5682, opened 2025-05-19
CQ: gui_core.forms: avoid code duplicity of GrassGUIApp(task).MainLoop()

pesekon2:
The same was done at the end of both `if-else` condition branches. Running it just once reduces the places where we need to fix it if something goes wrong.


echoix (approved):
I fully agree, but I expect static analysis/type checkers to have difficulty proving that task is defined when used. But, we see that the variable is assigned in each of condition's branches. So it should be totally fine. (Mypy won't like it for sure, pyright might be able, pyre or pyrefly I have no clue. I didn't start trying astral's ty yet).

PR #5695, opened 2025-05-20
v.surf.rst: Corrected defaults parameters and documentation

cwhite911:
I've corrected the default parameters for smoothing and dmin to match the code and updated the documentation.


petrasovaa (dismissed):

echoix (approved):
0.1 is the value of #define SMOOTH that is the default. For dmin, it corresponds to the code indeed. A little typo in "value" to fix

echoix (approved):

PR #5700, opened 2025-05-20
wxGUI: avoid creating nested list of errors (Graphical modeler)

landam:
This PR fixes regression caused by 3f98304d0d7bbc71727d07f3530370bcd7cdbbde:

Steps to reproduce:

1. create model with parameterized option (see attached [model.zip](https://github.com/user-attachments/files/20354205/model.zip))

2. run model

It fails with:

```py
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/gui/wxpython/gmodeler/panels.py", line 966, in OnRunModel
    self.model.Run(self._gconsole, self.OnModelDone, parent=self)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/martin/src/grass/dist.x86_64-pc-linux-gnu/gui/wxpython/gmodeler/model.py", line 668, in Run
    GError(parent=parent, message="\n".join(err))
                                  ~~~~~~~~~^^^^^
TypeError: sequence item 0: expected str instance, list found
```

### DIFF
--- a/gui/wxpython/gmodeler/model.py
+++ b/gui/wxpython/gmodeler/model.py
@@ -2673,7 +2673,7 @@ class ModelParamDialog(wx.Dialog):
 
     def GetErrors(self):
         """Check for errors, get list of messages"""
-        return [task.get_cmd_error() for task in self.tasks]
+        return [err for task in self.tasks for err in task.get_cmd_error()]
 
     def DeleteIntermediateData(self) -> bool:
         """Check if to delete intermediate data"""
--- a/python/grass/script/task.py
+++ b/python/grass/script/task.py
@@ -159,7 +159,7 @@ class grassTask:
                 return f
         raise ValueError(_("Flag not found: %s") % aFlag)
 
-    def get_cmd_error(self):
+    def get_cmd_error(self) -> list[str]:
         """Get error string produced by get_cmd(ignoreErrors = False)
 
         :return: list of errors


echoix (commented):
I would even add a type annotation for the return of the get_cmd_error() to make it clear:
https://github.com/OSGeo/grass/blob/7e5bcd6cb710d720635ef19df6cf35f632159f4c/python/grass/script/task.py#L162-L184

Like:
```python
    def get_cmd_error(self): -> list[str]
        """Get error string produced by get_cmd(ignoreErrors = False)

        :return: list of errors
        """
        errorList = []
        # determine if suppress_required flag is given
        for f in self.flags:
            if f["value"] and f["suppress_required"]:
                return errorList

        for p in self.params:
            if not p.get("value", "") and p.get("required", False):
                if not p.get("default", ""):
                    desc = p.get("label", "")
                    if not desc:
                        desc = p["description"]
                    errorList.append(
                        _("Parameter '%(name)s' (%(desc)s) is missing.")
                        % {"name": p["name"], "desc": desc}
                    )

        return errorList
```

landam:
> type annotation for the return of the get_cmd_error()

@echoix Done in c9a90b0dee

echoix (approved):
Looks fine, and the order of the two "for" in the comprehension looks correct (its supposed to be in the same order as when we write nested loops normally)

pesekon2 (approved):

PR #5706, opened 2025-05-20
lib/gpde: fix typos

luzpaz:
Found via codespell


echoix (approved):
Thanks for your great work! This week we are in a sprint with many many of the contributors all together, so your contributions are very welcome at this time!

PR #5710, opened 2025-05-21
db.connect: create function substitute_variable() to reduce code duplicity

pesekon2:
making a function for the repeated code reduces the places where we need to fix it if something goes wrong


pesekon2:
A discussion with C experts resulted in "not accepting" the PR. The function call would apparently give a superslight overhead to the module.

PR #5722, opened 2025-05-21
g.proj: upgrade to format=proj4 style text output

lindakarlovska:
This PR gets rid of the warning (initially found when using the Jupyter Notebook) when in the g.proj module the 'j' is deprecated and will be removed in a future release. Instead, the format=proj4 style is used.

The change was applied in the whole GRASS repository, the occurrences were found using 
`grep -Irni 'flags="[^"]*j[^"]*"'`

The info about deprecation can be found here: https://grass.osgeo.org/grass-devel/manuals/g.proj.html .

### DIFF
--- a/general/g.proj/testsuite/test_g_proj.py
+++ b/general/g.proj/testsuite/test_g_proj.py
@@ -60,7 +60,7 @@ class GProjTestCase(TestCase):
 
     def test_proj4_output(self):
         """Test if g.proj returns consistent PROJ4 output."""
-        module_flag = SimpleModule("g.proj", flags="j")
+        module_flag = SimpleModule("g.proj", flags="p", format="proj4")
         self.assertModule(module_flag)
         result_flag = module_flag.outputs.stdout
         self.assert_keys_in_output(result_flag, "+")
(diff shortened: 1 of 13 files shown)


scripts/r.import/r.import.py
```diff
@@ -150,7 +150,7 @@ def is_projection_matching(GDALdatasource):
     """Returns True if current location projection
     matches dataset projection, otherwise False"""
     try:
-        gs.run_command("r.in.gdal", input=GDALdatasource, flags="j", quiet=True)
+        gs.run_command("r.in.gdal", input=GDALdatasource, format="proj4", quiet=True)
```
pesekon2, line 153:
```suggestion
        gs.run_command("r.in.gdal", input=GDALdatasource, flags="j", quiet=True)
```

pesekon2 (changes requested):

scripts/r.import/r.import.py
pesekon2: I think this one slipped in by mistake. It's `r.in.gdal`

pesekon2 (approved):

PR #5733, opened 2025-05-21
GUI: make Python editor dockable

landam:
Currently Python simple editor is not dockable:

image

This PR makes the editor dockable in the similar way as eg. Graphical Modeler:

image


landam:
@pesekon2 

https://github.com/user-attachments/assets/bfac103f-39a0-4650-ad85-1449edaba61f

PR #5737, opened 2025-05-22
wxGUI/gmodeler: change syntax of variables (%variable -> %{variable})

### LINKED ISSUE #3870: [Bug] g.gui.gmodeler: having `%` in module's arguments makes the model invalid
## Describe the bug

If `%` is included in any of the model's modules' parameters, the validator complains that the model is invalid. This is caused by the fact that it thinks that it is a variable that is undefined. The model can run but this message is annoying (and wrong).

## To reproduce

Try to put `%` into some parameter. The good option where this can happen often is `v.db.select`/`v.extract` with the `where` option (`col1 LIKE 'A%'`).

## Idea for solution

Change the syntax for variables in the modeler. The pure `%` prefix is bad because it can fail in case of variables like `%d` and `%donkey` (one of them is pseudorandomly chosen; if it chooses `d` over `donkey`, it thinks that `onkey` is a string following the variable).

## System description
- Operating System: Linux (Ubuntu)
- GRASS GIS version: `main` as of 60f67276fbee985042f9fc571f824bb2c863ffbe

pesekon2:
fix #3870

BEWARE: it breaks older models that used the previous syntax - but as that syntax was buggy, the back-support is not the right way, see #3870


landam:
@pesekon2 The screetshot (g_gui_gmodeler_variables_raster.png) needs to be updated:

image

landam (approved):
Works as expected. 

BTW, I took to compress modified image (see https://github.com/OSGeo/grass/blob/main/doc/development/style_guide.md#images)

PR #5742, opened 2025-05-22
r.mapcalc: Support parallel computing by OpenMP

cyliang368:
This PR parallelizes a part of the r.mapcalc module using OpenMP. Note that raster3d is still not parallelized yet.

- [x] add note and benchmark to documentation
- [x] add openmp to CMakeLists.txt


petrasovaa:
*NOTE: this is not valid benchmark beyond 4 cores, see the one below*

I ran a benchmark on my desktop and it looks very good! I compared different data size and complexity of expression:
r_mapcalc_time
r_mapcalc_speedup
r_mapcalc_efficiency

My machine info:
```text
Architecture:             x86_64
  CPU op-mode(s):         32-bit, 64-bit
  Address sizes:          46 bits physical, 48 bits virtual
  Byte Order:             Little Endian
CPU(s):                   28
  On-line CPU(s) list:    0-27
Vendor ID:                GenuineIntel
  Model name:             Intel(R) Core(TM) i9-10940X CPU @ 3.30GHz
    CPU family:           6
    Model:                85
    Thread(s) per core:   2
    Core(s) per socket:   14
    Socket(s):            1
    Stepping:             7
    CPU(s) scaling MHz:   33%
    CPU max MHz:          4800.0000
    CPU min MHz:          1200.0000
```

Overall I don't see any slow down with more cores as @cyliang368 showed me before, so it may behave differently on different machines. I will do more testing but this looks very promising!

To clarify the plots, 'simple' means A+B, complex is `(A + B - 2 * B + A * B - A / B) / 2` and neighbors is `if ((A[5, 5] + B[-5, -5]) > 0, 1.6, 0.1)`. So the complexity of the expression makes a big difference and given there are often even larger expressions, this has potential to be even more useful.

echoix:
So, in the tests you ran, there's not much benefit after 4 cores. Does it apply to everything, or just the expressions tested?

petrasovaa:
> So, in the tests you ran, there's not much benefit after 4 cores. Does it apply to everything, or just the expressions tested?

I suspect much larger data and more complex expression (like with eval function), we would see benefit with couple more cores. But most parallel tools in GRASS in my experience behave like that.

cyliang368:
This reminds me of a similar behavior I observed while running benchmarks for `v.surf.rst` and `r.texture` on my laptop last year. I believe my laptop is not powerful enough, so more overheads occur when more cores are used. 

Overall, I think the benchmarks from @petrasovaa 's machine look good. I will add notes and these benchmarks to the document when I have time.

petrasovaa:
So I needed to run new benchmarks because I missed the fact that the code was already constrained to 4 cores max... So I removed locally the restriction and luckily the new results are not that much different:

r_mapcalc_time
r_mapcalc_speedup
r_mapcalc_efficiency

Based on this I would suggest to not constrain it, there may be some slow down for simple expressions, but it doesn't look that significant. Also, the more complex expression was able to take advantage of more than 4 cores.

petrasovaa:
Hm, I meant to test it on my fork, but apparently pushed it here...

petrasovaa:
It looks like `test_nmedian_bug_3296.py` is failing on windows. One test is crashing and other give incorrect results.

raster/r.mapcalc/map.c
```diff
@@ -532,26 +534,6 @@ int open_map(const char *name, int mod, int row, int col)
...
-
-        if (use_cats && !m->have_cats)
-            init_cats(m);
-
-        if (use_colors && !m->have_colors)
-            init_colors(m);
-
-        return i;
-    }
-
```
petrasovaa, line 554: @cyliang368 why has this part been completely removed?

cyliang368: I will get back to you tonight.

petrasovaa:
> It looks like `test_nmedian_bug_3296.py` is failing on windows. One test is crashing and other give incorrect results.

My guess is, it is connected to using static variables:
https://github.com/OSGeo/grass/blob/main/lib/calc/xnmedian.c#L47

On my fork I removed the reallocation so it allocates every time and the tests are running now. If that's indeed the cause the question is how to deal with that, @HuidaeCho, @cyliang368 any idea? I assume the constant allocation would slow it down. Maybe there could be a smarter way to allocate it.

raster/r.mapcalc/map.c
cyliang368: This part is removed because it causes read conflicts when multithreads are used.

Originally, this part is used to read stored and initialized maps when the same maps are shown multiple times in an expression (e.g. `b = a + a`). However, an issue arises when threads attempt to access the same map. This part redirects threads to read the same `fd` (specified L576) from the same `map` struct. However, threads can have different progress, so conflicts arise over the location of the file pointer (`fd`), and the reading order becomes messed up.

Removing this part allows every thread to create its own `map` structs, which are stored in `maps`. Every`map` struct is guaranteed to have a different `fd` because `fd` is initialized every time, even if the same map is used in other threads and expressions. This way can avoid the read conflict mentioned above.

Since a `map` struct only contains pointers and some metadata instead of arrays, this implementation consumes just a small amount of additional memory.

cyliang368:
> > It looks like `test_nmedian_bug_3296.py` is failing on windows. One test is crashing and other give incorrect results.
> 
> My guess is, it is connected to using static variables: https://github.com/OSGeo/grass/blob/main/lib/calc/xnmedian.c#L47
> 
> On my fork I removed the reallocation so it allocates every time and the tests are running now. If that's indeed the cause the question is how to deal with that, @HuidaeCho, @cyliang368 any idea? I assume the constant allocation would slow it down. Maybe there could be a smarter way to allocate it.

I haven't had a good idea about this error, which only occurs on Windows, now. 😢

raster/r.mapcalc/map.c
petrasovaa: Thanks for the explanation, so it is basically some kind of optimization that's not necessarily needed.

cyliang368:
> > It looks like `test_nmedian_bug_3296.py` is failing on windows. One test is crashing and other give incorrect results.
> 
> My guess is, it is connected to using static variables: https://github.com/OSGeo/grass/blob/main/lib/calc/xnmedian.c#L47
> 
> On my fork I removed the reallocation so it allocates every time and the tests are running now. If that's indeed the cause the question is how to deal with that, @HuidaeCho, @cyliang368 any idea? I assume the constant allocation would slow it down. Maybe there could be a smarter way to allocate it.

@petrasovaa To my knowledge, these static variables are shared within this function and are not thread-safe, which causes problems in multithreading. The implementation on your fork will need to allocate an array whenever this function is called. I don't have a good way to solve this disadvantage, and it could be a hard limitation of parallelization.

I am not familiar with this function, but I think it is seldom used, especially compared to `r.mapcalc`. Since the tests are all passed on your fork, and this PR can bring speedup to `r.mapcalc`, removing the static variable and merging this PR may be a good choice.

petrasovaa:
> @petrasovaa To my knowledge, these static variables are shared within this function and are not thread-safe, which causes problems in multithreading. The implementation on your fork will need to allocate an array whenever this function is called. I don't have a good way to solve this disadvantage, and it could be a hard limitation of parallelization.
> 
> I am not familiar with this function, but I think it is seldom used, especially compared to `r.mapcalc`. Since the tests are all passed on your fork, and this PR can bring speedup to `r.mapcalc`, removing the static variable and merging this PR may be a good choice.

In #6087 the static variables were removed so I believe this is good to go.

raster/r.mapcalc/evaluate.c
```diff
@@ -358,29 +440,50 @@ void execute(expr_list *ee)
 
     setup_maps();
 
+#if defined(_OPENMP)
+    threads = omp_get_max_threads();
```
nilason, line 444: Just taking a quick look at this, isn't number of threads set in main.c?

cyliang368: Yes, the number of threads basically is set in `main.c`. The way I parallelized `r.mapcalc` is to let each thread process a row. If the number of threads exceeds the number of rows in the rasters,  it's meaningless and causes an error. To avoid that, the number of threads must be adjusted after the program can access the number of rows. Therefore, this adjustment is written here instead of `main.c`.

nilason: Thanks!
(Now I checked what omp_get_max_threads actually does — all clear).

petrasovaa:
I'll merge this couple days unless somebody else wants to review it and needs more time.

nilason (approved):
Great job, thanks!

echoix:
If this could be merged before tomorrow 11:32 EDT (15:32 UTC) (in 14 hours), the new strings would be included in this week's translation update, otherwise it'll be next week.

petrasovaa:
@cyliang368 could you please provide a commit message that would explain the parallelization? Thanks!

cyliang368:
> @cyliang368 could you please provide a commit message that would explain the parallelization? Thanks!

parallelize r.mapcalc (2D) using OpenMP by making every thread calculate a corresponding row across rasters

raster/r.mapcalc/map.c
wenzeslaus: Can you look at the follow up in #6241, @cyliang368?

cyliang368:
I haven't had time recently. I will take a look when I have time.

wenzeslaus:
I'm getting a segfault and errors from Valgrind (Helgrind) when using the _rand_ function without `-s` with higher numbers of threads set with nprocs (>=4 or 3, but not 2).

```python
# Take your pick for the Valgrind command
grass --tmp-mapset ~/grassdata/nc_spm_08_grass7/ --exec valgrind r.mapcalc --o expr="test_elev_double_1 = double(rand(-15.0, 5.0))" nprocs=3
valgrind --tool=helgrind --trace-children=yes grass run --project ~/grassdata/nc_spm_08_grass7/user1/ r.mapcalc --o expr="test_elev_double_1 = double(rand(-15.0, 5.0))" nprocs=3
```

Other expressions give Valgrind at least some  errors, too, although not all of them fail with segfault.

marisn:
> I'm getting a segfault and errors from Valgrind (Helgrind) when using the _rand_ function without `-s` with higher numbers of threads set with nprocs (>=4 or 3, but not 2).

That is because r.mapcalc uses random number generator from lib/gis that is not thread safe and should not be called from multiple threads as it keeps internal state in static variables (see lib/gis/lrand48.c) that get updated by G__next() call leading to a race condition. Any of parallel modules directly or indirectly(!) calling any of random number generator functions have the same problem.

When parallelism is implemented in any of modules, underlying library calls must be evaluated to determine their thread safety. I'd vote that a comment should be added to each library function if it is safe or not safe to be called from multiple threads. As it is a huge task, it can be done step-by-step. E.g. all functions in lrand48.c can be marked as single thread only. Next topic, of course, is multiprocess safety – this applies to file reading/writing functions that are not designed to lock resources etc. GRASS is designed as a single process, single thread software and it shows.

Issue #5759, opened 2025-05-23
[Bug] GUI: where option is removed in Graphical modeler when properties dialog is re-open


landam:
## Describe the bug

When re-opening a tool dialog in Graphical modeler which contains `where` option a parameter value is not set correctly.

## To reproduce

1. Open graphical modeler
2. Add a tool with `where` option (eg. `v.db.select`)
3. Set tool parameters
4. Close dialog
5. Re-open dialog

The `where` option is not set. See the dialog statusbar in attached screeshots.

## Expected behavior

The `where` option contains previously set value. 

## Screenshots

When the tool is added to the model and parameters defined:

![Image](https://github.com/user-attachments/assets/5c68bf87-9eba-4727-88b9-bc0c779836e7)

When the dialog is re-open (no option changed):

![Image](https://github.com/user-attachments/assets/232e3128-d229-492d-9473-f9742f5060c8)

## System description

- Operating System: Linux
- GRASS GIS version: 8.4

PR #5760, opened 2025-05-23
GUI: do not remove parameter value on dialog update event

landam:
Currently when `map/input` option is changed than also all depending widgets (`layer`, `columns`, `where`) are updated. It includes also setting value to an empty string.

## Steps to reproduce

1. Launch ` v.db.select map=schools where="TAG = '001'" --ui`
2. Change `map` option

-> `where` option is removed

https://github.com/user-attachments/assets/456adaab-ebb1-4277-97dc-da9b2322b021

This PR changes current behaviour 

-> `where` option is kept

https://github.com/user-attachments/assets/dd5795b1-c990-45a0-bff1-ca508a91fe44

## Backround reasons

- keep decision to remove or update eg. `where` option on the user
- user may have multiple maps with the same column. When switching between maps, a `where` statement must be set again from the scratch (which is not desired in this case)
- solves #5759

### DIFF
--- a/gui/wxpython/gui_core/forms.py
+++ b/gui/wxpython/gui_core/forms.py
@@ -208,10 +208,6 @@ class UpdateThread(Thread):
 
             # @todo: replace name by isinstance() and signals
 
-            pBind = self.task.get_param(uid, element="wxId", raiseError=False)
-            if pBind:
-                pBind["value"] = ""
-
             # set appropriate types in t.* modules and g.list/remove element
             # selections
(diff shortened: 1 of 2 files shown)


landam:
@pesekon2 Also implemented for `ColumnSelect`:

https://github.com/user-attachments/assets/ba56f8e7-f3b9-40cb-8ef5-afb9b9164602

petrasovaa (approved):

PR #5772, opened 2025-05-24
r.univar: OpenMP private variables must be initialized

marisn:
Fix an issue reported by valgrind.

### DIFF
--- a/raster/r.univar/r.univar_main.c
+++ b/raster/r.univar/r.univar_main.c
@@ -381,6 +381,7 @@ static void process_raster(univar_stat *stats, thread_workspace *tw,
 #pragma omp parallel private(row, c_sum, c_sumsq, c_sum_abs)
     {
         int t_id = 0;
+        c_sum, c_sumsq, c_sum_abs = 0;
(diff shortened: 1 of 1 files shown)


marisn:
According to https://rookiehpc.org/openmp/docs/private/index.html

> [private](https://rookiehpc.org/openmp/docs/private/index.html) is the clause that contains the variables that each thread in the OpenMP [parallel](https://rookiehpc.org/openmp/docs/parallel/index.html) region will have a copy of. These copies are not initialised, specifically, they [do](https://rookiehpc.org/openmp/docs/do/index.html) not have the value that the original variable had at the time it was passed in the [private](https://rookiehpc.org/openmp/docs/private/index.html) clause.

echoix (approved):
According to the linked docs, private copies are not initialized like the original variable was.

echoix (commented):
Seems LLVM on macOS doesn't like that

echoix (approved):

nilason:
This cannot be backported to 8.4 without #5750.

PR #5809, opened 2025-06-01
docs: update grass-stable URL to grass-devel

neteler:
Update manual URLS from `grass-stable` -> `grass-devel`.

Addresses #5162 (partially)


neteler:
It was a `ag` and `sed` one-liner, should be easy.

What I did:

```sh
ag --markdown "grass.osgeo.org/grass-stable" -l | xargs sed -i 's+osgeo.org/grass-stable+osgeo.org/grass-devel+g'
```

wenzeslaus:
Wouldn't be better to have the "stable" URLs everywhere regardless of a version and then change them during the doc build if needed, e.g., by adding that one-liner to the CI doc build? That way, the source code can be the same on all branches.

neteler:
@wenzeslaus would that be somewhere here?
https://github.com/OSGeo/grass/blob/4c1e37ee4db92fd3ae7ffcac1bdde6bc234f2a35/.github/workflows/documentation.yml#L186

PR #5845, opened 2025-06-05
docs: Add format to example even when not required

wenzeslaus:
The Python example counts on format option being there, so we need to add that even if it is not required.
This allows for empty and not required format in v.info in #5844.


echoix (approved):
Reapproving

